PR #2, opened 2019-10-16
MF-84: Create Profile Widget for Patient Chart

fatmali:
Screen Shot 2019-10-16 at 12 36 35 PM

### DIFF
--- a/src/summary/banner/patient-banner.component.tsx
+++ b/src/summary/banner/patient-banner.component.tsx
@@ -1,0 +1,116 @@
+import React from "react";
+import styles from "./patient-banner.component.css";
+import { age } from "../profile/age-helpers";
+import dayjs from "dayjs";
+import ProfileSection from "../profile/profile-section.component";
+
+export default function PatientBanner(props: PatientBannerProps) {
+  const [showDemographics, setShowDemographics] = React.useState(false);
+
+  function getPatientNames() {
+    return `${props.patient.name[0].family.toUpperCase()}, 
+             ${props.patient.name[0].given.join(" ")}`;
+  }
+
+  function getPreferredIdentifier() {
+    return (
+      props.patient.identifier.find(id => id.use === "usual").value ||
+      props.patient.identifier[0].value
+    );
+  }
+
+  return (
+    <>
+      {showDemographics ? (
+        <div className={styles.profileWrapper}>
+          <div className={styles.collapseBtnWrapper}>
+            <button
+              className={`${styles.collapseBtn} omrs-btn omrs-rounded`}
+              onClick={() => setShowDemographics(false)}
+            >
+              collapse
+            </button>
+            <svg className="omrs-icon" fill="var(--omrs-color-inactive-grey)">
+              <use xlinkHref="#omrs-icon-chevron-down" />
+            </svg>
+          </div>
+          <ProfileSection patient={props.patient} match={props.match} />
+        </div>
+      ) : (
+        <div className={styles.patientBanner}>
+          {props.patient && (
+            <>
+              <div className={styles.demographics}>
+                <div className={`${styles.patientName} omrs-type-title-5`}>
+                  {getPatientNames()}
+                </div>
+                <div className={`${styles.otherDemographics}`}>
+                  <span
+                    className={`${styles.demographic} omrs-type-body-regular`}
+                  >
+                    {age(props.patient.birthDate)}
+                  </span>
+                </div>
+                <div className={`${styles.otherDemographics}`}>
+                  <span
+                    className={`${styles.desktopLabel} omrs-type-body-small`}
+                  >
+                    Born
+                  </span>
+                  <span
+                    className={`${styles.demographic} omrs-type-body-regular`}
+                  >
+                    {dayjs(props.patient.birthDate).format("DD-MMM-YYYY")}
+                  </span>
+                </div>
+                <div className={`${styles.otherDemographics}`}>
+                  <span
+                    className={`${styles.desktopLabel} omrs-type-body-small`}
+                  >
+                    Gender
+                  </span>
+                  <span
+                    className={`${styles.demographic} omrs-type-body-regular`}
+                  >
+                    {props.patient.gender}
+                  </span>
+                </div>
+                <div className={`${styles.otherDemographics}`}>
+                  <span
+                    className={`${styles.desktopLabel} omrs-type-body-small`}
+                  >
+                    Preferred ID
+                  </span>
+                  <span
+                    className={`${styles.demographic} omrs-type-body-regular`}
+                  >
+                    {getPreferredIdentifier()}
+                  </span>
+                </div>
+              </div>
+              <div className={styles.moreBtn}>
+                <button
+                  className={`${styles.moreBtn} omrs-unstyled`}
+                  onClick={() => setShowDemographics(true)}
+                >
+                  more
+                  <svg
+                    className="omrs-icon"
+                    fill="var(--omrs-color-inactive-grey)"
+                  >
+                    <use xlinkHref="#omrs-icon-chevron-down" />
+                  </svg>
+                </button>
+              </div>
+            </>
+          )}
+        </div>
+      )}
+    </>
+  );
+}
+
+type PatientBannerProps = {
+  match: any;
+  patient: fhir.Patient;
+};
--- a/src/summary/banner/patient-banner.component.tsx
+++ b/src/summary/banner/patient-banner.component.tsx
@@ -9,7 +9,7 @@ export default function PatientBanner(props: PatientBannerProps) {
 
   function getPatientNames() {
     return `${props.patient.name[0].family.toUpperCase()}, 
-             ${props.patient.name[0].given.join(" ")}`;
+             ${props.patient.name[0].given[1]}`;
   }
 
   function getPreferredIdentifier() {
--- a/src/summary/banner/patient-banner.component.tsx
+++ b/src/summary/banner/patient-banner.component.tsx
@@ -5,11 +5,85 @@ import dayjs from "dayjs";
 import ProfileSection from "../profile/profile-section.component";
 
 export default function PatientBanner(props: PatientBannerProps) {
-  const [showDemographics, setShowDemographics] = React.useState(false);
+  const [showingDemographics, setShowDemographics] = React.useState(false);
+
+  return (
+    <aside className={styles.patientBanner}>
+      {props.patient && (
+        <div className={styles.patientBanner}>
+          <div className={styles.demographics}>
+            <div className={`${styles.patientName} omrs-type-title-5`}>
+              {getPatientNames()}
+            </div>
+            <div className={`${styles.otherDemographics}`}>
+              <span className={`${styles.demographic} omrs-type-body-regular`}>
+                {age(props.patient.birthDate)}
+              </span>
+            </div>
+            <div className={`${styles.otherDemographics}`}>
+              <span className={`${styles.desktopLabel} omrs-type-body-small`}>
+                Born
+              </span>
+              <span className={`${styles.demographic} omrs-type-body-regular`}>
+                {dayjs(props.patient.birthDate).format("DD-MMM-YYYY")}
+              </span>
+            </div>
+            <div className={`${styles.otherDemographics}`}>
+              <span className={`${styles.desktopLabel} omrs-type-body-small`}>
+                Gender
+              </span>
+              <span className={`${styles.demographic} omrs-type-body-regular`}>
+                {props.patient.gender}
+              </span>
+            </div>
+            <div className={`${styles.otherDemographics}`}>
+              <span className={`${styles.desktopLabel} omrs-type-body-small`}>
+                Preferred ID
+              </span>
+              <span className={`${styles.demographic} omrs-type-body-regular`}>
+                {getPreferredIdentifier()}
+              </span>
+            </div>
+          </div>
+          <div className={styles.moreBtn}>
+            <button
+              className={`${styles.moreBtn} omrs-unstyled`}
+              onClick={toggleDemographics}
+            >
+              {showingDemographics ? "Close" : "Open"}
+            </button>
+            <svg
+              className={`omrs-icon`}
+              fill="var(--omrs-color-ink-medium-contrast)"
+            >
+              <use
+                xlinkHref={
+                  showingDemographics
+                    ? "#omrs-icon-chevron-up"
+                    : "#omrs-icon-chevron-down"
+                }
+              />
+            </svg>
+          </div>
+        </div>
+      )}
+      {showingDemographics && (
+        <div className={styles.patientProfile}>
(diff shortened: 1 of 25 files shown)


src/summary/profile/demographics-card.component.tsx
```diff
@@ -9,30 +8,25 @@ import dayjs from "dayjs";
...
-  const [currentPatient, setCurrentPatient] = React.useState(null);
-
-  React.useEffect(() => {
-    const subscription = getCurrentPatient().subscribe(patient =>
-      setCurrentPatient(patient)
-    );
-
-    return () => subscription.unsubscribe();
-  });
-
```
fatmali: Decided to have `profile-section` pass in `currentPatient` instead of repeating this code in every widget.

jonathandick: Shouldn’t we did this for all of patient chart?

fatmali: good point

src/summary/profile/identifiers-card.component.tsx
```diff
@@ -0,0 +1,46 @@
...
+    fontWeight: 500
+  };
+
+  return (
+    <SummaryCard name="Identifiers" match={props.match}>
+      {props.currentPatient &&
+        props.currentPatient.identifier.map(id => (
+          <SummaryCardRow linkTo="/" key={id.system}>
+            <SummaryCardRowContent>
+              <HorizontalLabelValue
```
jonathandick: I think it's important to make sure the preferred identifier is the first one in the list.

fatmali:
updated with @jonathandick comment;

joeldenning:
Please make any updates needed to match the 1.3.0 designs in Figma. Then let me know when this is ready for review. Specifically, the SummarySectionHeader component should be removed entirely now that it is gone from the designs.

src/summary/patient-chart-summary.css
```diff
@@ -5,3 +5,11 @@
 :global(.omrs-breakpoint-gt-phone) .patientSummary {
   padding: 1.25rem 1.5rem 0 1.5rem;
 }
+
+body {
```
joeldenning: This css rule should either go into the styleguide or should be removed. The reason is that this impacts the singular `<body>` element that the entire DOM and all the single-spa applications use. Putting it into esm-patient-chart means that the body would not have 0 margin before the patient chart is loaded, but then it *would* have 0 margin after you navigate to the patient chart. With css-loader (and css modules), once a css file is loaded, it is never unmounted. So then this margin 0 would apply to the body even after patient-chart's single-spa application is unmounted.

src/summary/patient-chart-summary.css
```diff
@@ -5,3 +5,11 @@
...
+  margin: 0;
+}
+.patientSummary {
+  background-color: var(--omrs-color-bg-lowest-contrast);
+  min-height: 90vh;
```
joeldenning: I don't like this -- it's better to just let content flow naturally and let scrolling be completely determined by content size and viewport height. Why do we need this?

src/summary/profile/contacts-card.component.tsx
```diff
@@ -0,0 +1,54 @@
+import React from "react";
+import SummaryCard from "../cards/summary-card.component";
+import SummaryCardRow from "../cards/summary-card-row.component";
+import SummaryCardRowContent from "../cards/summary-card-row-content.component";
+import VerticalLabelValue from "../cards/vertical-label-value.component";
+import { match } from "react-router";
+
+export default function ContactsCard(props: ContactsCardProps) {
+  function getAddress(address: fhir.Address) {
+    return `${address.city}, ${address.country} \n `;
```
joeldenning: Why is the ` \n ` needed?

src/summary/profile/contacts-card.component.tsx
```diff
@@ -0,0 +1,54 @@
...
+              <VerticalLabelValue
+                label={address.use}
+                value={getAddress(address)}
+              />
+            </SummaryCardRowContent>
+          </SummaryCardRow>
+        ))
+      ) : (
+        <SummaryCardRow>
+          <SummaryCardRowContent>None</SummaryCardRowContent>
```
joeldenning: Is this in the designs? We've been using emdash for when content is missing -- any reason to deviate from that here?

Or even should we just not render a SummaryCardRow at all when there is no address?

src/summary/profile/contacts-card.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+    );
+  });
+
+  it("displays contacts correctly", () => {
+    const wrapper = render(
+      <BrowserRouter>
+        <ContactsCard patient={patient} match={match}></ContactsCard>
+      </BrowserRouter>
+    );
+    expect(wrapper.queryByText(/Mobile/i)).not.toBeNull();
```
joeldenning: 👍  good tests

src/summary/profile/demographics-card.component.tsx
joeldenning: I slightly prefer subscribing to the patient in each component -- it doesn't cause a network request so it's totally fine from a perf perspective. A bit easier than plumbing the prop everywhere. Additionally, moving the subscription into the parent component means that whenever the patient re-renders, *all* of the patient chart components are re-rendered even if they don't need to. Whereas each one subscribing individually means that only the components that rely on the patient object re-render.

How about this - we create a hook called `useCurrentPatient` inside of esm-api and export it out for esm-patient-chart and other applications to use? Then in each component, you can just do the following:

```js
import { useCurrentPatient } from '@openmrs/esm-api';

function Foo() {
  const patient = useCurrentPatient();
  return <button>{patient.thing}</button>
}
```

Thoughts?

src/summary/profile/demographics-card.component.tsx
```diff
@@ -9,49 +8,46 @@ import dayjs from "dayjs";
...
           />
           <VerticalLabelValue
             label="Age"
-            value={currentPatient && age(currentPatient.birthDate)}
+            value={props.patient && age(props.patient.birthDate)}
           />
           <VerticalLabelValue
             label="Gender"
-            value={currentPatient && currentPatient.gender}
+            value={props.patient && props.patient.gender}
```
joeldenning: If we wanted to be really fancy, we could [upgrade to typescript 3.7 beta](https://devblogs.microsoft.com/typescript/announcing-typescript-3-7-beta/) and use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here.

```suggestion
            value={props.patient?.gender}
```

To do so, you'll also need to install [the babel optional chaining plugin](https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining), including adding it to the babelrc.

There is absolutely 0 need to do this in order for this PR to be approved, but thought I'd share because it's a very exciting and recent development for typescript in the last week.

src/summary/profile/identifiers-card.component.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const labelStyles = {
+    color: "var(--omrs-color-ink-medium-contrast)"
+  };
+
+  const valueStyles = {
+    color: "var(--omrs-color-ink-high-contrast)",
+    fontWeight: 500
+  };
+
+  function isPreferred(identifier) {
```
joeldenning: This function, as well as `preferredIdFirst`, could be pulled out of the IdentifiersCard function (maybe moved to the bottom of this whole file), since it's a pure function. There is a negligible, not important whatsoever, perf benefit to that. But I think that other advantage is just that it clarifies to maintainers that "this function doesn't rely on props or state."

src/summary/profile/identifiers-card.component.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  function preferredIdFirst(identifier1, identifier2) {
+    return isPreferred(identifier1) ? -1 : isPreferred(identifier2) ? 1 : 0;
+  }
+
+  return (
+    <SummaryCard name="Identifiers" match={props.match}>
+      {props.patient &&
+        props.patient.identifier
+          .sort((id_a, id_b) => preferredIdFirst(id_a, id_b))
```
joeldenning:
```suggestion
          .sort(preferredIdFirst)
```

^ same behavior

src/summary/profile/identifiers-card.component.tsx
```diff
@@ -0,0 +1,58 @@
...
+    color: "var(--omrs-color-ink-high-contrast)",
+    fontWeight: 500
+  };
+
+  function isPreferred(identifier) {
+    return identifier.use === "usual";
+  }
+
+  function preferredIdFirst(identifier1, identifier2) {
+    return isPreferred(identifier1) ? -1 : isPreferred(identifier2) ? 1 : 0;
```
joeldenning: Nested ternaries, especially without parentheses, are the inevitable fate of the react developer growing addicted to functional programming 😄 

At the very least, add parentheses to the inner ternary. However, also, I think that it is most readable just to avoid nested ternaries altogether, even though it's a few more lines of code:

```suggestion
  if (isPreferred(identifier1) {
    return -1
  } else if (isPreferred(identifier2)) {
    return 1
  } else {
    return 0
  }
```

src/summary/cards/horizontal-label-value.component.tsx
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import styles from "./horizontal-label-value.css";
+
+export default function HorizontalLabelValue(props) {
+  return (
+    <div className={styles.root} data-testid="horizontalLabelValue">
```
joeldenning: The test id doesn't appear to be used by anything. Also test ids are discouraged by react testing library except for exceptional cases.

```suggestion
    <div className={styles.root}>
```

src/summary/cards/horizontal-label-value.css
```diff
@@ -0,0 +1,10 @@
+sup {
```
joeldenning: CSS Modules are only local for the css selectors that are css classes. Your css here modifies *all `<sup>` elements that ever will exist in the app*. To fix, make sure you're only selecting `sup`'s within this component.

```suggestion
.root sup {
```

src/summary/profile/demographics-card.component.tsx
fatmali: I preferred subscribing to the patient in the parent component because of the following reasons:

1) the patient chart should, in fact, be re-rendered when a patient changes, I don't think there is a situation where we wouldn't want to re-render a patient chart component when loading a different patient, as all the components are dependent on the current patient. Having a single source of truth will make sure all components are displaying data for the same patient.

2) writing tests for these components are much easier, no need to keep mocking the `useCurrentPatient` function.

3) I feel like writing `useCurrentPatient` in each component is repetitive.

fatmali:
Hi @joeldenning ! Thanks for the review. I'll work on the suggested changes and push the changes.

src/summary/profile/identifiers-card.component.tsx
fatmali: 😄 thanks, I'll address this.

src/summary/profile/contacts-card.component.tsx
fatmali: This has not been addressed in the designs. I  think that using em-dash might be better to explicitly let a user know that there is no address/contact

src/summary/cards/vertical-label-value.component.tsx
```diff
@@ -16,11 +20,13 @@ export default function VerticalLabelValue(props: VerticalLabelValueProps) {
 }
 
 VerticalLabelValue.defaultProps = {
-  valueStyles: {}
+  valueStyles: {},
+  valueGlobalCssClasses: []
```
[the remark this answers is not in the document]
fatmali: I agree

src/summary/patient-chart-summary.css
fatmali: wow, that's interesting. I'll make the required changes

src/summary/cards/summary-card-row.css
```diff
@@ -1,4 +1,4 @@
-a.row:global(.omrs-unstyled) {
```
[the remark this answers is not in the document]
fatmali: Good point, no I didn't. I'll confirm.

src/summary/banner/patient-banner.component.tsx
```diff
@@ -0,0 +1,116 @@
...
+                    Preferred ID
+                  </span>
+                  <span
+                    className={`${styles.demographic} omrs-type-body-regular`}
+                  >
+                    {getPreferredIdentifier()}
+                  </span>
+                </div>
+              </div>
+              <div className={styles.moreBtn}>
```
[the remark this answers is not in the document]
fatmali: oops.

src/summary/profile/demographics-card.component.tsx
joeldenning: I disagree, but not enough to block approval/merging. Let's go forward with what you have.

src/summary/patient-chart-summary.css
joeldenning: The patient chart should not have a different color than the [canvas color](https://github.com/openmrs/openmrs-esm-styleguide/blob/master/src/canvas/canvas.css#L3) that is set throughout the app.

I'll talk to Greg about this to change it. It doesn't make sense to have different canvas colors for each page.

src/summary/profile/contacts-card.component.tsx
joeldenning:
```jsx
<div>
  <div>{address.street}</div>
  <div>{address.city}, {address.province}</div>
  <div>{address.country}</div>
</div>
```

joeldenning: New lines don't work in HTML unless your dom element is `display: inline`, which comes with its own challenges, so it's often easier just to break it up as separate divs that are `display: block`

src/summary/patient-chart-summary.css
gregoryjschmidt: Hi, agreed doesn't make sense to change canvas color through app.

I've been using the omrs-bg-lowest-contrast as the default canvas color in the mockups - for patient chart and search.

The omrs-color-background-canvas is the color from the hackathon designs, but was replaced when we designed the stylguide.

I'm looking @ storybook atm - and the following are the hackathon color styles no longer in use:

Primary blue
--omrs-color-primary-blue
Text black
--omrs-color-text-black
Inactive grey
--omrs-color-inactive-grey
Inactive grey two
--omrs-color-inactive-grey-two
Background canvas
--omrs-color-background-canvas
Critical alert red
--omrs-color-critical-alert-red

joeldenning: 👍 I have created https://github.com/openmrs/openmrs-esm-styleguide/pull/48 which removes those colors from the styleguide.

fatmali:
Hi @joeldenning, I updated this _painful_ PR with the changes you suggested, apologies you have to go through it again, will make future ones shorter :-)

PR #8, opened 2019-11-05
MF-91: Height and Weight Widget

Andu033:
Ticket: https://issues.openmrs.org/browse/MF-91


Andu033:
image

Andu033:
image

src/summary/cards/summary-card.component.tsx
```diff
@@ -3,8 +3,12 @@ import styles from "./summary-card.css";
 import { match } from "react-router";
 
 export default function SummaryCard(props: SummaryCardProps) {
+  const stylesCard: Styles = {
+    width: props.cardSize,
+    margin: "1rem 1rem 0 0"
+  };
   return (
-    <div className={`omrs-card ${styles.card}`} style={props.styles}>
```
fatmali: why not pass in the width and margin inside of `props.styles`?

src/summary/cards/summary-card.component.tsx
```diff
@@ -15,14 +19,14 @@ export default function SummaryCard(props: SummaryCardProps) {
...
-SummaryCard.defaultProps = {
-  styles: {}
-};
+SummaryCard.defaultProps = { cardSize: "21.75rem" };
 
 type SummaryCardProps = {
   name: string;
   match: match;
   children: React.ReactNode;
+  cardSize: String;
```
fatmali: this shouldn't be necessary, you can pass in the `width` as part of the `styles` prop

src/summary/documentation/date-helper.tsx
```diff
@@ -0,0 +1,6 @@
+var moment = require("moment");
```
fatmali: might be better to use `dayjs` instead as it is more lightweight than `moment`. Also `dayjs` is already used in the patient-chart to calculate age, so it's already included in the dependencies.

src/summary/documentation/dimensions-card.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+      for (let i = 0; i < ar.length; i += 3) {
+        res.push({
+          date: ageConvert(ar[i].effectiveDateTime),
+          cm: ar[i].valueQuantity.value,
+          kg: ar[i + 1].valueQuantity.value,
+          bmi: ar[i + 2].valueQuantity.value
+        });
+      }
+      setDimensions(res);
+    });
```
fatmali: You can now use the `createErrorHandler()` function from `@openmrs/esm-error-handling`

src/summary/documentation/dimensions-card.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+  React.useEffect(() => {
+    const abortController = new AbortController();
+    getPatientDimensions(
+      props.currentPatient.identifier[0].value,
+      abortController
+    ).then(response => {
+      const ar = response.data.entry
+        .filter(
+          el =>
+            el.resource.valueQuantity !== undefined &&
```
joeldenning: A javascript truthiness check would probably suffice here and maybe even be better than checking undefined explicitly:

```suggestion
            el.resource.valueQuantity &&
```

See https://developer.mozilla.org/en-US/docs/Glossary/Truthy

src/summary/documentation/dimensions-card-row.component.tsx
```diff
@@ -0,0 +1,35 @@
...
+        {props.content.cm}
+        {props.index === 0 && <p>cm</p>}
+      </div>
+      <div className={styles.text}>
+        {props.content.kg}
+        {props.index === 0 && <p>kg</p>}
+      </div>
+      <div className={styles.text}>
+        {props.content.bmi}
+        {props.index === 0 && <p>m/m</p>}
```
joeldenning: Should this be `kg/m`?? I'm looking at https://www.figma.com/file/YyVhWTNp89bNiOGadYrHBk/Patient-Chart---Master-(current-file)?node-id=1550%3A30065

```suggestion
        {props.index === 0 && <p>kg/m</p>}
```

src/summary/documentation/dimensions-card.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+          cm: ar[i].valueQuantity.value,
+          kg: ar[i + 1].valueQuantity.value,
+          bmi: ar[i + 2].valueQuantity.value
+        });
+      }
+      setDimensions(res);
+    });
+
+    return () => abortController.abort();
+  }, [props.currentPatient.identifier[0].value]);
```
joeldenning: Is there a way to search the observations api by patient id instead of identifier? If so that would be preferred.

src/summary/documentation/dimensions-card.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+        .filter(
+          el =>
+            el.resource.valueQuantity !== undefined &&
+            (el.resource.valueQuantity.unit === "cm" ||
+              el.resource.valueQuantity.unit === "kg" ||
+              el.resource.valueQuantity.unit === undefined)
+        )
+        .map(m => m.resource);
+      const res = [];
+      for (let i = 0; i < ar.length; i += 3) {
```
joeldenning: Why are we incrementing i by three every time? It seems like we're depending on the `data.entry` to be in a very specific order that may not hold true.

src/summary/documentation/dimensions-card.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+        res.push({
```
joeldenning: This is the perfect situation for using [Array.prototype.map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map). Let's use map instead of a for loop.

src/summary/documentation/dimensions-card.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const [dimensions, setDimensions] = React.useState([]);
+  const [showElements, setShowElements] = React.useState(3);
+
+  React.useEffect(() => {
+    const abortController = new AbortController();
+    getPatientDimensions(
+      props.currentPatient.identifier[0].value,
+      abortController
+    ).then(response => {
+      const ar = response.data.entry
```
joeldenning: Please name this variable with a more descriptive name.

src/summary/documentation/dimensions.resource.tsx
```diff
@@ -0,0 +1,11 @@
+import { openmrsFetch } from "@openmrs/esm-api";
+
+export function getPatientDimensions(
+  patientIdentifer: string,
+  abortController: AbortController
+) {
+  return openmrsFetch(
+    `/ws/fhir/Observation?patient.identifier=${patientIdentifer}`,
```
joeldenning: Let's try to retrieve this via `fhir.read()` instead of `openmrsFetch`. See https://wiki.openmrs.org/display/projects/openmrs-esm-api and https://github.com/FHIR/fhir.js

src/summary/documentation/dimensions-card.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+import HorizontalSectionCard from "./horizontal-section-card.component";
+import DimensionsSectionCard from "./dimensions-card-row.component";
+import { ageConvert } from "./date-helper";
+import ShowMoreCard from "./show-more-card.component";
+import style from "./dimensions-card.css";
+export default function DimensionsCard(props: DimensionsCardProps) {
+  const [dimensions, setDimensions] = React.useState([]);
+  const [showElements, setShowElements] = React.useState(3);
+
+  React.useEffect(() => {
```
joeldenning: This effect is rather complicated. I feel like we could simplify it substantially. Please do the following:

1. Paste an example response for the fhir observations endpoint so that I can see what the data looks like. I tried to find a patient with observations on openmrs-spa.org but wasn't able to.
2. Explain to me what you're trying to do with all the filters, maps, and pushes. Are you just trying to group the observations into Weight, Height, and BMI arrays?

src/summary/documentation/dimensions-card.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+    });
+
+    return () => abortController.abort();
+  }, [props.currentPatient.identifier[0].value]);
+  return (
+    <SummaryCard name="Height & Weight" match={props.match} cardSize="40rem">
+      <SummaryCardRow linkTo="/">
+        <SummaryCardRowContent>
+          <HorizontalSectionCard>
+            <div />
```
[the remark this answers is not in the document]
Andu033: If i don't put that div the grid will start with weight div and the columns will be aligned incorrectly

src/summary/documentation/dimensions-card.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+              <SummaryCardRowContent>
+                <DimensionsSectionCard content={el} index={index} />
+              </SummaryCardRowContent>
+            </SummaryCardRow>
+          );
+        })}
+      <SummaryCardRowContent>
+        <ShowMoreCard
+          func={() => {
+            setShowElements(dimensions.length + 3 - showElements);
```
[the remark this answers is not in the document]
Andu033: I used that to set the number of entries that are shown and to go back to 3 entries if you click on less. If i change that you will be unable to minimize back. I don't know what is the expected behaviour so I expected that more button become less button after extension.

src/summary/documentation/dimensions-card.component.tsx
```diff
@@ -0,0 +1,82 @@
...
+import VerticalLabelValue from "../cards/vertical-label-value.component";
+import SummaryCardRowContent from "../cards/summary-card-row-content.component";
+import HorizontalSectionCard from "./horizontal-section-card.component";
+import DimensionsSectionCard from "./dimensions-card-row.component";
+import { ageConvert } from "./date-helper";
+import ShowMoreCard from "./show-more-card.component";
+import style from "./dimensions-card.css";
+export default function DimensionsCard(props: DimensionsCardProps) {
+  const [dimensions, setDimensions] = React.useState([]);
+  const [showElements, setShowElements] = React.useState(3);
```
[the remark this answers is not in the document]
Andu033: Ok. I did.

src/summary/documentation/dimensions-card.component.tsx
Andu033: https://pastebin.com/PAzHTKkz
The fhir endpoint that return entries by code is not ready yet so I use the one that return all entries, I filter the ones that I want to use and then i group entries by date (are sorted by date as far as i know).

src/summary/documentation/dimensions-card.component.tsx
Andu033: I don't understand how to use it in this situation because i have to map 3 elements in one. Never did that with map.

src/summary/documentation/dimensions-card.component.tsx
Andu033: They are not ordered by date? This was an observation of mine and i saw that it works i don't know if it was a coincidence.

src/summary/documentation/dimensions-card.component.tsx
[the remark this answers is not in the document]
Andu033: I did a change for that.

src/summary/documentation/dimensions-card-row.css
```diff
@@ -0,0 +1,13 @@
+.gridContainer {
+  display: grid;
+  grid-template-columns: 16rem 8rem 8rem 8rem;
```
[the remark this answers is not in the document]
Andu033: I don't know. I ve tried some sizes until it fitted the design from figma.

src/summary/documentation/date-helper.tsx
Andu033: Thanks. I replaced that with timeago and works well but i could get only a date like : 2 years ago but in figma we have something like 04-Jan 2018. I'm doing something wrong?

src/summary/documentation/dimensions-card.component.tsx
Andu033: Ok. I added.

Andu033:
> Phewf, this is a lot of comments. One thing I'd like to say is that me being thorough on the review does not mean that you did a poor job on this - do not get discouraged smile. Over time as you submit more PRs and level up your skills, the PRs will go more smoothly. You did good work in this PR, and we'll iterate on it a couple times and get it merged.
> # An important rule for this PR
> 
>     * **Under no circumstances should you add any more features to this.** That's right, zero more features. The reason is that this PR is already big and hard for me to review. I don't want it to be a moving target where I have to rediscover what new features you've added to it.
> 
>     * You can work on new features in a separate branch that is based off of the `HeightAndWeight` branch. Keep all new work on that separate branch. Wait for this branch to merge into master before pushing that branch. Once this PR is merged, then you should pull master and merge that into your new branch. Then you can create another PR.
> 
> 
> In the future, I'd recommend submitting **far smaller PRs** to start out. I don't like having to give 43 comments in one review -- this one took me 45 minutes just to go through it. If you do smaller PRs, we can get into a rhythm where you understand the expectations for the code and it's not as time consuming to review it. Small PRs does not mean you can't work on the features ahead of what you've pushed for your current PR. Follow my instructions above ^ that explain how to use git branches to continue working on things while you're waiting for code review.

I commit the changes on this pr. Changes doesn't mean new features right? Or i should create a new branch and commit on that and do a new pr? I'm asking if it was a mistake to commit the requested modifications on this branch.

src/summary/documentation/dimensions.resource.tsx
Andu033: We have an example of use in our project?

src/summary/documentation/dimensions-card-row.css
```diff
@@ -0,0 +1,13 @@
...
+  padding: 10px;
+}
+.text {
+  display: inline;
```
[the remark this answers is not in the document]
Andu033: This was useless. I removed that.

Andu033:
image

Andu033:
This is how it looks like now

src/summary/documentation/date-helper.tsx
fatmali: @Andu033 You can have a check, if date is not today then use normal date formatting, if date is today then use time ago.

Andu033: @fatmali ok. I will try that. Thanks Fatma.

src/summary/documentation/dimensions-card.component.tsx
joeldenning: The algorithm here just seems quite arbitrary. There is no clear reason for `i+=3` to the reader of this code.

src/summary/documentation/dimensions.resource.tsx
joeldenning: No - here are the docs: https://github.com/FHIR/fhir.js

src/summary/documentation/dimensions-card.component.tsx
joeldenning: Make this effect simpler. The code here is too complicated and unclear

joeldenning:
I just commented on some of the existing comments - please address the unresolved comments.

src/summary/documentation/dimensions-card-row.css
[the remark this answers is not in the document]
joeldenning: We need a better answer than that in order to merge this. The css should work for tablet, mobile, and desktop

src/summary/documentation/dimensions-card.component.tsx
joeldenning: You need to explain what you're doing in a way that I can understand. Also your code should be more clear about what it's accomplishing and why.

joeldenning:
Here are some comments that still need to be addressed:

https://github.com/openmrs/openmrs-esm-patient-chart/pull/8#discussion_r343473628
https://github.com/openmrs/openmrs-esm-patient-chart/pull/8/files#diff-111c505a98e0661a15ce794c904de042

src/summary/documentation/dimensions-card-row.css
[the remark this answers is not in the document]
Andu033: I modified the way i'm doing the widget. I'm using now a table and flex combined with maxWidth t achieve responsive size. But now i have to wait for Fatma's pr to be merged. We are using same resource.

package.json
```diff
@@ -75,7 +75,8 @@
     "@openmrs/esm-api": "^1.0.8",
     "@types/react-router": "^5.0.3",
     "react": "^16.9.0",
-    "react-dom": "^16.9.0"
+    "react-dom": "^16.9.0",
+    "timeago.js": "^4.0.1"
```
fatmali: I don't see the point of this library anymore, to format the dates, you can use this helper here https://github.com/openmrs/openmrs-esm-patient-chart/blob/75a0defb19477cda383011a32a64ac0e1a8e87ec/src/summary/documentation/dimension-helpers.ts#L10

src/summary/cards/horizontal-section-card.component.tsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
```
fatmali: given that you are now using a `<table>` to display data, do we need this class?

src/summary/cards/horizontal-section-card.css
```diff
@@ -0,0 +1,9 @@
+.container {
```
fatmali: same applies for this, see comment above. is it needed?

src/summary/documentation/dimensions-card-level-one.css
```diff
@@ -0,0 +1,43 @@
+.table {
```
fatmali: @joeldenning these css classes have been repeated 3 times in different widgets to render data in a responsive table, do you think we should abstract this now?

src/summary/documentation/dimensions-card-row.component.tsx
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+import styles from "./dimensions-card-row.css";
+
+export default function DimensionsSectionCard(
+  props: DimensionsSectionCardProps
+) {
+  return (
+    <div className={styles.gridContainer}>
```
fatmali: @Andu033 what does this component do? isn't the table enough?

package.json
Andu033: Good point. I will delete that.

src/summary/cards/horizontal-section-card.component.tsx
Andu033: You are right. We don't need

src/summary/cards/horizontal-section-card.css
Andu033: You are right. We don't need

src/summary/documentation/dimensions-card-row.component.tsx
Andu033: Deleted that too

joeldenning:
Thanks fatma for helping drive this one forward

PR #9, opened 2019-11-08
MF-93: Notes UI

madvincy:


package.json
```diff
@@ -25,7 +25,7 @@
   ],
   "husky": {
     "hooks": {
-      "pre-commit": "pretty-quick --staged && concurrently -n eslint,test,typescript 'npm run lint' 'npm test' 'npm run typescript'"
+      "pre-commit": "pretty-quick --staged && npm run lint && npm test && npm run typescript"
```
fatmali: we need these to run concurrently, is there any reason why you removed it?

src/summary/cards/horizontal-label-value.component.tsx
```diff
@@ -11,13 +11,23 @@ export default function HorizontalLabelValue(props: HorizontalLabelValueProps) {
...
+        </div>
+      ) : null}
+      {props.detail ? (
+        <div
+          className={props.detailsClassName || "omrs-type-body-regular"}
+          style={props.detailsStyles}
+        >
+          {props.detail || "\u2014"}
+        </div>
+      ) : null}
```
fatmali: You can do this instead of using the ternary in this situation
```suggestion
      {props.detail && 
        <div
          className={props.detailsClassName || "omrs-type-body-regular"}
          style={props.detailsStyles}
        >
          {props.detail || "\u2014"}
        </div>
```

src/summary/cards/vertical-label-value.component.tsx
```diff
@@ -4,11 +4,15 @@ import styles from "./vertical-label-value.css";
...
+      <label
+        className={
+          props.labelClassName || `omrs-type-body-small ${styles.label}`
+        }
+      >
         {props.label}
       </label>
       <div
-        className={props.className}
+        className={props.valueClassName || props.className}
```
fatmali: I think we should choose one between `className` and `valueClassName`, having both might be confusing.

src/summary/history/allergy-card.component.tsx
```diff
@@ -37,6 +37,7 @@ export default function AllergyCard(props: AllergyCardProps) {
               <HorizontalLabelValue
                 label={allergy.resource.code.text}
                 labelClassName="omrs-bold"
+                detail={null}
```
fatmali: I don't like how this looks. Definitely a no. Make `detail` an optional param.

src/summary/history/conditions-card.component.tsx
```diff
@@ -32,6 +32,7 @@ export default function ConditionsCard(props: ConditionsCardProps) {
             label="Active Conditions"
             labelStyles={{ color: "var(--omrs-color-ink-medium-contrast)" }}
             value="Since"
+            detail={null}
```
fatmali: Same here

src/summary/history/encounter.resource.tsx
```diff
@@ -0,0 +1,235 @@
...
+  patientIdentifer: string,
+  abortController: AbortController
+) {
+  //   return openmrsFetch(
+  //     `/ws/fhir/Encounter?patient.identifier=${patientIdentifer}`,
+  //     { signal: abortController.signal }
+  //   );
+  return Promise.resolve(mockPatientEncounterSearchResponse);
+}
+const mockPatientEncounterSearchResponse = {
```
fatmali: perhaps you can move this to the __mock__ tests folder and can reuse it later for tests as well?

src/summary/history/notes-card.component.tsx
```diff
@@ -0,0 +1,78 @@
...
+export default function NotesCard(props: NotesCardProps) {
+  const [patientNotes, setPatientNotes] = React.useState(null);
+
+  React.useEffect(() => {
+    const abortController = new AbortController();
+
+    encounterResource(props.currentPatient.identifier[0].value, abortController)
+      .then(notes => setPatientNotes(notes))
+      .catch(error => {
+        setTimeout(() => {
```
fatmali: checkout how to use `createErrorHandler()` from `@openmrs/esm-error-handling`
```suggestion
        createErrorHandler();
```

package.json
joeldenning: I think that the syntax we're using with concurrently might not work on windows, but this syntax definitely does work on windows:

```suggestion
      "pre-commit": "pretty-quick --staged && concurrently 'npm:lint' 'npm:test' 'npm:typescript'"
```

src/summary/cards/vertical-label-value.component.tsx
joeldenning: Agreed 👍

package.json
joeldenning: I think the concurrently command as it currently is might have troubles working on windows. To fix, you can do this:

```suggestion
      "pre-commit": "pretty-quick --staged && concurrently npm:test npm:lint npm:typescript"
```

src/summary/history/notes-card.component.tsx
```diff
@@ -0,0 +1,78 @@
...
+      });
+
+    return () => abortController.abort();
+  }, [props.currentPatient.identifier[0].value]);
+
+  return (
+    <SummaryCard name="Notes" match={props.match} styles={{ width: "45.5rem" }}>
+      <SummaryCardRow>
+        <SummaryCardRowContent justifyContent="space-between">
+          <HorizontalLabelValue
```
joeldenning: In my opinion, we shouldn't use the HorizontalLabelValue component for the Notes header. The HorizontalLabelValue component is made for a situation where there are two values, but you need three. Additionally, "Date" is not a label for "Encounter type".

You should just write your own jsx for the header.

src/summary/history/notes-card.component.tsx
```diff
@@ -0,0 +1,78 @@
...
+                detail={null}
+              />
+              <VerticalLabelValue
+                label={note.resource.type[0].coding[0].display}
+                labelClassName="omrs-bold"
+                value={note.resource.location[0].location.display}
+                valueStyles={{ color: "var(--omrs-color-ink-medium-contrast)" }}
+              />
+              <HorizontalLabelValue
+                label={"Dr Mwangi, katherine"}
```
joeldenning: Why is this hardcoded?

src/summary/history/notes-card.component.tsx
madvincy: I have an idea in mind, what of having both label and value optional so as to generate 3 or multiple values

joeldenning: No. The HorizontalLabelValue component doesn't make sense here (it is not a header). We should change this to not use HorizontalLabelValue for the Notes header.

src/summary/history/encounter.resource.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+export function encounterResource(
+  patientIdentifer: string,
+  abortController: AbortController
+) {
+  return openmrsFetch(
+    `/ws/fhir/Encounter?patient.identifier=${patientIdentifer}`,
+    { signal: abortController.signal }
+  );
+  // return Promise.resolve(mockPatientEncounterSearchResponse);
```
joeldenning: Will merging this in break things? Does the fhir endpoint for encounters currently work?

We should always merge code in that works. If the fhir endpoint isn't ready, we should use the mocked data for now. We should not be merging in commented code.

src/summary/history/notes-card.component.tsx
```diff
@@ -0,0 +1,110 @@
...
+  React.useEffect(() => {
+    const abortController = new AbortController();
+
+    encounterResource(props.currentPatient.identifier[0].value, abortController)
+      .then(({ data }) => {
+        if (data.total > 0) {
+          setPatientNotes(getNotes(data.entry));
+        }
+      })
+      .catch(error => {
```
joeldenning: This is incorrect usage of createErrorHandler:

```suggestion
      .catch(createErrorHandler())
```

src/summary/history/notes-card.component.tsx
```diff
@@ -0,0 +1,110 @@
...
+      </SummaryCardFooter>
+    </SummaryCard>
+  );
+}
+function getNotes(notes) {
+  return notes.map(note => note.resource);
+}
+function getAuthorName(extension: any): string {
+  let authorName = "";
+  extension.filter(function(author) {
```
joeldenning: Array.prototype.filter is a [pure function](https://en.wikipedia.org/wiki/Pure_function). Your usage of filter here is impure, in that it modifies a variable (authorName) outside of the filter function. This is considered poor practice in javascript - when using pure functions like map, filter, and reduce, you should not have side effects.

Instead, here's how you can implement getAuthorName:

```js
const author = extension.find(ext => ext.url === "changedBy")
return author ? author.valueString.toUpperCase() : ""
```

src/summary/history/notes-card.component.tsx
```diff
@@ -0,0 +1,110 @@
...
+    if (author.url === "changedBy") {
+      authorName = author.valueString;
+    }
+  });
+  return authorName.toUpperCase();
+}
+function convertdate(date: string): string {
```
joeldenning: Javascript functions should be camel cased:

```suggestion
function convertDate(date: string): string {
```

src/summary/history/notes-card.component.tsx
```diff
@@ -0,0 +1,110 @@
...
+    return "Today   ".concat(unprocesseddate.format("h:mm a"));
+  } else if (unprocesseddate.format("YYYY") === dayjs().format("YYYY")) {
+    return unprocesseddate.format("DD-MMM h:mm a");
+  }
+  return unprocesseddate.format("DD-MMM-YYYY h:mm a");
+}
+
+type NotesCardProps = {
+  match: match;
+  currentPatient: any;
```
joeldenning: You can use `fhir.Patient` as a type (it comes from the `@types/fhir` package). See example [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/65b793aa7b30ab0a7b4bff620c4d1ce58f38c9df/src/summary/profile/contacts-card.component.tsx#L61).

This will strictly type the patient object and give you autocomplete for using properties on the patient.

src/summary/history/notes-card.component.tsx
```diff
@@ -0,0 +1,110 @@
...
+      authorName = author.valueString;
+    }
+  });
+  return authorName.toUpperCase();
+}
+function convertdate(date: string): string {
+  const unprocesseddate = dayjs(date);
+
+  if (unprocesseddate.format("DD-MMM-YYYY") === dayjs().format("DD-MMM-YYYY")) {
+    return "Today   ".concat(unprocesseddate.format("h:mm a"));
```
joeldenning: I see in the designs that they have columns within the Date field lined up vertically. Putting three space characters after `Today` will not ensure that.

If we want to ensure columns, we should separate the columns into two separate divs and ensure the same width for each of the columns. That might be more work than it's worth, so we could ask JJ/Greg how important it is to do that. To do that, we'd need to do something like this:

```jsx
return <div>Today</div><div>{unprocessedDate.format("h:mm a")}</div>
```

src/summary/history/notes-card.component.tsx
```diff
@@ -0,0 +1,110 @@
...
+  let authorName = "";
+  extension.filter(function(author) {
+    if (author.url === "changedBy") {
+      authorName = author.valueString;
+    }
+  });
+  return authorName.toUpperCase();
+}
+function convertdate(date: string): string {
+  const unprocesseddate = dayjs(date);
```
joeldenning: Javascript variables should be camel cased: `unprocessedDate`

src/summary/cards/horizontal-label-value.component.tsx
```diff
@@ -4,20 +4,24 @@ import styles from "./horizontal-label-value.css";
...
-        {props.specialKey && <sup>{"\u002A"}</sup>}
-      </label>
-      <div
-        title={props.label}
-        className={props.valueClassName || "omrs-type-body-regular"}
-        style={props.valueStyles}
-      >
-        {props.value || "\u2014"}
-      </div>
+      {props.label && (
```
joeldenning: Since this PR is no longer using HorizontalLabelValue, it should not modify it.

PR #10, opened 2019-11-08
Relationships widget

fatmali:


src/summary/profile/relationships-card.component.tsx
```diff
@@ -1,15 +1,55 @@
...
+import VerticalLabelValue from "../cards/vertical-label-value.component";
 
 export default function RelationshipsCard(props: RelationshipsCardProps) {
+  const [relationships, setRelationships] = React.useState(null);
+  React.useEffect(() => {
+    openmrsFetch(
+      `/ws/fhir/RelatedPerson?patient.identifier=${props.patient.identifier[0].value}`
+    )
+      .then(({ data }) => {
+        if (data.total > 0) {
```
joeldenning: Is this if statement needed? I think the more common thing would be to do `relationships.length > 0` on line 26, so that the emdash would still be shown if the length is zero.

src/summary/profile/relationships.card.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+  it("displays em-dash if there are no relationships", () => {
+    const wrapper = render(
+      <BrowserRouter>
+        <RelationshipsCard patient={patient} match={match}></RelationshipsCard>
+      </BrowserRouter>
+    );
+    expect(wrapper.getByText("\u2014")).not.toBeNull();
+  });
+
+  it("renders relationships correctly", async () => {
```
joeldenning: 👍 great test

src/summary/profile/relationships-card.component.tsx
```diff
@@ -1,15 +1,55 @@
...
+import { createErrorHandler } from "@openmrs/esm-error-handling";
+import SummaryCard from "../cards/summary-card.component";
+import SummaryCardRow from "../cards/summary-card-row.component";
+import SummaryCardRowContent from "../cards/summary-card-row-content.component";
+import VerticalLabelValue from "../cards/vertical-label-value.component";
 
 export default function RelationshipsCard(props: RelationshipsCardProps) {
+  const [relationships, setRelationships] = React.useState(null);
+  React.useEffect(() => {
+    openmrsFetch(
```
joeldenning: We've usually been putting these calls into a resource file for a couple reasons:

1. Easier mocking of the request

```js
jest.mock('./related-person.resource.tsx', () => ({
  getRelationships: jest.fn()
})
```

2. Reuse of any transformation logic that needs to be applied to the request/response objects when interacting with the RelatedPerson resource.

What are your thoughts on that? It looks like you've already solved (1) very easily just by mocking openmrsFetch itself. And for (2) we aren't really doing any transformation of the response object. So maybe okay not creating the resource files?

fatmali: hi @joeldenning , good points. I updated the PR and separated the call into a separate file to stick to the convention that we currently have.

src/summary/profile/relationships-card.component.tsx
fatmali: it's needed because if `total === 0` then there is no `entry` property.

PR #18, opened 2019-11-13
MF-96 Notes-level-2 User-interface and Reusable-Pagination

madvincy:
notes level 2
pagination end
pagination start
phone view
tabletview

### DIFF
--- a/src/summary/history/notes-card-style.css
+++ b/src/summary/history/notes-card-style.css
@@ -1,18 +1,32 @@
-.table {
-  width: 100%;
+.moreNotes {
+  display: flex;
+  align-items: center;
+  border: none;
+  padding: 0.625rem 1.625rem 0.625rem 1rem;
+  background-color: var(--omrs-color-bg-medium-contrast);
+  color: var(--omrs-color-ink-medium-contrast);
+  cursor: pointer;
+  text-decoration: none;
 }
 
-.tableRow {
+.tableNotesRow {
   border-bottom: 1px solid var(--omrs-color-bg-low-contrast);
 }
-
-.tableHeader,
-.tableData,
-.tableDate {
-  padding: 0.625rem;
+.svgUpper {
+  display: none;
 }
 
 .tableHeader {
+  padding: 0.625rem;
+}
+:global(.omrs-breakpoint-gt-phone) .card {
+  width: 21.75rem;
+  margin: 0.5rem 1rem;
+}
+.tableNotes {
+  width: 100%;
+}
+.tableNotesHeader {
   color: var(--omrs-color-ink-medium-contrast);
   font-size: 0.9375rem;
   text-transform: capitalize;
@@ -21,21 +35,53 @@
   font-family: "Work Sans";
 }
 
-.tableData {
+.tableNotesData,
+.tableNotesAuthor {
   font-size: 1.0625rem;
   text-align: center;
+  width: 33.3%;
 }
 
-:global(.omrs-breakpoint-lt-tablet) .tableDates {
-  width: 70%;
-}
-
-table.table .tableDate {
+.tableNotesDate {
   font-family: "Work Sans";
   font-weight: 500;
+  width: 33.3%;
+  padding: 0.625rem 1.625rem 0.625rem 1rem;
 }
 
 .location {
   color: var(--omrs-color-ink-medium-contrast);
+  font-family: "Roboto";
+}
+:global(.omrs-breakpoint-gt-phone) .tableNotesRow {
+  display: table-row;
+}
+
+:global(.omrs-breakpoint-lt-tablet) .tableNotesData,
+.tableNotesAuthor,
+.tableNotesRow {
+  display: block;
+}
+:global(.omrs-breakpoint-lt-tablet) .tableNotesAuthor {
+  color: var(--omrs-color-ink-medium-contrast);
+  text-align: left;
+  font-family: "Roboto";
+}
+:global(.omrs-breakpoint-lt-tablet) .tableNotesData {
+  text-align: left;
+  font-family: "Roboto";
+  font-weight: 500;
+  font-size: 1.0625rem;
+}
+:global(.omrs-breakpoint-lt-tablet) .tableNotesDate {
   font-family: "Work Sans";
+  text-align: left;
+  padding: 0;
+}
+:global(.omrs-breakpoint-lt-tablet) .tableNotes {
+  width: 100%;
+}
+:global(.omrs-breakpoint-lt-tablet) .tableNotesHeader,
+.tdLowerSvg {
+  display: none;
 }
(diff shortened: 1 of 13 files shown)


src/summary/history/patient-notes/notes-card-style.css
```diff
@@ -0,0 +1,140 @@
...
+table.table {
+  font-family: "Work Sans";
+  font-weight: 500;
+}
+
+.location {
+  color: var(--omrs-color-ink-medium-contrast);
+  font-family: "Work Sans";
+}
+@media only screen and (max-width: 760px),
```
joeldenning: We have reusable media queries in the styleguide. See https://styleguide.openmrs.org/?path=/story/openmrs-styleguide--responsive-breakpoints. The documentation there is sorta broken. So check out [the source code](https://github.com/openmrs/openmrs-esm-styleguide/blob/366426341bd8be5c0f44a920a4ea98c4f8296563/src/responsive-breakpoints/responsive-breakpoints.js) and [example](https://github.com/openmrs/openmrs-esm-styleguide/blob/master/src/responsive-breakpoints/responsive-breakpoints.stories.html#L29)

760px is not a responsive breakpoint that we support in the styleguide. Is there a reason why you chose it?

src/summary/history/patient-notes/notes-card-style.css
```diff
@@ -0,0 +1,140 @@
...
+    align-items: center;
+    border: none;
+    padding: 0.625rem 1.625rem 0.625rem 1rem;
+    background-color: white;
+    color: var(--omrs-color-ink-medium-contrast);
+    cursor: pointer;
+    text-decoration: none;
+  }
+
+  /* Hide table headers (but not display: none;, for accessibility) */
```
joeldenning: Can you explain why `display: none;` would be bad for accessibility here? Do we want it to be read to screen reader users, but not shown to visual users?

src/summary/level-two-routes.component.tsx
```diff
@@ -38,6 +39,16 @@ export function LevelTwoRoutes(props: LevelTwoRoutesProps) {
               />
             )}
           />
+          <Route
+            exact
+            path={`/patient/:patientUuid/chart/notes`}
+            render={routeProps => (
+              <PatientNotes
+                match={props.match}
```
joeldenning: `props.match` is the coming from the LevelTwoRoutes props, not from the actual route for PatientNotes. This is a problem. Follow [this pattern from the react-router docs](https://reacttraining.com/react-router/web/api/Route/render-func) instead

```suggestion
                {...routeProps}
```

src/summary/level-two-routes.component.tsx
```diff
@@ -3,6 +3,7 @@ import { match, Switch, Route } from "react-router";
 import { AllergyCardLevelTwo } from "./history/allergy-card-level-two.component";
 import { getCurrentPatient } from "@openmrs/esm-api";
 import DimensionsCardLevelTwo from "./documentation/dimensions-card-level-two.component";
+import PatientNotes from "./history/patient-notes/patient-notes-card-level-2.component";
```
joeldenning: It is easier to read React code that names imported variables the same way as the they are named inside of the component

```suggestion
import PatientNotesCardLevel2 from "./history/patient-notes/patient-notes-card-level-2.component";
```

src/summary/history/patient-notes/patient-notes-card-level-2.component.tsx
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import { match } from "react-router";
+import NotesCard from "./notes-card.component";
+import Styles from "./patient-notes-style.css";
+import { getEncounters } from "./encounter.resource";
+
+export default function PatientNotes(props: PatientNotesCardProps) {
```
joeldenning: Components should be named the same as the their filename. This makes it easy to find where components are defined:

```suggestion
export default function PatientNotesCardLevel2(props: PatientNotesCardProps) {
```

src/summary/history/patient-notes/patient-notes-card-level-2.component.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const properties = {
+    showMore: true,
+    width: "100.0rem",
+    Headers: {
+      date: "DATE",
```
joeldenning: It's slightly nicer to use `text-transform: uppercase;` rather than capitalize the words in text. The reason is that it doesn't require the translator to know that it should be uppercase, and also because screen readers read all caps words as acronyms, instead of as words. Which is quite annoying. Would be good to apply some CSS

src/summary/history/patient-notes/notes-card.component.tsx
```diff
@@ -0,0 +1,174 @@
...
+
+import dayjs from "dayjs";
+import { createErrorHandler } from "@openmrs/esm-error-handling";
+
+import { getEncounters } from "./encounter.resource";
+import SummaryCard from "../../cards/summary-card.component";
+import style from "./notes-card-style.css";
+import { Link } from "react-router-dom";
+
+export default function NotesCard(props: NotesCardProps) {
```
joeldenning: I do not like the idea of sharing the same component for LevelOne and LevelTwo routes. They are too different. For example, LevelTwo allows for sorting and pagination whereas LevelOne does not. Other differences include that the headers are capitalized in LevelTwo but not in LevelOne, and also that there is a Show More button in LevelOne and not in LevelTwo.

In my opinion, combining the two makes things significantly harder. Instead, we could create smaller components for reuse. Something like this:

```js
<PatientChartTable>
  <PatientChartHeader>
    <tr>Date</tr>
    <tr>Encounter Type, </tr>
  </PatientChartHeader>
  <PatientChartTableRow>
  </PatientChartTableRow>
</PatientChartTable>
```

You might need all of these components. But my point is that we're making abstractions in the wrong place right now and what you've built won't benefit other LevelTwoRoutes that are similar, such as Height + Weight or Vitals.

The component here has

src/summary/history/notes-card-style.css
```diff
@@ -30,7 +65,19 @@
   width: 70%;
 }
 
-table.table .tableDate {
+:global(.omrs-breakpoint-gt-phone) .card {
```
fatmali: Rather than setting fixed widths on bigger screens, I usually prefer to use flex. Flex offers the ability to have responsive behavior across different screen sizes.

src/summary/documentation/dimension-helpers.ts
```diff
@@ -22,3 +22,16 @@ export function formatDate(strDate: string) {
     return date.format("YYYY DD-MMM");
   }
 }
+export function formatNotesDate(date: string): string {
```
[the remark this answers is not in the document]
madvincy: It is slightly different from the others

madvincy:
Fixes made

PR #22, opened 2019-11-14
Fixed patient banner display

fatmali:


HerbertYiga:
Could this be having a ticket Id? And some description of what you did?

src/summary/level-two-routes.component.tsx
```diff
@@ -15,7 +15,7 @@ export function LevelTwoRoutes(props: LevelTwoRoutesProps) {
   });
 
   return (
-    <main className="omrs-main-content">
+    <>
```
joeldenning: Minor thing, but this fragment can be removed:

```jsx
return currentPatient && (
  <Switch>
  </Switch>
)
```

Other than that, though, it looks good.

src/root.component.tsx
```diff
@@ -26,15 +26,21 @@ function Root(props) {
...
-      />
+      <main className="omrs-main-content">
+        {showPatientSummary && (
+          <Route
+            path="/patient/:patientUuid/chart"
+            exact
+            component={PatientChartSummary}
+          />
+        )}
+        {showPatientSummary && (
```
joeldenning: I was under the impression that the level two routes are there own pages that don't require clicking on the banner first in order to see them. I do not think we should only show the level two routes if you've clicked on the banner.

src/root.component.tsx
```diff
@@ -26,15 +26,21 @@ function Root(props) {
...
-        component={PatientChartSummary}
-      />
-      <Route
-        path="/patient/:patientUuid/chart"
-        render={routeProps => <LevelTwoRoutes match={props.match} />}
-      />
+      <main className="omrs-main-content">
+        {showPatientSummary && (
+          <Route
+            path="/patient/:patientUuid/chart"
```
joeldenning: Conditionally rendering `<Routes>` is an antipattern - your routes are how your user gets to where they want to go. You do not want them to have to click on something in the UI in order for them to be able to use a route.

If the user is **exactly** on `/patient/:patientUuid/chart`, they should see the patient summary page. Additionally, the patient banner should show if the user clicks on it. Here's what I think the code should be here:

```jsx
    <BrowserRouter basename={window["getOpenmrsSpaBase"]()}>
      <Route path="/patient/:patientUuid/chart" render={routeProps => (
        <PatientBanner
          {...routeProps}
          patient={currentPatient}
          showPatientSummary={setShowPatientSummary}
         />
        )} />
      </Route>
      {/* this popover is not a route, since it's not represented by the URL. Instead, it's "transient state" that disappears when you reload the page */}
      {showPatientSummary && (
        <div className="patient-banner-popover">
          <PatientChartSummary patient={currentPatient} />
        </div>
      )}
      <main className="omrs-main-content">
        <Route
          path="/patient/:patientUuid/chart"
          exact
          component={PatientChartSummary}
        />
        <Route
          path="/patient/:patientUuid/chart"
          component={LevelTwoRoutes}
        />
      </main>
    </BrowserRouter>
```

src/root.component.tsx
```diff
@@ -26,15 +26,21 @@ function Root(props) {
...
+            exact
+            component={PatientChartSummary}
+          />
+        )}
+        {showPatientSummary && (
+          <Route
+            path="/patient/:patientUuid/chart"
+            render={routeProps => <LevelTwoRoutes match={props.match} />}
```
joeldenning:
```suggestion
            component={LevelTwoRoutes}
```

Or (equivalently):

```suggestion
            render={routeProps => <LevelTwoRoutes {...routeProps} />}
```

src/root.component.tsx
fatmali: Hi @joeldenning,the reason why we are conditionally displaying the route is when you click on the open patient banner button, the profile shows and everything else hides no matter the route. I thought this would be hacky, but I cant think of a more elegant solution for this situation.

joeldenning: That should be accomplished through z-index and modal. Not through hiding and showing the routes.

src/root.component.tsx
fatmali: Level two routes are their own pages and visible by default. Clicking on the banner hides them in order to display the profile.

src/root.component.tsx
fatmali: Hmm I like the idea of using a spread operator, but that means we will have to change the type of LevelTwoRoutesProps to the type of routeProps, which is okay I guess.

PR #26, opened 2019-11-19
fixed patient banner to have space after name

jonathandick:
Added space between , and given name on patient banner.


brandones:
LGTM

PR #27, opened 2019-11-21
Upgrading dependencies

joeldenning:


package.json
```diff
@@ -31,52 +31,55 @@
...
-    "eslint": "^6.2.1",
-    "eslint-config-prettier": "^6.1.0",
-    "eslint-config-ts-react-important-stuff": "^1.0.0",
-    "eslint-plugin-prettier": "^3.1.0",
-    "fork-ts-checker-webpack-plugin": "^1.5.0",
-    "husky": "^3.0.4",
+    "dayjs": "^1.8.17",
+    "eslint": "^6.6.0",
+    "eslint-config-prettier": "^6.7.0",
+    "eslint-config-ts-react-important-stuff": "^1.1.0",
```
joeldenning: This turns on the "Rules of hooks" lint errors and warnings. See https://github.com/joeldenning/eslint-config-react-important-stuff/releases/tag/v1.2.0

package.json
```diff
@@ -31,52 +31,55 @@
...
+    "eslint-plugin-prettier": "^3.1.1",
+    "fork-ts-checker-webpack-plugin": "^3.1.0",
```
joeldenning: See https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/releases - the breaking changes don't affect us.

package.json
```diff
@@ -31,52 +31,55 @@
...
+    "webpack-dev-server": "^3.9.0"
   },
   "dependencies": {
-    "@openmrs/esm-api": "^1.0.8",
-    "@types/react-router": "^5.0.3",
-    "react": "^16.9.0",
-    "react-dom": "^16.9.0"
+    "@openmrs/esm-api": "^1.1.0",
+    "@types/react-router": "^5.1.3",
+    "i18next": "^19.0.1",
```
joeldenning: the new version of openmrs-react-root-decorator has a peer dependency on i18next and react-i18next. See https://github.com/openmrs/openmrs-react-root-decorator/releases

src/openmrs-esm-patient-chart.tsx
```diff
@@ -14,3 +14,9 @@ export const bootstrap = lifecycles.bootstrap;
 export const mount = lifecycles.mount;
 export const unmount = lifecycles.unmount;
 export { backendDependencies } from "./openmrs-backend-dependencies";
+export const importTranslation = require.context(
```
joeldenning: Since @openmrs/react-root-decorator@3 turns on translations by default, we need to now export an `importTranslations` function from this module.

src/root.component.tsx
```diff
@@ -39,4 +39,7 @@ function Root(props) {
   );
 }
 
-export default openmrsRootDecorator({ featureName: "patient-chart" })(Root);
+export default openmrsRootDecorator({
+  featureName: "patient-chart",
+  moduleName: "@openmrs/esm-patient-chart"
```
joeldenning: See https://github.com/openmrs/openmrs-react-root-decorator/releases/tag/v2.0.0

package.json
brandones: Looks like we only upgraded to v1.1.0?

src/openmrs-esm-patient-chart.tsx
brandones: All of the modules with UI strings have to have this boilerplate? Any way to get rid of it?

package.json
joeldenning: I linked to the wrong package. There is `eslint-config-react-important-stuff` and `eslint-config-ts-react-important-stuff`. The latter has a dependency on the former. I linked to the former, though.

1.1.0 for ts-react-important-stuff is the latest version.

src/openmrs-esm-patient-chart.tsx
joeldenning: There is no way to get rid of it. Webpack's `require.context()` is the way to load all files within a directory, which we need to do with the translations directory.

PR #29, opened 2019-11-21
Adding backend dependencies

joeldenning:
Specifying these backend deps helps us communicate to developers why things aren't working. I don't know what the actual minimum version is for the fhir and rest java modules for the patient-chart to work. I know that these versions are a bit high because the version we have installed on openmrs-spa.org is lower than the specified versions here. However, I'm not sure how to calculate the correct version. I just copied these versions from esm-primary-navigation.


brandones:
Perhaps use the version numbers from openmrs-spa.org since those are lower?

joeldenning:
Good idea - updated

PR #31, opened 2019-11-21
Switching to useCurrentPatient hook

joeldenning:
See related openmrs/openmrs-esm-api#23

### DIFF
--- a/src/root.component.tsx
+++ b/src/root.component.tsx
@@ -3,28 +3,13 @@ import { BrowserRouter, Route } from "react-router-dom";
 import openmrsRootDecorator from "@openmrs/react-root-decorator";
 import PatientChartSummary from "./summary/patient-chart-summary.component";
 import PatientBanner from "./summary/banner/patient-banner.component";
-import { getCurrentPatient } from "@openmrs/esm-api";
 import { LevelTwoRoutes } from "./summary/level-two-routes.component";
 
 function Root(props) {
-  const [currentPatient, setCurrentPatient] = React.useState(null);
-  const [showPatientSummary, setShowPatientSummary] = React.useState(true);
-
-  React.useEffect(() => {
-    const subscription = getCurrentPatient().subscribe(patient => {
-      setCurrentPatient(patient);
-    });
-
-    return () => subscription.unsubscribe();
-  });
   return (
     <BrowserRouter basename={window["getOpenmrsSpaBase"]()}>
       <Route path="/patient/:patientUuid/chart">
-        <PatientBanner
-          match={props.match}
-          patient={currentPatient}
-          showPatientSummary={setShowPatientSummary}
-        />
+        <PatientBanner match={props.match} />
       </Route>
       <Route
         path="/patient/:patientUuid/chart"
--- a/src/summary/documentation/dimensions-card-level-two.component.tsx
+++ b/src/summary/documentation/dimensions-card-level-two.component.tsx
@@ -3,16 +3,25 @@ import styles from "./dimensions-card-level-two.css";
 import SummaryCard from "../cards/summary-card.component";
 import { getDimensions } from "./dimensions-card.resource";
 import { match } from "react-router";
+import { useCurrentPatient } from "@openmrs/esm-api";
 
 function DimensionsCardLevelTwo(props: DimensionsCardLevelTwoProps) {
   const [dimensions, setDimensions] = React.useState([]);
+  const [
+    isLoadingPatient,
+    patient,
+    patientUuid,
+    patientErr
+  ] = useCurrentPatient();
 
   React.useEffect(() => {
-    const sub = getDimensions(props.currentPatient.id).subscribe(dimensions =>
-      setDimensions(dimensions)
-    );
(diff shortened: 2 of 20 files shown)


src/root.component.tsx
```diff
@@ -3,28 +3,13 @@ import { BrowserRouter, Route } from "react-router-dom";
 import openmrsRootDecorator from "@openmrs/react-root-decorator";
 import PatientChartSummary from "./summary/patient-chart-summary.component";
 import PatientBanner from "./summary/banner/patient-banner.component";
-import { getCurrentPatient } from "@openmrs/esm-api";
 import { LevelTwoRoutes } from "./summary/level-two-routes.component";
 
 function Root(props) {
-  const [currentPatient, setCurrentPatient] = React.useState(null);
-  const [showPatientSummary, setShowPatientSummary] = React.useState(true);
```
joeldenning: This was doing nothing so I removed it.

src/root.component.tsx
```diff
@@ -3,28 +3,13 @@ import { BrowserRouter, Route } from "react-router-dom";
 import openmrsRootDecorator from "@openmrs/react-root-decorator";
 import PatientChartSummary from "./summary/patient-chart-summary.component";
 import PatientBanner from "./summary/banner/patient-banner.component";
-import { getCurrentPatient } from "@openmrs/esm-api";
 import { LevelTwoRoutes } from "./summary/level-two-routes.component";
 
 function Root(props) {
-  const [currentPatient, setCurrentPatient] = React.useState(null);
```
joeldenning: The child components are now using `useCurrentPatient` instead of this

src/summary/documentation/dimensions-card-level-two.component.tsx
```diff
@@ -3,16 +3,25 @@ import styles from "./dimensions-card-level-two.css";
...
+import { useCurrentPatient } from "@openmrs/esm-api";
 
 function DimensionsCardLevelTwo(props: DimensionsCardLevelTwoProps) {
   const [dimensions, setDimensions] = React.useState([]);
+  const [
+    isLoadingPatient,
+    patient,
+    patientUuid,
+    patientErr
+  ] = useCurrentPatient();
```
joeldenning: Getting patientUuid from hook instead of from props means that we can render this component **before** we have the patient object. This is called the [Fetch as you render approach](https://reactjs.org/docs/concurrent-mode-suspense.html#approach-1-fetch-on-render-not-using-suspense) to data fetching in react. The downside of the approach is what the docs call a "cascade effect" where you don't render the child components until the top level data is available. Switching to `useCurrentPatient` makes it more possible for us to avoid the "cascade effect" that has the negative effect. Note that it doesn't totally solve the cascade problem - it just makes it easier for us to avoid it where possible. In the case of dimensions and allergies and stuff, we unfortunately have to first load the patient before we even have the identifier needed to load the other stuff.

This pattern I'm doing here has been transferred over to a bunch of other components, too

src/summary/documentation/dimensions-card.component.tsx
```diff
@@ -6,17 +6,26 @@ import SummaryCardRow from "../cards/summary-card-row.component";
...
 
   React.useEffect(() => {
-    const sub = getDimensions(props.currentPatient.id).subscribe(dimensions => {
+    const sub = getDimensions(patientUuid).subscribe(dimensions => {
       setDimensions(dimensions);
     });
 
     return () => sub.unsubscribe();
-  }, []);
+  }, [patientUuid]);
```
joeldenning: This was a missing dependency in the React effect's array. It was causing the dimensions to be refetched every time that this component re-rendered.

src/summary/documentation/documentation-section.component.tsx
```diff
@@ -1,30 +1,16 @@
...
-
   return (
     <>
-      {currentPatient && (
-        <SummarySectionCards match={props.match}>
-          <VitalsCard match={props.match} patient={currentPatient} />
-          <DimensionsCard match={props.match} currentPatient={currentPatient} />
-        </SummarySectionCards>
-      )}
+      <SummarySectionCards match={props.match}>
```
joeldenning: These components can now handle being rendered without the patient being loaded yet

src/summary/history/history-section.component.tsx
```diff
@@ -2,38 +2,20 @@ import React from "react";
...
-    const subscription = getCurrentPatient().subscribe(patient =>
-      setCurrentPatient(patient)
-    );
-
-    return () => subscription.unsubscribe();
-  });
-
   return (
-    <>
-      {currentPatient && (
```
joeldenning: The child components can now handle being rendered before the current patient is loaded

fatmali:
Great work on this @joeldenning ! With Suspense still in beta, somethings might change, however, the code is extremely simplified and I love that.

joeldenning:
Yeah switching to suspense will be an option in the future. There are some downsides to suspense, too, though, including just being a complex thing to think about. I figured that for now doing the "fetch on render" is the best approach, and it might stay that way even after things shake out with data fetching + suspense.

PR #41, opened 2019-11-29
fixed api call based on changes to openmrs fhir module and fixed vita…

jonathandick:
…ls component test related to timezones


jonathandick:
The tests were failing because of timezone issues with dayjs. The original datetimes are in Kenyan EAT. When you create dates with dayjs, it seems to use your local timezone. As a result, they pass when you run in Kenya, but fail on PST which is 11 hours behind and can cause a morning datetime in kenya to become the day before is PST. We just need to be a little more careful about thinking about timezones when creating tests on dates.

PR #47, opened 2019-11-30
Programs

jonathandick:
Added a widget to list programs for which a patient has been enrolled. This widget will allow a user to navigate to program specific dashboards. It will not follow same pattern of Level1, Level2,Level 3 used for other widgets. The designs are pending but have discussed this with Greg Schmidt. 

In addition, I have move put programs next to conditions on the dashboard and moved allergies below them. Note that this ordering is somewhat temporary as in the near future, we will be refactoring patient chart to use a configuration which determines the layout of widgets. 

image

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -6,5 +6,9 @@
     ],
     "plugins": [
       "@babel/plugin-proposal-class-properties",
+      ["i18next-extract", {
+        "outputPath": "translations/{{locale}}.json",
+        "discardOldKeys": true
+      }]
     ]
   }
--- a/src/summary/history/conditions-card.component.tsx
+++ b/src/summary/history/conditions-card.component.tsx
@@ -10,6 +10,7 @@ import { createErrorHandler } from "@openmrs/esm-error-handling";
 import HorizontalLabelValue from "../cards/horizontal-label-value.component";
 import { useCurrentPatient } from "@openmrs/esm-api";
 import SummaryCardFooter from "../cards/summary-card-footer.component";
+import { useTranslation } from "react-i18next";
 
 export default function ConditionsCard(props: ConditionsCardProps) {
   const [patientConditions, setPatientConditions] = React.useState(null);
@@ -19,6 +20,7 @@ export default function ConditionsCard(props: ConditionsCardProps) {
     patientUuid,
     patientErr
   ] = useCurrentPatient();
+  const { t } = useTranslation();
 
   React.useEffect(() => {
     if (patient) {
@@ -35,7 +37,7 @@ export default function ConditionsCard(props: ConditionsCardProps) {
   }, [patient]);
 
   return (
-    <SummaryCard name="Conditions" match={props.match}>
+    <SummaryCard name={t("conditions", "Conditions")} match={props.match}>
       <SummaryCardRow>
         <SummaryCardRowContent>
           <HorizontalLabelValue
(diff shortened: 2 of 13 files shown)


src/summary/history/programs/programs.resource.tsx
```diff
@@ -0,0 +1,18 @@
+import { openmrsObservableFetch } from "@openmrs/esm-api";
+import { Observable } from "rxjs";
+import { map, take } from "rxjs/operators";
+
+
+export function performPatientProgramsSearch(
+  patientID: string
+): Observable<PatientPrograms[]> {
+  return openmrsObservableFetch(
+    `/ws/rest/v1/programenrollment?patient=${patientID}`
```
jonathandick: There is no fhir api for getting program related data. I have created this ticket,
https://issues.openmrs.org/projects/MF/issues/MF-121 to remind us to make the change.

src/summary/history/programs/programs.resource.tsx
```diff
@@ -0,0 +1,18 @@
...
+  ).pipe(
+    map(({ data }) => data["results"]),
+    take(3)
+  );
+}
+
+type PatientPrograms = {
```
fatmali: Whats the point of this type?

.babelrc
```diff
@@ -6,5 +6,9 @@
     ],
     "plugins": [
```
jonathandick: Added the translation plugin using npm i --save-dev babel-plugin-i18next-extract

src/summary/history/conditions-card.component.tsx
```diff
@@ -35,7 +37,7 @@ export default function ConditionsCard(props: ConditionsCardProps) {
   }, [patient]);
 
   return (
-    <SummaryCard name="Conditions" match={props.match}>
+    <SummaryCard name={t("conditions", "Conditions")} match={props.match}>
```
jonathandick: Per @joeldenning's help, you can use this function instead of the <Trans> element when translating a string being passed as a property. 

Note, this probable doesn't belong in this PR but was a test case and such a small amount of code that I'm including it here.

src/summary/history/programs/programs-card.component.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const [
+    isLoadingPatient,
+    patient,
+    patientUuid,
+    patientErr
+  ] = useCurrentPatient();
+  const { t } = useTranslation();
+
+  React.useEffect(() => {
+    const subscription = performPatientProgramsSearch(patientUuid).subscribe(
```
fatmali: nit: I think this function could be named simpler, perhaps `getPatientPrograms` or `fetchPatientPrograms`

src/summary/history/programs/programs-card.component.tsx
```diff
@@ -0,0 +1,72 @@
...
+              color: "var(--omrs-color-ink-medium-contrast)",
+              fontFamily: "Work Sans"
+            }}
+          />
+        </SummaryCardRowContent>
+      </SummaryCardRow>
+      {patientPrograms &&
+        patientPrograms.map(program => {
+          return (
+            <SummaryCardRow key={program.uuid} linkTo="">
```
fatmali: why the empty `linkTo`?

jonathandick: Because there's no level 2 yet. I will update once I create the level 2.

src/summary/history/programs/programs-card.component.tsx
jonathandick: Done. Changed to fetchPatientPrograms

src/summary/history/programs/programs.resource.tsx
jonathandick: updated

PR #50, opened 2019-12-03
removed vitals request

fatmali:


fatmali:
Commented out vitals requests and related tests.

PR #55, opened 2019-12-05
Dynamic dashboards

jonathandick:
Added the initial code to build a dashboard dynamically.

### DIFF
--- a/src/summary/documentation/vitals-card.component.tsx
+++ b/src/summary/documentation/vitals-card.component.tsx
@@ -7,7 +7,6 @@ import { formatDate } from "./dimension-helpers";
 import { createErrorHandler } from "@openmrs/esm-error-handling";
 import { useCurrentPatient } from "@openmrs/esm-api";
 import SummaryCardFooter from "../cards/summary-card-footer.component";
-import { link } from "fs";
 
 export default function VitalsCard(props: VitalsCardProps) {
   const [patientVitals, setPatientVitals] = React.useState(null);
--- a/src/summary/documentation/vitals-card.component.tsx
+++ b/src/summary/documentation/vitals-card.component.tsx
@@ -7,6 +7,7 @@ import { formatDate } from "./dimension-helpers";
 import { createErrorHandler } from "@openmrs/esm-error-handling";
 import { useCurrentPatient } from "@openmrs/esm-api";
 import SummaryCardFooter from "../cards/summary-card-footer.component";
+import { link } from "fs";
 
 export default function VitalsCard(props: VitalsCardProps) {
   const [patientVitals, setPatientVitals] = React.useState(null);
@@ -30,7 +31,8 @@ export default function VitalsCard(props: VitalsCardProps) {
     <SummaryCard
       name="Vitals"
       match={props.match}
-      styles={{ flex: 1, margin: ".5rem", width: "45.5rem", maxWidth: "46rem" }}
+      styles={{ width: "100%", maxWidth: "45rem" }}
+      link={`/patient/${patientUuid}/chart/vitals`}
     >
       <table className={styles.vitalsTable}>
         <thead>
--- a/src/summary/patient-chart-summary.component.tsx
+++ b/src/summary/patient-chart-summary.component.tsx
@@ -1,17 +1,49 @@
 import React from "react";
 import { RouteComponentProps } from "react-router";
 import styles from "./patient-chart-summary.css";
-import HistorySection from "./history/history-section.component";
-import DocumentationSection from "./documentation/documentation-section.component";
+import DimensionsCard from "./documentation/dimensions-card.component";
+import VitalsCard from "./documentation/vitals-card.component";
+import ConditionsCard from "./history/conditions-card.component";
+import AllergyCard from "./history/allergy-card.component";
+import NotesCard from "./history/notes-card.component";
+import ProgramsCard from "./history/programs/programs-card.component";
 
 export default function PatientChartSummary(props: PatientChartSummaryProps) {
+  const config = [
+    "conditions",
+    "programs",
+    "allergies",
+    "notes",
+    "vitals",
+    {
+      module: DimensionsCard
+    }
+  ];
+
+  const coreComponents = {
+    conditions: ConditionsCard,
+    programs: ProgramsCard,
+    allergies: AllergyCard,
+    notes: NotesCard,
+    vitals: VitalsCard,
+    heightAndWeight: DimensionsCard
+  };
+
   return (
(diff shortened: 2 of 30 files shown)


src/summary/patient-chart-summary.component.tsx
```diff
@@ -1,17 +1,49 @@
...
-import DocumentationSection from "./documentation/documentation-section.component";
+import DimensionsCard from "./documentation/dimensions-card.component";
+import VitalsCard from "./documentation/vitals-card.component";
+import ConditionsCard from "./history/conditions-card.component";
+import AllergyCard from "./history/allergy-card.component";
+import NotesCard from "./history/notes-card.component";
+import ProgramsCard from "./history/programs/programs-card.component";
 
 export default function PatientChartSummary(props: PatientChartSummaryProps) {
+  const config = [
```
jonathandick: This is just a placeholder for now. Will create a follow-up PR which utilizes esm-config.

brandones: I think it would be very difficult to make esm-module-config able to deal with a specification like this. I suggest using the spec [defined here](https://docs.google.com/document/d/1isZZrnR6I2ffl2qu1_prjBxVuQ5TP7zU3oHY76GeUVg/edit?pli=1#heading=h.xqmj7xdo2ybl).

src/summary/patient-chart-summary.css
```diff
@@ -7,3 +7,18 @@
...
+.patientChartCards {
+  padding: 3.25rem 0;
+  display: flex;
+  flex-wrap: wrap;
+  align-items: flex-start;
+  width: 800px;
+  justify-content: flex-start;
+}
+
+.h {
```
donaldkibet: A more descriptive name for the `.h` class

src/summary/documentation/vitals-card.component.tsx
```diff
@@ -7,6 +7,7 @@ import { formatDate } from "./dimension-helpers";
 import { createErrorHandler } from "@openmrs/esm-error-handling";
 import { useCurrentPatient } from "@openmrs/esm-api";
 import SummaryCardFooter from "../cards/summary-card-footer.component";
+import { link } from "fs";
```
donaldkibet: Why do we require this import `import { link } from "fs"`

fatmali: @jonathandick ?

src/summary/patient-chart-summary.component.tsx
```diff
@@ -1,17 +1,49 @@
...
+    "conditions",
+    "programs",
+    "allergies",
+    "notes",
+    "vitals",
+    {
+      module: DimensionsCard
```
fatmali: why not just stick to one type in the array?

src/summary/patient-chart-summary.css
fatmali: @jonathandick ?

src/summary/patient-chart-summary.css
```diff
@@ -7,3 +7,18 @@
   justify-content: center;
   padding-top: 3.25rem;
 }
+
+.patientChartCards {
+  padding: 3.25rem 0;
+  display: flex;
+  flex-wrap: wrap;
+  align-items: flex-start;
+  width: 800px;
```
fatmali: convert to rem

src/summary/patient-chart-summary.component.tsx
brandones: +1, I don't want to try and make esm-module-config accept polymorphic arrays

src/summary/patient-chart-summary.component.tsx
jonathandick: I'm confused by the google doc. How do we represent widgets which will be imported?

src/summary/patient-chart-summary.component.tsx
jonathandick: See above. How to represent cases in which widgets are from outside this repo?

src/summary/patient-chart-summary.component.tsx
```diff
@@ -1,17 +1,49 @@
...
-        <DocumentationSection match={props.match} />
+      <div className={styles.h}>
+        <div className={styles.patientChartCards}>
+          {config.map((widget, index) => {
+            let Component;
+            if (typeof widget === "string") {
+              Component = coreComponents[widget];
+            } else {
+              Component = widget["module"];
+            }
```
brandones: I suggest something like

```
const coreComponents = {
  "notes": NotesComponent,
  ...
}
const otherComponentArray = config.widgetDefinitions.map(dfn => {
  const module = SystemJS.import(dfn.esModule)
  const component = module[dfn.exportName]  // idk
  return { key: dfn.name, value: component }
})
const otherComponents = // massage otherComponents from an array of objects with "key" and "value" keys into a single object with the value of "key" as the keys and the value of "value" as the values
const components = coreComponents.mergeOrWhatever(otherComponents)  // maybe it's 
assign"?
config.map((widgetName, index) => {
  const Component = components[widgetName]
  return <Component props={props.match} key={index} />;
})
```

src/summary/patient-chart-summary.css
jonathandick: renamed to patientChartCardsContainer

src/summary/documentation/vitals-card.component.tsx
jonathandick: Not sure where that came from. Was it me that added? Either way, removing.

src/summary/patient-chart-summary.component.tsx
jonathandick: For now, I am going to leave as is. Per  @brandones, he still needs to add funcationlity to the config module for arrays. Additionally, he and I were discussing how to handle dynamic loading of modules referenced in config. I think it would be easiest if the config module could handle the systemjs calls (and handling promises) rather than forcing the module which is doing the import to handle that. Curious to here @joeldenning's take on this.

src/summary/patient-chart-summary.css
jonathandick: set to 50 rem

src/summary/patient-chart-summary.component.tsx
jonathandick: As described above, I'd like to keep as is for now. Once config module supports arrays (and hopefully dynamic loading of es6 modules), I will refactor code as necessary.

jonathandick:
@fatmali , are you ok with us merging this for now? I will address the config stuff in later PR.

fatmali:
Resolve the conflicts first.

PR #56, opened 2019-12-05
Adding webpack-bundle-analyzer and npm run analyze script

joeldenning:
See https://github.com/webpack-contrib/webpack-bundle-analyzer. Run `npm run analyze` to see this:

image

### DIFF
--- a/translations/en.json
+++ b/translations/en.json
@@ -1,3 +1,6 @@
 {
+  "Active Programs": "Active Programs",
+  "Since": "Since",
+  "care programs": "Care Programs",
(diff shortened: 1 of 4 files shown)


translations/en.json
```diff
@@ -1,3 +1,6 @@
 {
+  "Active Programs": "Active Programs",
```
joeldenning: Not sure why these keep on getting removed/added

jonathandick: This is the only one we are using the t function for. Perhaps related to that?

PR #63, opened 2019-12-18
initial form widget implementation

nkimaina:


src/summary/cards/summary-card.component.tsx
```diff
@@ -19,7 +19,9 @@ export default function SummaryCard(props: SummaryCardProps) {
           )}
         </div>
         <div className={styles.headerAdd}>
-          <button className={`omrs-unstyled ${styles.addBtn}`}>Add</button>
+          {!props.hideHeaderRightButton && (
```
jonathandick: I think we should do this in the opposite fashion, that is it should only be displayed if a route is provided such that adding will lead to that route

src/summary/forms/form-list.component.tsx
```diff
@@ -0,0 +1,46 @@
...
+      match={props.match}
+      styles={{ margin: "1.25rem, 1.5rem", width: "100%", maxWidth: "45rem" }}
+      hideHeaderRightButton={true}
+    >
+      {forms &&
+        forms.map(form => {
+          return (
+            <SummaryCardRow
+              key={form.uuid}
+              linkTo={`chart/formentry/${form.uuid}`}
```
jonathandick: this link should be a configuration and should allow someone to go to chart/forms/ampath-form-entry

src/summary/forms/form-list.component.tsx
```diff
@@ -0,0 +1,46 @@
+import React from "react";
+import { match } from "react-router";
+import SummaryCard from "../cards/summary-card.component";
+import SummaryCardRow from "../cards/summary-card-row.component";
+import { getAllForms, Form } from "./form-list.resource";
+export default function FormList(props: FormListProps) {
```
jonathandick: Instead of list, let's calls this FormsCard

PR #69, opened 2020-01-07
Breadcrumbs fix

fatmali:


src/summary/level-two-routes.component.tsx
```diff
@@ -8,44 +8,50 @@ import ProgramsLevelTwo from "./history/programs/programs-level-two.component";
 import MedicationsDetailedSummary from "./history/medications/medications-detailedSummary.component";
 import { AllergyCreateForm } from "./history/allergy-create-form.component";
 
-const levelTwoRoutes = [
+export const levelTwoRoutes = [
```
jonathandick: We will need to handle the case of dynamically added widgets + routes. We can do that in subsequent PR.

PR #71, opened 2020-01-09
Jj dyn

jonathandick:
Adding the ability to dynamically import react components + routes into esm-patient-chart from an outside cdn.


src/root.component.tsx
```diff
@@ -6,6 +6,44 @@ import PatientBanner from "./summary/banner/patient-banner.component";
...
+        if (m.status === "fulfilled") {
+          for (let [exportName, widget] of Object.entries(m.value.widgets)) {
+            if (widget.hasOwnProperty("routes")) {
+              widgetRoutes.push(widget["routes"]);
+            }
+          }
+        }
+      });
+      setWidgetRoutes(widgetRoutes);
+    });
```
brandones: What happens if one rejects?

src/summary/patient-chart-summary.component.tsx
```diff
@@ -10,44 +10,92 @@ import ProgramsCard from "./history/programs/programs-card.component";
...
+      medications: MedicationsSummary
+    };
+
+    const modulePromises = [];
+
+    config.map(c => {
+      if (c["esModule"]) {
+        modulePromises.push(System.import(c.esModule));
+      }
+    });
```
brandones: Let's remember to refactor this so that these promises only happen once. Probably just in the root component, and then the router passes the widgets to the pages.

src/summary/patient-chart-summary.component.tsx
```diff
@@ -10,44 +10,92 @@ import ProgramsCard from "./history/programs/programs-card.component";
...
+      conditions: ConditionsCard,
+      programs: ProgramsCard,
+      allergies: AllergyCard,
+      notes: NotesCard,
+      vitals: VitalsCard,
+      heightAndWeight: DimensionsCard,
+      medications: MedicationsSummary
+    };
+
+    const modulePromises = [];
```
brandones:
```suggestion
    // TODO: Refactor this so that these promises only happen once. 
    //   Probably just in the root component, and then the router passes the widgets to the pages.
    const modulePromises = [];
```

brandones:
Gotta fix CI. And if we're not gonna fix the promise duplication now, let's leave a note about it.

src/root.component.tsx
jonathandick: Good question. I'm not sure what to do. We could (1) do nothing; (2) warn the user that some data couldn't load; (3) try to reimport; (4) not show anything. I don't know what the best option is but I would lean towards (2). However, I'm not sure how a user would interpret this information but I think it's the best option to investigate. What do you think?

src/summary/patient-chart-summary.component.tsx
jonathandick: My understanding is that system.js will cache after importing. In subsequent, calls, I think of it (and this may be wrong) as more akin to doing a standard "import" although you're right, it does return a promise. From a performance perspective, I think it's safe. I'm also not sure how we would "save" the imports after being load in the root component. We could make them a prop to a child of root but that seems messy to me. It seemed cleaner to just use system.import again. What do you think?

src/summary/patient-chart-summary.component.tsx
```diff
@@ -49,13 +49,14 @@ export default function PatientChartSummary(props: PatientChartSummaryProps) {
...
     const modulePromises = [];
 
-    const widgets = [];
     config.map(c => {
       if (c["esModule"]) {
         modulePromises.push(System.import(c.esModule));
       }
     });
 
+    console.log(modulePromises);
```
fatmali: remove this `console.log`

src/summary/patient-chart-summary.component.tsx
```diff
@@ -66,14 +67,16 @@ export default function PatientChartSummary(props: PatientChartSummaryProps) {
         }
       });
 
+      console.log(moduleWidgets);
+
       config.map(c => {
         c.hasOwnProperty("esModule")
           ? widgets.push(moduleWidgets[c.exportName].root)
```
fatmali: This might lead to errors if  `moduleWidgets[c.exportName].root` is `undefined`

src/summary/patient-chart-summary.component.tsx
brandones: If you're right about caching, then your solution seems perfect :+1:

src/root.component.tsx
brandones: Agree on (2). Also is consistent with our previous decision about errors and honesty.

jonathandick: I've updated such that we create a toast for each module/widget that is unable to be loaded.

src/root.component.tsx
```diff
@@ -6,6 +6,52 @@ import PatientBanner from "./banner/patient-banner.component";
...
+        exportName: "programsWidget"
+      }
+    ];
+
+    //placeholder while config mdoule is completed
+    const config = [];
+
+    const modulePromises = [];
+
+    config.map(c => {
```
fatmali: use `forEach` instead of `map` here.

fatmali:
@jonathandick you have conflicting files, also is it possible to add a test or two for these changes?

PR #81, opened 2020-01-21
Beginning of workspace

fatmali:


src/widgets/allergies/allergy-form.component.tsx
```diff
@@ -13,6 +13,7 @@ import {
 import { createErrorHandler } from "@openmrs/esm-error-handling";
 import { useCurrentPatient } from "@openmrs/esm-api";
 import dayjs from "dayjs";
+import { WorkspaceItemProps } from "../../workspace/workspace.resource";
```
jonathandick: Will it be necessary for every component that wants to be placed in the workspace to import WorkspaceItemProps? It would be better if the component was not required to "know" about the workspace.

src/widgets/allergies/allergy-form.component.tsx
```diff
@@ -241,7 +242,11 @@ export function AllergyForm(props: AllergyFormProps) {
           background: "var(--omrs-color-bg-medium-contrast)"
         }}
       >
-        <form onSubmit={handleCreateFormSubmit} ref={formRef}>
+        <form
```
jonathandick: same thing here as above - is it necessary for this component to know about the workspace?

PR #85, opened 2020-01-27
Responsive workspace

fatmali:


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Patient dashboard microfrontend for the OpenMRS SPA",
   "main": "dist/openmrs-esm-patient-chart.js",
   "scripts": {
-    "start": "webpack-dev-server",
+    "start": "webpack-dev-server --host 0.0.0.0",
```
fatmali: this makes it possible to debug using a different device because the app is now broadcasted on the network rather than just localhost

src/widgets/cards/summary-card.css
```diff
@@ -6,7 +6,6 @@
 }
 
 :global(.omrs-breakpoint-gt-tablet) .card {
-  width: 21.75rem;
```
jonathandick: If we remove this, we will no longer have the ability to have two columns of widgets. I will try to think of how to redo this but please do so as well.

src/widgets/vitals/vitals-form.component.tsx
```diff
@@ -749,7 +748,6 @@ export function VitalsForm(props: vitalsFormProp) {
               ? styles.buttonStyles
               : `${styles.buttonStyles} ${styles.buttonStylesBorder}`
           }
-          style={{ position: "absolute", bottom: 0 }}
```
jonathandick: Per @donaldkibet , this is suppose to be absolute as the button is supposed to be at the bottom. Does this achieve the right design?

jonathandick:
I think we can merge this code. We can sort how to do two column layouts in a subsequent PR.

PR #86, opened 2020-01-27
MF-129: Allergy Level Three Widget

denniskigen:
Screenshot 2020-01-27 at 11 51 26

figma link https://www.figma.com/file/YyVhWTNp89bNiOGadYrHBk/Patient-Chart-Master-current-file?node-id=6427%3A189644

### DIFF
--- a/src/widgets/allergies/allergy-card-level-three.component.tsx
+++ b/src/widgets/allergies/allergy-card-level-three.component.tsx
@@ -1,0 +1,164 @@
+import React from "react";
+import { match } from "react-router";
+import { createErrorHandler } from "@openmrs/esm-error-handling";
+import { getPatientAllergyByPatientUuid } from "./allergy-intolerance.resource";
+import { useCurrentPatient } from "@openmrs/esm-api";
+import styles from "./allergy-card-level-three.css";
+import dayjs from "dayjs";
+import SummaryCard from "../cards/summary-card.component";
+
+export function AllergyCardLevelThree(props: AllergyCardLevelThreeProps) {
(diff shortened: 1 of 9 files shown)


src/widgets/allergies/allergy-card-level-three.component.tsx
```diff
@@ -0,0 +1,164 @@
...
+        .then(allergy => setAllergy(allergy.data))
+        .catch(createErrorHandler());
+
+      return () => abortController.abort();
+    }
+  }, [isLoadingPatient, patient]);
+
+  function displayAllergy() {
+    return (
+      <SummaryCard
```
fatmali: What happens if there are no allergies to display?

src/widgets/allergies/allergy-card-level-two.css
```diff
@@ -98,6 +98,9 @@
 
 .allergyFooter {
   text-align: center;
+  color: var(--omrs-color-ink-medium-contrast);
+  font-family: "Work sans";
+  font-size: 17px;
```
fatmali: use rem instead of px

denniskigen: Noted

src/widgets/allergies/allergy-card-level-three.component.tsx
denniskigen: Could this situation arise? This widget would only display after you've clicked on an existing allergy.

fatmali: ooh right, it's alright then.

PR #89, opened 2020-01-27
MF-128: Conditions Level Three Widget

denniskigen:
Screenshot 2020-01-27 at 16 03 58

figma link https://www.figma.com/file/YyVhWTNp89bNiOGadYrHBk/Patient-Chart-Master-current-file?node-id=6427%3A189644

### DIFF
--- a/src/summaries/level-two-routes.component.tsx
+++ b/src/summaries/level-two-routes.component.tsx
@@ -10,6 +10,7 @@ import { VitalsForm } from "../widgets/vitals/vitals-form.component";
 import ConditionsBriefSummary from "../widgets/conditions/conditions-brief-summary.component";
 import MedicationLevelTwo from "../widgets/medications/medication-level-two.component";
 import { AllergyCardLevelThree } from "../widgets/allergies/allergy-card-level-three.component";
+import ConditionsDetailedSummary from "../widgets/conditions/conditions-detailed-summary.component";
 
 export const levelTwoRoutes: PatientChartRoute[] = [
   {
@@ -56,6 +57,11 @@ export const levelTwoRoutes: PatientChartRoute[] = [
     url: "/patient/:patientUuid/chart/conditions",
     component: ConditionsBriefSummary,
     name: "Conditions"
+  },
+  {
+    url: "/patient/:patientUuid/chart/conditions/:conditionUuid",
+    component: ConditionsDetailedSummary,
+    name: "Condition"
   }
 ];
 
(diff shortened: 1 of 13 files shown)


src/summaries/level-two-routes.component.tsx
```diff
@@ -56,6 +57,11 @@ export const levelTwoRoutes: PatientChartRoute[] = [
     url: "/patient/:patientUuid/chart/conditions",
     component: ConditionsBriefSummary,
     name: "Conditions"
+  },
+  {
+    url: "/patient/:patientUuid/chart/conditions/:conditionUuid",
+    component: ConditionsDetailedSummary,
+    name: "Condition"
```
jonathandick: I think this should be "Conditions"

src/widgets/cards/summary-card.css
```diff
@@ -16,10 +16,19 @@
...
   color: var(--omrs-color-interaction-plus-one);
 }
 
+button.editBtn:hover {
+  color: var(--omrs-color-interaction-plus-one);
+}
+
 .header {
   display: grid;
   grid-template-columns: [one] 1fr [two] 68px;
```
jonathandick: Please don't use px, use rem.

src/widgets/conditions/conditions-detailed-summary.css
```diff
@@ -0,0 +1,51 @@
...
+  display: flex;
+}
+
+:global(.omrs-breakpoint-lt-tablet) .conditionSummary {
+  padding: 0.25rem 0.5rem 0 0.5rem;
+}
+
+.conditionCard {
+  margin: 0rem 1rem;
+  font-size: 1.063rem;
```
jonathandick: Why such an unusual font size?

src/summaries/level-two-routes.component.tsx
denniskigen: This route is for the detailed view when you click on an individual condition - so the singular form should be appropriate for it.

src/widgets/cards/summary-card.css
denniskigen: Ok. I'll change this.

src/widgets/conditions/conditions-detailed-summary.css
denniskigen: 17px in keeping with the design, though 16px (which neatly converts to 1rem) would probably be just as appropriate (and makes for a cleaner conversion). Should I go with 17px as per the design?

jonathandick: Please stick with design.

src/summaries/level-two-routes.component.tsx
jonathandick: In "Chart -> Conditions -> Renal Failure", the "Conditions" refers to the detailedSummary widget, so the link is to "Conditions". The choice of name within a breadcrumb is independent of the final node in the breadcrumb

src/widgets/conditions/conditions-detailed-summary.css
denniskigen: Ok. 1.063 rem it is.

PR #96, opened 2020-01-30
Past medication

kagai:


src/widgets/medications/medication-level-two.component.tsx
```diff
@@ -50,43 +51,142 @@ export default function MedicationLevelTwo(props: MedicationsOverviewProps) {
...
-          style={{ color: "var(--omrs-color-ink-medium-contrast)" }}
-          className={"omrs-type-body-large"}
+  function displayPastMedications() {
+    return (
+      <React.Fragment>
+        <SummaryCard
+          name={t("Medications - past", "Medications - past")}
+          match={props.match}
+          styles={{ width: "90%" }}
+          addBtnUrl={`/patient/${patientUuid}/chart/add`}
```
fatmali: does this route exist?

src/widgets/medications/medication-level-two.component.tsx
```diff
@@ -50,43 +51,142 @@ export default function MedicationLevelTwo(props: MedicationsOverviewProps) {
...
+                          <td>
+                            <span
+                              style={{
+                                fontWeight: 500,
+                                color: "var(--omrs-color-ink-high-contrast)"
+                              }}
+                            >
+                              {medication.drug.display}
+                            </span>
+                            {" \u2014 "} {medication.doseUnits.display}{" "}
```
fatmali: why so many whitespaces? is there a different way we can achieve this without having the whitespace inside html?

src/widgets/medications/medication-level-two.component.tsx
```diff
@@ -50,43 +51,142 @@ export default function MedicationLevelTwo(props: MedicationsOverviewProps) {
...
+                            <span
+                              style={{
+                                fontWeight: 500,
+                                color: "var(--omrs-color-ink-high-contrast)"
+                              }}
+                            >
+                              {getDosage(
+                                medication.drug.strength,
+                                medication.dose
+                              )}{" "}
```
fatmali: same here about whitespace

src/widgets/medications/medication-level-two.component.tsx
kagai: Removed.. it was a place holder thou'

src/widgets/medications/medication-level-two.component.tsx
kagai: Removed most of these

src/widgets/medications/medication-level-two.component.tsx
```diff
@@ -31,62 +32,165 @@ export default function MedicationLevelTwo(props: MedicationsOverviewProps) {
...
-            <td>Start Date</td>
-          </tr>
-        </thead>
-        <tbody>{patientMedications && parseMedication()}</tbody>
-      </table>
-      <div className={styles.medicationFooter}>
-        <p
-          style={{ color: "var(--omrs-color-ink-medium-contrast)" }}
-          className={"omrs-type-body-large"}
+  function displayPastMedications() {
```
drizzentic: L35 method has a lot of similarities `displayPastMedication`. Is there a reason we should explicitly have two methods for past and current medication?

kagai: I understand your concern, I tried that at first but it makes the logic complicated, the functions are similar but they have major differences that would make it complicated for no good reason.Let me know your thoughts on this.

donaldkibet:
> _No description provided._

lgtm

PR #97, opened 2020-01-30
MF-130: Implement a form for adding and editing conditions

denniskigen:
Screenshot 2020-01-30 at 12 14 19

Screenshot 2020-02-03 at 23 17 59

### DIFF
--- a/src/summaries/level-two-routes.component.tsx
+++ b/src/summaries/level-two-routes.component.tsx
@@ -68,7 +68,7 @@ export const levelTwoRoutes: PatientChartRoute[] = [
     name: "Order Medications"
   },
   {
-    url: "/patient/:patientUuid/chart/allergy/form/:allergyUuid?",
+    url: "/patient/:patientUuid/chart/allergies/form/:allergyUuid?",
     component: AllergyForm,
     name: "Allergy Form"
   },
--- a/src/summaries/level-two-routes.component.tsx
+++ b/src/summaries/level-two-routes.component.tsx
@@ -13,6 +13,7 @@ import { AllergyCardLevelThree } from "../widgets/allergies/allergy-card-level-t
 import ConditionsDetailedSummary from "../widgets/conditions/conditions-detailed-summary.component";
 import { MedicationOrderBasket } from "../widgets/medications/medication-order-basket.component";
 import { MedicationOrder } from "../widgets/medications/medication-order.component";
+import { ConditionsForm } from "../widgets/conditions/conditions-form.component";
 
 export const levelTwoRoutes: PatientChartRoute[] = [
   {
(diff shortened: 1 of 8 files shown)


ojwanganto:
@denniskigen could we also plan for a similar widget that provides a grid for inputting multiple conditions? I think that way a user will be saved from the need to add conditions one by one. Just a light comment though

src/widgets/conditions/conditions-form.component.tsx
```diff
@@ -0,0 +1,244 @@
...
+                    fontWeight: 500
+                  }}
+                >
+                  {patientCondition.resource.code.text}{" "}
+                </div>
+              </div>
+              <div className={`${styles.conditionHeader} omrs-bold`}>
+                <div
+                  className="omrs-type-body-regular"
+                  style={{ color: "#535355" }}
```
fatmali: use colors defined in the styleguide.

src/summaries/level-two-routes.component.tsx
```diff
@@ -28,6 +29,21 @@ export const levelTwoRoutes: PatientChartRoute[] = [
     component: AllergyCardLevelThree,
     name: "Allergy"
   },
+  {
+    url: "/patient/:patientUuid/chart/conditions",
+    component: ConditionsBriefSummary,
```
fatmali: this component was renamed to ConditionsOverview in master.

denniskigen: ConditionsBriefSummary is the level-two component. ConditionsOverview is the master card.

src/widgets/conditions/conditions-form.component.tsx
denniskigen: This particular color isn't included in the style guide I'm afraid.

denniskigen:
> @denniskigen could we also plan for a similar widget that provides a grid for inputting multiple conditions? I think that way a user will be saved from the need to add conditions one by one. Just a light comment though

Sure @ojwanganto, I'm open to such a discussion although I see that this wasn't catered for in the design. I'll share this sentiment with Greg.

denniskigen:
> This looks generally ok. However, I think that there is confusion over the naming conventions. The "level 1" card should be called conditions-summary.component.tsx and the component should be called ConditionsSummary. The level 2 component should be called conditions-detailed-summary.component.tsx and the component called ConditionsDetailedSummary.

#102 fixes this.

src/widgets/conditions/conditions-form.component.tsx
denniskigen: I've removed this

jonathandick:
Is this still a draft? Or is it ready to be reviewed for merging?

denniskigen:
> Is this still a draft? Or is it ready to be reviewed for merging?

It's still a draft primarily because the conditions resource is not available either via OpenMRS or via the FHIR resource. Presently, all the conditions data is just mock data.

jonathandick:
If that's the only reason, I think this can be converted to normal PR (not a draft) and we can merge

fatmali:
@denniskigen you can mark this as ready for review so it can get merged.

jonathandick:
I'm going to close this PR. Please push to new repo esm-patient-chart-widgets, once created

PR #101, opened 2020-02-03
Added navbars for top level patient chart and for summaries

jonathandick:


jonathandick:
These are based on the designs in figma:

image

A few things to note:

- The workspace is no longer functioning properly. I will fix in subsequent PR. 

- The nav is responsive in that it allows for a user to scroll horizontally if the size of the screen becomes to small to show all nav items.

- If you move from Summaries --> Overview to Results and click back to "Summaries", it will bring to the last "Summary" the user had seen.

PR #109, opened 2020-02-06
Multiple tabs 

drizzentic:
Fix issue where multiple tabs open on the workspace for vitals and medication


jonathandick:
LGTM. @fatmali, does this follow pattern you were thinking?

fatmali:
Yes, validations is a function that returns true or false on whether to add a new item to the workspace, if it returns a number, the workspace navigates to a tab at that index of the tabs array.

PR #111, opened 2020-02-07
App props as observable

nkimaina:


fatmali:
@nkimaina why do you need props for patient chart?

nkimaina:
@fatmali I need it access the 'mountParcel' method in the props, outside of patient chart. Down the line, components will need to access other app properties like configs. This can be done by propagating props from the root components all the way. This makes such components brittle, as one might miss to propagate the props breaking the link. It will likely cause regressions when adding new components. Exposing the props statically this way, prevents this. single-spa angular seed has provided this functionality, so I just copied it over.

fatmali:
Ok sounds good 👍

jonathandick:
I don't think this is the right pattern. Perhaps you should consider the context api for this? https://reactjs.org/docs/context.html

fatmali:
thinking about it, it worked on my machine, perhaps I should try pushing it to git later today and you try it without the workaround?

nkimaina:
@fatmali Did you manage to get it working? You can do a separate PR for it, then I will close this PR for now. 

@jonathandick That's a good idea. Given that we have a react way of doing it, I will close this pull request as react provides a way for using providers.

PR #113, opened 2020-02-07
Nav hoc

jonathandick:


jonathandick:
The build is failing because of an issue with importing lo-dash-es in the vitals-form. I'm not sure why that's happening or how to fix it. @donaldkibet I think you introduced this code, could you let me know your thoughts?

src/ui-components/chart-widget/chart-widget.component.tsx
```diff
@@ -0,0 +1,81 @@
...
+          `/patient/${patientUuid}/chart/${props.widgetConfig.name}` ? (
+          <Redirect
+            to={`/patient/${patientUuid}/chart/${props.widgetConfig.name}/${props.widgetConfig.defaultRoute}`}
+          />
+        ) : (
+          <Redirect to={props.paths[props.widgetConfig.name]} />
+        )}
+      </Route>
+
+      {props.widgetConfig.routes.map((item, index) => {
```
jonathandick: @fatmali , I'm struggling with where exactly routes should be declared. This component takes a config (see medications.component.tsx  below for en example) and: 1) creates a component with a nav bar and 2) declares the routes for that component. 

Should the routes be declared here? Should we be centrally declaring all routes in one file? The latter seems like a bit unwieldy but at least easy to figure out what routes are being declared. The former seems a little more logical, it's easier to associate a feature with routes. 

Interested to hear your thoughts.

donaldkibet:
> The build is failing because of an issue with importing lo-dash-es in the vitals-form. I'm not sure why that's happening or how to fix it. @donaldkibet I think you introduced this code, could you let me know your thoughts?

its is failing because `lodash-es` exposes ES Modules which requires transformation by babel. we can fix it by including the following config in jest.config.json file.
`"moduleNameMapper" :
 {"lodash-es": "lodash"}` or 
`"transformIgnorePatterns": [
  "<rootDir>/node_modules/(?!lodash-es)"
]`

jonathandick:
Thanks @donaldkibet, the fix above worked for jest.

I'm now getting a new test error for AllergyLevelThree (which I don't think is related to my code) . Could you take a look at that?  Or if it wasn't you who worked on this code originally, could you help me identify the right person?

jonathandick:
I have updated the nav structure in the following way. I think this does a reasonable job of categorizes things well and it's enough to go on:

image

src/ui-components/chart-widget/chart-widget.component.tsx
fatmali: Having all the routes in one file is going to be difficult to manage and to maintain. I think it would be much better to split it up into different files.

src/chart-review/orders/orders.component.tsx
```diff
@@ -0,0 +1,60 @@
...
+import OrdersChartOverview from "./orders-chart-overview.component";
+
+export default function Orders(props: any) {
+  let { patientUuid } = useParams();
+
+  function showOrders() {
+    newWorkspaceItem({
+      component: MedicationOrderBasket,
+      name: "Medication Order Basket",
+      props: { match: { params: {} } },
```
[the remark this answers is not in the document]
jonathandick: Yes. You can also just use a react router hook to get the same data, eg. useHistory, or useRouteMatch. See https://reacttraining.com/blog/react-router-v5-1/. 

In general, (I think) we should be moving away from using props.match and move to using hooks instead.

fatmali:
why is travis failing on this?

jonathandick:
I've addressed the changes above and I'm going to merge this PR. The following to be addressed in subsequent PRs:
1. Move Allergies/Conditions/Programs to top-level elements in the navigation structure
2. Add tests (and encourages devs to add tests to other components). I had to move test coverage down to 45% as we are now under 50% and I couldn't merge. 
3. Use the config module to set up configuration.

PR #117, opened 2020-02-13
MF-135: Medication level three component

kagai:


jonathandick:
Please post a picture of this.

src/widgets/medications/medication-level-three/medication-level-three.component.tsx
```diff
@@ -0,0 +1,216 @@
...
+            <thead>
+              <tr>
+                <td>Last updated</td>
+                <td>Last updated by</td>
+                <td>Last updated location</td>
+              </tr>
+            </thead>
+            <tbody>
+              <tr>
+                <td data-testid="last-updated">
```
fatmali: although `data-testid` is part of the `react-testing-library`, it's highly discouraged to use it because we are adding test attributes to the source code  that have no function aside from helping tests run. Say in future if we need to migrate away from the react-testing-library, it would involve a lot of work to change tests as well as some source code as well. I'd encourage you not to use it, and figure out a different way. One way would be to get an array of this row's children and use plain old indexes to access the indexes. cc @donaldkibet

jonathandick: @kagai any updates on this PR?

kagai: @jonathandick    screenshot

PR #119, opened 2020-02-14
Adding programs, allergies and conditions to main nav. Removing clini…

jonathandick:
…cal history


src/chart-review/encounters/encounters.component.tsx
```diff
@@ -1,42 +1,37 @@
...
   return new URLSearchParams(useLocation().search);
 }
 
 export default function Encounters(props: any) {
+  const match = useRouteMatch();
+
   const widgetConfig = {
     name: "encounters",
-    defaultTabIndex: 0,
+    defaultTabIndex: 1,
```
fatmali: what happens if multiple widgets have the same tab index?

jonathandick: That's not possible.  In this case, I've moved  allergies/conditions/programs to the main nav. In this config, you only choose a specific main nav item. Each component that this refers to, if it has a tab, can also set it's default item. So for example, the Results widget can also identify which defaulit would like to display.

PR #120, opened 2020-02-16
Programs summary and detail components

denniskigen:
Screenshot 2020-02-16 at 14 41 52

Level-three component (detailed summary) for programs as well as some modifications to the look of the level-two (summary) UI.


src/widgets/programs/programs-detailed-summary.component.tsx
```diff
@@ -0,0 +1,83 @@
...
+          </div>
+        </SummaryCard>
+      </>
+    );
+  }
+
+  return (
+    <>
+      {patientProgram && (
+        <div className={styles.programSummary}>{displayProgram()}</div>
```
fatmali: good work, why did you abstract the `propgram` into a function?

src/widgets/programs/programs.resource.tsx
```diff
@@ -13,7 +13,26 @@ export function fetchPatientPrograms(
...
-type PatientPrograms = {
+export function getPatientProgramByUuid(
+  programUuid: string
+): Observable<PatientProgram> {
+  return openmrsObservableFetch(
+    `/ws/rest/v1/programenrollment/${programUuid}`
+  ).pipe(map(({ data }) => mapToPatientProgram(data)));
+}
+
+function mapToPatientProgram(data: any): PatientProgram {
```
fatmali: you can get fancy and do this:
```suggestion
function mapToPatientProgram(data): PatientProgram {
      return {...data}
}
```
perhaps this shouldn't even be a function at all 😄

src/widgets/programs/programs-detailed-summary.component.tsx
denniskigen: Oversight on my part - I usually extract component logic when I need to conditionally display one component or another. In this case, there's really no need to extract the view logic into a function.

PR #121, opened 2020-02-16
Silence react-i18next warnings when running tests

denniskigen:
Screenshot 2020-02-16 at 22 21 06
Screenshot 2020-02-16 at 22 30 35

Silences annoying react-i18next Jest warning when running tests by mocking out the dep.

### DIFF
--- a/src/widgets/allergies/allergy-card-level-three.test.tsx
+++ b/src/widgets/allergies/allergy-card-level-three.test.tsx
@@ -11,6 +11,10 @@ const mockGetPatientAllergyByPatientUuid = getPatientAllergyByPatientUuid as jes
 const mockUseCurrentPatient = useCurrentPatient as jest.Mock;
 const mockUseRouteMatch = useRouteMatch as jest.Mock;
 
+jest.mock("react-i18next", () => ({
+  useTranslation: () => ({ t: key => key })
+}));
+
 jest.mock("./allergy-intolerance.resource", () => ({
   getPatientAllergyByPatientUuid: jest.fn()
 }));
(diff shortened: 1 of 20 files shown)


denniskigen:
> looks good, good work @denniskigen.

Thanks!

src/widgets/allergies/allergy-card-level-three.test.tsx
```diff
@@ -9,6 +9,10 @@ import { patient, mockAllergyResult } from "../../../__mocks__/allergy.mock";
 const mockGetPatientAllergyByPatientUuid = getPatientAllergyByPatientUuid as jest.Mock;
 const mockUseCurrentPatient = useCurrentPatient as jest.Mock;
 
+jest.mock("react-i18next", () => ({
```
donaldkibet: @denniskigen :thumbsup: just a suggestion rather than having to write this mock in every file we use translation can this work for our case, @fatmali what are your thoughts 
https://medium.com/@arivu_a/how-to-mock-functions-globally-in-jest-f267fedf7683

fatmali: nice find @donaldkibet , this looks simpler and much more elegant.

denniskigen: Even better @donaldkibet. I'll amend the PR.

denniskigen: A more concise way of mocking the i8next dependency globally is presented in their repo [here](https://github.com/i18next/react-i18next/tree/master/example/test-jest).

PR #124, opened 2020-02-18
Add config

jonathandick:
This is the initial introduction of a config via esm-module config. In a subsequent PR, I will add the ability to load dynamically non-core widgets.

### DIFF
--- a/src/chart-review/chart-review.component.tsx
+++ b/src/chart-review/chart-review.component.tsx
@@ -8,6 +8,7 @@ import Encounters from "./encounters/encounters.component";
 import Allergies from "./allergies/allergies.component";
 import Conditions from "./conditions/conditions.component";
 import Programs from "./programs/programs.component";
+import { useConfig } from "@openmrs/esm-module-config";
 
 export default function ChartReview(props: any) {
   const match = useRouteMatch();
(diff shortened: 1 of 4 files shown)


src/chart-review/chart-review.component.tsx
```diff
@@ -8,76 +8,86 @@ import Encounters from "./encounters/encounters.component";
...
-        path: `programs`,
-        component: () => {
-          return <Programs />;
-        }
+  const defaultPath = `/patient/${patientUuid}/chart/`;
+
+  function getConfigWidgets() {
+    const w = [];
+    config.widgets.map(widgetName => {
+      w.push(coreWidgets[widgetName]);
```
fatmali: map returns a new transformed array, no need of `w` here
```suggestion
return config.widgets.map(widgetName => coreWidgets[widgetName]);
```

src/chart-review/chart-review.component.tsx
```diff
@@ -8,76 +8,86 @@ import Encounters from "./encounters/encounters.component";
...
+    });
+    return w;
+  }
+
+  const coreWidgets = {
```
fatmali: can this be stored in a separate file ? Or is it going to come in as a config ?

jonathandick: Let me move to a separate file in subsequent PR.

src/chart-review/chart-review.component.tsx
jonathandick: Done

PR #130, opened 2020-02-24
Future appointments list ui.

drizzentic:
Add appointments widget to patient chart

### DIFF
--- a/src/chart-review/appointments/appointments.component.tsx
+++ b/src/chart-review/appointments/appointments.component.tsx
@@ -1,20 +1,19 @@
-import React from "react";
 import { useRouteMatch } from "react-router-dom";
 import ChartWidget from "../../ui-components/chart-widget/chart-widget.component";
-import PatientChartOverview from "./overview/patient-chart-overview.component";
+import React from "react";
 import AppointmentsOverview from "../../widgets/appointments/appointments-overview.component";
 
-export default function Summaries(props: any) {
+export default function Appointment(props: any) {
   const match = useRouteMatch();
 
   const widgetConfig = {
-    name: "summaries",
+    name: "appointments",
     defaultTabIndex: 0,
     tabs: [
       {
         name: "Overview",
         component: () => {
-          return <PatientChartOverview match={match} />;
+          return <AppointmentsOverview match={match} />;
         }
       }
     ]
--- a/src/chart-review/appointments/appointments.component.tsx
+++ b/src/chart-review/appointments/appointments.component.tsx
@@ -4,8 +4,6 @@ import React from "react";
 import AppointmentsOverview from "../../widgets/appointments/appointments-overview.component";
 
 export default function Appointment(props: any) {
-  const match = useRouteMatch();
-
   const widgetConfig = {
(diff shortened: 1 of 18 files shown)


src/chart-review/appointments/appointments.component.tsx
```diff
@@ -0,0 +1,23 @@
+import { useRouteMatch } from "react-router-dom";
+import ChartWidget from "../../ui-components/chart-widget/chart-widget.component";
+import React from "react";
+import AppointmentsOverview from "../../widgets/appointments/appointments-overview.component";
+
+export default function Appointment(props: any) {
+  const match = useRouteMatch();
+
+  const widgetConfig = {
```
fatmali: we can wait for #129 to be merged so that you can utilize the config module

src/widgets/appointments/appointments-overview.component.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { getAppointments } from "./appointments.resource";
+import { useCurrentPatient } from "@openmrs/esm-api";
+import dayjs from "dayjs";
+import SummaryCard from "../cards/summary-card.component";
+import SummaryCardFooter from "../cards/summary-card-footer.component";
+import styles from "./appointments-overview.css";
+export default function AppointmentsOverview(props: AppointmentsOverviewProps) {
+  const [patientAppointments, setPatientAppointments] = React.useState(null);
+  const [
+    sLoadingPatient,
```
fatmali:
```suggestion
    isLoadingPatient,
```

src/widgets/appointments/appointments-overview.component.tsx
```diff
@@ -0,0 +1,85 @@
...
+          linkTo={`/patient/${patientUuid}/chart/appointments`}
+        />
+      </SummaryCard>
+    );
+  }
+
+  return restAPIAppointmentsOverview();
+}
+
+type AppointmentsOverviewProps = { match: {} };
```
fatmali: why do you need this if you're using `useRouteMatch`

src/widgets/appointments/appointments.resource.tsx
```diff
@@ -0,0 +1,20 @@
+import { openmrsFetch } from "@openmrs/esm-api";
+import { map } from "rxjs/operators";
```
fatmali: are you using `map`?

drizzentic: Nop, i have cleaned up the code.

src/chart-review/appointments/appointments.component.tsx
drizzentic: Does this mean all the other widgets will have to be refactored to use the config module? If so why don't we merge this then we can have a single ticket to gather for the refactor of all widgets?

jonathandick: We're still figuring out this. In general these widgets need to be coded such that they have no dependencies outside of there repo with the one exception being 'react-router-dom'. That will be a webpack external and thus you can continue to use it's api.

PR #131, opened 2020-02-25
Dashboard import

jonathandick:
This PR introduces two new reusable components:
Widget - this takes a configuration object and imports a component from either core components or form an external library
Dashboard - this takes a configuration object including a layout and an array of widget-names and outputs an "organized" view (using css grid) of the widgets

I have been in discussions with various people about a reusable component which has multiple tabs and this will be in forthcoming PR. 

This PR also includes a new schema declaration which makes use of the above components. 

I am going to close PR #129 as this refactors and changes quite a lot from that PR. I also think that PR is much more challenging to understand. This branches off of the PR #129 branch.


src/chart-review/chart-review.component.tsx
```diff
@@ -1,87 +1,98 @@
...
+
+  function getExternalView(config, name: string): ViewType {
+    let i = config.widgetDefinitions.findIndex(item => item.name === name);
+    let view: ViewType = { name: name, component: null };
+    if (i !== -1) {
+      view.component = () => (
+        <Widget widgetConfig={config.widgetDefinitions[i]} />
+      );
+    }
+    i = config.dashboardDefinitions.findIndex(item => item.name === view);
```
fatmali: hard to understand whats happening here, `item.name === view`, is this a bug?

src/ui-components/core-views.tsx
```diff
@@ -0,0 +1,149 @@
...
+  summaryDashboard: {
+    name: "summaryDashboard",
+    title: "Overview",
+    layout: {
+      columns: 4
+    },
+    widgets: [
+      {
+        name: "conditionsOverview",
+        layout: { columnSpan: 2 }
```
nkimaina: It's nice that the new config allows you to specify the layout. We are really getting there with the configurable patient chart.

src/ui-components/dashboard/dashboard.component.tsx
```diff
@@ -0,0 +1,89 @@
...
+export default function Dashboard(props: DashboardProps) {
+  const [widgets, setWidgets] = React.useState([]);
+
+  React.useEffect(() => {
+    //loadDashboardFromConfig(props.dashboardConfig);
+
+    const widgets = [];
+    let widget;
+    props.dashboardConfig.widgets.forEach((widgetConfig, index) => {
+      widget = <Widget widgetConfig={widgetConfig} />;
```
nkimaina: I think variable widget is not being used. Could this be a repetition? Also using props.dashboardConfig.widgets.map would have resulted in fewer lines of code to maintain, hence fewer places to make mistakes.

src/chart-review/chart-review.component.tsx
jonathandick: there was a problem there. it should have been item.name === name. 
Changed to the following:
`   i = config.dashboardDefinitions.findIndex(
      dashboardDefinition => dashboardDefinition.name === name
    );
`

src/chart-review/allergies/allergies.component.tsx
```diff
@@ -1,12 +1,10 @@
 import React from "react";
```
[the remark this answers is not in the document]
jonathandick: I agree. I'll create ticket to address this in another PR.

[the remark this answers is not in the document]
jonathandick: I agree

src/ui-components/dashboard/dashboard.component.tsx
jonathandick: replaced with this: 
` setWidgets(
      props.dashboardConfig.widgets.map(widgetConfig => {
        widgetConfig.component = () => <Widget widgetConfig={widgetConfig} />;
        return widgetConfig;
      })
    );`

src/chart-review/chart-review.component.tsx
```diff
@@ -1,87 +1,101 @@
...
+          <Dashboard key={name} dashboardConfig={coreDashboards[name]} />
+        )
+      };
+    }
+    return;
+    //TODO: if(coreMultiDashboards[view])
+  }
+
+  function getExternalView(config, name: string): ViewType {
+    let i = config.widgetDefinitions.findIndex(item => item.name === name);
```
fatmali: this can be a little simplified :-)
```suggestion
    const widget = config.widgetDefinitions.find(widget => widget.name === name);
    const dashboard = config.dashboardDefinitions.find(dashboard => dashboard.name === name);
    if(widget){
         return { name, component: () => <Widget widgetConfig={widget} /> }
    } else if(dashboard) {
        return { name, component: () => <Dashboard dashboardConfig={dashboard}/>
    } else {
        return { name, component: null }
    }
    }
```

jonathandick: Agreed. changes made!

src/chart-review/chart-review.component.tsx
```diff
@@ -1,87 +1,101 @@
...
+          <Dashboard key={name} dashboardConfig={coreDashboards[name]} />
+        )
+      };
+    }
+    return;
+    //TODO: if(coreMultiDashboards[view])
+  }
+
+  function getExternalView(config, name: string): ViewType {
+    const widget = config.widgetDefinitions.findIndex(
```
fatmali: `find` and not `findIndex`

fatmali: @jonathandick ?

jonathandick: This was updated.

fatmali:
lgtm cc anyone can approve and merge this in @nkimaina @donaldkibet

src/ui-components/dashboard/dashboard.component.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  return (
+    <div className={styles.container}>
+      <div
+        className={styles.dashboard}
+        style={{ gridTemplateColumns: getColumnsLayoutStyle() || 2 }}
+      >
+        {widgets.length > 0 &&
+          widgets.map((widget, index) => {
+            let W = widget;
```
donaldkibet:
```suggestion
            let W = widget;
```

src/ui-components/dashboard/dashboard.css
```diff
@@ -0,0 +1,33 @@
+.container {
+  margin: 21px;
```
donaldkibet: convert to rem

jonathandick: My understanding is that for margins, you want to use pixels because this will be consistent across devices. If you use REM, you can end up with very small margins which is not desired in this case.

PR #133, opened 2020-02-26
Experimenting with cypress ui testing tool

drizzentic:
Cypress.io allows us to do end to end tests by simulating the process/steps users in a real system would go through while accessing the MF widgets and apps. 
In this branch, I have included sample tests for login and patient chart.


fatmali:
Good work @drizzentic. However, I'm not sure if we should end to end tests inside the patient chart as they are not related to just patient chart but all microfrontends. Perhaps this should be in a separate repo, that is cloned and run during the CI process and not during development?

drizzentic:
> Good work @drizzentic. However, I'm not sure if we should end to end tests inside the patient chart as they are not related to just patient chart but all microfrontends. Perhaps this should be in a separate repo, that is cloned and run during the CI process and not during development?

Sure, I did this as just a way of validating the tool as well as showing how simple we can achieve end to end tests. The strategy to be adopted is yet to be decided on. So when we have it in place then we shall know how to structure and administer this end to end tests. A good suggestion hadn't thought of that.

drizzentic:
> > Good work @drizzentic. However, I'm not sure if we should end to end tests inside the patient chart as they are not related to just patient chart but all microfrontends. Perhaps this should be in a separate repo, that is cloned and run during the CI process and not during development?
> 
> Sure, I did this as just a way of validating the tool as well as showing how simple we can achieve end to end tests. The strategy to be adopted is yet to be decided on. So when we have it in place then we shall know how to structure and administer this end to end tests. A good @fatmali 
 suggestion hadn't thought of that.

drizzentic:
@fatmali During the call I noted a suggestion from the team that we need to isolate the tests in a way they become app-specific. How would we adopt such a strategy? Would it mean that in the designated repository we have app-specific folders?

cypress/integration/login.spec.js
```diff
@@ -0,0 +1,16 @@
+describe("User should be able to login", () => {
```
jonathandick: We should move this test into the login MF and remove from here.

cypress/integration/patientchart/chart.spec.js
```diff
@@ -0,0 +1,47 @@
+describe("Patient chart", function() {
```
jonathandick: Remove the login portion of this test.
The remaining portion of this seems to test patient search. We should this into the esm-patient-home which presently houses the search functionality.

fatmali:
I'd suggest we handle it in a way that can do two types of integration testing i.e
1) internally within a mf, test integration between components
2) across the app, test integration between all mfs

For 1, it would make sense to have the tests inside the mf itself, as all of them are related to the MF.

For 2, it should probably be handled (might be already handled by the root config). Basically these tests would only check if a route displays an mf that it's expected to.

cypress/integration/patientchart/chart.spec.js
drizzentic: One challenge I am still trying to fix is the authentication process. Cypress has a feature called commands which we can use to define the login process and invoke it before any test is run. I tried out a few things and will reorganize the tests accordingly once I manage to get the commands working

drizzentic:
> @drizzentic I'd suggest we handle it in a way that can do two types of integration testing i.e
> 
> 1. internally within a mf, test integration between components
> 2. across the app, test integration between all mfs
> 
> For 1, it would make sense to have the tests inside the mf itself, as @jonathandick suggested because all of them would be related to the mf.
> 
> For 2, it should probably be handled (might be already handled by the root config). Basically these tests would only check if a route displays an mf that it's expected to. This also means that distros would handle their own testing, which is great because it would be difficult for us to test the different configurations that they may have e.g `loginSuccessUrl` etc.

@fatmali Both approaches are actually fine, the only catch would be to ensure that the tests are run only after the MF is deployed on the staging environment(if it passes then it can graduate to prod).

brandones:
@drizzentic Thank you for your work on this. Can we close this old PR, in agreement that it would be great to set up Cypress tests for the whole Single-SPA application?

PR #135, opened 2020-02-28
Tabbed view

jonathandick:
This introduce a new reusable widget for the patient charted: Tabbed View.  See below for an example:

image


src/view-components/tabbed-view/tabbed-view.component.tsx
```diff
@@ -0,0 +1,94 @@
...
+} from "react-router-dom";
+
+import styles from "./tabbed-view.css";
+import { getView, ViewType } from "../view-utils";
+
+function useQuery() {
+  return new URLSearchParams(useLocation().search);
+}
+
+export default function TabbedView(props: any) {
```
donaldkibet: Would be nice to have the props have a type

src/chart-review/chart-review.component.tsx
```diff
@@ -123,7 +77,8 @@ export default function ChartReview(props: any) {
                   >
                     <Link
                       to={
-                        defaultPath + item.path + (tabHistory[item.path] || "")
+                        tabHistory[item.path.substr(1)] ||
```
fatmali: why the index `1` specifically? what does the item at index `1` signify?

src/view-components/view-utils.tsx
```diff
@@ -0,0 +1,76 @@
...
+  return getCoreView(name, defaultPath) || getExternalView(config, name);
+}
+
+export function getCoreView(name: string, defaultPath: string): ViewType {
+  if (coreWidgets[name]) {
+    return coreWidgets[name];
+  }
+  if (coreDashboards[name]) {
+    return {
+      ...coreDashboards[name],
```
fatmali: 👏 👏 getting simpler and simpler :-)

src/chart-review/chart-review.component.tsx
jonathandick: Because the paths have a "/" in front, e.g. "/overview". See https://github.com/openmrs/openmrs-esm-patient-chart/blob/tabbed-view/src/view-components/core-views.tsx#L25 for an example config.

The tabHistory is using the  path without the "/" to store the last route.

src/view-components/view-utils.tsx
jonathandick: Thanks!

PR #136, opened 2020-03-02
Enabled workspace item to be launched from outside

nkimaina:


nkimaina:
Example of how to use it from a 3rd party module
https://github.com/AMPATH/ampath-esm-patient-chart-widgets/blob/master/src/work-space/work-space-controller.ts#L4

nkimaina:
@fatmali @jonathandick @brandones Is this the right approach?

fatmali:
I think we should probably move `newWorkspaceItem` and `getNewWorkspaceItem` to `openmrs-esm-api` since most modules already are using it.

brandones:
Nice work, @nkimaina . Just to dump a few thoughts:

Pros to keeping `newWorkspaceItem` etc in `patient-chart`:
- it is totally conceptually bound up with `patient-chart`, whereas `esm-api` should be kept generic

Pros to moving them to `esm-api`:
- No dynamic import of `patient-chart`
- No circular dependency between `patient-chart` and the custom widget-module

What about moving to a third dependency, `esm-patient-chart-widget-lib` or `esm-patient-chart-widgets`, which provides common stuff that patient chart widgets will want to use? This also avoids the circular dependency without creating the coupling with `esm-api`.

nkimaina:
@fatmali and @brandones 
These are good points.
It seems that as long as we want the patient chart extensible by injecting external components it, then we will need something to coordinate this, that is tightly coupled to the patient chart. 
esm-api already provides us with the current-patient-uuid and object, but it feels like it was built to handle api calls and framework agnostic functionality? I feel like the above functionality is tightly coupled to react and patient-chart, and therefore should reside in an external patient-chart-lib to handle this kind of coordination. 
@fatmali what would be involved in moving the functionality away from patient-chart?

fatmali:
@nkimaina 

1.  We will create the external patient chart lib to be an es module and export the functions out of its entry file.
2. Add the  external patient chart lib to the `import-map.json`
3.  Make the external patient chart lib to the externals array in webpack config for patient chart and all other widgets.

jonathandick:
Hey, to speed things up and given we have only a single api call to add, I would like to just place this in esm-api for now. We can move out later. It'll be a bit simpler to get this started since all the existing repos already use esm-api. Any strong objections to this prelim step?

brandones:
Ok by me

nkimaina:
@jonathandick that's okay. I will close this PR.

PR #137, opened 2020-03-04
Added ability to close tab

nkimaina:


nkimaina:
@fatmali please review

nkimaina:
@jonathandick @brandones @fatmali 
This enables a component launched in the workspace to request to be terminated. This happens when a user has saved a form, or a similar event.

src/workspace/workspace.component.tsx
```diff
@@ -76,20 +84,23 @@ export default function Workspace(props: any) {
...
+                key={i}
+                title={tab.name}
+                style={selectedTab === i ? {} : { display: "none" }}
+              >
+                <tab.component
+                  {...tab.props}
+                  entryStarted={() => setWorkBegan(i)}
+                  entrySubmitted={() => setWorkEnded(i)}
+                  entryCancelled={() => setWorkEnded(i)}
+                  closeComponent={() => onCloseTabRequest(i, tab)}
```
fatmali: why not use the `entryCancelled` prop for this purpose?

nkimaina: I thought about it:
entryCancelled/entrySubmitted is actually to notify the workspace manager, that it is safe to close a certain component hosted in the workspace. The intention might not to unload it from the workspace however. For e.g, after submitting a form, you want to display some information, and have the user close the tab after they are done reading the information. 
However in some case, you want to close(unload) a tab programmatically, and that's why we introduce a new event to close/unload it. 
I hope this makes sense. Let me know if you have any suggestions regarding this.

PR #140, opened 2020-03-06
made the card Head a static options bar

DAN-BIWOTT:


DAN-BIWOTT:
Alright, thank you. I've received a couple of feed-backs relative to this.
Let me make the necessary corrections. Thanks for the link also.

src/workspace/workspace-wrapper.component.css
```diff
@@ -9,15 +9,37 @@
...
+}
+
+@media only screen and (min-width: 769px) {
+  #cardBody {
+    margin-top: 5vh;
+  }
+}
+
+@media only screen and (max-width: 768px) {
+  #cardBody {
```
fatmali: in CSS, the `#` selector is used for `id` attribute while the `.` selector is used for `class` attribute. For consistency, use the class attribute instead as we have across the app.

src/workspace/workspace-wrapper.component.css
```diff
@@ -9,15 +9,37 @@
...
     opacity: 1;
     transform: scale(1);
   }
 }
+
+#optionsBar {
+  width: 100% !important;
+  background-color: #eff1f2;
+  position: fixed;
+  z-index: 2;
```
fatmali: z-index are always a red flag, and becomes extremely difficult to debug as the app grows and you have competing `z-index`. Why are you using the `z-index` here?

DAN-BIWOTT: The input fields scrolled above the title. But it's okay, i'll find a way around it without the z-index. Could you please also review at the next commit?

jonathandick:
Can you please post a picture of what you've done?

DAN-BIWOTT:
Screenshot from 2020-03-10 21-15-50

The workspace title is stationary during scrolls and all scrollbars have been removed. there's also a faint divider between the workspace navbar and vital navbar.

PR #141, opened 2020-03-06
Update core views

jonathandick:
This does a few things:

1. Migrates the core views to use the same structure as the schema.
2. Changes the definitions in core views to reference openmrs-esm-patient-chart-widgets repo rather than importing the widgets directly
3. Updates the schema definition for this module (changes some properties and imports defaults from the core-views file)


src/view-components/view-utils.tsx
```diff
@@ -3,75 +3,54 @@ import React from "react";
 import Dashboard from "./dashboard/dashboard.component";
 import TabbedView from "./tabbed-view/tabbed-view.component";
 
-import { coreWidgets, coreDashboards, coreTabbedViews } from "./core-views";
 import Widget from "./widget/widget.component";
+import {
+  widgetDefinitions as coreWidgetDefinitions,
```
nkimaina: We should probably rename widgetDefinitions to coreWidgetDefinitions, to avoid aliasing every time before use.

src/chart-review/chart-review.component.tsx
```diff
@@ -42,7 +41,7 @@ export default function ChartReview(props: any) {
   }
 
   React.useEffect(() => {
-    const views: ViewType[] = config.primaryNavbar.map(item => {
+    const views: View[] = config.primaryNavbar.map(item => {
       let view = getView(item.view, config, defaultPath + item.path);
       if (view && view.component) item.component = view.component;
```
fatmali: nit: I prefer having if statements with curly braces. No matter how small.

src/view-components/view-utils.test.tsx
```diff
@@ -0,0 +1,128 @@
...
+import "@testing-library/jest-dom/extend-expect";
+
+import { getView } from "./view-utils";
+import {
+  widgetDefinitions as coreWidgetDefinitions,
+  dashboardDefinitions as coreDashboardDefinitions,
+  tabbedViewDefinitions as coreTabbedViewDefinitions
+} from "./core-views";
+
+declare var System;
```
fatmali: mock System  globally in the jest config

src/view-components/view-utils.tsx
jonathandick: Sounds good

src/chart-review/chart-review.component.tsx
jonathandick: I can try to change. It may be that prettier is doing this automatically.

src/view-components/view-utils.test.tsx
jonathandick: done

jonathandick:
Changes addressed. I'm going to merge this.

PR #145, opened 2020-03-16
Parcel config

jonathandick:
We have a requested use case where a widget created in an external library requires the ability to mount a parcel from within that widget. In order to do so, the external library must have access to the mountParcel function from single-spa for the esm-patient-chart application. 

This PR introduces the ability to declare within a config that a widget requires the single-spa mount parcel function via a schema property called "usesSingleSpaContext" . For now, I'm only including the mountParcel function from SingleSpaContext  only because that is only what's required by the use case. Technically, SingleSpaContext also makes available single-spa itself.  

See documentation here about SingleSpaContext: https://single-spa.js.org/docs/ecosystem-react/

### DIFF
--- a/src/view-components/widget/widget.component.tsx
+++ b/src/view-components/widget/widget.component.tsx
@@ -8,7 +8,7 @@ export default function Widget(props) {
 
   React.useEffect(() => {
     loadWidgetFromConfig(props.widgetConfig);
-  }, [props.widgetConfig, loadWidgetFromConfig]);
+  }, [props.widgetConfig]);
 
   function loadWidgetFromConfig(widgetConfig: WidgetConfig) {
     let Component: FunctionComponent<ComponentProps>;
--- a/src/view-components/widget/widget.component.tsx
+++ b/src/view-components/widget/widget.component.tsx
@@ -1,6 +1,6 @@
 import React, { FunctionComponent, useContext } from "react";
 //@ts-ignore The present types for single-spa-react are not updated yet for 2.9 which has SingleSpaContext
-import s, { SingleSpaContext } from "single-spa-react";
+import { SingleSpaContext } from "single-spa-react";
 
 export default function Widget(props) {
   const [widget, setWidget] = React.useState(null);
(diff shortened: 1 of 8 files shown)


src/view-components/widget/widget.test.tsx
```diff
@@ -80,4 +89,21 @@ describe(`<Widget />`, () => {
...
+        widgetConfig={{
+          name: "test",
+          esModule: "@openmrs/esm-patient-chart-widgets",
+          usesSingleSpaContext: true
+        }}
+      />
+    );
+
+    const element = await waitForElement(() => queryByText("Test Widget"));
+    expect(element).not.toBeNull();
```
nkimaina: Is there a way to check that the context has indeed been injected? something like:
```
expect(element.prop. singleSpaContext.mountParcel).not.toBeNull();
```
This will ensure that the spa context will always be passed when people do modifications to the widget component.

src/view-components/widget/widget.component.tsx
```diff
@@ -1,45 +1,60 @@
 import React, { FunctionComponent, useContext } from "react";
-import { AppPropsContext } from "../../app-props-context";
+//@ts-ignore The present types for single-spa-react are not updated yet for 2.9 which has SingleSpaContext
+import s, { SingleSpaContext } from "single-spa-react";
```
fatmali: what's `s`?

src/view-components/widget/widget.test.tsx
jonathandick: I can't figure out how to get props use react testing library. The way we could test this would be to see that if the element returned had the prop "singleSpaContext" with value of {mountParcel: function}. @fatmali do you know to test for props?

src/view-components/widget/widget.component.tsx
jonathandick: s is single-spa-react but we're not use it so I have removed it.

jonathandick:
I'm going to merge this for now. Will update tests once able to .

PR #146, opened 2020-03-16
Spanish translation

kant:


brandones:
Aren't we managing these translations through Transifex? @jonathandick ?

kant:
@brandones Some translations are done in Transifex...

brandones:
But not these ones? Why not?

PR #153, opened 2020-04-20
Mf 172

nkimaina:


nkimaina:
Added scrolling of nav bar header items if items don't fit on the screen. Selecting of an item, then refreshing the page scrolls to that item if it's not in view. Added pager buttons to help scrolling in non-touch devices. 
Screenshot 2020-04-20 at 12 39 28
Screenshot 2020-04-20 at 12 39 13
Screenshot 2020-04-20 at 12 38 57

nkimaina:
Hey @FlorianRappl What do you think of the above? It's to improve user experience on tablets, kind of how material design does it?

FlorianRappl:
I think it looks great. Only concern I have is the size of these buttons. They may need to be slightly larger to have a good / large enough hitbox.

src/chart-review/chart-review.component.tsx
```diff
@@ -28,6 +28,20 @@ export default function ChartReview(props: any) {
   const [selected, setSelected] = React.useState(getInitialTab());
   const [tabHistory, setTabHistory] = React.useState({});
 
+  // NAV HEADER
+  const [navHeaderOverflowed, setNavHeaderOverflowed] = React.useState(false);
```
FlorianRappl, line 32: I think the three states here should be put into a single one using `useReducer`. They are connected and usually not set individually (e.g., once its overflowed its left, right, or both, otherwise its usually none).

Alternatively, a single state using an enum could be good:

```ts
enum HeaderOverflow {
  none = 0,
  left = 1,
  right = 2,
  both = 3,
};
```

where `HeaderOverflow.both` is just there for convenience, as it is `HeaderOverflow.left | HeaderOverflow.right`.

src/chart-review/chart-review.css
```diff
@@ -55,3 +64,19 @@
...
+
+.leftHeaderPagination {
+  padding: 0;
+  background-color: white;
+  cursor: pointer;
+  border: 0;
+  border-right: 1px solid grey !important;
+}
+
+.rightHeaderPagination {
```
FlorianRappl, line 76: This is quite a repetition from `leftHeaderPagination`. Maybe make  rule `.leftHeaderPagination, .rightHeaderPagination { ... }` that contains the shared part?

src/chart-review/chart-review.component.tsx
nkimaina: sounds good

nkimaina:
@FlorianRappl did the requested changes

FlorianRappl:
LGTM!

PR #156, opened 2020-05-06
Update README.md

brandones:
from https://wiki.openmrs.org/display/projects/openmrs-esm-patient-chart

I also added some links and cleaned up the copy a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 [![Build Status](https://travis-ci.org/openmrs/openmrs-esm-patient-chart.svg?branch=master)](https://travis-ci.org/openmrs/openmrs-esm-patient-chart)
 
-# Openmrs ESM Patient Chart
+# OpenMRS ESM Patient Chart
 
 ## What is this?
 
--- a/README.md
+++ b/README.md
@@ -10,9 +10,8 @@ patient's information, such as vitals, demographic and relationships.
 
 ## How do I install this?
 
-Requirements: Java 8, Latest NodeJS & NPM, Latest Git, Latest OpenMRS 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,10 +1,217 @@
...
 
-More info on the [wiki][]
+openmrs-esm-patient-chart is a patient dashboard microfrontend for the 
+OpenMRS SPA. It provides a simple dashboard with cards detailing the 
+patient's information, such as vitals, demographic and relationships.
 
-[wiki]: https://wiki.openmrs.org/display/projects/openmrs-esm-patient-chart
+## How do I install this?
+
+Requirements: Java 8, Latest NodeJS & NPM, Latest Git, Latest OpenMRS
```
FlorianRappl, line 13: Why is Java 8 a requirement?

README.md
```diff
@@ -1,10 +1,217 @@
 [![Build Status](https://travis-ci.org/openmrs/openmrs-esm-patient-chart.svg?branch=master)](https://travis-ci.org/openmrs/openmrs-esm-patient-chart)
 
-Openmrs ESM Patient Chart
-=========
+# Openmrs ESM Patient Chart
```
FlorianRappl, line 3: OpenMRS?

FlorianRappl: (not sure here - sometimes we use something like `openmrs-esm-patient-chart`, but here it seems we want it in title case)

brandones:
```suggestion
# OpenMRS ESM Patient Chart
```

README.md
brandones: It isn't, except for OpenMRS itself. @PermissionError can correct me if I'm wrong.

PR #160, opened 2020-05-18
MF-222: Added ability to specify patient widgets configs

nkimaina:
This allows one to specify configs in the patient chart widgets module, and merge the configs with the patient chart widgets configs. Requires esm-patient-chart-widget module to be released.

### DIFF
--- a/src/chart-review/chart-review.component.tsx
+++ b/src/chart-review/chart-review.component.tsx
@@ -12,8 +12,11 @@ import styles from "./chart-review.css";
 import { useConfig } from "@openmrs/esm-module-config";
 
 import { getView, View } from "../view-components/view-utils";
-import { ChartConfig, Navbar } from "../root.component";
 import { VisitButton } from "@openmrs/esm-patient-chart-widgets";
+import {
+  ChartConfig,
+  Navbar
+} from "../config-schemas/openmrs-esm-patient-chart-schema";
 
 export default function ChartReview(props: any) {
   const match = useRouteMatch();
@@ -22,6 +25,7 @@ export default function ChartReview(props: any) {
   const { patientUuid } = useParams();
   const { view: viewPath } = useParams();
   const config = useConfig<ChartConfig>();
+
   const defaultPath = `/patient/${patientUuid}/chart`;
   const [views, setViews] = React.useState<View[]>([]);
   const [navbarItems, setNavbarItems] = React.useState<Navbar[]>([]);
--- a/src/config-schemas/openmrs-esm-patient-chart-schema.ts
+++ b/src/config-schemas/openmrs-esm-patient-chart-schema.ts
@@ -4,12 +4,7 @@ import {
   coreDashboardDefinitions,
   coreTabbedViewDefinitions
 } from "../view-components/core-views";
-
-import {
-  widgetCommonConfigSchema,
-  WidgetCommonConfig,
-  ExternalPatientChartConfigs
-} from "@openmrs/esm-patient-chart-widgets";
+import { widgetCommonConfigSchema } from "./widget-common-config";
 
 export const esmPatientChartSchema = {
   primaryNavbar: {
@@ -107,9 +102,7 @@ export const esmPatientChartSchema = {
     default: coreTabbedViewDefinitions
   },
   externalPatientChartConfigs: {
-    dictionaryElements: {
-      skipValidations: true
-    }
+    skipValidations: true
   },
   widgetCommonConfig: {
     ...widgetCommonConfigSchema
@@ -146,8 +139,6 @@ export type ChartConfig = {
     title: string;
     navbar: Navbar;
   }[];
-  externalPatientChartConfigs: ExternalPatientChartConfigs;
-  widgetCommonConfig: WidgetCommonConfig;
 };
 
 export type Navbar = { label: string; path: string; view: string };
--- a/src/config-schemas/openmrs-esm-patient-chart-schema.ts
+++ b/src/config-schemas/openmrs-esm-patient-chart-schema.ts
@@ -3,7 +3,12 @@ import {
   coreWidgetDefinitions,
   coreDashboardDefinitions,
   coreTabbedViewDefinitions
-} from "./view-components/core-views";
+} from "../view-components/core-views";
+
+import {
+  ChartWidgetsConfig,
+  chartWidgetsConfigSchemas
+} from "@openmrs/esm-patient-chart-widgets";
 
 export const esmPatientChartSchema = {
   primaryNavbar: {
@@ -99,5 +104,40 @@ export const esmPatientChartSchema = {
       }
     },
     default: coreTabbedViewDefinitions
-  }
+  },
+  ...chartWidgetsConfigSchemas
 };
+
+export type ChartConfig = ChartWidgetsConfig & {
+  primaryNavbar: Navbar[];
+  widgetDefinitions: {
+    name: string;
+    esModule?: string;
+    label?: string;
+    path?: string;
+  };
+
+  dashboardDefinitions: {
+    name: string;
+    title: string;
+    layout: { columns: number };
+    widgets: {
+      name: string;
+      esModule: string;
+      label: string;
+      path: string;
+      layout: {
+        rowSpan: number;
+        columnSpan: number;
+      };
+    }[];
+  };
+
+  tabbedDashboardDefinitions: {
+    name: string;
+    title: string;
+    navbar: Navbar;
+  }[];
+};
+
+export type Navbar = { label: string; path: string; view: string };
--- a/src/config-schemas/openmrs-esm-patient-chart-schema.ts
+++ b/src/config-schemas/openmrs-esm-patient-chart-schema.ts
@@ -5,7 +5,11 @@ import {
   coreTabbedViewDefinitions
 } from "../view-components/core-views";
 
-import { ChartWidgetsConfig } from "@openmrs/esm-patient-chart-widgets";
+import {
+  widgetCommonConfigSchema,
+  WidgetCommonConfig,
+  ExternalPatientChartConfigs
+} from "@openmrs/esm-patient-chart-widgets";
 
 export const esmPatientChartSchema = {
   primaryNavbar: {
@@ -102,10 +106,13 @@ export const esmPatientChartSchema = {
     },
     default: coreTabbedViewDefinitions
   },
-  widgetsConfigs: {
+  externalPatientChartConfigs: {
     dictionaryElements: {
       skipValidations: true
     }
+  },
+  widgetCommonConfig: {
+    ...widgetCommonConfigSchema
   }
 };
 
@@ -139,7 +146,8 @@ export type ChartConfig = {
     title: string;
     navbar: Navbar;
   }[];
-  widgetsConfigs: ChartWidgetsConfig;
+  externalPatientChartConfigs: ExternalPatientChartConfigs;
+  widgetCommonConfig: WidgetCommonConfig;
 };
 
 export type Navbar = { label: string; path: string; view: string };
--- a/src/config-schemas/openmrs-esm-patient-chart-schema.ts
+++ b/src/config-schemas/openmrs-esm-patient-chart-schema.ts
@@ -5,10 +5,7 @@ import {
   coreTabbedViewDefinitions
 } from "../view-components/core-views";
 
-import {
-  ChartWidgetsConfig,
-  chartWidgetsConfigSchemas
-} from "@openmrs/esm-patient-chart-widgets";
+import { ChartWidgetsConfig } from "@openmrs/esm-patient-chart-widgets";
 
 export const esmPatientChartSchema = {
   primaryNavbar: {
@@ -105,10 +102,14 @@ export const esmPatientChartSchema = {
     },
     default: coreTabbedViewDefinitions
   },
-  ...chartWidgetsConfigSchemas
+  widgetsConfigs: {
+    dictionaryElements: {
+      skipValidations: true
+    }
+  }
 };
 
-export type ChartConfig = ChartWidgetsConfig & {
+export type ChartConfig = {
   primaryNavbar: Navbar[];
   widgetDefinitions: {
     name: string;
@@ -138,6 +139,7 @@ export type ChartConfig = ChartWidgetsConfig & {
     title: string;
     navbar: Navbar;
   }[];
+  widgetsConfigs: ChartWidgetsConfig;
 };
 
 export type Navbar = { label: string; path: string; view: string };
--- a/src/config-schemas/openmrs-esm-patient-chart-schema.ts
+++ b/src/config-schemas/openmrs-esm-patient-chart-schema.ts
@@ -101,9 +101,6 @@ export const esmPatientChartSchema = {
     },
     default: coreTabbedViewDefinitions
   },
-  externalPatientChartConfigs: {
-    skipValidations: true
-  },
   widgetCommonConfig: {
     ...widgetCommonConfigSchema
   }
--- a/src/config-schemas/vitals-config-schema.ts
+++ b/src/config-schemas/vitals-config-schema.ts
@@ -1,0 +1,49 @@
+import { validators } from "@openmrs/esm-module-config";
+
+export const vitalsConfigSchema = {
+  triageFormConfig: {
+    formUuid: {
+      validators: [validators.isString],
+      default: "a000cb34-9ec1-4344-a1c8-f692232f6edd"
+    },
+    encounterTypeUuid: {
+      validators: [validators.isString],
+      default: "67a71486-1a54-468f-ac3e-7091a9a79584"
+    },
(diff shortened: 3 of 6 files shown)


brandones:
Nice, thanks for taking a crack at this! I'd been putting off thinking about it because it might be kind of tricky.

This approach only works for configuring widgets that are in patient-chart-widgets, which seems like a problem.

Another approach would be to make widgets responsible for managing their own configurations directly with esm-module-config. But that raises the question of how to provide different configurations for the same widget being used in different places. e.g. maybe there's a vitals widget that needs to be shown on both the nurse and doctor dashboards but with different configurations.

Another approach would be to add support to esm-module-config for unvalidated config subtrees. e.g. your config schema could define a key `widgetConfig` for which esm-module-config doesn't attempt any kind of validation. Then esm-patient-chart could just pass the resultant config object for that key through to the widget, and let the widget worry about whether or not it's valid. We could also expose a validation API for esm-module-config, so that the widgets could validate the configs they receive from esm-patient-chart.

Yeah, I think I like this last approach best. Thoughts?

A corollary challenge, however, is that MFEs are singletons -- even if we have a config API that has natural support for configuring widgets differently in different places, it's not clear to me how we should have the widget actually use those different configs. Also, I glean from [this comment](https://github.com/single-spa/single-spa/issues/525#issue-606491196) that SIngle-SPA does not support displaying the same MFE in multiple places in the same route, so that's sort of a hard restriction on widget reuse.

nkimaina:
@brandones  These are good points. I feel like we need a call to discuss the various approaches. What do you think?

nkimaina:
After thinking about it some more: Your last approach could work and could solve a lot of unforeseen use-cases. We can generalize it as:
1. Provide a validation api, that allows one to validate the schema just it time before using it. The advantage, widgets will not need to deal with providing the config before hand.  The trade-off here is: person providing the config won't know whether the config is valid until the component actually attempts to use it.
2. Generalize config identification. Can be done by exposing api to get config by identifier. 

The results of the above would look like: 
 `useConfig(<identifier> , <schemaObject>)`

What do you think?

brandones:
Cool! But I was thinking of an API like

- `useConfig<T>(<identifier>): T` and
- `validate(<scemaObject>, <config: T>): void`

So that patient-chart can fetch each widget's config using `useConfig`, pass the config to the widget, and then the widget can run `validate` on the received config during initialization. This allows the widget to own the config schema, and for validation to happen at widget initialization time, meaning no errors get to production.

nkimaina:
How difficult is it to make these changes? It's something that I could help you if you point me to the right direction.

nkimaina:
Would a third party module, say ampath-patient-widget-module, also use this pattern? Will every widget have the key widgetConfig?
e.g to define the config for a hiv-drug-summary widget, could I do:  
`{
widgetConfig: {
hivSummary: {
...
}
}
}`

brandones:
The changes to `esm-module-config` should be pretty straightforward. To sketch them:

### skipValidation

The config schema syntax we want to allow is something like
```js
{
  foo: {
    skipValidation: true
  }
}
```

- add a test to `module-config.test.ts` which checks that getConfig uses but does not validate elements with `skipValidation: true`. This test should use a schema like the one above, pass in a config that has `foo: { what: "ever", you: { want: true } }` or something, and check that the resultant `config.foo` looks like that passed-in config.
- make the test pass, 
- make another test, this one to check if skipValidation works within an array schema (defined by arrayElements)
- make it pass

### validate

Just wrap `validateConfig`. Write a test or two.

### useConfig(<identifier>)

Whoops, I think I was wrong about this being necessary. I think the widget configs will just be obtained by `patient-chart` using the normal `useConfig`.

## Schema structure

I'm thinking that the patient chart schema will have something like

```js
  widgetDefinitions: {
    arrayElements: {
      name: { validators: [isString]},
      ...
      config: { skipValidation: true }
  };
```

And then the resulting config will have arbitrary objects at `widgetDefintions[i].config`, and it will just pass that object to its respective widget. That widget will validate that object against a schema that only contains what it actually needs to configure --

```js
// vitals config schema
{
  concepts: {
    arrayElements: {
      validators: [isConceptReferenceTerm]
    }
  },
  maxEncountersToDisplay: { validators: [isInteger, withinRange(1, 50)]
}
```

I've taken some creative license but hopefully you get the gist :)

src/config-schemas/openmrs-esm-patient-chart-schema.ts
```diff
@@ -99,5 +101,45 @@ export const esmPatientChartSchema = {
       }
     },
     default: coreTabbedViewDefinitions
+  },
+  widgetsConfigs: {
+    dictionaryElements: {
+      skipValidations: true
+    }
```
brandones, line 108: Oh wow, I'm realizing now that I read this that I asked you to do a bunch of totally unnecessary work. I'm really sorry, @nkimaina . I think the following would be better (putting it up with `widgetDefinitions`):

```js
widgetDefintions: {
  arrayElements: {
      name: { validators: [validators.isString] },
      esModule: { validators: [validators.isString] },
      usesSingleSpaContext: { validators: [validators.isBoolean] }
      config: { validators: [validators.isObject] }
    },
    default: coreWidgetDefinitions
  }
}
```

nkimaina: No worries. I thought about it some more. Correct me if am wrong:
1. With the above approach, I would need to override a lot of default configs just to change a single config entry, like temperature concept for instance. I would need to override configs for how the widget is rendered, which module it comes from etc. As the widget definition configs changes over time, I would need to be aware of those changes and provide overrides for them as well. 
2. Some of the configs are used across multiple widgets. For example, the vitals concepts would be used by the vitals-summary widget, as well as vitals form, some external widget etc. 

In fact, we might not have done extra work, but might have provided a way to provide external configs. I would prefer to rename the 'widgetConfigs' to 'externalConfigs'. The existing widgetDefinitions are good enough as they describe how to load and render widgets. The other configs might describe how to fetch data for the widgets e.t.c. 
Generally, I think we need to prefer flat/shallow configs, and try as much as possible to avoid deeply nested configs as it just makes it complicated to patch/override 'leaf' configs, and makes them harder to read/work with. 
What are you thoughts on all this?

brandones: >     1. With the above approach, I would need to override a lot of default configs just to change a single config entry, like temperature concept for instance. I would need to override configs for how the widget is rendered, which module it comes from etc. As the widget definition configs changes over time, I would need to be aware of those changes and provide overrides for them as well.

Yes, this is a really good observation.

>     2. Some of the configs are used across multiple widgets. For example, the vitals concepts would be used by the vitals-summary widget, as well as vitals form, some external widget etc.
> 
> In fact, we might not have done extra work, but might have provided a way to provide external configs. I would prefer to rename the 'widgetConfigs' to 'externalConfigs'.

Ooh, interesting, yes. Can we call it `widgetCommonConfig`? But it still can be like
```js
widgetCommonConfig: { validators: [isObject] }
```

an example config would be
```json
"widgetCommonConfig": {
  "temperatureConcept": "1jflaw-jeir2039-jfqa"
}
```

> The existing widgetDefinitions are good enough as they describe how to load and render widgets. The other configs might describe how to fetch data for the widgets e.t.c.

I imagine that at some point we will want widget-specific config. e.g.

```json
"widgetDefinitions": [
  {
    "name": "vitals-summary",
    ...
    "config": {
      "numEncounters": 6
    }
  }
]

> Generally, I think we need to prefer flat/shallow configs, and try as much as possible to avoid deeply nested configs as it just makes it complicated to patch/override 'leaf' configs, and makes them harder to read/work with.

Note that configs are "deep merged," so
```js
provide({ a: { b: { c: 0 }, d: { e: 0 } } })
provide({ a: { b: { c: 1 } } })
```
results in
```js
{ a: { b: { c: 1 }, d: { e: 0 } } }
```

However, config arrays are not merged in any way, so in this case your point stands. You've made a solid case for a common widget config.

In the general case, as with any denormalized structure, "flattening" requires the introduction of references (e.g. foreign keys) that add a layer of indirection and make things harder to think about. I think the normalized/nested approach should be preferred, except in cases where the same config information is needed by other parts of the config. I could imagine introducing some kind of support for array merging, as well. Once there's a really solid use case.

nkimaina: sounds good. I have added the common config property. I still think it's a good idea to allow one to inject extrnal configs without being restricted by current patient chart schema. I have introduced the property 'externalPatientChartConfigs' that they can use for this purpose.

brandones: How would that be used?

src/config-schemas/openmrs-esm-patient-chart-schema.ts
```diff
@@ -3,7 +3,13 @@ import {
...
   coreDashboardDefinitions,
   coreTabbedViewDefinitions
-} from "./view-components/core-views";
+} from "../view-components/core-views";
+
+import {
+  widgetCommonConfigSchema,
+  WidgetCommonConfig,
+  ExternalPatientChartConfigs
+} from "@openmrs/esm-patient-chart-widgets";
```
brandones, line 12: The point of the mechanism to pass config objects from `patient-chart` to widgets is to avoid any dependency on `patient-chart-widgets`. Can you remove this dependency? Those things should live here in `patient-chart`.

src/config-schemas/openmrs-esm-patient-chart-schema.ts
nkimaina: I am trying to use it here:
https://github.com/AMPATH/ampath-esm-patient-chart-widgets/blob/4e6e81092d04c3b5ea10249a65c2f068b2b78212/src/config/programs.schema.ts#L2

and loading the config here:
https://github.com/AMPATH/ampath-esm-root-config/blob/0cc6bbe9d026ea1b7cdfafb45f275ed21a38ff8e/src/configs/programs.config.json#L1

The above configs are used by multiple widgets.

src/config-schemas/openmrs-esm-patient-chart-schema.ts
nkimaina: I have moved them over. It's introduced separation between Types for the config, and Schema for the config. That reason is patient-chart imports patient-chart-widget, and therefore it's inevitable. They are defined in 2 different modules, but I guess we can leave with that.

src/chart-review/chart-review.component.tsx
```diff
@@ -12,8 +12,11 @@ import styles from "./chart-review.css";
 import { useConfig } from "@openmrs/esm-module-config";
 
 import { getView, View } from "../view-components/view-utils";
-import { ChartConfig, Navbar } from "../root.component";
 import { VisitButton } from "@openmrs/esm-patient-chart-widgets";
```
nkimaina, line 15: @brandones just to bring to your attention that I think it's inevitable that patient-chart will be tightly coupled to the patient-chart-widgets. As you can see the import happens here for this VisitButton, and I think we are going to see more of this as the app grows.

src/config-schemas/openmrs-esm-patient-chart-schema.ts
brandones: patient-chart statically imports patient-chart-widgets? Where? It shouldn't...

src/chart-review/chart-review.component.tsx
brandones: Well, for one, the system for managing visits in the patient chart was never really designed, so I would expect it to change.

And two, I think this is an implementation mistake. These Visit* things are not widgets. They are components of the patient chart. They should be part of the patient chart.

I'm confident we can maintain this decoupling, and that doing so will keep the code cleaner and the system easier to understand, maintain, and extend.

nkimaina: I agree. We need distinguish the two very clearly. The documentation will go a long way in helping with it. It also needs to come out very clearly why the two modules should not be tightly coupled, and the trade-offs involved. Until now, I thought it was intentional that patient-chart and patient-chart-widget are tightly coupled, but managed in different repos to make it possible/encourage these widgets to be re-used and extended in 3rd party patient-chart-widgets modules. I hope we continue this discussion about the current design, because as you can see it is easy to mis-understand the design, and can get out of hand real quick. 
That said, what's pending right now? I was hoping to address the issue of loading configs for 3rd party configs in this PR.

brandones: Continuing the conversation on Slack....

src/config-schemas/openmrs-esm-patient-chart-schema.ts
brandones: Why is this different from `widgetCommonConfig`?

brandones: And this word "external"... what is this config external to?

nkimaina: > And this word "external"... what is this config external to?

It means it's a config that will be used by distribution specific esm-widget i.e. will not be used by either the patient-chart or patient-chart-widget. We rely on patient-chart to load these configs for us, because these 'distribution specific' widgets don't have an entry point, so it's a bit tricky to set-up. Anyways, we can remove it in this PR but this is kind of a blocker for me as I am trying to work on integrating AMPATH specific widgets right now, and I can't do it if I don't have a way to load configs for it.

nkimaina: > Why is this different from `widgetCommonConfig`?

It's different because these are provided by 3rd party (patient chart widget modules created by different distributions). For e.g in the example that I shared, it shows how to configure a way to group forms by program. It's an approach that I have seen being used in Kenya, but may not be true for every other distribution/implementation in openmrs. We could add it though, if anyone else is/will be interested in using this approach.

brandones: Right, I think you would just include those configs in the widget-specific config:
```
"widgetDefinitions": [
  {
    "name": "pacman",
    "esModule": "@ampath/esm-pacman-widget",
    "config": {
      "numGhosts": 4
    }
  }
]
```

src/config-schemas/widget-common-config.ts
```diff
@@ -0,0 +1,5 @@
+import { vitalsConfigSchema } from "./vitals-config-schema";
+
+export const widgetCommonConfigSchema = {
+  vitalsConfig: vitalsConfigSchema
+};
```
brandones, line 5: Wait, what? This isn't common to widgets at all. This is vials-specific. The vitals widget schema shouldn't exist in patient-chart, it should be in patient-chart-widgets, with the vitals widget.

src/config-schemas/vitals-config-schema.ts
```diff
@@ -0,0 +1,49 @@
+import { validators } from "@openmrs/esm-module-config";
+
+export const vitalsConfigSchema = {
+  triageFormConfig: {
```
brandones, line 4: What's the triage form?

brandones, line 4: What's the triage form?

src/config-schemas/widget-common-config.ts
```diff
@@ -0,0 +1,5 @@
+import { vitalsConfigSchema } from "./vitals-config-schema";
+
+export const widgetCommonConfigSchema = {
+  vitalsConfig: vitalsConfigSchema
```
brandones, line 4: Wait, what? This isn't common to widgets at all. This is vials-specific. The vitals widget schema shouldn't exist in patient-chart, it should be in patient-chart-widgets, with the vitals widget.

src/config-schemas/vitals-config-schema.ts
nkimaina: It's the encounter type and formuuid that is used to with the vitals form. Right now it's hardcoded and hence the form can't save in other distributions.

src/config-schemas/widget-common-config.ts
nkimaina: I had it in the patient-chart-widget, but you said that I move it over to patient-chart. I can move it over. It's used by multiple widgets and that's why it's in the common configs. 
The approach that you wan't us to use the merging feature will result in duplication in the patient chart config i.e 
```
widgetConfig: [{
name: "vitals-summary",
commonConfig: {
"vitalsConfg": { concepts...}
},
{
name: "vitals-detail",
commonConfig: {
"vitalsConfg": { concepts...}
},
{
name: "vitals-overview",
commonConfig: {
"vitalsConfg": { concepts...}
},
{
name: "vitals-form",
commonConfig: {
"vitalsConfg": { concepts...}
}
]
```

See how the merging solution is repetitive?

nkimaina: Would it be easier if you just did this feature or some else? I think we are not reaching an understanding and it's no longer fun to work on this feature. Some of these issues that are being raised came-up earlier on and I thought we reached an understanding?

PR #162, opened 2020-05-28
Decentralized Activation

FlorianRappl:
Refactored to use `decentralized-activation`.

**Important**:
The module name has to be renamed with a suffix of `-app`. As outlined last week we should also rename the repository then.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,9 @@
   "version": "1.0.0",
   "description": "Patient dashboard microfrontend for the OpenMRS SPA",
   "main": "dist/openmrs-esm-patient-chart.js",
+  "types": "src/index.ts",
+  "license": "MPL-2.0",
+  "homepage": "https://github.com/openmrs/openmrs-esm-patient-chart#readme",
   "scripts": {
     "start": "webpack-dev-server --host 0.0.0.0",
     "start-https": "webpack-dev-server --https",
@@ -31,7 +34,6 @@
       "pre-commit": "pretty-quick --staged && concurrently npm:lint npm:typescript npm:coverage"
     }
   },
-  "types": "src/openmrs-esm-patient-chart.tsx",
   "devDependencies": {
     "@babel/compat-data": "~7.9.0",
     "@babel/core": "^7.8.7",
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -0,0 +1,19 @@
...
+  "../translations",
+  false,
+  /.json$/,
+  "lazy"
+);
+
+function setupOpenMRS() {
+  return {
+    lifecycle: () => import("./openmrs-esm-patient-chart"),
+    activate: location => routeRegex(/^patient\/.+\/chart/, location)
```
brandones, line 15: Perhaps we can get rid of the single-spa dependency by taking advantage of strings as activation functions and lists of activation functions (since [single-spa 5.3.0](https://github.com/single-spa/single-spa/releases/tag/v5.3.0))?

```suggestion
    activate: ["patient-chart", "patient_chart"]
```

package.json
```diff
@@ -86,18 +90,11 @@
     "@openmrs/esm-api": "^1.3.0",
     "@openmrs/esm-module-config": "^1.2.0",
     "@openmrs/esm-patient-chart-widgets": "^1.0.2",
-    "@types/lodash-es": "^4.17.3",
-    "@types/react-router": "^5.1.4",
+    "@openmrs/esm-root-config": "^1.2.0",
```
brandones, line 93: I reckon we should be able to get rid of this dependency for all these PRs now?

brandones: Or should we do that later?

FlorianRappl: Right now we cannot get rid of it. It transports the helper functions.

We'd need to find a new location for the two helper functions (or just copy them).

src/index.ts
FlorianRappl: That would be cool. Do these strings also work with our base route? We are not at "/", but rather at "/openmrs/spa".

brandones: Oh dang, that's right... perhaps the initial script should do something like
```
           function preprocessActivator(activator) {
             if (Array.isArray(activator)) {
               return activator.map(preprocessActivator);
             } else if (typeof activator === "string") {
               return window.getOpenmrsSpaBase() + activator;
             } else {
               return activator
             }
           }
           registerApplication(appName, result.lifecycle, preprocessActivator(result.activate));
```

What do you think?

FlorianRappl: Looks great to me!

I'll create a new PR for the openmrs-module-spa repo; I think the first approach with the `preprocessActivator` is the way to go.

I'd also include the translations from the openmrs-esm-root-config. Then its time to retire the repo. It served its purpose.

brandones:
Thanks @FlorianRappl ! I'll leave these to you to merge since I'm not sure if they need to be coordinated with other changes to not break CI.

PR #167, opened 2020-06-12
Fix the package name and version

mddubey:
As part of this PR following things are done:
- [x] Change the package name to be `@openmrs/esm-patient-chart-app`. 
- [x] Fix the version to be 0.0.1.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
-  "name": "openmrs-esm-patient-chart-app",
-  "version": "1.0.0",
+  "name": "@openmrs/esm-patient-chart-app",
+  "version": "0.0.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


mddubey:
Hello @FlorianRappl,
Thanks for your response. We discussed this in the MF dev standup on Thursday. We decided to go ahead with `0.0.1`, since it is not yet stable and production ready. We decided to not make a `major release`(1.x.x) yet, as it would mean it is stable and production ready.

FlorianRappl:
Yes but starting with only a patch level is also not right. Usually for a first release that is pre-prod you go with 0.1.0.

mddubey:
Thanks @FlorianRappl for the comment.

I have changed the version to be `0.1.0`. I don't think we discussed any specific version, the idea was not to go with a major release. We just discussed `0.0.1` as one example.

PR #169, opened 2020-06-25
MF-222  Create configurations for vitals, height and weight widgets

brandones:
https://issues.openmrs.org/browse/MF-222

see https://github.com/openmrs/openmrs-esm-patient-chart-widgets/pull/115

depends on https://github.com/openmrs/openmrs-esm-module-config/pull/49


brandones:
Thanks @FlorianRappl . That makes sense to me. And actually it helped me catch an incorrect typing in the schema declaration file. I've left the shorthand syntax in a few places where I think it's easier to read, e.g. `useState<JSX.Element[]>([])`.

PR #170, opened 2020-07-07
MF-247: Configuration and translations for the immunization widget.

mddubey:
As part of this PR following things are done:
- [x] Add translations for Immunization Widget, Patient Banner and Visit Dashboards.
- [x] Change the Nav bar to have translation keys which could be used to provide translations.


translations/en.json
```diff
@@ -4,5 +4,62 @@
   "Medications - current": "Medications - current",
   "Medications - past": "Medications - past",
   "Since": "Since",
-  "care programs": "Care Programs"
+  "care programs": "Care Programs",
+  "Immunizations": "Immunizations",
+  "vaccine": "VACCINE",
+  "recent vaccination": "RECENT VACCINATION",
```
FlorianRappl, line 10: Things like only uppercase should be enforced via styling rules - not the translation phrases.

mddubey: @FlorianRappl, This is actually the translation we are putting for the vaccine and recent vaccinations. However, we will make sure to use `uppercase` at the places this text is used.

FlorianRappl: Yes sure, but the translation is rather "recent vaccination", right @mddubey ? If you want to make it uppercase then this is only for design purposes - which is why CSS exist.

PR #171, opened 2020-07-13
Add attachments overview component to patient dashboard tab

ivange94:


FlorianRappl:
I guess we need to update the type defs here?

The build is failing:

```plaintext
src/view-components/core-views.tsx:282:15 - error TS2322: Type '{ columnSpan: number; }' is not assignable to type '{ columns: number; }'.
  Object literal may only specify known properties, and 'columnSpan' does not exist in type '{ columns: number; }'.
282     layout: { columnSpan: 1 }
                  ~~~~~~~~~~~~~
  src/view-components/dashboard/dashboard.component.tsx:65:3
    65   layout?: {
         ~~~~~~
    The expected type comes from property 'layout' which is declared here on type 'DashboardConfig'
```

ivange94:
> I guess we need to update the type defs here?
> 
> The build is failing:
> 
> ```
> src/view-components/core-views.tsx:282:15 - error TS2322: Type '{ columnSpan: number; }' is not assignable to type '{ columns: number; }'.
>   Object literal may only specify known properties, and 'columnSpan' does not exist in type '{ columns: number; }'.
> 282     layout: { columnSpan: 1 }
>                   ~~~~~~~~~~~~~
>   src/view-components/dashboard/dashboard.component.tsx:65:3
>     65   layout?: {
>          ~~~~~~
>     The expected type comes from property 'layout' which is declared here on type 'DashboardConfig'
> ```
> 
> cc @ivange94

I think this was later changed from `columnSpan` to just `column`. I learned those steps during my on-boarding with @jonathandick but I don't fully understand what it does though. Because despite having that error, everything seemed to work fine on the browser. Do you mind explaining to me why I need to make that change in 3 places just to get that widget appear as a tab on the summary page or is there a doc that explains this?

The error should be gone now though.

FlorianRappl:
> Do you mind explaining to me why I need to make that change in 3 places just to get that widget appear as a tab on the summary page or is there a doc that explains this?

I did not create `patient-chart` nor `patient-chart-widgets` (nor the configuration for it). So I cannot give you any reasoning here. All I can tell is that this will change more or less *drastically* soon and that

PR #172, opened 2020-07-14
MF-219: Add chart summary translations

denniskigen:
https://issues.openmrs.org/browse/MF-219

This PR introduces translation keys for the patient chart summary widgets. It also discards the `babel-plugin-i18next-extract` plugin.

Important to note that the keys are specified in **camelCase** format.


brandones:
@denniskigen  Sorry if we've answered this question before, but why aren't we using camelCase keys here?

denniskigen:
> @denniskigen Sorry if we've answered this question before, but why aren't we using camelCase keys here?

Primarily - it's because tests that made TextMatcher-based assertions would fail when using camelCase keys.

brandones:
Oh that's right. And we weren't able to figure out how to make react-i18next return the default string instead of the key in the tests?

denniskigen:
> Oh that's right. And we weren't able to figure out how to make react-i18next return the default string instead of the key in the tests?

Yeah

brandones:
> And if we did, that would mean a large scale refactor to change all of those assertions.

Don't all the lookups use the default string (which should equal the English translation)?

denniskigen:
> > And if we did, that would mean a large scale refactor to change all of those assertions.
> 
> Don't all the lookups use the default string (which should equal the English translation)?

Oh yeah, you're right.

brandones:
Actually, I just [added a test](https://github.com/openmrs/openmrs-esm-login/pull/82) to esm-login demonstrating that translation works as expected there—the keys are camelCase but the tests can query by the default string. Maybe look at the configuration of that package to see how i18next should be set up for tests? Probably something about the mock?

denniskigen:
> Actually, I just [added a test](https://github.com/openmrs/openmrs-esm-login/pull/82) to esm-login demonstrating that translation works as expected there—the keys are camelCase but the tests can query by the default string. Maybe look at the configuration of that package to see how i18next should be set up for tests? Probably something about the mock?

So camelCased keys work fine when using the `Trans` component. The problem occurs when using the `useTranslation` hook. We have many instances where the `useTranslation` hook is a better fit than the `Trans` component. Consider this snippet from [RecordDetails](https://github.com/openmrs/openmrs-esm-patient-chart-widgets/blob/master/src/ui-components/cards/record-details-card.component.tsx) for example:

```js
export default function RecordDetails(props: DetailsProps) {

  return (
    <SummaryCard
      name="Details"
      styles={{
        width: "100%",
        backgroundColor: "var(--omrs-color-bg-medium-contrast)"
      }}
    >
      <div
        style={props.styles}
        className={`omrs-type-body-regular ${styles.detailsCard}`}
      >
        {props.children}
      </div>
    </SummaryCard>
  );
}
```

It appears to me that the `useTranslation` hook would the appropriate tool for translating the "Details" header. Wrapping the string in a `t` function would give you this: `<Summary Card name={t("Details")} ...>`.

If we use `t{"details", "Details"}` then a test that asserts for the presence of "Details" would fail.

brandones:
Oh yeah, that's right! Thank you, I'm sorry I'd forgotten that detail.

But it looks like we can circumvent that problem by fixing the mock. See https://github.com/openmrs/openmrs-esm-login/blob/5c6954967c4f6291aca9462c88e8317b5c6ba7c4/__mocks__/react-i18next.js#L38

instead of

```javascript
useMock.t = (k) => k;
```

it could be something like
```javascript
useMock.t = (k, o) => o?.defaultValue || (typeof o === "string" ? o : k);
```

(this reflects the API of `t`, which is kind of confusing—see the [API docs](https://www.i18next.com/overview/api#t) vs the [Essentials page](https://www.i18next.com/translation-function/essentials#passing-a-default-value), and related [Issue](https://github.com/i18next/i18next-scanner/issues/102)).

denniskigen:
Great, just tested it out and seems to work great. I've updated the PR.

brandones:
Awesome. Okay, I reviewed the widgets repo PR. The only big thing is that, as I commented [there](https://github.com/openmrs/openmrs-esm-patient-chart-widgets/pull/124#discussion_r465821011), we should use CSS text-transform for capitalization; we should definitely not manage different translations (`_upper` and `_lower`) just for capitalization.

Thanks so much for all your work on this. 46 files worth of string changes are a lot.

PR #179, opened 2020-09-24
Enable widgets as extensions and host the medications/drugorder widget via this mechanism

manuelroemer:
Enables the patient chart to alternatively render widgets via the extension system. This is an alternative to the current way and does *not* inhibit existing functionality.

For now, the configuration is kept rather simple. Instead of specifying an `esModule` in the widget configuration, one can instead define an `extensionSlotName`. If done, the patient chart will render an extension slot with that name into which other modules can then place the widget itself.

This PR also makes the medication widget use this extension system. For the moment, nothing will be displayed, but that will change as soon as the [drugorder widget](https://github.com/openmrs/openmrs-esm-drugorder) is deployed.

### DIFF
--- a/src/view-components/core-views.tsx
+++ b/src/view-components/core-views.tsx
@@ -23,12 +23,12 @@ export const coreWidgetDefinitions: WidgetConfig[] = [
 
   {
     name: "MedicationsOverview",
-    esModule: "@openmrs/esm-patient-chart-widgets"
+    extensionSlotName: "patient-chart-dashboard-medications"
   },
 
   {
     name: "MedicationsSummary",
-    esModule: "@openmrs/esm-patient-chart-widgets"
+    extensionSlotName: "patient-chart-dashboard-medications"
   },
 
   {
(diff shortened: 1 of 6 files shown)


src/view-components/core-views.tsx
```diff
@@ -23,12 +23,12 @@ export const coreWidgetDefinitions: WidgetConfig[] = [
 
   {
     name: "MedicationsOverview",
-    esModule: "@openmrs/esm-patient-chart-widgets"
+    extensionSlotName: "patient-chart-dashboard-medications"
```
manuelroemer, line 26: I'm not sure whether we want to include this already because nothing will be displayed until the drugorder widget is deployed. We can also remove/delay these changes here until that happened.
What do you think? 

@FlorianRappl

FlorianRappl: I'd say lets keep it. We'll bring that online soon!

FlorianRappl (approved):
You rock!

Great addition with the slot in the config. Makes it easy to transition.

PR #182, opened 2020-11-09
Context workspace integration

manuelroemer:
There will be additional changes coming in the future (e.g. regarding styling). For now, this PR simply enables the extension context mechanism for prototyping/development.


src/root.component.tsx
```diff
@@ -8,10 +8,33 @@ import styles from "./root.css";
...
+    currentWorkspaceExtensionSlot,
+    setCurrentWorkspaceExtensionSlot
+  ] = useState<React.FC<ExtensionSlotReactProps>>();
+  const [workspaceTitle, setWorkspaceTitle] = useState("");
+
+  useNavigationContext({
+    type: "workspace",
+    handler: (link, state: { title?: string }) => {
+      setCurrentWorkspaceExtensionSlot(() => (
+        <ExtensionSlotReact extensionSlotName={link} state={state} />
```
manuelroemer, line 31: Actually, now that I look at these changes, I'm wondering whether we _should_ pass the `patientUuid` here. For example, it could be done like this:

```ts
const [,,patientUuid] = useCurrentPatient();

 useNavigationContext({
    type: "workspace",
    handler: (link, state: { title?: string }) => {
      setCurrentWorkspaceExtensionSlot(() => (
        <ExtensionSlotReact extensionSlotName={link} state={{ patientUuid, ...state }} />
                                                             // ^ Relevant changes here in state
    ....
```

Does that make any sense? It would provide a reasonable default and thus wouldn't require any `switchTo` users to always have to know and pass the correct `patientUuid` to the extension.

FlorianRappl: Hm yes and no.

It makes sense somehow, but in general we may not know the patient UUID. This particular context provider would be able to get that in, but if we use the basket "standalone" we can / will only be able to identify the patent UUID via the URL parameter, right?

FlorianRappl: I would not introduce the `useCurrentPatient` here. Does it even work? I mean this component opens the browser router, which has access to the routes yielding the current patient UUID. Since the component is thus not in a React Context the patient UUID should not be retrievable here.

src/root.component.tsx
```diff
@@ -46,6 +69,14 @@ function Root(props) {
             </div>
           </div>
         </main>
+        <ContextWorkspace
+          title={workspaceTitle}
+          extensionSlot={currentWorkspaceExtensionSlot}
+          clearExtensionSlot={() => {
```
FlorianRappl, line 75: We could place this in a `useCallback` hook to avoid respecifying it per render.

src/root.component.tsx
manuelroemer: Hm, actually, I think that my thoughts on this topic are not exactly correct. I initially thought that this would make sense if you cannot fetch the correct patient uuid at your `switchTo` location, but if that's not possible, you cannot navigate to the correct extension slot URL anyway...

manuelroemer: Oh, you were faster than me - yeah, I think that my comment should be ignored. We'll have to pass the UUID through the link anyway, so there's really not much point in providing this default value here.

FlorianRappl (approved):
LGTM - thanks a lot 🍻 !

PR #184, opened 2020-11-09
Adaptations to the context workspace (Enable closing from within / Enable scrolling)

manuelroemer:
* Pass callbacks to the hosted context workspace content to enable closing the workspace from within
* Enable scrolling inside the workspace


FlorianRappl:
The build fails here. There are some errors with the dependencies. Most likely we need to rebuild the package-lock.

manuelroemer:
Should work now. I also had to add "esnext" to the tsconfig, otherwise the build would not work locally.

FlorianRappl:
Yeah I think we should remove these usages in 3.0.9 - I've noted them down.

FlorianRappl (approved):
LGTM - thanks :beers:!

PR #185, opened 2020-11-10
Reorganize core views

denniskigen:
This PR: 

- Moves the `Allergies` widget up higher in the UI so it sits alongside the `Care Programs` widget and underneath the `Conditions` and `Immunizations` widget. It also assigns it `columnSpan: 2` so it fills up half of the page width. 

The rationale for this is that each of the widgets mentioned above takes up half of the available page width. With the current widget order, there's an empty space next to the `Care Programs` widget. The `Allergies` widget is placed at the bottom where the surrounding widgets are full-width. This makes the UI look jarring and inconsistent.

Before:

Screenshot 2020-11-10 at 09 28 36

After:

Screenshot 2020-11-10 at 09 27 40

### DIFF
--- a/src/view-components/core-views.tsx
+++ b/src/view-components/core-views.tsx
@@ -122,6 +122,12 @@ export const coreDashboardDefinitions: DashboardConfig[] = [
         basePath: "programs"
       },
       {
+        name: "AllergiesOverview",
+        esModule: "@openmrs/esm-patient-chart-widgets",
+        layout: { columnSpan: 2 },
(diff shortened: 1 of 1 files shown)


brandones (approved):
Seems fine, but will be obviated by the new designs soon (if not obviated already)

denniskigen:
Precisely.I need this for the new designs

brandones:
Oh, then perfect :ok_hand:

PR #186, opened 2020-11-11
Fixed the patientUuid not being passed to extensions all the time

manuelroemer:
As mentioned by @jonathandick on Slack, the drugorder had an issue fetching the patient data on certain steps. That's due to a missing dependency in a `useEffect` hook. This fixes the problem.


FlorianRappl (approved):
Awesome - thanks:)

PR #187, opened 2020-11-13
Update to latest config spec

brandones:
Interdependent with https://github.com/openmrs/openmrs-esm-core/pull/27 .

Once all of these are approved, I will merge the esm-core PR, release it, and update the dependency.


denniskigen (approved):
LGTM, thanks @brandones

PR #194, opened 2020-12-04
Add slot to the top of every dashboard

brandones:
Mostly just for the workshop tomorrow—people can drop "Hello world" widgets right into the patient chart.

### DIFF
--- a/src/view-components/dashboard/dashboard.component.tsx
+++ b/src/view-components/dashboard/dashboard.component.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import styles from "./dashboard.css";
 import Widget, { WidgetConfig } from "../widget/widget.component";
+import { ExtensionSlot } from "@openmrs/esm-react-utils";
 
 export default function Dashboard(props: DashboardProps) {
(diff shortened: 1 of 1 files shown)


jonathandick (approved):

src/view-components/dashboard/dashboard.component.tsx
```diff
@@ -27,34 +28,50 @@ export default function Dashboard(props: DashboardProps) {
...
-            >
-              {widget}
-            </div>
-          );
-        })}
+    <>
+      <div className={styles.container}>
+        <div className={styles.dashboard} style={{ gridTemplateColumns: 1 }}>
+          <ExtensionSlot
+            extensionSlotName={`patient-chart-${sanitize(
```
FlorianRappl, line 35: I'm not sure sanatize makes sense here. After all, this name has to match what was given in the widgets. If a user enters something wrong in the config it should not show up, or?

brandones: This code should go away in the not-too-distant future, once we redo patient-chart to be fully extension-based.

PR #196, opened 2020-12-08
MF-387: Add translation strings for error tile

denniskigen:
Required in https://github.com/openmrs/openmrs-esm-patient-chart-widgets/pull/160.


FlorianRappl (approved):
LGTM!

PR #207, opened 2021-02-05
Reorganize core views

denniskigen:
Moves the vitals and biometrics widgets to the top of the dashboard. These two are the widgets that receive the most focus in the designs. I think it makes sense that they are located higher up at least in the interim before we get to v1.0.

Screenshot 2021-02-05 at 14 47 45

Tangentially, I think it's worth looking into why the vitals widget takes so long to load.

### DIFF
--- a/src/view-components/core-views.tsx
+++ b/src/view-components/core-views.tsx
@@ -104,6 +104,18 @@ export const coreDashboardDefinitions: DashboardConfig[] = [
     },
     widgets: [
       {
+        name: "VitalsOverview",
+        extensionSlotName: "vitals-widget",
+        layout: { columnSpan: 2 },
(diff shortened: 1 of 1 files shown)


FlorianRappl (approved):
LGTM!

brandones:
@denniskigen what's going on with this? Can we clean up the conflict and merge? cc @FlorianRappl

denniskigen:
Updated @brandones @FlorianRappl

FlorianRappl:
Sorry @denniskigen  - since this one is about to be removed / replaced we should not accept it.

See https://github.com/openmrs/openmrs-esm-patient-chart/tree/feature/monorepo

denniskigen:
Oh, neat @FlorianRappl

PR #213, opened 2021-03-31
fix workspace rendering bug and widgets rerendering

nmalyschkin:
This fixes the workspace bug (workspace was not rendered a second time) as well as the issue, that widgets were unmounted/mounted multiple times.


FlorianRappl:
Since `Widget` is gone in the merge coming in tomorrow this PR cannot be accepted.

PR #220, opened 2021-04-15
MF-500: New structure for the actions button on patient-banner extensions.

vasharma05:
**What does this PR do?**
This PR changes the structure of the CustomOverflowMenu that was created earlier. The new structure allows adding the menu items from different esm-modules. Previously, CustomOverflowMenuItem component was being used, which couldn't be used in every esm-module, so the new structure allows the freedom from this component and adding the class into either button/link in the new extensions will do the work.

### DIFF
--- a/packages/esm-patient-banner-app/package-lock.json
+++ b/packages/esm-patient-banner-app/package-lock.json
@@ -1,0 +1,9073 @@
+{
+  "name": "@openmrs/esm-patient-banner-app",
+  "version": "3.0.0",
+  "lockfileVersion": 2,
+  "requires": true,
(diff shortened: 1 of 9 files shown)


mherman22 (approved):

HerbertYiga:
@vasharma05  thanks for working on this, it would be great to have the screen shots of before and after your changes

vasharma05:
The screenshots will be the same, because I haven't added any new thing. I have just changed the extension which was already present, used in the Actions menu in the patient banner.

FlorianRappl (approved):
LGTM!

PR #224, opened 2021-04-21
test results in monorepo

nmalyschkin:
I've previously merged the test results widget code as is, so this PR is only the adaptive changes needed to get it working in the monorepo.


packages/esm-patient-test-results-app/src/withDashboardRouting.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { Switch, Route, BrowserRouter, useParams } from "react-router-dom";
+
+export default function withDashboardRouting(WrappedComponent) {
+  const WrappedWithParams = (props) => {
+    const params = useParams();
+    return <WrappedComponent {...props} {...params} />;
+  };
+
+  return ({ basePath, patient, patientUuid }) => (
+    <BrowserRouter basename={`${window.spaBase}${basePath}/test-results`}>
```
FlorianRappl, line 11: I think this whole component should change.

The `wrappedWithParams` is imho no longer required. Using the general navigation context also is now deprecated.

packages/esm-patient-test-results-app/src/overview/recent-overview.component.tsx
```diff
@@ -2,10 +2,12 @@ import React from "react";
 
 import Button from "carbon-components-react/lib/components/Button";
 import DataTableSkeleton from "carbon-components-react/lib/components/DataTableSkeleton";
+
+import { switchTo, useCurrentPatient } from "@openmrs/esm-framework";
```
FlorianRappl, line 6: Would not use `useCurrentPatient`. The props already supply the current patient.

packages/esm-patient-test-results-app/src/withDashboardRouting.tsx
nmalyschkin: okay, so what would be the proper alternative to extract parameter?

FlorianRappl: The props are passed in. There is no need to extract parameters.

nmalyschkin: Do you mean, that the complete is passed through as a prop and I can extract the parameters inside my component?
In need the uuids encoded in the URL in any case or the routing would not work.

FlorianRappl: There is no URL anymore. You open the workspace by attaching an extension.

nmalyschkin: This is not a workspace anymore but a dashboard (hence the name `withDashboardRouting`).
I used [esm-patient-notes-app](https://github.com/openmrs/openmrs-esm-patient-chart/blob/feature/monorepo/packages/esm-patient-notes-app/src/notes/notes.component.tsx) as reference.

FlorianRappl: Then this was even the situation beforehand; props such as `patientUuid` as passed in directly. No need to do anything.

FlorianRappl: The reason for the notes provider was that there are components within components that require complex objects such as the patient, which are not transported via the URL. While prop drilling could be done these more complex objects could also be transported via a context.

PR #225, opened 2021-04-21
use useExtensionSlotMeta to extend patient chart dashboard with exten…

nmalyschkin:
depends on https://github.com/openmrs/openmrs-esm-core/pull/99

Instead of configuring new dashboards in the patient chart app, dashboards can be added through extension meta data. This way, if the test-results MF is not loaded, the dashboard is not configured and not displayed.

Any thoughts?


packages/esm-patient-test-results-app/src/index.ts
```diff
@@ -35,6 +29,17 @@ function setupOpenMRS() {
           columnSpan: 4,
         },
       },
+      {
+        id: "test-results-summary-dashboard",
+        slot: "patient-chart-dashboard-slot",
+        load: getAsyncLifecycle((() => {}) as any, options),
```
FlorianRappl, line 35: I think this could be used to display the link in the nav bar.

FlorianRappl (commented):
Yeah idea is sound and is actually also what I've been experimenting with.

In my view the component could be the navbar link. This way the dashboard MF itself is responsible for rendering the nav entry and the chart MF can do the routing and layouting of the dashboard.

packages/esm-patient-test-results-app/src/index.ts
nmalyschkin: But does the dashboard MF need a _custom_ way of defining a link in the nav bar, do we have a use-case where we wouldn't use the default?

I don't like leaving the load function empty either. I thought we could set `meta.config.type` to `custom`, in which case the component would be displayed instead of a `GridView` or `TabbedView`

FlorianRappl: Well, already the localization part is kind of magic right now and only works via meta and magic. Plus we don't know - some dashboards may want to display dynamic information (e.g., a badge) there.

So yes I'd say that would be the way forward.

Maybe the link could still come from the outside, and only the link's contents would be defined by the component.

nmalyschkin: Alright, sounds good!

We don't have a way for an ExtensionSlot to define a fallback, in case a Extension does not provide a component, right? So even "default" behavior would be set explicitly in this case.

nmalyschkin: With default behavior I mean: display a link with the title from the meta object.

FlorianRappl: Yes correct!

nmalyschkin:
The way it is implemented right now, we support both ways: extensions and config.

packages/esm-patient-chart-app/src/index.ts
```diff
@@ -44,6 +44,7 @@ function setupOpenMRS() {
     extensions: [
       {
         id: "patient-chart-nav-items",
+        slot: "nav-menu-slot",
```
FlorianRappl, line 47: This should be attached dynamically. We only want to have this attached when the patient chart is rendered.

nmalyschkin: oh okay, I was already wondering why this was missing.
Well the attach mechanism isn't working for me – I'll have look why.

FlorianRappl (approved):
LGTM!

PR #229, opened 2021-04-27
minor build script and description fixes

nmalyschkin:


packages/esm-patient-common-lib/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@openmrs/esm-patient-common-lib",
   "version": "3.0.0",
   "license": "MPL-2.0",
-  "description": "Patient notes microfrontend for the OpenMRS SPA",
+  "description": "Library for common patient chart components",
   "browser": "dist/openmrs-esm-patient-common-lib.js",
   "main": "src/index.ts",
   "source": true,
```
FlorianRappl, line 8: Great - can we also remove most of the script parts? For instance, `openmrs develop` is only for MFs.

I think in general also the `dependencies` (etc.) should be adjusted.

FlorianRappl (commented):
LGTM - just some further cleanup in the package.json

packages/esm-patient-common-lib/package.json
```diff
@@ -33,24 +28,11 @@
...
   },
   "dependencies": {
-    "@carbon/charts-react": "^0.41.14",
-    "@carbon/icons-react": "^10.18.0",
-    "@typescript-eslint/eslint-plugin": "^4.15.1",
-    "@typescript-eslint/parser": "^4.15.1",
-    "carbon-components-react": "^7.25.0",
-    "d3": "^5.16.0",
-    "lodash-es": "^4.17.15",
     "systemjs-webpack-interop": "^1.2.1"
```
FlorianRappl, line 31: Do we still need this? I mean I don't see a webpack build for this one (and since we bundle it in we don't have an issue with paths anyway).

nmalyschkin: I guess we don't need this either then:
https://github.com/openmrs/openmrs-esm-patient-chart/blob/feature/monorepo/packages/esm-patient-common-lib/src/set-public-path.ts

FlorianRappl (approved):
LGTM!

PR #232, opened 2021-04-29
Remove errant semi-colon from markup

denniskigen:
Removes an errant semicolon from the `ConditionsOverview` markup.


FlorianRappl (approved):
LGTM!

PR #238, opened 2021-05-02
Enhancements to the visit notes form

denniskigen:
- Use the correct slot name in the attach function. This fixes an issue where clicking the `Add` button would not trigger the form opening in a workspace.
- Add translation keys and their corresponding strings.
- Amend the empty results pane to show both the failed search query and the error message.
- Small changes to the rendering logic for the diagnosis search field.
- Display a toast with a success message upon successfully saving a visit note - this is just a halfway solution, we should render an alert per the [designs](https://app.zeplin.io/project/5f7223cfda10f94d67cec6d0/screen/60182928c320622924b940ff).
- Memoize debounced search function.
- Disable the submit button when submitting.
- Improved type annotations.


FlorianRappl (approved):
Great changes, thanks a lot @denniskigen !

PR #245, opened 2021-05-06
Enhancements to the Programs form

denniskigen:
This PR makes the following enhancements to the programs form:

- Carbonize the UI in keeping with the designs.
- Enable launching the form in the new workspace.
- Disable the submit button while submitting.
- Close the workspace and render a toast with a success message upon successfully saving a program enrollment.
- Validate `Completion date` input so that the allowed dates range between the `Enrollment date` and the current date.

Screenshot 2021-05-06 at 09 47 09


FlorianRappl (approved):
LGTM - thanks a lot! 🍻

PR #247, opened 2021-05-07
Remove appui dependencies

denniskigen:
Previously, we made calls to an appui endpoint (`/ws/rest/v1/appui/session`) for fetching session data. 

Following @corneliouzbett's [work](https://github.com/openmrs/openmrs-module-webservices.rest/pull/418), we can now leverage the `useSessionUser` hook and reliably expect it to provide both the `currentProvider` and the `sessionLocation`.

This PR removes all of the existing appui-reliant implementations and refactors the components that used them to make use of the useSessionUser hook instead.


FlorianRappl (approved):
LGTM - thanks a lot !

PR #250, opened 2021-05-09
Loop call to fetch appointments fixed

vasharma05:
**What does this PR do?**

This PR has a fix for clearing out the loop call to fetch appointments from **ws/rest/v1/appointments/search** for a patient.
In the code snippet below for appointments overview widget,
image
 
 Whenever useEffect (line35) runs the API call to fetch the appointments(line 39), which then saves the data in the `appointments` state using `setAppointments`(line 40), the re-rendering happens and the startDate variable is updated to the new Date-Time. When the `startDate` value updates with the latest time, the useEffect re-runs, due to startDate as one of its dependencies and this never-ending loop calls the endpoint again and again.(you can check the network tab in the dev tools on the patient-chart summary page to verify this.)
 
 I have removed the `startDate` from the dependencies of the useEffect function in this PR.

### DIFF
--- a/packages/esm-patient-appointments-app/src/appointments/appointments-overview.component.tsx
+++ b/packages/esm-patient-appointments-app/src/appointments/appointments-overview.component.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import dayjs from 'dayjs';
 import styles from './appointments-overview.scss';
 import Add16 from '@carbon/icons-react/es/add/16';
@@ -23,12 +23,16 @@ interface AppointmentOverviewProps {
   patientUuid: string;
 }
 
+function getStartDate() {
+  return dayjs().format();
+}
+
 const AppointmentsOverview: React.FC<AppointmentOverviewProps> = ({ patientUuid }) => {
   const { t } = useTranslation();
   const appointmentsToShowCount = 5;
   const [appointments, setAppointments] = React.useState(null);
   const [error, setError] = React.useState(null);
-  const startDate = dayjs().format();
+  const startDate = useMemo(getStartDate, []);
   const displayText = t('appointments', 'appointments');
   const headerTitle = t('appointments', 'Appointments');
 


packages/esm-patient-appointments-app/src/appointments/appointments-overview.component.tsx
```diff
@@ -42,7 +42,7 @@ const AppointmentsOverview: React.FC<AppointmentOverviewProps> = ({ patientUuid
 
       return () => abortController.abort();
     }
-  }, [patientUuid, startDate]);
```
FlorianRappl, line 45: Why don't we "fix" start date instead? I think right now every re-render would update the start date. Shouldn't it be a constant instead?

vasharma05: We can make it a state instead, it won't update on every re-render. What you say?

FlorianRappl: Yeah that would be a suitable workaround (actually it can also just be a memo - so it does not matter if you use `useState` and `useMemo` except that the latter would not require array destructing and thus is a bit simpler to look out / understand).

However, what happens then if we re-mount? My question is - shouldn't the appointment start be the actual appointment start date?

FlorianRappl: I'd say let's not discuss this.

Let's just make `const startDate = useMemo(getStartDate, []);` where

```
function getStartDate() {
  return dayjs().format();
}
```

is a global utility.

vasharma05: Sure! Thanks @FlorianRappl ! I'll do it and pushing it asap.

vasharma05:
@FlorianRappl please review the new push.

packages/esm-patient-chart-app/src/visit/widget/visits-overview.component.tsx
```diff
@@ -0,0 +1,238 @@
...
+  TableRow,
+  TableExpandRow,
+  TableExpandedRow,
+  TableExpandHeader,
+} from 'carbon-components-react/es/components/DataTable';
+import { useTranslation } from 'react-i18next';
+import { EmptyState, ErrorState } from '@openmrs/esm-patient-common-lib';
+import { Visit, getVisitsForPatient, createErrorHandler, OpenmrsResource } from '@openmrs/esm-framework';
+import { fetchEncounterByUuid } from './encounter.resource';
+import { StructuredListBody } from 'carbon-components-react';
```
FlorianRappl, line 22: Import from submodule please.

packages/esm-patient-chart-app/src/visit/widget/visits-overview.component.tsx
```diff
@@ -0,0 +1,238 @@
...
+interface EncounterViewProps {
+  encounter: OpenmrsResource;
+}
+
+const RenderEncounter: React.FC<EncounterViewProps> = ({ encounter }) => {
+  const { t } = useTranslation();
+  const [encounterDetails, setEncounterDetails] = useState<OpenmrsResource | null>(null);
+  const { uuid, encounterType } = encounter;
+
+  const formatVisitTime = (inputDate: any) => {
```
FlorianRappl, line 35: This function can be extracted out of the component. No need to re-allocate on every render.

packages/esm-patient-chart-app/src/visit/widget/visits-overview.component.tsx
```diff
@@ -0,0 +1,238 @@
...
+    return dayjs(inputDate).format('HH:MM');
+  };
+
+  useEffect(() => {
+    if (uuid) {
+      const sub = fetchEncounterByUuid(uuid).subscribe(setEncounterDetails, createErrorHandler());
+      return () => sub.unsubscribe();
+    }
+  }, [uuid]);
+  console.log(encounterDetails, encounterType.display, 'encounterDetails');
```
FlorianRappl, line 45: Remove

packages/esm-patient-chart-app/src/visit/widget/visits-overview.component.tsx
```diff
@@ -0,0 +1,238 @@
...
+
+  return (
+    <div className={styles.encounterRow}>
+      <p className={styles.leftSide}>{encounterType.display}</p>
+      <div className={styles.rightSide}>
+        {encounterDetails && (
+          <React.Fragment>
```
FlorianRappl, line 52: Use the shorthand `<>`

packages/esm-patient-chart-app/src/visit/widget/visits-overview.component.tsx
```diff
@@ -0,0 +1,238 @@
...
+  headerTitle,
+  visits,
+  tableRows,
+  showAllVisits,
+  toggleShowAllVisits,
+}) => {
+  const { t } = useTranslation();
+  const displayText = t('startVisit', 'Visit');
+
+  const tableHeaders = [
```
FlorianRappl, line 88: Can be moved out of the component.

packages/esm-patient-chart-app/src/visit/widget/visits-overview.component.tsx
```diff
@@ -0,0 +1,238 @@
...
+    { key: 'date', header: 'Date' },
+    { key: 'time', header: 'Time' },
+    { key: 'location', header: 'Location' },
+    { key: 'visitType', header: 'Visit Type' },
+  ];
+
+  const launchStartVisitPrompt = React.useCallback(
```
FlorianRappl, line 95: Honestly, the whole thing can be made a "constant" function by moving it out of the component.

There is no need to re-allocate this on every render.

packages/esm-patient-chart-app/src/visit/widget/visits-overview.component.tsx
```diff
@@ -0,0 +1,238 @@
...
+  useEffect(() => {
+    if (patientUuid) {
+      const abortController = new AbortController();
+      getVisitsForPatient(patientUuid, abortController).subscribe(({ data }) => {
+        setvisits(data.results);
+      }, createErrorHandler());
+    }
+  }, [patientUuid]);
+
+  const formatVisitDate = (inputDate: any) => {
```
FlorianRappl, line 204: Extract out of the component.

packages/esm-patient-chart-app/src/visit/widget/visits-overview.component.tsx
```diff
@@ -0,0 +1,238 @@
...
+    return dayjs(inputDate).format('DD-MMM-YYYY');
+  };
+  const formatVisitTime = (inputDate: any) => {
```
FlorianRappl, line 207: Extract out of the component.

packages/esm-patient-chart-app/src/visit/widget/visits-overview.component.tsx
```diff
@@ -0,0 +1,238 @@
...
+                      <TableExpandRow {...getRowProps({ row })}>
+                        {row.cells.map((cell) => (
+                          <TableCell key={cell.id}>{cell.value}</TableCell>
+                        ))}
+                      </TableExpandRow>
+                      {row.isExpanded && (
+                        <tr className={styles.expandedRow}>
+                          <td colSpan={headers.length + 1}>
+                            {tableRows[index].encounters.length ? (
+                              tableRows[index].encounters.map((encounter) => <RenderEncounter encounter={encounter} />)
```
FlorianRappl, line 144: Missing `key`.

vasharma05:
Ohh sorry @FlorianRappl my mistake I just mixed the 2 branches and didn't even check it out after pushing. Really sorry for the mishap. The other changes is regarding the `Past Visit Widget` in the patient chart summary, which is in progress, but stuck in a few things in the ux-design channel. I'll just revert back all the mishap.

Thanks alot for reviewing the changes and sorry for your time.

vasharma05:
@FlorianRappl I have done the required change and checked it thorough. Please check it now.

FlorianRappl (approved):
LGTM!

PR #252, opened 2021-05-10
Add ability to pass patient info as prop to form-entry

donaldkibet:
### Description

1. Add ability to pass patient info as a prop to form-entry
2. Fix bug on the depreciated esm-api `getCurrentPatient` function being unavailable
3. Add form-entry link to the left side-nav

### DIFF
--- a/packages/esm-patient-forms-app/src/index.ts
+++ b/packages/esm-patient-forms-app/src/index.ts
@@ -1,4 +1,11 @@
-import { defineConfigSchema, getAsyncLifecycle, messageOmrsServiceWorker } from '@openmrs/esm-framework';
+import {
+  defineConfigSchema,
+  getAsyncLifecycle,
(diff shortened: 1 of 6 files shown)


packages/esm-patient-forms-app/src/index.ts
```diff
@@ -9,11 +16,6 @@ function setupOpenMRS() {
     pattern: '.+/ws/rest/v1/form.*',
   });
 
-  messageOmrsServiceWorker({
```
FlorianRappl, line 12: Why remove it?

FlorianRappl (approved):

packages/esm-patient-forms-app/src/index.ts
donaldkibet: Got deleted by mistake while rebasing, returned it in

PR #254, opened 2021-05-11
Updated to webpack 5

vasharma05:
Minor changes in `webpack.config.js` in accordance with the webpack 5.

### DIFF
--- a/packages/esm-patient-allergies-app/webpack.config.js
+++ b/packages/esm-patient-allergies-app/webpack.config.js
@@ -23,7 +23,6 @@ module.exports = (env, argv = {}) => ({
     filename,
     libraryTarget: 'system',
     path: resolve(__dirname, 'dist'),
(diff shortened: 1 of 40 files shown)


vasharma05:
The warning regarding the `system` property in the parser you said we can skip, gave the error while building the the medications app, so removing the `system` prop from the `parser` in the `webpack.config.js` from all packages to not throw error in other apps as well.

packages/esm-patient-allergies-app/webpack.config.js
```diff
@@ -27,9 +27,7 @@ module.exports = (env, argv = {}) => ({
   module: {
     rules: [
       {
-        parser: {
-          system: false,
-        },
+        parser: {},
```
FlorianRappl, line 30: No need for the empty object, just remove this part completely.

FlorianRappl (approved):
LGTM - thanks! :beers:

PR #258, opened 2021-05-18
Exlude Sass files from formatting

denniskigen:
Sass files should not be formatted by Prettier.

Thanks to @nmalyschkin for figuring out that this was what was screwing up Prettier locally.


nmalyschkin (approved):
Great! ;)

PR #261, opened 2021-05-19
More improvements to the programs form

denniskigen:
It takes a while to load a patient's eligible programs in the Programs field dropdown when you first launch the programs form. This PR adds a loading spinner which disappears once the programs have been loaded. It also renders a tile with an error message in the case where the patient is already enrolled in all of the available programs.


FlorianRappl (approved):
LGTM - thanks!

PR #264, opened 2021-05-20
use esm-framework usePagination hook

donaldkibet:
### Description

Since `usePagination` hook has been moved to  [esm-react-utils](https://github.com/openmrs/openmrs-esm-core/blob/master/packages/framework/esm-react-utils/src/usePagination.ts) duplication would occur on [esm-common-lib](https://github.com/openmrs/openmrs-esm-patient-chart/blob/master/packages/esm-patient-common-lib/src/usePagination.ts)


FlorianRappl (approved):
Great - thanks!

PR #266, opened 2021-05-24
Fix forms pagination arrow positioning

donaldkibet:
### Description

1. Fixed `nextPage` button not displaying correctly
2. Removed duplicated `Empty` component

Before
Screenshot 2021-05-24 at 14 35 50

After
Screenshot 2021-05-24 at 14 40 03


FlorianRappl (approved):
LGTM - thanks!

PR #270, opened 2021-05-25
testresults range select

nmalyschkin:
image

### DIFF
--- a/packages/esm-patient-test-results-app/src/trendline/RangeSelector.tsx
+++ b/packages/esm-patient-test-results-app/src/trendline/RangeSelector.tsx
@@ -3,6 +3,7 @@ import Tab from 'carbon-components-react/es/components/Tab';
 import * as React from 'react';
 
 import styles from './trendline.scss';
+import { useTranslation } from 'react-i18next';
 
 export enum RangeTypes {
   DAY_1,
@@ -14,7 +15,7 @@ export enum RangeTypes {
   ALL,
 }
 
-export const deduceRange = (type: RangeTypes, upperRange: Date, lowerRange: Date): [Date, Date] | [] => {
+export function deduceRange(type: RangeTypes, upperRange: Date, lowerRange: Date): [Date, Date] | [] {
   switch (type) {
     case RangeTypes.DAY_1:
       return [new Date(Date.parse(upperRange.toString()) - 1 * 24 * 3600 * 1000), upperRange];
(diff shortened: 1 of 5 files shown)


packages/esm-patient-test-results-app/src/trendline/RangeSelector.tsx
```diff
@@ -0,0 +1,49 @@
...
+    case RangeTypes.ALL:
+    default:
+      return [lowerRange, upperRange];
+  }
+};
+
+export const RangeSelector: React.FC<{ setRange: (range: RangeTypes) => void }> = ({ setRange }) => {
+  return (
+    <Tabs light selected={6} className={styles['range-tabs']}>
+      <Tab label="1 day" onClick={() => setRange(RangeTypes.DAY_1)}></Tab>
```
FlorianRappl, line 40: Can we localize the labels?

Also some repetition could be saved if you just put it in an array.

Finally, the `<Tab />` seems self-closed. There is no content.

packages/esm-patient-test-results-app/src/trendline/RangeSelector.tsx
```diff
@@ -0,0 +1,49 @@
...
+  DAY_1,
+  DAY_5,
+  MONTH_1,
+  MONTH_6,
+  YEAR_1,
+  YEAR_5,
+  ALL,
+}
+
+export const deduceRange = (type: RangeTypes, upperRange: Date, lowerRange: Date): [Date, Date] | [] => {
```
FlorianRappl, line 17: Classic example of a `function`.

packages/esm-patient-test-results-app/src/trendline/RangeSelector.tsx
```diff
@@ -0,0 +1,49 @@
...
+  switch (type) {
+    case RangeTypes.DAY_1:
+      return [new Date(Date.parse(upperRange.toString()) - 1 * 24 * 3600 * 1000), upperRange];
```
FlorianRappl, line 20: Instead of the `type` we could also just get in the time span. Seems more flexible and straight forward to me.

All would be / could be just max int.

nmalyschkin: We select time spans but set time ranges, therefore we need the upper (and lower) range information. Also time ranges can be computed starting from the current day or the day of the last data point, which should not concern the selector component (separation of concern).

FlorianRappl: Right now it always starts from upperRange apparently.

Either way, if you want to keep the `RangeTypes` I'd prefer to use it like that:

```js
function computeSpan(type) {
  switch (type) {
         case RangeTypes.DAY_1:
            return 1 * 24 * 3600 * 1000;
          case RangeTypes.DAY_5:
            return 5 * 24 * 3600 * 1000;
          case RangeTypes.MONTH_1:
            return 30 * 24 * 3600 * 1000;
          case RangeTypes.MONTH_6:
            return 182 * 24 * 3600 * 1000;
          case RangeTypes.YEAR_1:
            return 365 * 24 * 3600 * 1000;
          case RangeTypes.YEAR_5:  
            return 5 * 365 * 24 * 3600 * 1000;
          default:
            return undefined;
  }
}

export function deduceRange(type: RangeTypes, upperRange: Date, lowerRange: Date): [Date, Date] {
  const span = computeSpan(type);

  if (span) {
    return [new Date(Date.parse(upperRange.toString()) - span), upperRange];
  } else {
    return [lowerRange, upperRange];
  }
}
```

nmalyschkin: Refactored it, hope to your liking

FlorianRappl (approved):
LGTM! :beers:

PR #272, opened 2021-05-26
MF-564: Restore vitals widget tests

denniskigen:
- Fix outdated tests for the components that make up the vitals widget.
- Also, fix an incorrect spec in the `AllergiesOverview` test suite.


FlorianRappl (approved):
LGTM - thanks! :beers:

PR #275, opened 2021-05-27
fix ui-action z-index

donaldkibet:
### Description

Before
Screenshot 2021-05-27 at 14 59 25

After
Screenshot 2021-05-27 at 15 00 19


FlorianRappl (approved):
Works for me! Thanks :beers:

PR #277, opened 2021-05-31
add new trendline stylings and date formats, add zebra styling

nmalyschkin:
https://issues.openmrs.org/browse/MF-610
https://issues.openmrs.org/browse/MF-611

### DIFF
--- a/packages/esm-patient-test-results-app/src/trendline/trendline.component.tsx
+++ b/packages/esm-patient-test-results-app/src/trendline/trendline.component.tsx
@@ -88,12 +88,11 @@ const Trendline: React.FC<{
   /**
    * reorder svg element to bring line in front of the area
    */
-  React.useEffect(() => {
-    if (patientData)
-      setTimeout(() => {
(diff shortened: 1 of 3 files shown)


packages/esm-patient-test-results-app/src/trendline/trendline.component.tsx
```diff
@@ -85,6 +85,17 @@ const Trendline: React.FC<{
     [setRange, upperRange],
   );
 
+  /**
+   * reorder svg element to bring line in front of the area
+   */
+  React.useEffect(() => {
+    if (patientData)
+      setTimeout(() => {
```
FlorianRappl, line 93: Why is this deferred? Would a `useLayoutEffect` be better here?

Also don't forget that `querySelector` may return `null`. Maybe use `?.`? It could be that future versions of Carbon implement that differently...

nmalyschkin: `useLayoutEffect` was not on my radar – you're totally right, this is the right tool for the job.

PR #278, opened 2021-05-31
Add ability to launch workspace in desktop and tablet modes

donaldkibet:
### Description

At the moment workspace can only be launched in tablet mode. this PR introduces the ability to launch work in both desktop and tablet mode.

### Screenshoot

workspace


packages/esm-patient-chart-app/src/root.component.tsx
```diff
@@ -10,7 +10,6 @@ export default function Root() {
     <BrowserRouter basename={spaRoot}>
       <SideMenu />
       <Route path={dashboardPath} component={PatientChart} />
-      <Route path={basePath} component={ContextWorkspace} />
```
FlorianRappl, line 13: I don't agree with this change.

The context for the workspace should work globally in patient chart and not be confined to the patient chart component.

packages/esm-patient-chart-app/src/ui-components/patient-chart.component.scss
```diff
@@ -56,5 +57,14 @@ $actionPanelExpandedOffset: $actionNavOffset + $actionPanelOffset;
   display: flex;
   width: 100%;
   padding-right: $actionNavOffset;
-  flex-direction: column;
+  flex-direction: row;
+}
+
+.inActiveDesktopWorkspace{
```
FlorianRappl, line 63: This is rather inflexible CSS design. It's usually not how its done. You also don't include this with a notion of "desktop", but rather, for instance "activeInlineWorkspace".

Also usually this is where CSS media queries shine.

packages/esm-patient-chart-app/src/ui-components/patient-chart.component.tsx
```diff
@@ -19,6 +22,7 @@ const PatientChart: React.FC<RouteComponentProps<PatientChartParams>> = ({ match
   const { patientUuid, view, subview } = match.params;
   const [loading, patient] = useCurrentPatient(patientUuid);
   const state = useMemo(() => ({ patient, patientUuid }), [patient, patientUuid]);
+  const { active } = useWorkspace();
```
FlorianRappl, line 25: I am not sure if we need all this.

CSS should be sufficient. There are plenty of selectors in CSS (e.g., `:empty`, adjacent children, ...) which should make it possible to support the desktop view without any functional changes.

The problem when introducing functional changes just to achieve layout is that this won't scale. And it introduces spaghetti code. I'd appreciate if we either stick to CSS only, or properly encapsulate the changes (effectively also not requiring changes to the existing code). The existing code should be free of any "desktop" or "mobile" considerations. Exception: The real set of functionalities is different between desktop and mobile.

packages/esm-patient-chart-app/src/ui-components/patient-chart.component.scss
```diff
@@ -58,3 +60,7 @@ $actionPanelExpandedOffset: $actionNavOffset + $actionPanelOffset;
   padding-right: $actionNavOffset;
   flex-direction: column;
 }
+
+.marginRight {
+  margin-right: 35%;
```
FlorianRappl, line 65: This does not feel right. It should be implicitly right.

Usually, this is where flex box shines.

packages/esm-patient-chart-app/src/ui-components/patient-chart.component.tsx
```diff
@@ -19,10 +20,11 @@ const PatientChart: React.FC<RouteComponentProps<PatientChartParams>> = ({ match
...
 
   const mainClassName = `
     omrs-main-content 
-    ${styles.chartContainer} 
+    ${styles.chartContainer} ${active && styles.marginRight}
```
FlorianRappl, line 27: Not sure if we need this. It should be automatically layouted well. Flexbox to the rescue!! 🦸

packages/esm-patient-chart-app/src/ui-components/patient-chart.component.scss
donaldkibet: Tried using flex but since we are using `position: fixed` it is taken out of the normal flow and is therefore not applicable

packages/esm-patient-chart-app/src/ui-components/patient-chart.component.tsx
```diff
@@ -8,6 +8,7 @@ import { useVisitDialog } from '../hooks/useVisitDialog';
 import { RouteComponentProps } from 'react-router-dom';
 import { ExtensionSlot, useCurrentPatient } from '@openmrs/esm-framework';
 import ActionMenu from './action-menu.component';
+import { useWorkspace } from '../hooks/useWorkspace';
```
FlorianRappl, line 11: Not needed anymore.

FlorianRappl (approved):
LGTM!

packages/esm-patient-chart-app/src/ui-components/patient-chart.component.scss
FlorianRappl: Yes - we should remove these fixed positions. Just saying.

(My tendency towards flexbox was not "keep everything as is", but rather "lets change it to apply a better layouting mechanism"...)

PR #282, opened 2021-06-07
Conditions form enhancements

denniskigen:
This PR improves the conditions form as follows:

- Set the padding of items in the search results panel to `0.875rem`.
- Set the `max-height` of the search results panel to 14rem.
- Set the background color of a condition in the search results panel to `$ui-03` on hover.

These changes tie in the UI of this form together with that of the visit notes form for an overall consistent UX.

### DIFF
--- a/packages/esm-patient-conditions-app/src/conditions/conditions-form.scss
+++ b/packages/esm-patient-conditions-app/src/conditions/conditions-form.scss
@@ -3,16 +3,17 @@
 @import "~carbon-components/src/globals/scss/mixins";
 
 .condition {
-  padding: 0.625rem;
+  padding: 0.875rem;
 }
 
 .conditionsList {
   background-color: $ui-02;
+  max-height: 14rem;
+  overflow-y: scroll;
 }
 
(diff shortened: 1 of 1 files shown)


HerbertYiga:
cc @FlorianRappl

packages/esm-patient-conditions-app/src/conditions/conditions-form.scss
```diff
@@ -3,16 +3,17 @@
...
 
 .condition {
-  padding: 0.625rem;
+  padding: 0.875rem;
 }
 
 .conditionsList {
   background-color: $ui-02;
+  max-height: 14rem;
+  overflow-y: scroll;
```
FlorianRappl, line 12: AFAIK this will *always* show the scrollbar. Is this intentional? Should it rather be `auto`?

denniskigen: No, it's not. Changed to `auto`.

FlorianRappl (approved):
:beers:!

PR #284, opened 2021-06-07
Set vertical overflow behaviour to auto.

denniskigen:
Preferable to `overflow-y: scroll` as we don't want scrollbars to be shown when content is not being clipped.


FlorianRappl (approved):
:beers:!

PR #293, opened 2021-06-18
Fix vitals header record button toggling the header view

donaldkibet:
### Description

1. At the moment when you click on the vitals-header `Record Vitals` button, it opens the vitals and biometrics and at the same time toggles the vitals to display the vitals details. This PR prevents this behavior.
2. Also removed the unnecessary `z-index` on the patient header and force contents to be pushed downwards when vitals header status is expanded.

### Screenshoot

vitals


FlorianRappl (approved):
LGTM! Thanks :beers:

PR #297, opened 2021-06-22
Offline Visits

manuelroemer:
Adds offline visit handling in the patient chart.


packages/esm-patient-chart-app/src/offline.ts
```diff
@@ -0,0 +1,95 @@
...
+interface OfflineVisit extends NewVisitPayload {
+  uuid: string;
+}
+
+const visitSyncType = 'visit';
+
+export function setupCacheableRoutes() {
+  messageOmrsServiceWorker({
+    type: 'registerDynamicRoute',
+    pattern: '.+/openmrs/ws/fhir2/R4/Patient/.+',
```
FlorianRappl, line 25: What I was wondering: Do we need the `/openmrs/` part in here? Wouldn't `.+/ws/fhir2/...` be sufficient? See also line 30 (consistency). And I think it may be more "reliable" (i.e., if this prefix / path changes the actual module is still named correctly).

packages/esm-patient-chart-app/src/ui-components/patient-chart.component.tsx
```diff
@@ -19,15 +20,17 @@ interface PatientChartParams {
 const PatientChart: React.FC<RouteComponentProps<PatientChartParams>> = ({ match }) => {
   const { patientUuid, view, subview } = match.params;
   const [loading, patient] = useCurrentPatient(patientUuid);
+  const sessionUser = useSessionUser();
   const state = useMemo(() => ({ patient, patientUuid }), [patient, patientUuid]);
   const { active } = useWorkspace();
 
   const mainClassName = `
```
FlorianRappl, line 27: Should that have new lines in there? I'd say let's make it a single line: ``const mainClassName = `omrs-main-content ${styles.chartContainer}`;``.

FlorianRappl (commented):
LGTM - thanks!

packages/esm-patient-chart-app/src/offline.ts
manuelroemer: You are right, that was just an oversight. Thanks for mentioning it!

packages/esm-patient-chart-app/src/offline.ts
```diff
@@ -0,0 +1,100 @@
...
+
+async function createOfflineVisitForPatient(patientUuid: string, location: string) {
+  const offlineVisit: OfflineVisit = {
+    uuid: generateOfflineUuid(),
+    patient: patientUuid,
+    startDatetime: new Date(),
+    location,
+    // TODO: This UUID belongs to the "Facility Visit" type.
+    //       This should be replaced with the dedicated offline visit as soon as it exists in the BE.
+    visitType: '7b0f5697-27e3-40c4-8bae-f4049abfb4ed',
```
manuelroemer, line 79: I've just added another temporary visit UUID for now. I think we can merge this now and just replace this UUID with the offline visit's UUID when it's there. Then this PR doesn't stay open for an indefinite time.

FlorianRappl (approved):
LGTM! 🍻

PR #298, opened 2021-06-26
Update allergy form to the latest design

donaldkibet:
### Description

1. Updated allergy form to use the latest designs

### Screenshoot

allergies


packages/esm-patient-allergies-app/src/allergies/allergies-form/allergy-form.component.tsx
```diff
@@ -0,0 +1,298 @@
...
+    (value: boolean, id: string, event: React.ChangeEvent<HTMLInputElement>) => {
+      value
+        ? setPatientReactions((prevState) => [...prevState, id])
+        : setPatientReactions((prevState) => prevState.filter((reaction) => reaction !== id));
+    },
+    [],
+  );
+
+  const handleAllergenTypeChange = React.useCallback((index: number) => {
+    if (index === 0) setAllergenType(AllergenType.DRUG);
```
FlorianRappl, line 104: Use block statements (`if (...) { ... }`) and in this case its either a `switch` or `if ... else if ...` construct. These cases are all exclusive.

packages/esm-patient-allergies-app/src/allergies/allergies-form/allergy-form.component.tsx
```diff
@@ -0,0 +1,298 @@
...
+        closeWorkspace();
+
+        showToast({
+          kind: 'success',
+          description: t('allergySaved', 'Allergy saved successfully'),
+        });
+      }
+      (err) => {
+        createErrorHandler();
+        setStatus(StateTypes.ERROR);
```
FlorianRappl, line 140: That we have multiple setX calls here is a sign of a join state. Sure you want to have it like that? If so, a `useReducer` would make more sense.

packages/esm-patient-allergies-app/src/index.ts
```diff
@@ -74,6 +76,16 @@ function setupOpenMRS() {
         online: { showAddAllergy: true },
         offline: { showAddAllergy: false },
       },
+      {
+        id: patientAllergiesFormWorkspace,
```
FlorianRappl, line 80: Should the workspace also work offline? If no (i.e., if it should not be shown when offline) then all is fine, otherwise `offline: true` should be used.

FlorianRappl (commented):
Great work! :beers:

FlorianRappl (approved):
:beers:

PR #299, opened 2021-06-27
Prevent hitting the api on multiple time for the same resource (Observations endpoint)

donaldkibet:
### Description

1. At the moment the vitals-header state component makes three API requests on every page navigation action, which is not ideal. Not sure if this is the best approach to fix this, but I have removed the config object as a dependency on fetch hitting the observation endpoint. Below is a gif on the network tab of before and after this change.

###. Screenshots

before

before-1

after

after

### DIFF
--- a/packages/esm-patient-vitals-app/src/vitals/vitals-header/vital-header-state.component.tsx
+++ b/packages/esm-patient-vitals-app/src/vitals/vitals-header/vital-header-state.component.tsx
@@ -22,6 +22,7 @@ interface VitalHeaderProps {
 const VitalHeader: React.FC<VitalHeaderProps> = ({ patientUuid, showRecordVitals }) => {
   const { t } = useTranslation();
   const config = useConfig();
+  const { concepts } = React.useMemo(() => config, [config]);
   const [vital, setVital] = useState<PatientVitals>();
   const [displayState, setDisplayState] = useState<ViewState>({
     view: 'Default',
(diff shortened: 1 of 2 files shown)


packages/esm-patient-vitals-app/src/vitals/vitals-header/vital-header-state.component.tsx
```diff
@@ -22,6 +22,7 @@ interface VitalHeaderProps {
 const VitalHeader: React.FC<VitalHeaderProps> = ({ patientUuid, showRecordVitals }) => {
   const { t } = useTranslation();
   const config = useConfig();
+  const { concepts } = React.useMemo(() => config, [config]);
```
FlorianRappl, line 25: I find it odd that we receive a changed config. Should we improve the `useConfig` hook? I think it should directly give us something that won't change between re-renders.

packages/esm-patient-vitals-app/src/vitals/vitals-header/vital-header-state.component.tsx
```diff
@@ -44,14 +45,14 @@ const VitalHeader: React.FC<VitalHeaderProps> = ({ patientUuid, showRecordVitals
...
-      const subscription = performPatientsVitalsSearch(config.concepts, patientUuid, 10).subscribe((vitals) => {
+    if (patientUuid && concepts) {
+      const subscription = performPatientsVitalsSearch(concepts, patientUuid, 10).subscribe((vitals) => {
         setVital(vitals[0]);
         setIsLoading(false);
       }, createErrorHandler);
       return () => subscription.unsubscribe();
     }
-  }, [patientUuid, config.concepts]);
+  }, [patientUuid]);
```
FlorianRappl, line 55: Still, if we changed concepts we should re-fetch.

My question is; why did concepts change in the first place?

donaldkibet: Concepts changed because the `useConfig` updates the concepts. I will close this PR since the best place to make the changes is on the `useConfig` hook to return values that don't change over re-renders

PR #300, opened 2021-06-28
Add ability to specify other allergen

donaldkibet:
### Description

1. At the moment, when a user selects another allergen, there is no ability to specify which allergen that is, this PR Adds that ability in.
2. Hide the Other text box, until a user selects other option

### Screenshoot

allergyform-enhancement


FlorianRappl (approved):
LGTM!

PR #303, opened 2021-07-01
Remove Medication Orders as a default clinical view

donaldkibet:
### Description

At the moment the medication order is set as a default clinical view, since we shouldn't be specifying this kind of configuration, am removing it.


FlorianRappl (approved):
LGTM!

PR #304, opened 2021-07-01
MF-637: Fix design issues removed active visit underline

donaldkibet:
### What does this PR do?

1. Remove the underline from the active visit tag
2. Remove ability to redirect to form builder in the event no forms are available and display a meaningful message

### Screenshoot

Screenshot 2021-07-01 at 22 07 32

### DIFF
--- a/packages/esm-patient-common-lib/src/empty-state/empty-state.component.tsx
+++ b/packages/esm-patient-common-lib/src/empty-state/empty-state.component.tsx
@@ -24,6 +24,7 @@ export const EmptyState: React.FC<EmptyStateProps> = (props) => {
         </Trans>
       </p>
       <p className={styles.action}>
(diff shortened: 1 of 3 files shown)


FlorianRappl (approved):
Awesome! :beers:

packages/esm-patient-common-lib/src/empty-state/empty-state.component.tsx
```diff
@@ -8,7 +8,6 @@ import { EmptyDataIllustration } from './empty-data-illustration.component';
 export interface EmptyStateProps {
   headerTitle: string;
   displayText: string;
-  launchForm?(): void;
```
FlorianRappl, line 11: This was removed, but the MFs using it have not been adjusted.

PR #305, opened 2021-07-01
Found modules with unresolved backend dependencies.

alaboso:
- Found modules with unresolved backend dependencies.
- fhir module missing dependency where fhir2 module is in use

### DIFF
--- a/packages/esm-patient-allergies-app/src/index.ts
+++ b/packages/esm-patient-allergies-app/src/index.ts
@@ -14,7 +14,7 @@ const importTranslation = require.context('../translations', false, /.json$/, 'l
 
 const backendDependencies = {
   'webservices.rest': '^2.2.0',
-  fhir: '^1.4.2',
+  'fhir2': '^1.2.0',
 };
 
 const frontendDependencies = {
--- a/packages/esm-patient-appointments-app/src/index.ts
+++ b/packages/esm-patient-appointments-app/src/index.ts
@@ -6,7 +6,7 @@ const importTranslation = require.context('../translations', false, /.json$/, 'l
 
 const backendDependencies = {
   'webservices.rest': '^2.2.0',
-  fhir: '^1.4.2',
+  'fhir2': '^1.2.0',
 };
 
 const frontendDependencies = {
--- a/packages/esm-patient-attachments-app/src/index.ts
+++ b/packages/esm-patient-attachments-app/src/index.ts
@@ -6,7 +6,7 @@ const importTranslation = require.context('../translations', false, /.json$/, 'l
 
 const backendDependencies = {
   'webservices.rest': '^2.2.0',
-  fhir: '^1.4.2',
+  'fhir2': '^1.2.0',
 };
 
 const frontendDependencies = {
--- a/packages/esm-patient-banner-app/src/index.ts
+++ b/packages/esm-patient-banner-app/src/index.ts
@@ -4,7 +4,7 @@ const importTranslation = require.context('../translations', false, /.json$/, 'l
 
 const backendDependencies = {
   'webservices.rest': '^2.2.0',
-  fhir: '^1.4.2',
+  'fhir2': '^1.2.0',
 };
 
 const frontendDependencies = {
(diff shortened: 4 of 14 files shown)


packages/esm-patient-allergies-app/src/index.ts
```diff
@@ -14,7 +14,7 @@ const importTranslation = require.context('../translations', false, /.json$/, 'l
 
 const backendDependencies = {
   'webservices.rest': '^2.2.0',
-  fhir: '^1.4.2',
+  'fhir2': '^1.2.0',
```
FlorianRappl, line 17: Is this indeed the case?

First of all: Some of these potentially don't require fhir and / or webservices.rest. I'd like to know which ones. If we improve the marked dependencies we should do it correctly.

alaboso: Yes - The call to `ws/rest/v1/module?v=custom:(uuid,version)` yields 
`{"results":[...{"uuid":"fhir2","version":"1.1.0"}...]}` - there is `fhir2` and not the expected `fhir` in the response.

I have gone through the esms and this is what I found:

```
esm-patient-allergies-app
	fhir - performPatientAllergySearch
	rest - getPatientAllergyByPatientUuid && getAllergyAllergenByConceptUuid
	
esm-patient-appointments-app
	rest - createAppointment, getAppointments, getAppointmentsByUuid
	
esm-patient-attachments-app
	rest - getAttachmentByUuid, getAttachments
	
esm-patient-banner-app
	rest - fetchPatientRelationships()
	fhir - Just for tests (mocks)
	
esm-patient-biometrics-app
	rest - useVitalsSignsConceptMetaData()
	fhir - getPatientBiometricObservations()
	
esm-patient-clinical-view-app
	rest - 
	fhir - 
	
esm-patient-conditions-app
	ret - searchConditionConcepts()
	fhir - performPatientConditionsSearch(), getConditionByUuid()
	
esm-patient-forms-app
	rest - fetchAllForms(), fetchPatientEncounters()
	fhir - 
	
esm-patient-immunizations-app
	rest - searchImmunizationsConceptSetByMapping(), getImmunizationsConceptSetByUuid()
	fhir - performPatientImmunizationsSearch(), savePatientImmunization()
	
esm-patient-medications-app
	rest - getPatientEncounterId(), getDrugByName(), getDurationUnits()
	fhir - 
	
esm-patient-notes-app
	rest - getEncounterObservableRESTAPI(), fetchEncounterByUuid()
	fhir - searchEncounterByPatientIdentifierWithMatchingVisit(), getEncounterByUuid(), getEncounters()
	
esm-patient-programs-app
	rest - fetchActiveEnrollments(), fetchAvailablePrograms(), fetchLocations()
	fhir - 
	
esm-patient-test-results-app
	rest - loadPresentConcepts()
	fhir - fhirObservationRequests()??
	
esm-patient-vitals-app
	rest - getPatientsLatestVitals(), savePatientVitals()
	fhir - performPatientsVitalsSearch()
```

@samuelmale @ebambo

FlorianRappl: > Yes - The call to ws/rest/v1/module?v=custom:(uuid,version) yields
{"results":[...{"uuid":"fhir2","version":"1.1.0"}...]} - there is fhir2 and not the expected fhir in the response.

Of course, I believe you that *you* installed fhir2, but my question was - does that *indeed* work **exclusively** with fhir2?

alaboso: No. `fhir2` *or* `fhir*` should work but since `fhir2` is a replacement (enhancement) for `fhir` we can't deploy both....

FlorianRappl: Alright. Understood - but we can't have exclusive dependencies. So it's either we say "fhir" or "fhir2". Why wasn't this module properly versioned (I mean the version is there to actually avoid having a package with another name, which essentially is the same - just a different iteration ...)?

cc @jonathandick

jonathandick:
My understanding is that the fhir module was essentially non functional and not worth  the effort to refactor. Fhir2 is a redo. I dint think anyone should ever be using original fhir module

FlorianRappl:
Thanks for the clarification @jonathandick - in this case we should follow @alaboso proposal and replace every occurrence of `fhir` with `fhir2`.

PR #306, opened 2021-07-05
Manage form view states with useReducer

denniskigen:
Use `useReducer` hooks to manage view states in forms. I think these implementations are easier to reason about than their equivalent `useState` hook based implementations.

Other changes:
- Increase timeouts for debounced searches to 500ms.
- Set the default onset date in the `Conditions` form to the current date (`new Date())`.
- Render an error notification when the user tries to enrol a patient in a program using the Programs form but the patient is already enrolled in all of the available programs. Previously, we were rendering this error in a `Tile` that would appear under the `Select program` field.
- Rename `convertToObsPayLoad` to `convertToObsPayload`.


FlorianRappl (approved):
LGTM! Thanks :beers:

PR #307, opened 2021-07-07
Use useVisit in more locations and transitively pass the current visit to esm-form-entry so that forms are associated with (offline) visits

manuelroemer:
For offline visits being available to forms, there must still be some work done. Specifically, offline visits must be known to the `useVisits` hook. As soon as that hook returns the current offline visit as the `currentVisit`, forms automatically work with the current offline visit.


FlorianRappl (approved):
LGTM - thanks!

PR #308, opened 2021-07-08
Added descriptor to offline visit queue.

manuelroemer:
Allows visits to be used as dependencies.


FlorianRappl (approved):
LGTM - thanks!

PR #314, opened 2021-07-29
MF-522: See all button under forms widget opens the clinical views page

NischithJShetty:
This commit is a fix for the below ticket:
https://issues.openmrs.org/browse/MF-522

### DIFF
--- a/packages/esm-patient-forms-app/src/forms/form-view-see-all.component.tsx
+++ b/packages/esm-patient-forms-app/src/forms/form-view-see-all.component.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react';
+import React, { useState, useEffect, useMemo } from 'react';
 import Search from 'carbon-components-react/es/components/Search';
 import debounce from 'lodash-es/debounce';
 import isEmpty from 'lodash-es/isEmpty';
@@ -71,13 +71,15 @@ const filterFormsByName = (formName: string, forms: Array<Form>) => {
 const FormView: React.FC<FormViewProps> = ({ forms, patientUuid, patient }) => {
   const { t } = useTranslation();
   const { currentVisit } = useVisit(patientUuid);
-  const [searchTerm, setSearchTerm] = React.useState<string>(null);
-  const [allForms, setAllForms] = React.useState<Array<Form>>(forms);
+  const [searchTerm, setSearchTerm] = useState<string>(null);
+  const [allForms, setAllForms] = useState<Array<Form>>(forms);
   const { results, goTo, currentPage } = usePagination(allForms.sort(sortFormLatestFirst), 10);
 
-  const handleSearch = React.useMemo(() => debounce((searchTerm) => setSearchTerm(searchTerm), 300), []);
+  const handleSearch = (event) => {
+    setSearchTerm(event.target.value);
+  };
 
-  React.useEffect(() => {
+  useEffect(() => {
     setAllForms(!isEmpty(searchTerm) ? filterFormsByName(searchTerm, forms) : forms);
   }, [searchTerm, forms]);
 
@@ -123,7 +125,7 @@ const FormView: React.FC<FormViewProps> = ({ forms, patientUuid, patient }) => {
         labelText=""
         className={styles.formSearchInput}
         placeholder={t('searchForForm', 'Search for a form')}
-        onChange={(evnt) => handleSearch(evnt.target.value)}
+        onChange={handleSearch}
       />
       <>
         {!isEmpty(searchTerm) && !isEmpty(allForms) && (
--- a/packages/esm-patient-forms-app/src/forms/form-view-see-all.component.tsx
+++ b/packages/esm-patient-forms-app/src/forms/form-view-see-all.component.tsx
@@ -1,6 +1,5 @@
-import React, { useState, useEffect, useMemo } from 'react';
+import React, { useState, useEffect, useMemo, FunctionComponent } from 'react';
 import Search from 'carbon-components-react/es/components/Search';
-import debounce from 'lodash-es/debounce';
 import isEmpty from 'lodash-es/isEmpty';
 import styles from './form-view.component.scss';
 import { attach, navigate, usePagination, useVisit, Visit } from '@openmrs/esm-framework';
@@ -68,7 +67,7 @@ const filterFormsByName = (formName: string, forms: Array<Form>) => {
   return forms.filter((form) => form.name.toLowerCase().search(formName.toLowerCase()) !== -1);
 };
 
-const FormView: React.FC<FormViewProps> = ({ forms, patientUuid, patient }) => {
+const FormView: FunctionComponent<FormViewProps> = ({ forms, patientUuid, patient }) => {
   const { t } = useTranslation();
   const { currentVisit } = useVisit(patientUuid);
   const [searchTerm, setSearchTerm] = useState<string>(null);
--- a/packages/esm-patient-forms-app/src/forms/form-view-see-all.component.tsx
+++ b/packages/esm-patient-forms-app/src/forms/form-view-see-all.component.tsx
@@ -4,7 +4,7 @@ import debounce from 'lodash-es/debounce';
 import isEmpty from 'lodash-es/isEmpty';
 import styles from './form-view.component.scss';
 import { attach, navigate, usePagination, useVisit, Visit } from '@openmrs/esm-framework';
-import { PatientChartPagination } from '@openmrs/esm-patient-common-lib';
+import { PatientChartPagination } from './pagination-see-all.component';
 import { useTranslation } from 'react-i18next';
 import { Form } from '../types';
 import DataTable, {
@@ -73,7 +73,7 @@ const FormView: React.FC<FormViewProps> = ({ forms, patientUuid, patient }) => {
   const { currentVisit } = useVisit(patientUuid);
   const [searchTerm, setSearchTerm] = React.useState<string>(null);
   const [allForms, setAllForms] = React.useState<Array<Form>>(forms);
-  const { results, goTo, currentPage } = usePagination(allForms.sort(sortFormLatestFirst), 5);
+  const { results, goTo, currentPage } = usePagination(allForms.sort(sortFormLatestFirst), 10);
 
   const handleSearch = React.useMemo(() => debounce((searchTerm) => setSearchTerm(searchTerm), 300), []);
 
@@ -166,8 +166,8 @@ const FormView: React.FC<FormViewProps> = ({ forms, patientUuid, patient }) => {
               pageNumber={currentPage}
               totalItems={allForms.length}
               currentItems={results.length}
-              pageUrl={`./forms-see-all.component.tsx?patientUuid=${patientUuid}`}
-              pageSize={5}
+              pageUrl={`$\{openmrsSpaBase}/patient/${patientUuid}/chart/summary`}
+              pageSize={10}
               onPageNumberChange={({ page }) => goTo(page)}
             />
           </>
--- a/packages/esm-patient-forms-app/src/forms/form-view.component.tsx
+++ b/packages/esm-patient-forms-app/src/forms/form-view.component.tsx
@@ -166,7 +166,7 @@ const FormView: React.FC<FormViewProps> = ({ forms, patientUuid, patient }) => {
               pageNumber={currentPage}
               totalItems={allForms.length}
               currentItems={results.length}
-              pageUrl={`./forms-see-all.component.tsx?patientUuid=${patientUuid}`}
+              pageUrl={`patient/forms/${patientUuid}/forms-see-all`}
               pageSize={5}
               onPageNumberChange={({ page }) => goTo(page)}
             />
--- a/packages/esm-patient-forms-app/src/forms/form-view.component.tsx
+++ b/packages/esm-patient-forms-app/src/forms/form-view.component.tsx
@@ -166,7 +166,7 @@ const FormView: React.FC<FormViewProps> = ({ forms, patientUuid, patient }) => {
               pageNumber={currentPage}
               totalItems={allForms.length}
               currentItems={results.length}
-              pageUrl={`patient/forms/${patientUuid}/forms-see-all`}
+              pageUrl={`chart/patient/${patientUuid}/forms`}
               pageSize={5}
               onPageNumberChange={({ page }) => goTo(page)}
             />
--- a/packages/esm-patient-forms-app/src/forms/forms-see-all.component.tsx
+++ b/packages/esm-patient-forms-app/src/forms/forms-see-all.component.tsx
@@ -1,0 +1,134 @@
+import React from 'react';
+import dayjs from 'dayjs';
+import first from 'lodash-es/first';
+import ContentSwitcher from 'carbon-components-react/es/components/ContentSwitcher';
+import Switch from 'carbon-components-react/es/components/Switch';
+import FormView from './form-view-see-all.component';
+import styles from './forms.component.scss';
+import { EmptyState, ErrorState } from '@openmrs/esm-patient-common-lib';
+import { useTranslation } from 'react-i18next';
+import { fetchAllForms, fetchPatientEncounters } from './forms.resource';
+import { filterAvailableAndCompletedForms } from './forms-utils';
+import { Encounter, Form } from '../types';
+import EmptyFormView from './empty-form.component';
+import { fhir } from '@openmrs/esm-framework';
+
+enum FormViewState {
+  recommended = 0,
+  completed,
+  all,
+}
+
+interface FormsProps {
+  patientUuid: string;
+  patient: fhir.Patient;
+}
+
+const Forms: React.FC<FormsProps> = ({ patientUuid, patient }) => {
+  async function setPatient(uUid) {
+    patient = (await fhir.read<fhir.Patient>({ type: 'Patient', patient: uUid })).data;
+  }
+
+  const query = new URLSearchParams(location.search);
+  patientUuid = query.get('patientUuid');
+  setPatient(patientUuid);
+  const { t } = useTranslation();
+  const displayText = t('forms', 'Forms');
+  const headerTitle = t('forms', 'Forms');
+  const [error, setError] = React.useState(null);
+  const [forms, setForms] = React.useState<Array<Form>>([]);
+  const [encounters, setEncounters] = React.useState<Array<Encounter>>([]);
+  const [completedForms, setCompletedForms] = React.useState<Array<Form>>([]);
+  const [selectedFormView, setSelectedFormView] = React.useState<FormViewState>(FormViewState.all);
+  const [filledForms, setFilledForms] = React.useState<Array<Form>>([]);
+
+  React.useEffect(() => {
+    fetchAllForms().subscribe((forms) => setForms(forms), setError);
+  }, []);
+
+  React.useEffect(() => {
+    const fromDate = dayjs(new Date()).startOf('day').subtract(500, 'day');
+    const toDate = dayjs(new Date()).endOf('day');
+    fetchPatientEncounters(patientUuid, fromDate.toDate(), toDate.toDate()).subscribe(
+      (encounters) => setEncounters(encounters),
+      setError,
+    );
+  }, [patientUuid]);
+
+  React.useEffect(() => {
+    const availableForms = filterAvailableAndCompletedForms(forms, encounters);
+    const completedForms = availableForms.completed.map((encounters) => {
+      encounters.form.complete = true;
+      encounters.form.lastCompleted = encounters.encounterDateTime ? encounters.encounterDateTime : null;
+      return encounters.form;
+    });
+    setCompletedForms(completedForms);
+  }, [forms, encounters]);
+
+  React.useEffect(() => {
+    const filledForms = forms.map((form) => {
+      completedForms.map((completeForm) => {
+        if (completeForm.uuid === form.uuid) {
+          form.complete = true;
+          form.lastCompleted = completeForm.lastCompleted ? completeForm.lastCompleted : null;
+        }
+      });
+      return form;
+    });
+    setFilledForms(filledForms);
+  }, [forms, completedForms]);
+
+  const RenderForm = () => {
+    return (
+      <div className={styles.formsWidgetContainer}>
+        <div className={styles.formsHeaderContainer}>
+          <h4 className={`${styles.productiveHeading03} ${styles.text02}`}>{headerTitle}</h4>
+          <div className={styles.contextSwitcherContainer}>
+            <ContentSwitcher
+              className={styles.contextSwitcherWidth}
+              onChange={(event) => setSelectedFormView(event.name as any)}
+              selectedIndex={selectedFormView}>
+              <Switch name={FormViewState.recommended} text="Recommended" />
+              <Switch name={FormViewState.completed} text="Completed" />
+              <Switch name={FormViewState.all} text="All" />
+            </ContentSwitcher>
+          </div>
+        </div>
+        <div style={{ width: '100%' }}>
+          {selectedFormView === FormViewState.completed && (
+            <FormView
+              forms={completedForms}
+              patientUuid={patientUuid}
+              patient={patient}
+              encounterUuid={first<Encounter>(encounters)?.uuid}
+            />
+          )}
+          {selectedFormView === FormViewState.all && (
+            <FormView
+              forms={filledForms}
+              patientUuid={patientUuid}
+              patient={patient}
+              encounterUuid={first<Encounter>(encounters)?.uuid}
+            />
+          )}
+          {selectedFormView === FormViewState.recommended && (
+            <EmptyFormView action={t('noRecommendedFormsAvailable', 'No recommended forms available at the moment')} />
+          )}
+        </div>
+      </div>
+    );
+  };
+
+  return (
+    <>
+      {filledForms.length > 0 ? (
+        <RenderForm />
+      ) : (
+        <EmptyState displayText={displayText} headerTitle={t('helpText', 'Contact system Admin to configure form')} />
+      )}
+      {error && <ErrorState error={error} headerTitle={headerTitle} />}
+    </>
+  );
+};
+
+export default Forms;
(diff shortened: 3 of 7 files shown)


packages/esm-patient-forms-app/src/forms/forms-see-all.component.tsx
```diff
@@ -0,0 +1,134 @@
...
+}
+
+interface FormsProps {
+  patientUuid: string;
+  patient: fhir.Patient;
+}
+
+const Forms: React.FC<FormsProps> = ({ patientUuid, patient }) => {
+  async function setPatient(uUid) {
+    patient = (await fhir.read<fhir.Patient>({ type: 'Patient', patient: uUid })).data;
```
FlorianRappl, line 29: This is fishy on multiple levels.

Please rewrite using a side effect and changing a proper state.

packages/esm-patient-forms-app/src/forms/forms-see-all.component.tsx
```diff
@@ -0,0 +1,134 @@
...
+          form.complete = true;
+          form.lastCompleted = completeForm.lastCompleted ? completeForm.lastCompleted : null;
+        }
+      });
+      return form;
+    });
+    setFilledForms(filledForms);
+  }, [forms, completedForms]);
+
+  const RenderForm = () => {
```
FlorianRappl, line 81: Hm not sure why we create a component in the component -> just render directly

packages/esm-patient-forms-app/src/index.ts
```diff
@@ -38,6 +38,8 @@ function setupOpenMRS() {
   defineConfigSchema(moduleName, {});
 
   return {
+    lifecycle: getAsyncLifecycle(() => import('./forms/forms-see-all.component'), options),
+    activate: 'forms-see-all.component',
```
FlorianRappl, line 42: This is called "route" and I think the route should be more adequate for this scenario.

FlorianRappl (commented):
I think we are on a great way here. Really top notch!

packages/esm-patient-forms-app/src/forms/forms-see-all.component.tsx
NischithJShetty: Sure. I had adopted this from a different component but there is a need to optimise it further. I will make the necessary corrections.

packages/esm-patient-forms-app/src/forms/forms-see-all.component.tsx
NischithJShetty: Will change this as well

packages/esm-patient-forms-app/src/index.ts
NischithJShetty: I could have missed the topic on "route" in the 3.0 docs. It would be helpful if you could point me to the OpenMRS docs about "route". Thank you.

FlorianRappl: `activate` is the old name, `route` is the new name. (See https://openmrs.github.io/openmrs-esm-core/#/main/creating_a_microfrontend - there is `load` and `route`, while you use `lifecycle` and `activate` -- these two are outdated and should be replaced with the new names)

Independent of the property name the exact route should be adjusted, too. As written, a route called "forms-see.all.component` is not good. A route should be something like `all-forms` or `forms/all` etc.

packages/esm-patient-forms-app/src/forms/forms-see-all.component.tsx
NischithJShetty: I have now removed the earlier function and replaced it with a 'State Hook'

packages/esm-patient-forms-app/src/forms/forms-see-all.component.tsx
NischithJShetty: I have removed the extra component in the second commit

packages/esm-patient-forms-app/src/forms/form-view.component.tsx
```diff
@@ -166,7 +166,7 @@ const FormView: React.FC<FormViewProps> = ({ forms, patientUuid, patient }) => {
               pageNumber={currentPage}
               totalItems={allForms.length}
               currentItems={results.length}
-              pageUrl={`$\{openmrsSpaBase}/patient/${patientUuid}/chart/`}
+              pageUrl={`./forms-see-all.component.tsx?patientUuid=${patientUuid}`}
```
FlorianRappl, line 169: That is a super strange URL. Please make that a nice URL... Also make the patient UUID part of the path (i.e., a segment).

packages/esm-patient-forms-app/src/forms/forms-see-all.component.tsx
```diff
@@ -0,0 +1,122 @@
...
+  recommended = 0,
+  completed,
+  all,
+}
+
+const Forms = () => {
+  const query = new URLSearchParams(location.search);
+  const patientUuid = query.get('patientUuid');
+  const [patient, setPatient] = useState<fhir.Patient>();
+  getObjectFHIR(patientUuid).then(setPatient);
```
FlorianRappl, line 26: Super fishy - this will be done in each render!

Use a side-effect ...

packages/esm-patient-forms-app/src/forms/forms-see-all.component.tsx
NischithJShetty: Optimised this further with Effect Hook

packages/esm-patient-forms-app/src/index.ts
NischithJShetty: Have made the changes accordingly, Thank you.

packages/esm-patient-forms-app/src/forms/form-view.component.tsx
NischithJShetty: I have updated this with a new URL

packages/esm-patient-forms-app/src/forms/forms-see-all.component.tsx
NischithJShetty: The suggested changes have been made.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+const Forms = () => {
+  const urlPathArray = window.location.pathname.split('/');
+  const pathUuid = urlPathArray[urlPathArray.length - 2];
+  const [patientUuid, setUUID] = React.useState(null);
+  const [patient, setPatient] = React.useState<fhir.Patient>();
+
+  React.useEffect(() => {
+    getObjectFHIR(patientUuid).then(setPatient);
+    setUUID(pathUuid);
```
FlorianRappl, line 30: Why is this needed? I think the UUID is already known and does not change. Also let's not use `window.location` to read that out... `useHistory` and get the matching param instead.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const headerTitle = t('forms', 'Forms');
+  const [error, setError] = React.useState(null);
+  const [forms, setForms] = React.useState<Array<Form>>([]);
+  const [encounters, setEncounters] = React.useState<Array<Encounter>>([]);
+  const [completedForms, setCompletedForms] = React.useState<Array<Form>>([]);
+  const [selectedFormView, setSelectedFormView] = React.useState<FormViewState>(FormViewState.all);
+  const [filledForms, setFilledForms] = React.useState<Array<Form>>([]);
+
+  React.useEffect(() => {
+    fetchAllForms().subscribe((forms) => setForms(forms), setError);
```
FlorianRappl, line 44: What's that with `subscribe`? Why don't we `unsubscribe`? Why no simple `openmrsFetch`?

Please avoid using these observables.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
```diff
@@ -0,0 +1,128 @@
...
+  React.useEffect(() => {
+    const fromDate = dayjs(new Date()).startOf('day').subtract(500, 'day');
+    const toDate = dayjs(new Date()).endOf('day');
+    fetchPatientEncounters(patientUuid, fromDate.toDate(), toDate.toDate()).subscribe(
+      (encounters) => setEncounters(encounters),
+      setError,
+    );
+  }, [patientUuid]);
+
+  React.useEffect(() => {
```
FlorianRappl, line 56: To be honest looks more like a `useMemo` to me.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
```diff
@@ -0,0 +1,128 @@
...
+    const filledForms = forms.map((form) => {
+      completedForms.map((completeForm) => {
+        if (completeForm.uuid === form.uuid) {
+          form.complete = true;
+          form.lastCompleted = completeForm.lastCompleted ? completeForm.lastCompleted : null;
+        }
+      });
+      return form;
+    });
+    setFilledForms(filledForms);
```
FlorianRappl, line 76: Same here I think this is `useMemo`

packages/esm-patient-forms-app/src/index.ts
```diff
@@ -38,6 +38,12 @@ function setupOpenMRS() {
   defineConfigSchema(moduleName, {});
 
   return {
+    pages: [
+      {
+        load: getAsyncLifecycle(() => import('./forms/forms-see-all'), options),
+        route: 'patient/forms',
```
FlorianRappl, line 44: Miss, e.g., the patient UUID etc.

According to the hyperlink we have something like `patient/forms/${patientUuid}/forms-see-all`, but here I just see `patient/forms`. Is the `forms-see-all` not required? The UUID should be. But the URL format indicates that the segment for patient UUID is actually the form UUID.

Why not have

```
/patient/${patientUuid}/forms
```

NischithJShetty: I thought even if a part (begining part) of the string specified with route matches the URL it will load the page. If it is possible to assign a value like `route: 'patient/${patientUuid}/forms'` , I will do that.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
NischithJShetty: Is this `useHistory` from "react-router-dom" ? and does this involve wrapping the page with the <Router>  component? could you please point me to a script that has used `useHistory` inside the 3.0 framework?

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
NischithJShetty: This file is a adaptation of [forms.component.tsx](https://github.com/openmrs/openmrs-esm-patient-chart/blob/master/packages/esm-patient-forms-app/src/forms/forms.component.tsx) and I have only made minor alterations to it to and have not tried to create the widget from scratch. Most of the code directly comes from the original file which was used to build the Form Entry widget and is now a part of the "openmrs-esm-patient-chart" repo.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
NischithJShetty: This file is a adaptation of forms.component.tsx and I have only made minor alterations to it to fix the small bug and have not tried to create the widget from scratch. Most of the code directly comes from the original file which was used to build the Form Entry widget and is now a part of the "openmrs-esm-patient-chart" repo.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
NischithJShetty: The explation to this is the same as above and the answer can probably be given by the original committer. Do you think we should go ahead and change all these files too? It would be interesting to find out the logic used though. Should the task of rewrting the whole widget be undertaken now or just try to sort out this minor bug for the time being is a decision to be made. Thanks.

packages/esm-patient-forms-app/src/forms/form-view-see-all.component.tsx
```diff
@@ -0,0 +1,185 @@
...
+    );
+  };
+  return (
+    <div className={styles.formContainer}>
+      <Search
+        id="searchInput"
+        labelText=""
+        className={styles.formSearchInput}
+        placeholder={t('searchForForm', 'Search for a form')}
+        onChange={(evnt) => handleSearch(evnt.target.value)}
```
FlorianRappl, line 126: If you change the definition of taking `value -> foo(value)` to `evt -> foo(evt.target.value)` then there is no need for the second function; and you gain performance.

So please change the definition of `handleSearch` and pass it in here directly (i.e., `onChange={handleSearch}`)

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
FlorianRappl: It is from `react-router-dom`. There should be plenty of examples using `useHistory` / `react-router-dom` in this repository. Please look for it yourself (hint: there should be a search functionality in whatever IDE you are using).

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
FlorianRappl: We should reuse components - not copy them. This is a no go that results in barely maintainable code that is bloated and contains bugs.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
FlorianRappl: We should reuse components - not copy them. This is a no go that results in barely maintainable code that is bloated and contains bugs.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
FlorianRappl: We should reuse components - not copy them. This is a no go that results in barely maintainable code that is bloated and contains bugs.

packages/esm-patient-forms-app/src/index.ts
FlorianRappl: At Amazon you are also not going to a page "see-all-products" but "products". I am not sure why you don't follow general principles (i.e., things that you see around in the world, where best practices are already followed), but come up with things like that.

The routing mechanism can also use a regular expression. Have a look at the patient chart microfrontend to see how that can be done.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
NischithJShetty: Exactly my point here! It was impossible to **reuse** components here. Let me explain the origins of this bug. If you go to [form-view.component.tsx](https://github.com/openmrs/openmrs-esm-patient-chart/blob/master/packages/esm-patient-forms-app/src/forms/form-view.component.tsx) and observe Ln #169, the variable "pageUrl" has been hardcoded. Then again on Ln #170 the variable "pageSize" has been **hardcoded**. This was the key component required in this situation and it was impossible re-use it. So I had to go for the closest work around. The other option is rewrting everything which is clearly beyond the scope of this ticket. Now, we need to understand why reusable components with hard coded variables are part of our repo in the first place and make corrections there.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
NischithJShetty: Exactly my point here! It was impossible to reuse components here. Let me explain the origins of this bug. If you go to form-view.component.tsx and observe Ln #169, the variable "pageUrl" has been hardcoded. Then again on Ln #170 the variable "pageSize" has been hardcoded. This was the key component required in this situation and it was impossible re-use it. So I had to go for the closest work around. The other option is rewrting everything which is clearly beyond the scope of this ticket. Now, we need to understand why reusable components with hard coded variables are part of our repo in the first place and make corrections there.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
NischithJShetty: As I have explained before the existing component defined in [form-view.component.tsx](https://github.com/openmrs/openmrs-esm-patient-chart/blob/master/packages/esm-patient-forms-app/src/forms/form-view.component.tsx) has hardcoded variables on Ln# 169 and Ln# 170 making reusability beyond scope. Please suggest the right solutions going forward.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
NischithJShetty: But putting the old codes aside, I will rewrite certain aspects. I will make changes based on your suggestions and come up with a new commit. Thank you.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
NischithJShetty: But putting the old codes aside, I will rewrite certain aspects. I will make changes based on your suggestions and come up with a new commit. Thank you.

packages/esm-patient-forms-app/src/index.ts
NischithJShetty: Done. Thanks!

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
NischithJShetty: I have implemented the other changes but in spite of a lot of attempts “useHistory” was giving errors as reported in “https://stackoverflow.com/questions/64838587/how-to-properly-use-usehistory-from-react-router-dom” so could not make use of it. A proper solution to the error could not be found anywhere, so continued with the old implementation.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
NischithJShetty: This is done and I have replaced the older function with "openmrsFetch"

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
NischithJShetty: I have made other changes but probably "useMemo" is not needed here as the fuction gets called only once. Please let me know if I am missing something.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
NischithJShetty: Even this function gets invoked only once and so probably "useMemo" is not needed.

packages/esm-patient-forms-app/src/index.ts
NischithJShetty: I have changed the path format.

packages/esm-patient-forms-app/src/forms/form-view-see-all.component.tsx
```diff
@@ -0,0 +1,185 @@
...
+};
+
+const FormView: React.FC<FormViewProps> = ({ forms, patientUuid, patient }) => {
+  const { t } = useTranslation();
+  const { currentVisit } = useVisit(patientUuid);
+  const [searchTerm, setSearchTerm] = React.useState<string>(null);
+  const [allForms, setAllForms] = React.useState<Array<Form>>(forms);
+  const { results, goTo, currentPage } = usePagination(allForms.sort(sortFormLatestFirst), 10);
+
+  const handleSearch = React.useMemo(() => debounce((searchTerm) => setSearchTerm(searchTerm), 300), []);
```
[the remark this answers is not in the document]
NischithJShetty: I have fixed this. Thanks!

packages/esm-patient-forms-app/src/forms/form-view-see-all.component.tsx
NischithJShetty: This is also done.

NischithJShetty:
@FlorianRappl I have fixed most of the issues but faced some problems with one and I have mentioned that above. Please review and let me know if further changes are required. Thank you.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
FlorianRappl: Yes you are missing that the `setCompletedForms` is causing a re-render.

Instead, use just a `useMemo`.

Your pattern (bad practice):

```js
const [state, setState] = React.useState();
React.useEffect(() => {
  // long computation giving "result"
  setState(result);
}, []);
```

Instead we can do:

```js
const state = React.useMemo(() => {
  // long computation giving "result"
  return result;
}, []);
```

much more efficient and it's clear from context what goes on...

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
FlorianRappl: Please read (and understand) my other comment carefully... Thanks!

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
NischithJShetty: Thanks for explaining. My observation was that even when `setCompletedForms` gets called enforcing a re-render the below block of code does not get executed because the dependacy array will prevent its execution:

`useEffect(() => {
    const fromDate = dayjs(new Date()).startOf('day').subtract(500, 'day');
    const toDate = dayjs(new Date()).endOf('day');
    fetchPatientEncounters(patientUuid, fromDate.toDate(), toDate.toDate()).then(setEncounters);
  }, [patientUuid]);`

I tried replacing this block with `useMemo` but it was not working probably because there is a `fetch` call happening from within this function and the useMemo fails to re-assign the value returned by the "async"  `fetch` function. But there is no problem when I use `useState` functionality. I will try implementing "useMemo" again if you want.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
NischithJShetty: Even here because of the same reason above (since the value of "completedForms" is dependent upon "encounters") the returned value from async "fetch" was not get re-assigned  when useMemo was used. Please let me know if I need to try again and implement as "useMemo" itself.

NischithJShetty:
Hi @FlorianRappl 

I have stated this as individual comments alongside the particular isssues as well. I did make a lot of attempts to replace `useState` with `useMemo` but it does not seem to be working in this situation. Even when I go to the official React docs there seems to be no mention of using `useMemo` in the place of `useState`. The docs don't really talk about using `useMemo` to initialise state variables. Please let me know if there are any docs detailing the usage of `useMemo` to initialise/ change component state.

When I changed `useState` to `useMemo` it was not updating the variable after a call to the  `openmrsFetch` function  which is `async`. I think when there is a `fetch` call being made its easier to use `useState` than `useMemo`. I could not find any reliable doc suggesting the use of `useMemo` in the place of `useState`. In spite of multiple attempts I could not make use of `useMemo` in this situation.

Sorry about this as I have tried everything possible and hope you are fine if we continue to use `useState` in this situation. 

Thank you.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
FlorianRappl: > I tried replacing this block with useMemo but it was not working probably because there is a fetch call happening from within this function and the useMemo fails to re-assign the value returned by the "async" fetch function.

Yes, that is a good point. I did not see this beforehand, but the operation is async. I'd still say the code there can be improved (that I could not immediately see there having a resource being fetched just confirms this), but either way - let's leave it as is right now. Thanks for trying!

FlorianRappl:
> Even when I go to the official React docs there seems to be no mention of using useMemo in the place of useState.

It is no replacement and I have not written this. Please compare again with my original example (and where the duality that has been used is the right replacement). Here, we have an async operation, which indeed is a side-effect. Ideally, however, you will reduce the number of re-renders (as you always should).

In any case let's not lose more time here.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
```diff
@@ -0,0 +1,169 @@
...
+import styles from './forms.component.scss';
+import { EmptyState } from '@openmrs/esm-patient-common-lib';
+import { useTranslation } from 'react-i18next';
+import { toFormObject } from './forms.resource';
+import { filterAvailableAndCompletedForms } from './forms-utils';
+import { Form } from '../types';
+import EmptyFormView from './empty-form.component';
+import { getObjectFHIR } from './get-FHIR-object';
+import { openmrsFetch } from '@openmrs/esm-framework';
+import _ from 'lodash';
```
FlorianRappl, line 16: Please use `lodash-es` and a direct submodule import.

You see that already in the same module, e.g.:

```js
import first from 'lodash-es/first';
```

packages/esm-patient-forms-app/src/forms/pagination-see-all.component.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import Pagination from 'carbon-components-react/es/components/Pagination';
+import styles from '@openmrs/esm-patient-common-lib/src/pagination/pagination.component.scss';
+import { useTranslation } from 'react-i18next';
+import { ConfigurableLink } from '@openmrs/esm-framework';
+import { usePaginationInfo } from '@openmrs/esm-patient-common-lib/src/pagination/usePaginationInfo';
```
FlorianRappl, line 6: `@openmrs/esm-patient-common-lib` is fully imported. Just use `import { ...} from '@openmrs/esm-patient-common-lib'`.

packages/esm-patient-forms-app/src/forms/forms-see-all.tsx
NischithJShetty: The corrections have been made.

packages/esm-patient-forms-app/src/forms/pagination-see-all.component.tsx
NischithJShetty: The corrections have been made.

FlorianRappl (approved):
LGTM - lets go with it!

PR #319, opened 2021-08-09
Carbonised start visit components

donaldkibet:
### Description

1. Carbonised visits widget i.e. start prompt, end prompt, start visit component and editing past visit 

### Screenshoot

visit


packages/esm-patient-chart-app/src/visit/past-visit-overview.component.tsx
```diff
@@ -0,0 +1,194 @@
...
+  );
+
+  const rowData: Array<DataTableRow> = useMemo(
+    () =>
+      patientPastVisits.length
+        ? patientPastVisits.map((visit, index) => {
+            return {
+              id: `${index}`,
+              startDate: dayjs(visit.startDatetime).format(dateFormat),
+              visitType: visit.visitType.display,
```
HerbertYiga, line 93: cc @FlorianRappl  @brandones

gracepotma: What is the concern here Herbert?

gracepotma:
@manuelroemer please review :) (Can't add you as a reviewer yet for some reason; I think you have to accept our invite to this repo)

packages/esm-patient-chart-app/src/visit/visit-prompt/start-visit.component.tsx
```diff
@@ -0,0 +1,66 @@
...
+          'You can add past visit, update past visit or add new past visit, Click on one of the buttons below',
+        );
+
+  const handleOpenEditPastVisit = useCallback(() => {
+    attach('patient-chart-workspace-slot', 'past-visits-overview');
+    closeModal();
+  }, [closeModal]);
+
+  const handleOpenStartVisitForm = useCallback(() => {
+    attach('patient-chart-workspace-slot', 'start-visit-workspace-form'), closeModal();
```
manuelroemer, line 36: Can you replace the `,` here with `;`?

packages/esm-patient-chart-app/src/visit/past-visit-overview.component.tsx
```diff
@@ -0,0 +1,194 @@
...
+import OverflowMenu from 'carbon-components-react/es/components/OverflowMenu';
+import OverflowMenuItem from 'carbon-components-react/es/components/OverflowMenuItem';
+import { useTranslation } from 'react-i18next';
+import dayjs from 'dayjs';
+import { ErrorState } from '@openmrs/esm-patient-common-lib';
+import { attach, detach, getStartedVisit, getVisitsForPatient, VisitMode, VisitStatus } from '@openmrs/esm-framework';
+import styles from './past-visit-overview.component.scss';
+import Button from 'carbon-components-react/es/components/Button';
+
+// TO DO Abstract this logic for state machines to a hook
```
manuelroemer, line 25: I think that this is fine as-is, actually. If you want to, you can go ahead and trim down the interfaces, e.g. remove `Pending`, `Error` and `Resolved` in favor of:

```ts
interface Action {
  type: ActionTypes;
}
```

manuelroemer (commented):
Awesome job here @donaldkibet! That is a lot of work done in one swoop 🎉 

Since this is quite a big one I'd be glad if either @brandones and/or @FlorianRappl also did a review. I'm sure that I may have missed one or the other thing.

packages/esm-patient-chart-app/src/visit/past-visit-overview.component.tsx
```diff
@@ -0,0 +1,194 @@
...
+        status: 'pending',
+      };
+    case ActionTypes.resolved:
+      return { status: 'resolved' };
+    case ActionTypes.error:
+      return { status: 'error' };
+    default:
+      return state;
+  }
+}
```
brandones, line 63: Note that this is basically an identity function, `x => x`. That's a good hint that we can simplify.

There are two routes we could take this:

1. We could replace the `useReducer` line on line 71 with `const [viewState, setViewState] = useState<ViewState>('pending');`, get rid of the `viewStateReducer` function and all the ActionTypes, and replace `dispatch` calls with `setViewState`.

OR

2. We could go all-in on the reducer pattern. The reducer would manage the status *as well as* the data and error states. You'd delete lines 72 and 73 and make this reducer something like

```suggestion
function reducer(state: PastVisitOverviewState, action: Action): PastVisitOverviewState {
  switch (action.type) {
    case ActionTypes.pending:
      return {
        status: 'pending',
        ...state
      };
    case ActionTypes.resolved:
      return {
        status: 'resolved',
        patientPastVisits: action.payload,
        error: null
      };
    case ActionTypes.error:
      return {
        status: 'error',
        patientPastVisits: null,
        error: action.payload
      };
    default:
      return state;
  }
}
```

and then

```typescript
const [{ status, patientPastVisits, error }, dispatch ] = useReducer(reducer, { status: 'pending' });
```

Hopefully that's clear!

packages/esm-patient-chart-app/src/visit/visit-form/start-visit-helper.ts
```diff
@@ -0,0 +1,27 @@
+import dayjs from 'dayjs';
+
+export const convertTime12to24 = (time12h, timeFormat: string) => {
```
brandones, line 3:
```suggestion
type amPm = 'AM' | 'PM';

export const convertTime12to24 = (time12h, timeFormat: amPm) => {
```

?

packages/esm-patient-chart-app/src/visit/visit-form/start-visit-helper.ts
```diff
@@ -0,0 +1,27 @@
...
+  let [hours, minutes] = time12h.split(':');
+
+  if (hours === '12') {
+    hours = '00';
+  }
```
brandones, line 8:
```suggestion
  if (hours === '12' && timeFormat === 'AM') {
    hours = '00';
  }
```

?

packages/esm-patient-chart-app/src/visit/visit-form/start-visit-helper.ts
```diff
@@ -0,0 +1,27 @@
...
+  const visitDate = dayjs(date).format('DD/MM/YYYY');
+  switch (type) {
+    case 'year':
+      return dayjs(visitDate).year();
+    case 'month':
+      return dayjs(visitDate).month();
+    case 'date':
+      return dayjs(visitDate).date();
+  }
+};
```
brandones, line 27: Is`getDatePart(<type>, <date>)` always simply `dayjs(<date>).<type>()`? In which case we can get rid of this function entirely and just use `dayjs` directly?

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -0,0 +1,189 @@
...
+interface StartVisitFormProps {
+  isTablet: boolean;
+  patientUuid: string;
+}
+
+const StartVisitForm: React.FC<StartVisitFormProps> = ({ isTablet, patientUuid }) => {
+  const { t } = useTranslation();
+  const locations = useLocations();
+  const sessionUser = useSessionUser();
+  const isPM = useCallback(() => new Date().getHours() >= 12, []);
```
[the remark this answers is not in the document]
brandones: Yes. Have a read through https://kentcdodds.com/blog/usememo-and-usecallback sometime, it's very helpful. You could also simply inline it below,

```typescript
const [timeFormat, setTimeFormat] = useState<string>(new Date().getHours() >=12 ? 'PM' : 'AM');
```

it's very simple and has basically no overhead. In fact, it simplifies the whole component since it's one less function I'm trying to keep track of when reading.

packages/esm-patient-chart-app/src/visit/visit-form/visit-type-overview.component.tsx
```diff
@@ -0,0 +1,70 @@
...
+  useEffect(() => {
+    if (!isEmpty(searchTerm)) {
+      const results = visitTypes.filter(
+        (visitType) => visitType.display.toLowerCase().search(searchTerm.toLowerCase()) !== -1,
+      );
+      setSearchResults(results);
+    } else {
+      setSearchResults(visitTypes);
+    }
+  }, [searchTerm, visitTypes]);
```
brandones, line 32:
```suggestion
  const searchResults = useMemo(() => {
    if (!isEmpty(searchTerm)) {
      return visitTypes.filter(
        (visitType) => visitType.display.toLowerCase().search(searchTerm.toLowerCase()) !== -1,
      );
    } else {
      return visitTypes;
    }
  }, [searchTerm, visitTypes]);
```

packages/esm-patient-chart-app/src/visit/visit-form/visit-type-overview.component.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  const handleSearch = useMemo(() => debounce((searchTerm) => setSearchTerm(searchTerm), 300), []);
```
brandones, line 34:
```suggestion
  const handleSearch = () => debounce((searchTerm) => setSearchTerm(searchTerm), 300);
```

I don't thiiiiiiink there's any reason to use `useMemo` here

brandones:
Great work @donaldkibet ! There's a few things we should clean up before merging.

packages/esm-patient-chart-app/src/visit/visit-form/visit-type-overview.component.tsx
FlorianRappl: Important: The suggestion is wrong. `useMemo` accepts a callback to derive a value, so the suggested change would always call `debounce` and create a new debounced handler from it.

Instead it is:

```js
const handleSearch = debounce((searchTerm) => setSearchTerm(searchTerm), 300);
```

Presumably, the `useMemo` is indeed not really useful / worth it in here, but it all depends on what calling `debounce` does under the hood.

brandones: Whoops! That's right. That's also what we have in [patient search](https://github.com/openmrs/openmrs-esm-patient-management/blob/c150ccd37f52bbee28c9063ac5accd82b0a0f4c9/packages/esm-patient-search-app/src/patient-search/patient-search.component.tsx#L62). If we need to optimize I think we'd [do so with `useCallback`](https://stackoverflow.com/a/55616626/1464495).

FlorianRappl: Hm in general yes, but in this particular case I'd argue no. If you'd use `useCallback(debounce(...))` then you'd already walk through the potentially expensive function call. Even worse - whatever you created there gets most likely thrown away unless dependencies changed. In this case `useMemo` would be preferred, but as written this depends on the cost of calling `debounce` (or phrased differently "the cost of constructing a debounced function").

packages/esm-patient-chart-app/src/hooks/useVisitDialog.tsx
```diff
@@ -1,36 +1,31 @@
...
+      switch (type) {
+        case 'start':
+          return setVisitDialogType({ type: 'start', state });
+        case 'prompt':
+          return setVisitDialogType({ type: 'prompt', state });
+        case 'end':
+          return setVisitDialogType({ type: 'end', state });
+        case 'close':
+          return setVisitDialogType({ type: 'close', state });
+      }
```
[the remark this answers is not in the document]
brandones: @donaldkibet ?

packages/esm-patient-chart-app/src/visit/past-visit-overview.component.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+interface PastVisitOverviewProps {
+  patientUuid: string;
+}
+
+const PastVisitOverview: React.FC<PastVisitOverviewProps> = ({ patientUuid }) => {
+  const { t } = useTranslation();
+  const [{ status, patientPastVisits, error }, dispatch] = useReducer(reducer, {
+    status: 'pending',
+    patientPastVisits: [],
```
brandones, line 83:
```suggestion
    patientPastVisits: null,
```

Or else make it set to `[]` in the "error" action of the reducer, instead of `null`.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -0,0 +1,187 @@
...
+          dayjs(visitDate).month(),
+          dayjs(visitDate).date(),
+          hours,
+          minutes,
+        ),
+        visitType: visitType,
+        location: selectedLocation,
+      };
+
+      saveVisit(visitPayload, new AbortController()).subscribe(
```
brandones, line 72: This looks to me like a memory leak—subscribing with no `unsubscribe()`. Maybe you can use `first()`, as suggested [here](https://medium.com/javascript-everyday/no-need-to-unsubscribe-rxjs-operators-will-help-you-out-f8b8ce7bf26a)?

packages/esm-patient-chart-app/src/visit/past-visit-overview.component.tsx
```diff
@@ -0,0 +1,202 @@
...
+  patientUuid: string;
+}
+
+const PastVisitOverview: React.FC<PastVisitOverviewProps> = ({ patientUuid }) => {
+  const { t } = useTranslation();
+  const [{ status, patientPastVisits, error }, dispatch] = useReducer(reducer, {
+    status: 'pending',
+    patientPastVisits: null,
+  });
+  const dateFormat = 'DD-MMM-YYYY';
```
brandones, line 85: We should be using the locale-appropriate date format rather than providing one explicitly. You should be able to get the current locale with

```typescript
const { t, i18n } = useTranslation();
i18n.language
```

brandones: (this applies to all the explicit date formats in this PR)

packages/esm-patient-chart-app/src/visit/visit-prompt/end-visit.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+      },
+      (error) => {
+        showToast({
+          kind: 'error',
+          description: t('visitEndError', 'Error ending current visit'),
+        });
+      },
+    );
+
+    return () => sub && sub.unsubscribe();
```
brandones, line 48: Similar problem to the memory leak from before, since `endCurrentVisit()()` (i.e. the returned function) is never going to be called. You can probably solve these in the same way, however you figure out to do it.

brandones (changes requested):
It's so close!!!

packages/esm-patient-chart-app/src/visit/past-visit-overview.component.tsx
donaldkibet: I have ended up using the specifying the locale from `i18n`

```sh
 toLocaleDateString(locale, { dateStyle: 'medium' })
```

brandones (approved):
Awesome work @donaldkibet !

PR #324, opened 2021-08-12
widened vitals, immunization, allergies and biometric widgets

KorirC:
### Description 
widened vitals, immunization, allergies and biometric widgets [MF-697](https://issues.openmrs.org/browse/MF-697)


donaldkibet (approved):
LGMT

PR #328, opened 2021-08-17
bug fixes on vitals and clinical view addition

walteronoh:
### Description
[MF-701](https://issues.openmrs.org/browse/MF-701) - Set add buttons to only open one workspace at a time on vitals and clinical view widgets.

mf-701-fix_bug_on_vitals_and_clinical_view_addition

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,8 @@
     "sass-loader": "^10.1.0",
     "typescript": "^4.0.3",
     "webpack": "^5.36.2",
-    "webpack-cli": "^4.7.0"
+    "webpack-cli": "^4.7.0",
+    "d3-selection": "^3.0.0"
   },
   "dependencies": {}
 }
--- a/packages/esm-patient-biometrics-app/src/clear-workspace.ts
+++ b/packages/esm-patient-biometrics-app/src/clear-workspace.ts
@@ -1,0 +1,5 @@
+import { detachAll } from '@openmrs/esm-framework';
+
+export const clearWorkspace = (extensionSlotName) => {
+  detachAll(extensionSlotName);
+};
--- a/packages/esm-patient-clinical-view-app/src/clear-workspace.ts
+++ b/packages/esm-patient-clinical-view-app/src/clear-workspace.ts
@@ -1,0 +1,5 @@
+import { detachAll } from '@openmrs/esm-framework';
+
+export const clearWorkspace = (extensionSlotName) => {
+  detachAll(extensionSlotName);
+};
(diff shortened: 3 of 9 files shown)


packages/esm-patient-clinical-view-app/src/clinical-view-overview/clinical-view-overview.component.tsx
```diff
@@ -23,6 +24,7 @@ const ClinicalViewOverview: React.FC<ClinicalViewOverviewProps> = ({ patientUuid
   }, [patient, patientUuid]);
 
   const launchClinicalViewForm = React.useCallback(() => {
+    clearWorkspace('patient-chart-workspace-slot');
```
donaldkibet, line 27: @walteronoh am not sure if this is the best approach to solve this. What happens to forms that have been minimised in the workspace or forms with pending data.

package.json
```diff
@@ -67,7 +67,8 @@
     "sass-loader": "^10.1.0",
     "typescript": "^4.0.3",
     "webpack": "^5.36.2",
-    "webpack-cli": "^4.7.0"
+    "webpack-cli": "^4.7.0",
+    "d3-selection": "^3.0.0"
```
manuelroemer, line 71: Ah, right, about this one: Ideally we don't add this dependency here. I need to talk to Florian about this in general (#320 is still outstanding due to the same issue).

For the moment though, can you successfully build/run everything without adding this dependency? All the other PRs so far seem to function well without it.

packages/esm-patient-biometrics-app/src/clear-workspace.ts
```diff
@@ -0,0 +1,5 @@
+import { detachAll } from '@openmrs/esm-framework';
+
+export const clearWorkspace = (extensionSlotName) => {
```
manuelroemer, line 3: This currently just renames the `detachAll` function without doing anything else. -> I'd remove this entirely and just call `detachAll` directly (assuming that it's still called in the future which, considering @donaldkibet's comment, probably won't be the case).

packages/esm-patient-clinical-view-app/src/clear-workspace.ts
```diff
@@ -0,0 +1,5 @@
+import { detachAll } from '@openmrs/esm-framework';
+
+export const clearWorkspace = (extensionSlotName) => {
```
manuelroemer, line 3: The same here. Generally though, if you have two functions doing exactly the same you can move them to `esm-patient-common-lib`. Then you don't have to copy/paste them. But again, in this specific case I'd just remove them entirely.

packages/esm-patient-vitals-app/src/clear-workspace.ts
```diff
@@ -0,0 +1,5 @@
+import { detachAll } from '@openmrs/esm-framework';
+
+export const clearWorkspace = (extensionSlotName) => {
```
manuelroemer, line 3: Also this one.

NischithJShetty:
@walteronoh 
Please note that the directory structure of the `Biometrics` and `Vitals` widgets have changed and you will now have to make changes to the new files in place of some of the file changes that have currently happened in this PR. This would be the best way to resolve the file conflicts.

More details at: 
https://github.com/openmrs/openmrs-esm-patient-chart/pull/342#issuecomment-911332413

Thanks.

PR #330, opened 2021-08-17
MF-521: Re-order function does not move past medications to active medications

KorirC:
### Description
Re-order function to move past medications to active medications and display without user having to refresh the page [MF-521](https://issues.openmrs.org/browse/MF-521)

### DIFF
--- a/packages/esm-patient-medications-app/src/medications-summary/medications-summary.component.tsx
+++ b/packages/esm-patient-medications-app/src/medications-summary/medications-summary.component.tsx
@@ -5,6 +5,7 @@ import styles from './medications-summary.scss';
 import MedicationsDetailsTable from '../components/medications-details-table.component';
 import { useTranslation } from 'react-i18next';
 import { usePatientOrders } from '../utils/use-current-patient-orders.hook';
+import { useEffect } from 'react';
 
 export interface MedicationsSummaryProps {
   patientUuid: string;
@@ -12,9 +13,11 @@ export interface MedicationsSummaryProps {
 
 export default function MedicationsSummary({ patientUuid }: MedicationsSummaryProps) {
   const { t } = useTranslation();
-  const [activePatientOrders] = usePatientOrders(patientUuid, 'ACTIVE');
+  const [activePatientOrders,fetchActivePatientOrders] = usePatientOrders(patientUuid, 'ACTIVE');
   const [pastPatientOrders] = usePatientOrders(patientUuid, 'any');
-
+  useEffect(() => {
+    fetchActivePatientOrders();
+  });
   return (
     <>
       <h1 className={styles.productiveHeading03}>{t('medications', 'Medications')}</h1>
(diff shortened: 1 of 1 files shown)


donaldkibet (changes requested):
@KorirC seems to me you have pushed changes for carbonising immunisation, update the PR with the correct changes

packages/esm-patient-medications-app/src/medications-summary/medications-summary.component.tsx
```diff
@@ -12,9 +12,11 @@ export interface MedicationsSummaryProps {
 
 export default function MedicationsSummary({ patientUuid }: MedicationsSummaryProps) {
   const { t } = useTranslation();
-  const [activePatientOrders] = usePatientOrders(patientUuid, 'ACTIVE');
+  const [activePatientOrders, fetchActivePatientOrders] = usePatientOrders(patientUuid, 'ACTIVE');
   const [pastPatientOrders] = usePatientOrders(patientUuid, 'any');
-
+  useEffect(() => {
+    fetchActivePatientOrders();
+  });
```
manuelroemer, line 19: This seems wrong to me - the effect would refetch the dependencies on every single render of the component. That is way too often. When exactly do you want to (re)fetch the orders? If it's only once during the initial render, you should add an empty array `[]` to the `useEffect`'s deps.

packages/esm-patient-medications-app/src/medications-summary/medications-summary.component.tsx
```diff
@@ -12,9 +12,11 @@ export interface MedicationsSummaryProps {
 
 export default function MedicationsSummary({ patientUuid }: MedicationsSummaryProps) {
   const { t } = useTranslation();
-  const [activePatientOrders] = usePatientOrders(patientUuid, 'ACTIVE');
+  const [activePatientOrders, fetchActivePatientOrders] = usePatientOrders(patientUuid, 'ACTIVE');
   const [pastPatientOrders] = usePatientOrders(patientUuid, 'any');
-
+  useEffect(() => {
+    fetchActivePatientOrders();
+  }, [activePatientOrders]);
```
manuelroemer, line 19: I feel bad for commenting on the same thing again, so sorry about that!
The issue is that this new version doesn't really make much sense. Reason: `usePatientOrders` gives you two things: The current order array (`activePatientOrders`) and a function for refetchig (`fetchActivePatientOrders`).
What you're now doing is the following: You immediately refetch the orders when the component is mounted (twice, actually - the hook also does that). The refetch function does some work with the backend and then updates `activePatientOrders`. And with `activePatientOrders` being updated your effect triggers again, starting a new refetch which then agains triggers a new one and a new one. -> You end up in an infinite fetching loop.

I just had a look at the ticket and I think that it's actually a bit more difficult and cannot be solved this easily. What you are really trying to do is: When new orders have been placed in the order basket, you want to refetch the data in the summary tables. This means that the medications summary needs somehow know when orders have been finalized by the basket. This is tricky because the two components are separated and don't really have a shared parent (they are both extensions in different places).

Now, there are multiple ways to solve this, some that come to my mind right now are:
1. Dispatch an event in the order basket notifying about when it's closed. Listen to the event in the medication table component and refetch the medications when the event arrives.
2. We already have a global store for the order basket (`order-basket-store.ts`). The store could be enhanced such that it accepts callbacks where other MFs can register to be notified about changes in the basket.

There are probably other ways too, but these two might be simple enough to get started with this ticket. I would personally prefer number 2 since the store already is present anyway.

PR #337, opened 2021-08-23
MF-439 'See All' should take user to new page (specifically for: Vitals, Biometrics & Notes widgets)

NischithJShetty:
This PR deals with three widgets (Vitals. Biometrics and Notes) and  addresses the issue of redirecting the 'See all' link of the widgets to the main widget page from the summary page. It also makes sure that the widgets are paginated. 

For instance, below is an image of the Vitals widget before and after the change:
image


NischithJShetty:
@manuelroemer 
This PR is very similar to the earlier PR #321 reviewd by you and creates pagination in the three widgets mentioned above, along with the proper redirection of the 'See all' link. I have made changes just like in  #321 and have mostly restructured the existing scripts to facilitate the addition of pagination. Hope you are free to review this PR. Thank you.

packages/esm-patient-biometrics-app/src/biometrics/biometrics-summary.component.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import BiometricsBase from './biometrics-base.component';
+
+interface BiometricsProps {
```
HerbertYiga, line 5: cc @FlorianRappl

packages/esm-patient-biometrics-app/src/biometrics/pagination-see-all.component.tsx
```diff
@@ -0,0 +1,52 @@
...
+          <Pagination
+            className={styles.pagination}
+            page={pageNumber}
+            pageSize={pageSize}
+            pageSizes={pageSizes}
+            totalItems={totalItems}
+            onChange={onPageNumberChange}
+          />
+        </div>
+      )}
```
manuelroemer, line 49: Only commenting once, but there are also other files with the same structure:
In the common lib, there exists the `PatientChartPagination` component which has the exact same structure. Can you reuse that one instead of the duplicate code here?

NischithJShetty: True. I could not use that component from common lib as it had a hard coded attribute (on Ln. 36). This file that I have created is the exact copy but without the hard coded attribute. I did not want to change the original common lib component as it could be getting used else where.

manuelroemer: Feel free to change that! 😉 If something gets messed up, the build pipeline should catch it before it can cause any issues.

If it's only about the hardcoded "See all", you can just do the following (which is very very unlikely to break something):

```ts
import React from 'react';
import Pagination from 'carbon-components-react/es/components/Pagination';
import styles from './pagination.component.scss';
import { ConfigurableLink } from '@openmrs/esm-framework';
import { usePaginationInfo } from './usePaginationInfo';

interface PatientChartPaginationProps {
  currentItems: number;
  totalItems: number;
  pageNumber: number;
  pageSize: number;
  onPageNumberChange?: any;
  pageUrl?: string;
  linkText?: string; // <- Changed. Make sure to make this prop optional.
}

export const PatientChartPagination: React.FC<PatientChartPaginationProps> = ({
  totalItems,
  pageSize,
  onPageNumberChange,
  pageNumber,
  pageUrl = '',
  currentItems,
  linkText, // <- Changed. 
}) => {
  const { itemsDisplayed, pageSizes } = usePaginationInfo(pageSize, totalItems, pageNumber, currentItems);

  return (
    <>
      {totalItems > 0 && (
        <div className={styles.paginationContainer}>
          <div className={styles.paginationLink}>
            {itemsDisplayed}
            {pageUrl && (
              <ConfigurableLink to={pageUrl} className={styles.configurableLink}>
                {linkText ?? t('seeAll', 'See all')} // <- Changed. Have a fallback for current users who don't need a special link text.
              </ConfigurableLink>
            )}
          </div>
          <Pagination
            className={styles.pagination}
            page={pageNumber}
            pageSize={pageSize}
            pageSizes={pageSizes}
            totalItems={totalItems}
            onChange={onPageNumberChange}
          />
        </div>
      )}
    </>
  );
};
```

packages/esm-patient-biometrics-app/src/biometrics/biometricsPagination.component.tsx
```diff
@@ -0,0 +1,78 @@
...
+  tableHeaders,
+}) => {
+  const { results, goTo, currentPage } = usePagination(tableRows, pageSize);
+
+  return (
+    <div>
+      <TableContainer>
+        <DataTable rows={results} headers={tableHeaders} isSortable={true} size="short" useZebraStyles>
+          {({ rows, headers, getHeaderProps, getTableProps }) => (
+            <Table {...getTableProps()}>
```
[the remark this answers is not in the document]
NischithJShetty: You are right. The way it's described in the carbon design docs is different. To be honenst what I have done here is just worked on the pagination bit, all other table structure comes from the previous committer. When I look at it, the tables of all our widgets have been defined this way in our repos (they have been tested and working fine). I agree that we should make it look as it is defined in the Carbon docs and rewrite all our tables according to the official guidelines sometime. But for now hopefully we can keep it this way.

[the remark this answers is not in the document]
manuelroemer: Yes, that's fine with me as long as it works correctly. I was just wondering about it - thanks for clarifying this!

packages/esm-patient-biometrics-app/src/biometrics/pagination-see-all.component.tsx
NischithJShetty: This has been implemented. Thanks.

packages/esm-patient-biometrics-app/src/biometrics/biometricsPagination.component.tsx
```diff
@@ -0,0 +1,78 @@
+import React, { FunctionComponent } from 'react';
```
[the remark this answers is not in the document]
NischithJShetty: I have made changes to make things consistent with the rest of the repo. Thanks.

NischithJShetty:
@manuelroemer 
I have made a new commit after implementing all the suggestions. Thank you.

PR #344, opened 2021-08-25
Mf 740:Changed the design of the  success message

alexkemboi4127:
After recording the vitals and biometrics on the form and hit save, a success message is displayed to show that the vitals have been recorded. The message should match the design.The design now looks as shown below:
Screenshot from 2021-08-25 16-03-32


manuelroemer (approved):
LGTM, thanks!

PR #368, opened 2021-09-07
Fixed the margins in the patient banner

vasharma05:
### What does this PR include?
I have pushed the changes regarding the margin fixes in the patient banner. The results are as follows:
CC: @denniskigen 

Before: 
image

After: 
image

### DIFF
--- a/packages/esm-form-entry-app/tsconfig.json
+++ b/packages/esm-form-entry-app/tsconfig.json
@@ -18,7 +18,7 @@
       "es2018",
       "dom"
     ],
-    "skipLibCheck": true,
+    "skipLibCheck": true
   },
   "angularCompilerOptions": {
     "disableTypeScriptVersionCheck": true,
--- a/packages/esm-form-entry-app/tsconfig.json
+++ b/packages/esm-form-entry-app/tsconfig.json
@@ -17,7 +17,8 @@
     "lib": [
       "es2018",
       "dom"
-    ]
+    ],
+    "skipLibCheck": true,
   },
   "angularCompilerOptions": {
     "disableTypeScriptVersionCheck": true,
--- a/yarn.lock
+++ b/yarn.lock
@@ -358,19 +358,19 @@
     source-map "^0.5.0"
 
 "@babel/core@^7.1.0", "@babel/core@^7.11.1", "@babel/core@^7.11.4", "@babel/core@^7.11.6", "@babel/core@^7.7.5", "@babel/core@^7.8.6":
-  version "7.15.0"
-  resolved "https://registry.yarnpkg.com/@babel/core/-/core-7.15.0.tgz#749e57c68778b73ad8082775561f67f5196aafa8"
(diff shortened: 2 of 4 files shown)


FlorianRappl (approved):

yarn.lock
```diff
@@ -17122,9 +17220,9 @@ typescript@4.1.5:
   integrity sha512-6OSu9PTIzmn9TCDiovULTnET6BgXtDYL4Gg4szY+cGsc3JP1dQL8qvE8kShTRx1NIw4Q9IBHlwODjkjWEtMUyA==
 
 typescript@^4.0.2, typescript@^4.0.3, typescript@^4.2.4:
-  version "4.3.5"
-  resolved "https://registry.yarnpkg.com/typescript/-/typescript-4.3.5.tgz#4d1c37cc16e893973c45a06886b7113234f119f4"
-  integrity sha512-DqQgihaQ9cUrskJo9kIyW/+g0Vxsk8cDtZ52a3NGh0YNTfpUSArXSohyUGnvbPazEPLu398C0UxmKSOrPumUzA==
+  version "4.4.2"
```
FlorianRappl, line 17223: By updating TypeScript it seems like you did not introduce a replaecment for `MediaStreamError`.

See here: https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1029

Please do that replacement.

vasharma05: @FlorianRappl the error comes up in the ngx-webcam(in the `node_modules` folder), I went for the following things:

1. Finding a replacement for `MediaStreamError`, which I was not able to find. I found `MediaRecordErrorEvent` which is newly added, but can't say that it's the correct one.
2. Since this error comes in the `ngx-webcam` package, I thought of upgrading it from 0.2.2 to 0.3.x, thinking there might be changes related to this error, but it was not the way I thought it would be.

What else do you think I can do to tackle this?

Right now, I am thinking to revert back to Typescript 4.3 instead of 4.4, what do you think?

FlorianRappl: If its only in a lib you can just disable lib checking.

https://stackoverflow.com/questions/40262513/how-to-disable-suppress-errors-from-libraries-in-typescript

vasharma05: Great! That worked!
I didn't know about muting the libraries, before!
Thanks!

vasharma05: @FlorianRappl , just last question, should I just disable lib checking in the `esm-form-entry`(using ngx-webcam) or should I put it in the global tsconfig.json?
I think it should be in `esm-form-entry`'s tsconfig instead of global, just wanted to confirm beforehand.

FlorianRappl: Yes, let's keep it as minimal as possible. Only esm-form-entry for now.

packages/esm-form-entry-app/tsconfig.json
```diff
@@ -17,7 +17,8 @@
     "lib": [
       "es2018",
       "dom"
-    ]
+    ],
+    "skipLibCheck": true,
```
FlorianRappl, line 21: Please make it a valid JSON.

vasharma05: So sorry! Removed the `,`.

PR #372, opened 2021-09-08
MF-699 : Fix New patient search shows previous patient workspace

donaldkibet:
## Requirements

## Summary

This PR fixes issues where workspace remained opened with previous patient info after changing a patient through search

## Screenshots

MF-699

## Other

*None.*


packages/esm-patient-chart-app/src/ui-components/patient-chart.component.tsx
```diff
@@ -23,6 +23,10 @@ const PatientChart: React.FC<RouteComponentProps<PatientChartParams>> = ({ match
   const state = useMemo(() => ({ patient, patientUuid }), [patient, patientUuid]);
   const { windowSize, openWindows } = useContextWorkspace();
 
+  useEffect(() => {
+    !loading && detachAll('patient-chart-workspace-slot');
```
FlorianRappl, line 27: Hm this is neither explanatory nor robust (for instance if we would switch to SWR and revalidate the data this would not work).

Why not close the workspaces when the patient chart is unmounted?

packages/esm-patient-chart-app/src/ui-components/patient-chart.component.tsx
```diff
@@ -24,8 +25,16 @@ const PatientChart: React.FC<RouteComponentProps<PatientChartParams>> = ({ match
   const { windowSize, openWindows } = useContextWorkspace();
 
   useEffect(() => {
-    !loading && detachAll('patient-chart-workspace-slot');
-  }, [loading]);
+    const handler = () => {
```
FlorianRappl, line 28: This is not really what we want, right?

I mean why / how should the `patientUuid` be different than the `currentPatientUuid`? If this is the case, then the original `patientUuid` is *wrong* - so this is the thing that should be fixed.

gracepotma:
Hey @donaldkibet, anything you need to keep this work moving?

gracepotma:
Could @denniskigen review this? Or perhaps @brandones? Would be great to merge this.

packages/esm-patient-chart-app/src/ui-components/patient-chart.component.tsx
brandones: Florian's comment makes sense to me, @donaldkibet . Can you investigate why that's happening?

donaldkibet: I have investigated the issue, and it seems that once you change the patient, the event listener fires first before the useEffect hook, this results in having the `patientUuid` being the previous one and the `currentPatientUuid` being the current once that has changed. subsequently the useEffect runs again since patientUuid has changed and this results in having both `patientUuid` and `currentPatientUuid` being the same.

Since this isn't the ideal situation, i have used `window.document.referrer` to get the previousPatientUuid and compare that to the `currentPatientUuid` retrieved from `location.pathname` if this two don't match that indicates that the patient details have changed and we close the opened workspace. 


```
useEffect(() => {
    const handler = () => {
      const previousPatientUuid = getPatientUuidFromUrl(window.document.referrer);
      const currentPatientUuid = getPatientUuidFromUrl(location.pathname);
      if (previousPatientUuid !== currentPatientUuid) {
        detachAll('patient-chart-workspace-slot');
      }
    };
    window.addEventListener('popstate', handler);
    return () => window.removeEventListener('popstate', handler);
  }, [patientUuid]);
```

brandones: Okay, that makes sense. Yeah I understood your fix but it's good to have a clearer understanding of the problem. Could you link to the event listener and useEffect hook that you're talking about? Do you think there's a way we can get them to fire in the correct order?

PR #374, opened 2021-09-09
Order entry UI enhancements

denniskigen:
## Requirements

## Summary

This PR provides a whole host of enhancements to the Order Entry UI. These enhancements are driven by the desire to bring the UI/UX as close as possible to the available designs. More specifically, these changes include:

- Adding empty states to the medications summary component. This ensures that both the active medications and past medications widgets can show empty states when they don't have data.
- Changing button labels on the `MedicationOrder` form from `Cancel` to `Discard order` and `Save` to `Save order`.
- Changing the width of `ButtonSet`s so they cover their container's full horizontal width. Additionally, I've also changed the height of the buttons to `4rem` per the designs.
- Adding an empty state to the Order Basket per the designs.
- Adding a new header to the Order Basket with the same styling as the rest of the patient chart widget table headers. This header is only shown in the tablet viewport as specified in the designs.
- Modifying the display of drug and dosage information in order basket item tiles to match the designs.
- Subtle tweaks to button icon colours.
- Changing the color of primary elements from (blue - #0f62fe) to ($brand-teal-01). This ensures consistency with the rest of the UI. 
- Adding test coverage to some of the components that make up the Order Entry widget.


denniskigen:
Setting this to draft to accommodate any changes suggested by Ciaran here https://splashy-tray-3e1.notion.site/Order-entry-UI-mods-c3a30e8364fc46ef9bf5237b73e7f9a9.

packages/esm-patient-medications-app/src/order-basket/order-basket-item-list.component.tsx
```diff
@@ -1,16 +1,19 @@
 import React from 'react';
+import { Tile } from 'carbon-components-react/es/components/Tile';
 import OrderBasketItemTile from './order-basket-item.component';
 import styles from './order-basket-item-list.scss';
 import { useTranslation } from 'react-i18next';
 import { OrderBasketItem } from '../types/order-basket-item';
 
 export interface OrderBasketItemListProps {
+  isTablet: boolean;
```
FlorianRappl, line 9: Boolean props should be made optional, thus they could be used like

```jsx
<Foo bar /> // set to true
<Foo /> // set to false
<Foo bar={var} /> // explicit, e.g., due to a state being passed in
```

packages/esm-patient-medications-app/src/order-basket/order-basket-item-list.scss
```diff
@@ -1,7 +1,70 @@
...
+  margin-bottom: 1rem;
+  @include carbon--type-style("productive-heading-03");
+  color: $text-02;
+}
+
+.heading:after {
+  content: "";
+  display: block;
+  width: 2rem;
+  padding-top: 0.188rem;
```
FlorianRappl, line 28: A lot of magic numbers where. Why 0.188? Why not 0.185 or so?

packages/esm-patient-medications-app/src/order-basket/order-basket-item-list.scss
```diff
@@ -1,7 +1,70 @@
...
 .orderCategoryHeading {
-  @extend .productiveHeading01;
+  @extend .bodyShort01;
   color: $text-02;
   margin: $spacing-05 0;
 }
+
+.desktopTile {
+  text-align: center;
+  border: 1px solid #e0e0e0;
```
FlorianRappl, line 40: $ui-03

packages/esm-patient-medications-app/src/order-basket/order-basket-item.scss
```diff
@@ -1,21 +1,49 @@
 @import "../root";
 
+.clickableTileDesktop {
+  margin-bottom: 0.5rem;
+}
+
+.clickableTileTablet {
+  &:not(:last-of-type) {
+    border-bottom: 0.063rem solid $ui-03;
```
FlorianRappl, line 9: 0.063rem? I mean are we going to moon here;)? Where is this very specific number coming from. Why not 0.065 or so?

packages/esm-patient-medications-app/src/order-basket/order-basket-search-results.component.tsx
```diff
@@ -47,62 +49,60 @@ export default function OrderBasketSearchResults({
...
-              }}
-            />
+            <>
+              <Pagination
+                page={page}
+                pageSize={pageSize}
+                pageSizes={[10, 20, 30, 40, 50]}
+                totalItems={searchResults.length}
+                onChange={({ page, pageSize }) => {
+                  setPage(page);
```
FlorianRappl, line 98: Is there a reason to set both values? Shouldn't we only set the value that changed?

packages/esm-patient-medications-app/src/order-basket/order-basket-search-results.scss
```diff
@@ -1,5 +1,64 @@
...
+
+  &:not(:last-of-type) {
+    margin-bottom: 1rem;
+  }
+}
+
+.divider {
+  border: 0;
+  border-top: 1px solid $color-gray-30;
+  width: 66.66666667%;
```
FlorianRappl, line 24: If you want to be exact just have 66.666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666666...%

I guess overall 66.67 or so would be exact enough. In general though I am not sure we need to be that exact. A 66% would be sufficient, too.

packages/esm-patient-medications-app/src/order-basket/order-basket-search.scss
```diff
@@ -15,3 +15,7 @@
   box-shadow: 0 2px 6px 0 rgba(0, 0, 0, 0.3);
   z-index: 1;
 }
+
+.searchPopupContainer :global(.bx--search-input:focus) {
+  outline: 2px solid #ff832b;
```
FlorianRappl, line 20: What's that color?

packages/esm-patient-medications-app/src/order-basket/order-basket.scss
```diff
@@ -3,3 +3,16 @@
...
+  max-width: 50%;
+  height: 4rem;
+}
+
+.spinner :global(.bx--loading__stroke) {
+  stroke: $brand-teal-01;
+}
+
+.activeMedicationsContainer {
+  border: 1px solid #e0e0e0;
```
FlorianRappl, line 17: $ui-03

FlorianRappl (commented):
I think we are on a great way here. Thanks a lot! :beers:

packages/esm-patient-medications-app/src/order-basket/order-basket-item-list.scss
denniskigen: Screenshot 2021-09-12 at 12 54 04 

3px ~= 0.188rem

packages/esm-patient-medications-app/src/order-basket/order-basket-item.scss
denniskigen: 😄 just strict (maybe too overly so) conformance to the measurements outlined in the designs https://app.zeplin.io/project/60d59321e8100b0324762e05/screen/60d5e127606309ba4cfdafea. 

Screenshot 2021-09-12 at 13 03 26

`1px bottom border` ~= 0.0.625rem

Would you prefer to have these really small measurements in px than rem?

packages/esm-patient-medications-app/src/order-basket/order-basket-search.scss
denniskigen: It's a flavour of orange. It's not in `esm-core` yet, I'd planned to add it in a follow up PR and update this.

Screenshot 2021-09-12 at 13 11 54

packages/esm-patient-medications-app/src/order-basket/order-basket-item-list.scss
FlorianRappl: Then use 3px. It does not seem that there is a relation to the used primary font size.

packages/esm-patient-medications-app/src/order-basket/order-basket-item.scss
FlorianRappl: If you use px or rem depends on if there is a dependency on the primary font size or not. If such odd numbers come out it usually means that there is no dependency. If for instance we have 4px or 8px with a primary font size of 16px then chances are high that 0.25 or 0.5rem are meant, but overall this is a discussion (or specification) with the designer.

packages/esm-patient-medications-app/src/order-basket/order-basket-search.scss
FlorianRappl: It should be added. In general, all used colors should be properly defined by the designer. We should not need to add new colors - the palette should be settled by now.

packages/esm-patient-medications-app/src/order-basket/order-basket-search-results.component.tsx
denniskigen: Dunno why the original author went for this implementation. I see this approach is being used in multiple patient chart widgets. That said, Ciaran actually wants this pagination removed altogether (I can file a ticket for that).

Screenshot 2021-09-12 at 19 30 12

FlorianRappl: Sounds good to me!

brandones:
Ack, sorry this got stale, @denniskigen . If you can fix up the conflicts I can review.

denniskigen:
This is up to date now @brandones

packages/esm-patient-medications-app/src/order-basket/order-basket-item-list.component.tsx
```diff
@@ -1,16 +1,19 @@
 import React from 'react';
+import { Tile } from 'carbon-components-react/es/components/Tile';
 import OrderBasketItemTile from './order-basket-item.component';
 import styles from './order-basket-item-list.scss';
 import { useTranslation } from 'react-i18next';
 import { OrderBasketItem } from '../types/order-basket-item';
 
 export interface OrderBasketItemListProps {
+  isTablet?: boolean;
```
brandones, line 9: I would really rather we just

```typescript
const isTablet = useLayoutType() === "tablet";
```

rather than thread this prop around.

packages/esm-patient-medications-app/src/order-basket/order-basket-search.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { screen, render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import OrderBasketSearch from './order-basket-search.component';
+
+const testProps = {
+  encounterUuid: '',
+  onSearchResultClicked: () => {},
+};
+
+describe('OrderBasketSearch: ', () => {
```
brandones, line 11: This test might be one layer more granular than needs to exist. I would just test the whole search component together (box & results).

packages/esm-patient-medications-app/src/order-basket/order-basket.component.tsx
```diff
@@ -117,62 +127,67 @@ const OrderBasket = connect<OrderBasketProps, OrderBasketStoreActions, OrderBask
...
-            <div className={styles.orderBasketContainer}>
-              <OrderBasketItemList
-                orderBasketItems={items}
-                onItemClicked={(order) => openMedicationOrderFormForUpdatingExistingOrder(items.indexOf(order))}
-                onItemRemoveClicked={(order) => {
-                  const newOrders = [...items];
-                  newOrders.splice(items.indexOf(order), 1);
-                  setItems(newOrders);
-                }}
+        {(() => {
```
brandones, line 130: Not sure how I feel about this pattern, but alright.

PR #382, opened 2021-09-17
Wrap patient chart in SWRConfig context

denniskigen:
## Requirements

## Summary

This PR adds the [SWR](https://swr.vercel.app/) dependency to the patient chart and wraps the `PatientChart` component in a [SWRConfig context](https://swr.vercel.app/docs/global-configuration) that specifies some sensible global configuration default options for all SWR hooks.

### DIFF
--- a/packages/esm-patient-chart-app/src/root.component.tsx
+++ b/packages/esm-patient-chart-app/src/root.component.tsx
@@ -6,6 +6,12 @@ import { BrowserRouter, Route } from 'react-router-dom';
 import { basePath, dashboardPath, spaRoot } from './constants';
 import styles from './root.scss';
 import { ContextWindowSizeProvider } from './hooks/useContextWindowSize';
+import { SWRConfig } from 'swr';
+
+const swrConfiguration = {
+  // Maximum number of retries when the backend returns an error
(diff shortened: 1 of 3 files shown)


packages/esm-patient-chart-app/src/root.component.tsx
```diff
@@ -6,14 +6,22 @@ import { BrowserRouter, Route } from 'react-router-dom';
...
+};
 
 export default function Root() {
   return (
     <BrowserRouter basename={spaRoot}>
       <ContextWindowSizeProvider>
         <div className={styles.patientChartWrapper}>
           <SideMenu />
-          <Route path={dashboardPath} component={PatientChart} />
+          <SWRConfig value={{ ...swrConfiguration }}>
```
FlorianRappl, line 22: Any reason for re-creating the object? Would `value={swrConfiguration}` be sufficient?

The other question is why we only wrap one route - would it make sense to wrap the whole patient chart wrapper (or even the router) in the config?

denniskigen: 1. Oops.
2. Fair enough

FlorianRappl (approved):

PR #384, opened 2021-09-21
MF-783: Session Check Fixup

vasharma05:
## Requirements

- [ Yes] My work conforms to the [**OpenMRS 3.0 Styleguide**](https://om.rs/styleguide).
- [ Yes] I checked for feature overlap with [**existing widgets**](https://om.rs/directory).

## Summary
It is a fixup for the error that comes up stating "Cannot read property UUID of null" in the patient chart. 

## Screenshots

*None.*

## Related Issue
https://issues.openmrs.org/browse/MF-783

## Other

*None.*


packages/esm-patient-notes-app/src/notes/visit-notes-form.component.tsx
```diff
@@ -208,7 +208,7 @@ const VisitNotesForm: React.FC<VisitNotesFormProps> = ({ patientUuid, isTablet }
   );
 
   const handleSubmit = React.useCallback(
-    (event: SyntheticEvent<HTMLFormElement>) => {
+    (event) => {
```
HerbertYiga, line 211: cc @FlorianRappl

packages/esm-patient-chart-app/src/ui-components/patient-chart.component.tsx
```diff
@@ -19,13 +19,13 @@ interface PatientChartParams {
 const PatientChart: React.FC<RouteComponentProps<PatientChartParams>> = ({ match }) => {
   const { patientUuid, view, subview } = match.params;
   const [loading, patient] = useCurrentPatient(patientUuid);
-  const sessionUser = useSessionUser();
+  const sessionUser = null;
```
manuelroemer, line 22: That seems like a leftover from testing (it should still be `useSessionUser`, not always `null`).

packages/esm-patient-notes-app/src/notes/visit-notes-form.component.tsx
```diff
@@ -208,7 +208,7 @@ const VisitNotesForm: React.FC<VisitNotesFormProps> = ({ patientUuid, isTablet }
   );
 
   const handleSubmit = React.useCallback(
-    (event: SyntheticEvent<HTMLFormElement>) => {
```
manuelroemer, line 211: Is there a reason for removing the type? Having a type here is better then an implicit `any` imo.

packages/esm-patient-chart-app/src/ui-components/patient-chart.component.tsx
vasharma05: Oh yes, so sorry about that. I'll push the changed asap

packages/esm-patient-notes-app/src/notes/visit-notes-form.component.tsx
vasharma05: It was giving an error in this file, actually I think there's an update in carbon's Button, because I'm encountering many errors in management repo too.

manuelroemer: Probably not in Carbon's Button directly, but we no longer import the components from the `.../es/...` submodules which results in us suddenly having TypeScript typings for the components. -> TS can now complain in locations where we had no typings before.
I myself am not getting an error in this location. Maybe you could try pulling from master/updating packages if the error persists on your side. The issue may also just be that the `HTMLFormElement` type conflicts with the button's `onClick` handler type. As a quick fix, you could replace `SyntheticEvent<HTMLFormElement>` with `SyntheticEvent` since the `event.preventDefault` call is the only thing required by the handler.
In any case, fixing the typing instead of removing them is, whenever possible, better.

manuelroemer (approved):
LGTM, thanks!

PR #387, opened 2021-09-22
(enhancement) carbonise appointment form

donaldkibet:
## Requirements

## Summary

1. Carbonise appointment form
2. Fix failing test on visits and medication

## Screenshots

appointment-form

## Related Issue

*None.*

## Other

*None.*


packages/esm-patient-appointments-app/src/appointments/appointments.helpers.ts
```diff
@@ -0,0 +1,15 @@
+export type amPm = 'AM' | 'PM';
+
+export const convertTime12to24 = (time12h, timeFormat: amPm) => {
```
manuelroemer, line 3: If I remember correctly we already have a similar function somewhere else (I think it had something to do with visits). If so, I'd move it to the common-lib to avoid duplication.

packages/esm-patient-appointments-app/src/hooks/useAppointmentService.tsx
```diff
@@ -0,0 +1,77 @@
...
+      },
+    );
+  }, []);
+
+  return useMemo(() => {
+    return { status, services, error };
+  }, [error, services, status]);
+};
+
+export default useAppointmentService;
```
manuelroemer, line 77: Not something that has to be changed now, but maybe something to consider depending on @denniskigen's migration plans:
I'd generally suggest to use SWR for every *new* data fetching hook that is added to the project to avoid later migrations and, most importantly, to reduce the number of different data fetching variants that we already have in the codebase. It would (probably) also make the hook a lot smaller. ;)

packages/esm-patient-appointments-app/src/appointments/appointments-form.scss
```diff
@@ -0,0 +1,70 @@
...
+@import "~carbon-components/src/globals/scss/mixins";
+@import "~carbon-components/scss/globals/scss/vendor/@carbon/elements/scss/type/styles";
+@import "~carbon-components/src/globals/scss/vendor/@carbon/layout/scss/generated/spacing";
+
+.productiveHeading02 {
+  color: $color-gray-70;
+  @include carbon--type-style("productive-heading-02");
+}
+
+$openmrs-background-grey: #ededed;
```
manuelroemer, line 12: Not 100% sure what the decision was here - should such aliases be added to the individual SCSS files or globally in the styleguide? 

CC @FlorianRappl

packages/esm-patient-appointments-app/src/appointments/appointments-form.component.tsx
```diff
@@ -1,279 +1,230 @@
...
+    const { startDate, timeFormat, startTime, serviceType, serviceUuid, comments, locationUuid } = formData;
+    const [hours, minutes] = convertTime12to24(startTime, timeFormat);
+    const startDatetime = new Date(
+      dayjs(startDate).year(),
+      dayjs(startDate).month(),
+      dayjs(startDate).date(),
+      hours,
+      minutes,
+    );
+    dayjs();
```
manuelroemer, line 69: Not a dayjs expert: Is this required or left in by accident?

manuelroemer (commented):
Nice work @donaldkibet! 🚀

packages/esm-patient-appointments-app/src/appointments/appointments-form.component.tsx
donaldkibet: This was left by accident

packages/esm-patient-appointments-app/src/appointments/appointments-form.scss
donaldkibet: This colour isn't provided yet in the colour palette, once this[PR] (https://github.com/openmrs/openmrs-esm-core/pull/184) is merged will remove this locally defined colour

packages/esm-patient-appointments-app/src/hooks/useAppointmentService.tsx
denniskigen: I'll likely change this as part of the SWR migration but as this code doesn't exist yet, I'd prefer we don't do the change here. I'll base my work off of this code.

manuelroemer: Okay, great! cc @donaldkibet

packages/esm-patient-appointments-app/src/appointments/appointments-form.scss
denniskigen: Not 100% sure either. I tried addressing this in https://github.com/openmrs/openmrs-esm-core/pull/184, but I'm not really sure how to move that forward.

packages/esm-patient-appointments-app/src/hooks/useAppointmentService.tsx
manuelroemer: Oh wait, it looks like @donaldkibet already changed this.
@denniskigen With the code already being changed, would you prefer the changes being rolled back? If not, is there anything else that would have to be done in this area?

packages/esm-patient-appointments-app/src/hooks/useAppointmentService.tsx
```diff
@@ -0,0 +1,31 @@
...
+    }
+    if (data) {
+      return 'resolved';
+    }
+  }, [data, error]);
+
+  return useMemo(() => {
+    return { status: status, services: data?.data, error };
+  }, [error, data, status]);
+};
```
manuelroemer, line 29: If we decide to keep this change you can slim this down even further:

```suggestion
const useAppointmentService = () => {
  return useSWR<{ data: Array<Service> }, Error>(
    '/ws/rest/v1/appointmentService/all/full',
    openmrsFetch,
  );
};
```

This can be used like this (i.e. there is no need for keeping a custom status around as it can easily be "calculated" from the SWR result):

```ts
const { data, error } = useAppointmentService();

// status === 'pending' can for example be replaced with this:
const isPending = !error && !data;
```

packages/esm-patient-appointments-app/src/hooks/useAppointmentService.tsx
denniskigen: Not particularly, this looks good enough to me!

packages/esm-patient-appointments-app/src/hooks/useAppointmentService.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+const useAppointmentService = () => {
+  const { data, error } = useSWR<{ data: Array<Service> }, Error>(
+    '/ws/rest/v1/appointmentService/all/full',
+    openmrsFetch,
+  );
+
+  return useMemo(() => {
+    return { data, error };
+  }, [error, data]);
```
manuelroemer, line 14: We're close!

There should be no need to memo these values though, you should really just be able to write:

```ts
const useAppointmentService = () => {
  return useSWR<{ data: Array<Service> }>(
    '/ws/rest/v1/appointmentService/all/full',
    openmrsFetch,
  );
}
```

react-swr will take care to eliminate any unnecessary renders.

donaldkibet: done

denniskigen (approved):
LGTM! Thanks Donald!

PR #388, opened 2021-09-22
MF 707

herbertUG:
## Requirements

## Summary
This PR fixes the issue where we have the workspace height being cut out, it also fixes the issue with

## Screenshots
Screenshot 2021-09-22 at 1 13 31 PM

## Related Issue
*None.*

## Other
*None.*

### DIFF
--- a/packages/esm-patient-chart-app/src/workspace/context-workspace.scss
+++ b/packages/esm-patient-chart-app/src/workspace/context-workspace.scss
@@ -23,10 +23,10 @@
 /* Desktop view */
 :global(.omrs-breakpoint-gt-tablet) .contextWorkspaceContainer {
   z-index: 1;
+  background-color: $ui-02;
   margin-right: 50px;
   flex: 1;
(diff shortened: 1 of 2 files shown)


packages/esm-patient-chart-app/src/workspace/context-workspace.scss
```diff
@@ -17,16 +17,16 @@
   bottom: 0;
   z-index: 10;
   background-color: $ui-01;
-  overflow-y: auto;
+  overflow:hidden;
```
brandones, line 20:
```suggestion
  overflow: hidden;
```

brandones:
Why did you modify the tablet view CSS?

Also I don't think your screenshot demonstrates the fix, since you're looking at a form that is longer than the viewport. The problem is specifically with forms that are shorter than the viewport. Please include screenshots/gifs that demonstrate the fix more clearly. Include tablet mode, if you're also trying to change something about tablet mode.

PR #389, opened 2021-09-22
Add SWR test helpers

denniskigen:
## Requirements

## Summary

Add test helpers specific to SWR. Specifically, this PR adds a reusable custom render method that wraps the render function in an SWR context. 

~~It also re-exports `@testing-library/react` functions for convenience.~~ (Edit: Bad idea, removed)

## Issue

https://issues.openmrs.org/browse/MF-786


tools/test-helpers.tsx
```diff
@@ -14,3 +30,6 @@ export function waitForLoadingToFinish() {
     timeout: 4000,
   });
 }
+
+export * from '@testing-library/react';
```
manuelroemer, line 34: I'm not sure about this reexport. Does it really solve/simplify that much? If I'm not mistaken this only shifts imports around and:
* will make testing-library imports less explicit/discoverable
* require us to write paths like `../../../../tools/test-helpers` instead of a simple `@testing-library/react` (if we don't add workarounds)
* require us to migrate all current imports to the new one (otherwise it's inconsistent).

manuelroemer (commented):
The swr part looks good to me, thanks! Just one question from my side about the testing-library exports.

tools/test-helpers.tsx
denniskigen: Not really, tbh. All I need to do is to wrap the `render` function in an SWR wrapper. 
How about defining a `customRender` function in the test-helpers file that wraps the @testing-library render in an SWRConfig wrapper. 

```tsx
const swrWrapper = ({ children }) => {
  return (
    <SWRConfig
      value={{
        dedupingInterval: 0,
        provider: () => new Map(),
      }}>
      {children}
    </SWRConfig>
  );
};

export const customRender = (ui, options?) => render(ui, { wrapper: swrWrapper, ...options });
```

I'd then use this in place of `render` for tests that use SWR. 

```tsx
function renderAppointments() {
  customRender(<AppointmentsBase {...testProps} />);
}
```

`render` then remains available for whoever wishes to render a component that doesn't use SWR.

manuelroemer: That sounds good to me! Maybe it makes sense to call it something like `renderWithSwrWrapper` or `renderSwr` so that it also sounds like an "extension" of the default `render` (your call really, just making suggestions).

manuelroemer (approved):
LGTM!

PR #393, opened 2021-09-23
MF-713: Test results need to show ALL data for a test, not just from a specific panel

manuelroemer:
## Requirements

## Summary

Proposal to fix MF-713. Also upgrades carbon-charts to fix MF-757 and issues in the trendlines. 

## Screenshots

grafik

## Issue
https://issues.openmrs.org/browse/MF-713
https://issues.openmrs.org/browse/MF-757

## Other

@gracepotma This fix has an interesting side-effect. The test results suddenly display a lot of other panels, e.g. these:
grafik

I don't have deep knowledge of test results as this is literally the first time I'm working with them. Is this acceptable/something that *should* happen?

### DIFF
--- a/packages/esm-patient-test-results-app/src/loadPatientTestData/loadPatientData.ts
+++ b/packages/esm-patient-test-results-app/src/loadPatientTestData/loadPatientData.ts
@@ -76,7 +76,9 @@ async function reloadData(patientUuid: string) {
 
     if (memRef) {
       memRef[0][memRef[1]] = entry;
-    } else {
+    }
+
+    if (obsByClass[entry.conceptClass]) {
       obsByClass[entry.conceptClass].push(entry);
     }
   });
@@ -111,7 +113,6 @@ async function reloadData(patientUuid: string) {
 
 function loadPatientData(patientUuid: string): [PatientData | undefined, Promise<PatientData>] {
   const [cachedPatientData, shouldReload] = getUserDataFromCache(patientUuid);
-
   return [cachedPatientData, shouldReload.then((reload) => (reload ? reloadData(patientUuid) : cachedPatientData))];
 }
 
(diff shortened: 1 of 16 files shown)


packages/esm-patient-test-results-app/src/loadPatientTestData/loadPatientData.ts
```diff
@@ -76,7 +76,9 @@ async function reloadData(patientUuid: string) {
 
     if (memRef) {
       memRef[0][memRef[1]] = entry;
-    } else {
+    }
+
+    if (obsByClass[entry.conceptClass]) {
```
brandones, line 81: This file is not very easy to understand, but I think I'm getting there. Could you explain a bit what's going on here—why this change works?

manuelroemer: > This file is not very easy to understand

I know what you mean - understanding the things going on here took way longer than they should have for these few lines. 

In a nutshell, the function is responsible for fetching and aggregating all of the test result data. It essentially fetches everything available (to some degree, I think) and then places the fetched test result entries into (a) the concept group and (b) the members of lab sets (which are, from my understanding, test results with other test result children).
Previously there was the issue that for single test results, not all results were shown. But they were already fetched, the issue was just that they weren't added to the correct concept category (because they were already added to the LabSet members and because of that the `else` case never triggered before). By removing the `else`, all results always get added to the correct concept category and are appropriately displayed.

manuelroemer: I think that this will also have a follow-up PR to solve the other issue which we talked about in the community standup (the timezone issue and the issue that the LabSets don't display *all* available test results), but from my side we can already merge this one as it solves one of the three issues.

brandones: Sounds good, thanks @manuelroemer . And if at any point while working on this you think of better names for these variables, that would be welcome :)

brandones (approved):

PR #428, opened 2021-10-14
Pius/add extension slot to workspace header

pirupius:
## Summary

This PR adds an extension slot to the context workspace `HeaderGlobalBar`. The extension gives flexibility to add optional additions to the header. 

Our use case in OHRI gives users an option to collapse long forms into a minified view. The video shows an implementation where we added a toggle button.

## Screenshots

https://user-images.githubusercontent.com/15266028/137250078-edea70e6-d975-4327-8cb3-c95e5fdd4a64.mov


gracepotma:
(While I don't agree with the example UX of adding the toggle to the top workspace nav, we all definitely agree we need another slot here :) )

PR #431, opened 2021-10-14
Update deps, remove jsxBracketSameLine rule, run prettier

brandones:
The eslint-prettier plugin doesn't recognize the jsxBracketSameLine rule in `prettier.config.js`. I am not particularly attached to that rule.


brandones:
Merging this to fix the build

PR #455, opened 2021-11-11
03-922: Fix test results overview card border styling

denniskigen:
## Requirements

## Summary

Fixes the border styling of the test result overview cards so that:

- A 2px solid teal border is applied to the active card - the card whose Trendline/Timeline view is currently open. If no card is active, then the topmost card in the layout should be considered the active/selected card. Also, this styling should only be applied when in the Desktop viewport. See Zeplin screen https://zpl.io/aRR4KLN.

- All the cards have a 1px solid grey border around them in the Test Results dashboard view so they match the rest of the patient chart widgets.

Unrelated changes:

- Fix the date format of the ordered date.

## Video

Note how the active state shifts from card to card as I toggle the Trendline / Timeline views. Also, note how the teal styling is not applied when I switch to a tablet viewport.

https://user-images.githubusercontent.com/8509731/141290129-b4e95a66-224a-4361-8afc-b79aaa3fad75.mp4

## Issue

https://issues.openmrs.org/browse/O3-922


brandones:
@denniskigen It looks to me like the reason there's no highlight in the tablet design is because there's no Trendline/Timeline view at all, so there's no need to indicate which panel the Trendline/Timeline corresponds to. Looking at the designs, it seems like in tablet mode, the Trendline/Timeline opens [full-screen](https://app.zeplin.io/project/60d59321e8100b0324762e05/screen/60d5e8575dfb1e04f29e85d2), with the panel at the top.

So for now let's just make sure that the highlight is there whenever there's a split.

denniskigen:
@brandones, that's how I've implemented the feature. The video doesn't demonstrate this as well as it could, but if you watch from 21 seconds onwards, you'll see that the highlight disappears in tablet mode. As you rightly suggest, the widget should span the entire viewport width when in Tablet mode. That'll come in a different PR in the not too distant future.

packages/esm-patient-test-results-app/src/overview/common-overview.tsx
```diff
@@ -126,14 +142,22 @@ const CommonOverview: React.FC<CommonOverviewProps> = ({
           const allNormalResults = !data.some((result) => abnormalInterpretation.includes(result.interpretation));
           const patientSummaryDashboardData = data.slice(0, DashboardResultsCount);
           return (
-            <Card allNormalResults={allNormalResults} key={uuid}>
+            <Card
+              allNormalResults={allNormalResults}
+              key={uuid}
+              isPatientSummaryDashboard={isPatientSummaryDashboard}
```
brandones, line 148: Yikes, I know this is preexising but `isPatientSummaryDashboard` smells fishy to me. That's wayyy too much knowledge for an extension to have about the outside world. What's the actual semantic meaning, within the extension itself? It it like a more minimal form of the extension, or a more compact view, or a mode in which it is able to appear on a dashboard alongside other widgets?

packages/esm-patient-test-results-app/src/overview/common-overview.tsx
```diff
@@ -126,14 +142,22 @@ const CommonOverview: React.FC<CommonOverviewProps> = ({
...
+              isActiveCard={
+                (activeCard && activeCard === uuid) ||
+                (!activeCard && uuid === overviewData[0][overviewData[0].length - 1])
```
brandones, line 151: I don't understand this. There's a default active card? And it's... whatever is the last card? I'm sure I have that wrong

brandones (approved):
Looks great, but I'd love to get the semantic issue about `isPatientSummaryDashboard` fixed up.

EDIT: Oh wait sorry I forgot I had a question, let's get that figured out too. I suspect there should at least be a comment there, if the logic is good.

packages/esm-patient-test-results-app/src/overview/common-datatable.component.tsx
```diff
@@ -0,0 +1,67 @@
...
+              {headers.map((header) => (
+                <TableHeader key={header.key} {...getHeaderProps({ header })} isSortable>
+                  {header.header}
+                </TableHeader>
+              ))}
+            </TableRow>
+          </TableHead>
+          <TableBody>
+            {rows.map((row, i) => (
+              <TypedTableRow key={row.id} interpretation={data[i]?.interpretation} {...getRowProps({ row })}>
```
brandones, line 54: I think things that live in `helpers` that are only used once should just be moved to the file where they are used. I would rather `TypedTableRow` be defined in this same file.

packages/esm-patient-test-results-app/src/overview/common-overview.component.tsx
```diff
@@ -0,0 +1,139 @@
...
+  hideToolbar: true;
+}
+
+type Only<T, U> = {
+  [P in keyof T]: T[P];
+} & {
+  [P in keyof U]?: never;
+};
+
+type Either<T, U> = Only<T, U> | Only<U, T>;
```
brandones, line 36: This doesn't work if `T` contains any properties that `U` contains: [TS Playground #1](https://www.typescriptlang.org/play?#code/C4TwDgpgBA8gdgGxAHgCoBooFUB8UC8UA3gFBRQDaAClAJZxQDWEIA9gGZSoC6AXF9W4BuEgF8oAMmJlKNekxYds3APz84EAG4QATiNEiSoSFACitYAAtdaTLgKxEKDNjwAfR0mRZMqHIfpgXXYAQwBjaABBaXIQ-gAjVlYECBC4GXiEpJS0sRISQODw6AAhGKhMiuzU9PIwrOSavKNwKNYdMsJzKxtIzBL-fLDWOABnYCh2JP5I9s7iKDioYB0AVwhMStCEUehREiA)

We can fix it by adding an Omit, see [TS Playground #2](https://www.typescriptlang.org/play?#code/C4TwDgpgBA8gdgGxAHgCoBooFUB8UC8UA3gFBRQDaAClAJZxQDWEIA9gGZSoC6AXF9W4BuEgF8oAMmJlKNekxYdYAW1rBkWTMzadUObgH5+cCADcIAJxGiRJUJCgBRNQAtLaTLgKxEKDNjwAHx8kDUw9W3pgS3YAQwBjaABBaXJY-gAjVlYECFi4GQzM7Nz8sRISKJiE6AAhVKgixpK8gvJ44pzW8rtwZNYLesJnYDcLZCTMWpxbeNY4AGdgKHSoJIGh4hX+YAsAVwhMJriEBehREjnF5ab1we8iRp39w6gOqBOzqAurpcbYgAmAEZ+HdNo9VrsDt9LvM-hlAQAmUEbB5PKBQ86w67-AEAZhR90Ij3emJhvxugIALITwdsMS8js8DphSS8YUA)

```suggestion
// Defining a disjoint union type `Either`
type Only<T, U> = {
  [P in keyof T]: T[P];
} & {
  [P in keyof Omit<U, keyof T>]?: never;
};

type Either<T, U> = Only<T, U> | Only<U, T>;
```

brandones: I know it doesn't matter for this particular use case, but a good Disjoint Union type is a nice thing to have around :)

packages/esm-patient-test-results-app/src/overview/common-overview.component.tsx
```diff
@@ -0,0 +1,139 @@
...
+  }, [patientUuid]);
+
+  if (!overviewData.length)
+    return <EmptyState headerTitle={t('testResults', 'Test Results')} displayText={t('testResults', 'test results')} />;
+
+  return (
+    <>
+      {(() => {
+        const cards = overviewData.map(([title, type, data, date, uuid]) => (
+          <article
```
brandones, line 69: Nice, now that I'm reviewing the Mozilla docs for `article`, I'm realizing that maybe we should be using that more often in extensions.

packages/esm-patient-test-results-app/src/overview/common-datatable.component.tsx
```diff
@@ -0,0 +1,67 @@
...
+        description={description}
+        {...getTableContainerProps()}
+      >
+        {toolbar}
+        <Table {...getTableProps()} isSortable useZebraStyles>
+          <colgroup className={styles.columns}>
+            <col span={1} />
+            <col span={1} />
+            <col span={1} />
+          </colgroup>
```
brandones, line 42: What's this about?

brandones (approved):
Looks good. Would have rather had the refactor in a separate PR, for future reference. Nice work though.

packages/esm-patient-test-results-app/src/overview/common-datatable.component.tsx
denniskigen: Was intended to get control over the DataTable column widths. A Carbon style override would likely do the trick just as well.

packages/esm-patient-test-results-app/src/overview/common-datatable.component.tsx
denniskigen: I agree. I can fix that in another PR.

denniskigen:
Yeah, I got sucked into the proverbial wormhole with this one. I'll update the ticket description as well as the PR's so it outlines the extra scope that this ended up covering. Thanks for reviewing!

PR #465, opened 2021-11-24
O3-922: Make active card highlight viewport-dependent

denniskigen:
## Requirements

## Summary

https://github.com/openmrs/openmrs-esm-patient-chart/pull/455 added a teal border to the active card in the test results overview. One omission from that PR was that this highlight should only appear in the desktop viewport. This PR adds a check that enforces that condition. 

## Issue

https://issues.openmrs.org/projects/MF/issues/O3-922

### DIFF
--- a/packages/esm-patient-test-results-app/src/desktop-view/desktop-view.component.tsx
+++ b/packages/esm-patient-test-results-app/src/desktop-view/desktop-view.component.tsx
@@ -40,7 +40,15 @@ const deduceViewState = ({ panelUuid, testUuid, type = 'none' }): ViewState => {
   }
 };
 
-const DesktopView: React.FC<Record<string, any>> = ({ patientUuid, panelUuid, testUuid, type, basePath }) => {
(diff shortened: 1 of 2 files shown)


brandones:
I continue to think, as I wrote in #455 , that if both the left column of cards and the right column of graphed results is visible, the active card in the left must be highlighted. On tablet, the results graph is full-screen, so the left column of cards is never visible when one is "active," so it doesn't matter whether it's highlighted or not. So I'd advocate closing this. It will be irrelevant once the results graph behaves correctly; and highlighting the card is the more "generally correct" behavior (i.e., if Ciaran decided that actually we should keep the screen split in tablet mode, we would definitely want to be highlighting the active card).

denniskigen:
Totally right @brandones, this was intended as short term fix that would most certainly by obviated by the incoming tablet design revisions.

brandones:
Cool, thanks @denniskigen . Unless it's urgently needed for some reason, I'll just close it out.

PR #481, opened 2021-12-16
bug fix - fix wrong bmi calculations

donaldkibet:
## Requirements

## Summary

- This PR fixes issue with the calculation of BMI
- Update the biometrics chart background to match designs

#### Before
Screenshot 2021-12-16 at 11 30 15

#### After
Screenshot 2021-12-16 at 11 31 40

## Issue

*None.*

## Other

*None.*

### DIFF
--- a/packages/esm-patient-biometrics-app/src/biometrics/biometrics.resource.tsx
+++ b/packages/esm-patient-biometrics-app/src/biometrics/biometrics.resource.tsx
@@ -1,6 +1,7 @@
 import useSWR from 'swr';
 import { fhirBaseUrl, FHIRResource, useConfig, openmrsFetch } from '@openmrs/esm-framework';
 import { calculateBMI } from './biometrics-helpers';
+import { toDateWithoutSeconds } from '@openmrs/esm-patient-common-lib';
 
 export const pageSize = 100;
 
@@ -25,9 +26,9 @@ export function useBiometrics(patientUuid: string) {
     data:
       data?.data?.total > 0
         ? formatDimensions(
-            filterByConceptUuid(observations, concepts.muacUuid),
             filterByConceptUuid(observations, concepts.heightUuid),
             filterByConceptUuid(observations, concepts.weightUuid),
+            filterByConceptUuid(observations, concepts.muacUuid),
           )
         : null,
     isError: error,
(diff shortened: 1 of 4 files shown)


packages/esm-patient-biometrics-app/src/biometrics/biometrics.resource.tsx
```diff
@@ -1,6 +1,7 @@
 import useSWR from 'swr';
 import { fhirBaseUrl, FHIRResource, useConfig, openmrsFetch } from '@openmrs/esm-framework';
 import { calculateBMI } from './biometrics-helpers';
+import { toDateWithoutSeconds } from '@openmrs/esm-patient-common-lib';
```
jonathandick, line 4: is this leveraging the recent PR from brandon?

brandones: It's not. I didn't write a function that removes seconds from a date.

Why were multiple lines being generated in the first place? i.e. why does some data seem to have multiple copies that differ by seconds?

donaldkibet: The reason why some data has multiple copies is because it is being capture at different times, the difference is in milliseconds.

brandones: Oh okay, so the data is all submitted to the server at the same moment, and the time difference is between the server saving each thing to the database. That is very goofy.

The way you're solving this is okay. A more formally correct way to do it would be to check whether the difference between times is less than some value. Then it behaves correctly even in the unlikely scenario where the millisecond difference runs over a minute boundary. It would also be more expressive of the intent of the code (though it wasn't hard to understand what you did).

donaldkibet: @brandones i have removed the date manipulation function to be done correctly on a different PR

brandones: Ok @donaldkibet can you update the PR title to reflect what this PR now does (removes biometrics from vitals display, as best I can tell)?

donaldkibet: @brandones is there further feedback on this?

brandones: Does it fix the wrong BMI calculations? Looking at the code I don't understand how it does that.

brandones (approved):

brandones:
Looks like this broke the build. Can you take a look, @donaldkibet ? https://github.com/openmrs/openmrs-esm-patient-chart/actions/runs/1663863145

It's weird, because all the actions for this PR passed. Something similar happened yesterday with a PR from @denniskigen . I wonder if something is up with the GitHub Actions.

PR #491, opened 2022-01-08
O3-905  Patient chart tests depend on time zone

brandones:
## Requirements

## Summary

Without this, four tests fail when run from a computer on Pacific Time. With this change, the tests pass regardless of the time zone of the computer running them.

## Issue

https://issues.openmrs.org/browse/O3-905


denniskigen (approved):
Nifty!

PR #492, opened 2022-01-10
O3-683: Independently scrollable chart body and workspace

denniskigen:
## Requirements

## Summary

This ensures that the user can scroll independently between the patient chart body and the workspace when a form is open.

## Video

> Note that because the detach function is currently borked, closing or discarding an active form from the workspace would not work. This is why I have to override the active form each time I launch a new one in the video below.

https://user-images.githubusercontent.com/8509731/148730910-1e6e32e7-c415-4e8d-867c-1e0ce005da2b.mp4

## Issue

https://issues.openmrs.org/projects/MF/issues/O3-683

### DIFF
--- a/packages/esm-patient-chart-app/src/root.scss
+++ b/packages/esm-patient-chart-app/src/root.scss
@@ -1,5 +1,6 @@
-@import "carbon-components/scss/globals/scss/typography.scss";
 @import "~@openmrs/esm-styleguide/src/vars";
+@import "~carbon-components/src/globals/scss/vars";
+@import "~carbon-components/src/globals/scss/mixins";
 
 .patientChartWrapper {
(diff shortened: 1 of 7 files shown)


gracepotma:
@denniskigen what happens when BOTH the patient chart AND the form need to be scrolled? 
I know the user is unblocked right now thanks to this PR, *IF* their mouse has a scroll wheel or they have a trackpad. But if they don't have a scrollwheel on their mouse, how would they control which screen side to scroll? 

Shouldn't there be a scroll bar for both sections, like this: 
image

I'll create a separate ticket for this so we can mark this one done.

denniskigen:
Do the concerns raised above still apply, @gracepotma?

PR #499, opened 2022-01-12
(enhancement) Add ability for form-entry to display scheduled appointments

donaldkibet:
## Requirements

## Summary

- This PR seeks to add the ability to view schedule appointment once the RTC date is selected. However this brings up a new challenge of how to support this feature with implementation. that don't have etl to process this information. In the broader sense some aspects of form-entry are dependent on etl.
@enyachoke @jonathandick 

## Screenshots
Screenshot 2022-01-12 at 12 11 29

## Issue

*None.*

## Other

*None.*

### DIFF
--- a/packages/esm-form-entry-app/src/app/services/monthly-scheduled-resource.service.ts
+++ b/packages/esm-form-entry-app/src/app/services/monthly-scheduled-resource.service.ts
@@ -1,15 +1,16 @@
-import { HttpClient, HttpHeaders } from '@angular/common/http';
 import { Injectable } from '@angular/core';
+import { HttpClient } from '@angular/common/http';
+
+import { MonthlyScheduleParams } from '../types';
+import { WindowRef } from '../window-ref';
 
 @Injectable()
 export class MonthlyScheduleResourceService {
-  constructor(private http: HttpClient) {}
+  constructor(private http: HttpClient, private windowRef: WindowRef) {}
 
-  public getMonthlySchedule(params) {
-    const requestHeader = new HttpHeaders().set('Authorization', 'Basic ZGtpYmV0Ol9Eb25hbGQyMDE5Kg==');
-    return this.http.get(
-      'https://ngx.ampath.or.ke/etl-latest/etl/get-monthly-schedule?startDate=2022-01-31&endDate=2022-02-04&locationUuids=18c343eb-b353-462a-9139-b16606e6b6c2&limit=5&programType=781d85b0-1359-11df-a1f1-0026b9348838,781d897a-1359-11df-a1f1-0026b9348838,96047aaf-7ab3-45e9-be6a-b61810fe617d,c19aec66-1a40-4588-9b03-b6be55a8dd1d,f7793d42-11ac-4cfd-9b35-e0a21a7a7c31,334c9e98-173f-4454-a8ce-f80b20b7fdf0,96ba279b-b23b-4e78-aba9-dcbd46a96b7b,781d8880-1359-11df-a1f1-0026b9348838&groupBy=groupByPerson,groupByAttendedDate,groupByRtcDate',
-      { headers: requestHeader },
-    );
+  public getMonthlySchedule(params: MonthlyScheduleParams) {
+    const etlUrl = `/etl-latest/etl/get-monthly-schedule?startDate=${params.startDate}&endDate=${params.endDate}&locationUuids=${params.locationUuids}&limit=${params.limit}&programType=${params.programType}&groupBy=groupByPerson,groupByAttendedDate,groupByRtcDate`;
+    const url = this.windowRef.nativeWindow.openmrsBase.concat(etlUrl);
+    return this.http.get(url);
   }
 }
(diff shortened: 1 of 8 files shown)


jonathandick (changes requested):
@donaldkibet unfortunately, this looks too ampath specific. Would it be possible to create this as an extension within the ampath repo but create the ability to load it into the form? @enyachoke any ideas on how to implement this?

jonathandick:
@donaldkibet any updates on the plan for this? I realize that because this is in angualr we can't simply create an extension. The problem is that the route you're using to request data to populate this is specific to ampath. Perhaps there's a way to make that a configurable property? Though I think that's a last resort. Do you have any other ideas for how to handle this?

brandones:
> in angualr we can't simply create an extension

We should be able to do this; it is part of the point of the extension system.

But yes, interested in hearing @donaldkibet 's ideas about what to do with this.

donaldkibet:
> @donaldkibet any updates on the plan for this? I realize that because this is in angualr we can't simply create an extension. The problem is that the route you're using to request data to populate this is specific to ampath. Perhaps there's a way to make that a configurable property? Though I think that's a last resort. Do you have any other ideas for how to handle this?

The easy route is to pass config into the form-entry extension and configure the data-source like this one for appointment scheduling. Alternatively add configurability to the form-engine itself since the one provided is tightly coupled with react

brandones:
> Alternatively add configurability to the form-engine itself since the one provided is tightly coupled with react

You don't mean the config stuff in esm-framework, do you? You can use the config anywhere with

```ts
getConfigStore("@openmrs/my-module").subscribe(store => {
  if (store.loaded && store.config) {
    doStuff(store.config);
  }
);
```

So creating a wrapper for Angular or any other framework should be cake. The only reason the React wrapper (`useConfig`) is as complicated as it is is because it also
1) Gets the module name from context
2) Uses Suspense to avoid rendering until the config loads
3) Automatically pulls the right config whether it is being used in an extension or not

packages/esm-form-entry-app/src/app/services/monthly-scheduled-resource.service.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import { MonthlyScheduleParams } from '../types';
+import { WindowRef } from '../window-ref';
+
+@Injectable()
+export class MonthlyScheduleResourceService {
+  constructor(private http: HttpClient, private windowRef: WindowRef) {}
+
+  public getMonthlySchedule(params: MonthlyScheduleParams) {
+    const etlUrl = `/etl-latest/etl/get-monthly-schedule?startDate=${params.startDate}&endDate=${params.endDate}&locationUuids=${params.locationUuids}&limit=${params.limit}&programType=${params.programType}&groupBy=groupByPerson,groupByAttendedDate,groupByRtcDate`;
```
jonathandick, line 12: @donaldkibet this line remains a problem (I think). This is specific to ampath and needs to be configuration. I will also file a jira ticket to faciliate creation of a similar endpoint based on the bahmni appointments data model.

donaldkibet: @jonathandick i have made the etl filed configurable

jonathandick (approved):
Just one last change and then you can merge

packages/esm-form-entry-app/src/config-schema.ts
```diff
@@ -0,0 +1,16 @@
+import { Type } from '@openmrs/esm-framework';
+
+export const configSchema = {
+  dataSources: {
+    monthlySchedule: {
+      _type: Type.Boolean,
+      _default: false,
+      _description: 'Whether to inject monthly scheduled appointment data source to form-entry engine',
```
brandones, line 8:
```suggestion
      _description: 'Whether to use monthly scheduled appointment data source in form-entry engine. Requires `baseEtlUrl`.',
```

packages/esm-form-entry-app/src/config-schema.ts
```diff
@@ -0,0 +1,16 @@
...
+    },
+  },
+  baseEtlUrl: {
+    _type: Type.String,
+    _default: '/etl-latest/etl/',
+    _description: 'Custom URL to load resources required for form-entry',
```
brandones, line 14:
```suggestion
    _description: 'Custom URL to load resources required for configurable form-entry features (under `dataSources`).',
```

donaldkibet:
> Just one last change and then you can merge

@jonathandick please point to the final bits that needs change

packages/esm-form-entry-app/src/config-schema.ts
```diff
@@ -0,0 +1,17 @@
...
+export const configSchema = {
+  dataSources: {
+    monthlySchedule: {
+      _type: Type.Boolean,
+      _default: false,
+      _description:
+        'Whether to use monthly scheduled appointment data source in form-entry engine. Requires `baseEtlUrl`.',
+    },
+  },
+  baseEtlUrl: {
```
jonathandick, line 12: Please change name of baseETLUrl to "appointmentsResourceUrl" and then this can be merged

jonathandick (approved):

packages/esm-form-entry-app/src/config-schema.ts
brandones: In that case @donaldkibet  you'll also want to change the `_default` to point directly at the appointments resource, instead of being the base URL, and change the usages appropriately.

donaldkibet:
@jonathandick @brandones should this go in now

PR #503, opened 2022-01-13
O3-976: Share Patient Chart Common Lib

FlorianRappl:
## Requirements

## Summary

Right now the common lib is duplicated for each MF of patient-chart. Now, with dynamic module sharing in place we should actually add it to the dynamically shared libs.

## Screenshots

*None.*

## Issue

See [O3-976](https://issues.openmrs.org/browse/O3-976?filter=-1).

## Other

For whatever reason I cannot log in locally. Always get unauthenticated as response when entering the credentials. This was the same behavior as online at https://openmrs-spa.org/openmrs/spa/login. So I guess its a current backend issue.

### DIFF
--- a/packages/esm-generic-patient-widgets-app/package.json
+++ b/packages/esm-generic-patient-widgets-app/package.json
@@ -43,6 +43,7 @@
   },
   "peerDependencies": {
     "@openmrs/esm-framework": "3.x",
+    "@openmrs/esm-patient-common-lib": "3.x",
     "carbon-components": "10.x",
     "carbon-icons": "7.x",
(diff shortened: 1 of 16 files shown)


manuelroemer (approved):
LGTM!

> For whatever reason I cannot log in locally. Always get unauthenticated as response when entering the credentials.

I had the same issue with the `admin` account, but `admin2` (same PW) still works for me. Maybe that also works for you.

FlorianRappl:
> I had the same issue with the admin account, but admin2 (same PW) still works for me. Maybe that also works for you.

Works - thanks!

PR #505, opened 2022-01-14
O3-780: Add bundle size monitoring to PR actions

FlorianRappl:
## Requirements

## Summary

We should be tracking bundle size changes. We want a tool that comments on every PR with the bundle size change caused by the PR.

## Screenshots

*None.*

## Issue

[O3-780](https://issues.openmrs.org/browse/O3-780)

## Other

*None.*

### DIFF
--- a/.github/workflows/size.yml
+++ b/.github/workflows/size.yml
@@ -1,0 +1,22 @@
+name: Report bundle size
+
+on:
+  pull_request:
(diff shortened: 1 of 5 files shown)


FlorianRappl:
(the change should be OK, its fine that the report bundle size fails right now as the file is not yet in `master`)

PR #509, opened 2022-01-18
O3-1011: Fix Start Visit form time picker validator regex

denniskigen:
## Requirements

## Summary

The regular expression supplied to the pattern prop of the TimePicker input in the Start Visit form isn't working as it should. It's currently excluding some valid dates. 

This commit fixes the regex so it matches all valid 12-hour clock values.

## Screenshots

> This tooltip was incorrectly shown for a visit started at 3:14 PM.

Screenshot 2022-01-12 at 15 16 53


jonathandick (approved):
LGTM

PR #529, opened 2022-01-27
(bug) fix test-results not showing on encounters widget

donaldkibet:
## Requirements

## Summary

- update visit custom representation to full, this is because for instance not using drug order endpoint, this will cause the REST API to fail e.g AMRS.
- Add test-results view on the encounters widget to display test-results.
- Add configurable concepts for encounters instead of using fixed concept name on visit-detail component.
- Updated the global toggle button on encounters to be `All Encounters` and for individual encounter to simply `encounters`
- Add concept name for each clinical type as shown in the gif below.

## Screenshots

encounters

## Issue

*None.*

## Other

*None.*


jonathandick (approved):
LGTM

PR #532, opened 2022-01-31
(enhancement) squash console test results error, export visit-details as extension slot

donaldkibet:
## Requirements

## Summary

- export `visit-detail` component as extension to be used on active-visit, the code is similar, therefore no need to have duplicate code doing the same thing.  [see here](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-active-visits-app/src/visits-summary).
- squash console error when opening `test-results` as seen here below 
Screenshot 2022-02-07 at 16 50 03

## Screenshots

*None.*

## Issue

*None.*

## Other

*None.*

### DIFF
--- a/packages/esm-patient-chart-app/src/index.ts
+++ b/packages/esm-patient-chart-app/src/index.ts
@@ -148,6 +148,14 @@ function setupOpenMRS() {
           moduleName,
         }),
       },
+      {
+        id: 'visit-detail-overview',
+        slot: 'visit-detail-overview-slot',
+        load: getAsyncLifecycle(() => import('./visit/visits-widget/past-visits-components/visit-detail.component'), {
+          featureName: 'visit-details',
+          moduleName,
+        }),
+      },
     ],
   };
 }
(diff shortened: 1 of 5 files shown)


jonathandick (commented):
@donaldkibet could you add some text to the description of this PR for why you are doing what you're doing (especially regarding the extension slot).

packages/esm-patient-chart-app/src/index.ts
```diff
@@ -148,6 +148,14 @@ function setupOpenMRS() {
           moduleName,
         }),
       },
+      {
+        id: 'visit-detail-overview',
```
brandones, line 152:
```suggestion
        id: 'visit-detail',
```

?

Calling something "xyz-detail-overview" doesn't really make sense to me. Is it a detail view, or is it an overview? They're opposites.

brandones:
This changes some UI, right? Adds an extension to a slot somewhere? A screenshot would be very helpful.

PR #537, opened 2022-02-04
Bump patch version on release

ibacher:
## Requirements

## Summary

I don't think this is the *best* way to do things, but the current system is even more broken, so let's adopt the "finalise version at release" strategy.

## Screenshots

*None.*

## Issue

*None.*

## Other

*None.*


brandones (approved):
Along with https://github.com/openmrs/openmrs-esm-patient-management/pull/96 , looks good!

PR #554, opened 2022-02-15
03-1104: Hide component forms

donaldkibet:
## Requirements

## Summary

- Hide component forms as these forms aren't required by to be viewed but are referenced form

## Screenshots

*None.*

## Issue

*None.*

## Other

*None.*

### DIFF
--- a/packages/esm-patient-forms-app/src/hooks/use-forms.ts
+++ b/packages/esm-patient-forms-app/src/hooks/use-forms.ts
@@ -8,8 +8,8 @@ import { isFormFullyCached } from '../offline-forms/offline-form-helpers';
 export function useFormEncounters(cachedOfflineFormsOnly = false) {
   return useSWR([formEncounterUrl, cachedOfflineFormsOnly], async () => {
     const res = await openmrsFetch<ListResponse<FormEncounter>>(formEncounterUrl);
-    // Only show published forms
-    const forms = res.data?.results?.filter((form) => form.published) ?? [];
+    // show published forms and hide component forms
+    const forms = res.data?.results?.filter((form) => form.published && !form.name.includes('component')) ?? [];
 
     if (!cachedOfflineFormsOnly) {
       return forms;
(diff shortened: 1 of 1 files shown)


packages/esm-patient-forms-app/src/hooks/use-forms.ts
```diff
@@ -8,8 +8,8 @@ import { isFormFullyCached } from '../offline-forms/offline-form-helpers';
 export function useFormEncounters(cachedOfflineFormsOnly = false) {
   return useSWR([formEncounterUrl, cachedOfflineFormsOnly], async () => {
     const res = await openmrsFetch<ListResponse<FormEncounter>>(formEncounterUrl);
-    // Only show published forms
-    const forms = res.data?.results?.filter((form) => form.published) ?? [];
+    // show published forms and hide component forms
+    const forms = res.data?.results?.filter((form) => form.published && !form.name.includes('component')) ?? [];
```
denniskigen, line 12: Screenshot 2022-02-15 at 22 46 03

[String.prototype.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) makes case-sensitive matches. Won't the check fail for `Component_POC`? 

Perhaps it should be `!form.name.includes('Component_POC')` instead.

donaldkibet: That will work for the spa environment, the naming convention adopted is to always name the component forms starting with `component` key word. This `!form.name.includes('Component_POC')` won't work for `component_cdm...` as shown here below
Screenshot 2022-02-16 at 09 29 38

consequently i have updated for this to work on both ends `!formName.toLowerCase().includes('component')`

packages/esm-patient-forms-app/src/hooks/use-forms.ts
```diff
@@ -67,3 +67,5 @@ function mapToFormCompletedInfo(
     };
   });
 }
+
+const removeComponentForm = (formName: string): boolean => !formName.toLowerCase().includes('component');
```
ZacButko, line 71: This function name doesn't accurately describe what it does. It would be better called something like 'isNotComponent'.
However, for boolean values it's better to put it in the positive, so it would be better format the function as 'isComponent', then do the negation on line 12.

packages/esm-patient-forms-app/src/hooks/use-forms.ts
denniskigen: Nice, how about a one-liner like:
 
`const forms = res.data?.results?.filter((form) => form.published && !/component_/i.test(form.name)) ?? [];`

denniskigen (approved):
LGTM. Thanks, @donaldkibet! Please feel free to merge.

PR #557, opened 2022-02-16
[O3-1060] Data Timeline FilterSet

ZacButko:
## Requirements

## Summary

Introduces a FilterSet component which is able to take a set of concepts and filter out individual leaf tests or entire concept nodes. This uses a new FilterContext which is shared with the temporary MultipleTimelines component to render.

Compare to target spec https://app.zeplin.io/project/60d59321e8100b0324762e05/screen/60d5ea5bf120a6089c660a68

A lot of the data loading logic is temporary and meant to be replaced by a new endpoint

## Screenshots

FilterSet

## Issue

Relates to https://issues.openmrs.org/browse/O3-1060
Still left to do:
- search function
- combine results into one timeline
- better page layout

## Other

Had to silence failing test in `side-menu.component.test.txt` in order to push PR


denniskigen (commented):
Really like the look of this, @ZacButko! Nice work.

packages/esm-patient-test-results-app/src/hiv/HIVCareAndTreatment.tsx
```diff
@@ -0,0 +1,53 @@
...
+import concepts from './concepts';
+import FilterSet from '../filter/FilterSet';
+import FilterContext, { FilterProvider } from '../filter/FilterContext';
+import usePatientResultsData from '../loadPatientTestData/usePatientResultsData';
+import { usePatient } from '@openmrs/esm-framework';
+import { MultiTimeline } from '../timeline/Timeline';
+import { EmptyState, ErrorState } from '@openmrs/esm-patient-common-lib';
+
+const DataLoader = ({ sortedObs }) => {
+  // this is a bad way to do this. Appoogies in advance
```
ibacher, line 12:
```suggestion
  // this is a bad way to do this. Apologies in advance
```

packages/esm-patient-test-results-app/src/hiv/HIVCareAndTreatment.tsx
```diff
@@ -0,0 +1,53 @@
...
+const HIVCareAndTreatment = () => {
+  const patient = usePatient();
+  const patientUuid = patient?.patient?.id;
+  const { sortedObs, loaded, error } = usePatientResultsData(patientUuid);
+  return (
+    <FilterProvider>
+      {!loaded && <InlineLoading />}
+      {error && <ErrorState error={error} headerTitle="Data Load Error" />}
+      {loaded && !error && sortedObs && !!Object.keys(sortedObs).length && (
+        <>
```
ibacher, line 32: Wouldn't it be cleaner and more appropriate to put the `FilterProvider` here in the tree and refactor things so we don't need the weird `DataLoader` component? E.g.,

```ts
    <>
      {!loaded && <InlineLoading />}
      {error && <ErrorState error={error} headerTitle="Data Load Error" />}
      {loaded && !error && sortedObs && !!Object.keys(sortedObs).length && (
        <FilterProvider sortedObs={sortedObs}>
```

ibacher (commented):
@ZacButko Thanks for this! This probably needs a deeper look than what I've done, but here's a few quick thoughts.

packages/esm-patient-test-results-app/src/hiv/concepts.ts
```diff
@@ -0,0 +1,58 @@
+const concepts = {
+  display: 'Bloodwork',
+  concept: 'ConvSet', // convienence set
+  obs: [],
```
dkayiwa, line 4: Given that the set itself will not have obs, do we need the above `obs: []`?

packages/esm-patient-test-results-app/src/hiv/concepts.ts
```diff
@@ -0,0 +1,58 @@
+const concepts = {
+  display: 'Bloodwork',
+  concept: 'ConvSet', // convienence set
```
dkayiwa, line 3: Since you already have the `display`, are you using `concept: 'ConvSet'` for anything?

packages/esm-patient-test-results-app/src/hiv/concepts.ts
```diff
@@ -0,0 +1,58 @@
...
+  obs: [],
+  subSets: [
+    {
+      display: 'Hematology',
+      concept: 'ConvSet',
+      obs: [],
+      subSets: [
+        {
+          display: 'Complete Blood Count',
+          concept: 'LabSet',
```
dkayiwa, line 13: Since you already have the `display`, are you using `concept: 'LabSet'` for anything?

packages/esm-patient-test-results-app/src/hiv/concepts.ts
```diff
@@ -0,0 +1,58 @@
...
+          obs: [
+            {
+              display: 'Hematocrit',
```
dkayiwa, line 16: Would you want the `value` here?

dkayiwa: In addition to the value, do you also need the `obsDatetime`?

packages/esm-patient-chart-app/src/view-components/side-menu.component.test.tsx
```diff
@@ -12,7 +12,7 @@ jest.mock('@openmrs/esm-framework', () => ({
 
 jest.mock('./side-menu.component.scss', () => ({}));
 
-describe('sidemenu', () => {
+describe.skip('sidemenu', () => {
```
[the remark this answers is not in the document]
ZacButko: Hopefully not! This test was failing for me on master. Not sure what caused it, but it was preventing me from pushing this PR. 

Does anyone happen to know what's causing this test to fail?

packages/esm-patient-test-results-app/src/hiv/HIVCareAndTreatment.tsx
```diff
@@ -0,0 +1,53 @@
...
+    if (tests.length && !Object.keys(state?.checkboxes).length) {
+      initialize(Object.fromEntries(tests.map((test) => [test, false])), concepts);
+    }
+  }, [sortedObs, initialize, state]);
+  return null;
+};
+
+const HIVCareAndTreatment = () => {
+  const patient = usePatient();
+  const patientUuid = patient?.patient?.id;
```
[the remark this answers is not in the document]
ZacButko: Thanks! Somehow I missed that patientUuid was already provided from usePatient

packages/esm-patient-test-results-app/src/hiv/concepts.ts
ZacButko: No, it doesn't look like `concept` is currently being used. I will delete it.

ZacButko: Although, It might help later if we want to group findings from a convenience set together on the timeline. Is this difficult to provide also? I think this is cheap and might provide some good benefits for FE if it is existing.

packages/esm-patient-test-results-app/src/hiv/concepts.ts
ZacButko: No, we don't need to pass a key if it is empty

packages/esm-patient-test-results-app/src/hiv/concepts.ts
ZacButko: Currently there is no use for this on the frontend. However it could provide some additional information if we later decide to extend the feature. I will remove this for now.

packages/esm-patient-test-results-app/src/hiv/concepts.ts
ZacButko: Yes, value would go here. I am not familiar yet with the observation data structure, but all data necessary to populate the timeline would need to be included here.

packages/esm-patient-test-results-app/src/timeline/useTimelineData.ts
```diff
@@ -98,3 +96,78 @@ export const useTimelineData = (patientUuid: string, panelUuid?: string) => {
...
+      };
+    // look for the specified panelUuid. If none is specified, just take any obs
+    const obs = Object.entries(sortedObs) || [];
+
+    //   if (!panelData)
+    //     return {
+    //       data: { parsedTime: {} as ReturnType<typeof parseTime> },
+    //       loaded,
+    //       error: new Error('panel data missing'),
+    //     };
```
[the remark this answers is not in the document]
ZacButko: Marie Kondo is the best Kondo. Turns out I wasn't using the entire function :)

packages/esm-patient-test-results-app/src/hiv/HIVCareAndTreatment.tsx
ZacButko: Let's try it and see how it works

packages/esm-patient-chart-app/src/view-components/side-menu.component.test.tsx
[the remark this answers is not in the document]
ZacButko: ● sidemenu › is rendered when viewport != tablet

    TypeError: window.matchMedia is not a function

packages/esm-patient-test-results-app/src/hiv/concepts.ts
dkayiwa: I do not want this to be based on the observation data structure (which is not client friendly in the first place), but rather on what you need to accomplish your task. 😊

packages/esm-patient-test-results-app/src/hiv/concepts.ts
dkayiwa: I generally tend to avoid the temptation of including/implementing what is not currently needed. I postpone it to the future when the times comes for it. You will be shocked at many unused features in software, which were added with the hope that they may be needed in future, but they never ever get needed. 😊

packages/esm-patient-test-results-app/src/hiv/concepts.ts
dkayiwa: You got it right that it is cheap to add. But given the fact that we are developing for resource constrained environments, where the more data we return to the client, the more client side memory and bandwidth we consume, i would like to return only the bare minimum that you need for now. We can always easily add new fields as we need them. On the backend, am implementing this in the REST webservices module which has more flexible and rapid release cycles than the core platform.

packages/esm-patient-test-results-app/src/hiv/concepts.ts
ibacher: We'd probably need: `obsDateTime`, `value`, `interpretation` at a minimum. Possibly `encounter` and `status` (where it exists). IIRC, the current timeline is built on top of the FHIR API. The main difference there is we provide the the reference ranges (from the concept) along-side the actual value (which is important for the visual indicators generated for concerning results).

packages/esm-patient-test-results-app/src/hiv/concepts.ts
ZacButko: Ok, let's drop it for now and think about adding later

packages/esm-patient-test-results-app/src/hiv/concepts.ts
ZacButko: Sounds good

packages/esm-patient-test-results-app/src/hiv/concepts.ts
ZacButko: @dkayiwa  Let me start to de-structure the current timeline so I can tell you exactly what fields will be needed here.

packages/esm-patient-chart-app/src/view-components/side-menu.component.test.tsx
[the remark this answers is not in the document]
denniskigen: I'm surprised that I don't get this test failure when I run tests on your branch on my machine. Zero context w.r.t to a solution - just thought I'd mention this.

Screenshot 2022-02-17 at 21 04 28
.

packages/esm-patient-test-results-app/src/hiv/concepts.ts
bmamlin: +1 to @ibacher's list. In addition...

For the concept (row):
* A uuid or url to the obs concept (you may only need a localized name for display, but concept names – especially short names for use in tables – are not unique)
* `units` (already part of the requirements for data to be displayed)

For the obs (values):
* `comment` (if not null) should always be included and shown (via footnote, tooltip, or other drill-down), since it may contain 
* A uuid or url to the obs, so it's straightforward to get additional detail of the obs when a user interacts with a specific value in the table

packages/esm-patient-chart-app/src/view-components/side-menu.component.test.tsx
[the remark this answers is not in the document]
ZacButko: (Magically??) Resolved on my local machine. I uncommented this test in latest commit.

packages/esm-patient-test-results-app/src/index.ts
```diff
@@ -9,6 +9,13 @@ import { createDashboardLink } from '@openmrs/esm-patient-common-lib';
 import { configSchema } from './config-schema';
 import { dashboardMeta } from './dashboard.meta';
 
+export const hivMeta = {
+  name: 'Hiv-care-and-treatment',
+  slot: 'Hiv dashboard',
```
donaldkibet, line 14:
```suggestion
  slot: 'hiv-dashboard-slot',
```

packages/esm-patient-test-results-app/src/hiv/concepts.ts
ZacButko: We will refine the endpoint and endpoint contents on subsequent PR. For this PR - FilterSet component I think it is ok to proceed without this being finalized. Thanks for the initial endpoint @dkayiwa

packages/esm-patient-test-results-app/src/index.ts
donaldkibet: convention to name slots with a `slot` suffix

donaldkibet (approved):
🚀

packages/esm-patient-test-results-app/src/index.ts
ZacButko: I've noticed that the kebab case syntax is common throughout the project. However I noticed in the Getting Started docs it says that the human readable name is preferred for Slot names? http://o3-dev.docs.openmrs.org/#/main/extensions?id=naming-slots Perhaps @brandones  can clarify

donaldkibet: Okay 👍🏽

jonathandick: As a user of the implementer tools, I find it much easier to identify that something is a slot when it's name ends in "-slot". I believe that's been our convention. @brandones?

brandones: Yeah I think I was foolish for me not to push against this harder. The obvious solution is that the implementer tools should just make it clearer what is a slot and what is not. If the best way to do this is to append "slot" when displaying the names, that should be done automatically by the tooling, not as a convention by humans who can and will screw it up.

jonathandick: For the time being, until this change is made by the implementer tools, could we keep this convention (and get this PR merged)?

brandones: Fine by me

PR #560, opened 2022-02-16
(feat) Better dashboard titles in breadcrumbs menu

denniskigen:
## Requirements

## Summary

This improves the display of dashboard titles in the breadcrumbs menu. More specifically, it improves how dashboard titles containing more than one word are displayed. It introduces a convention where the dashboard `name` is an underscore delimited string (e.g. `clinical_views`). Then in the `title` property of the `registerBreadcrumbs` invocation, we replace the underscores from the string with spaces and capitalize the result. 

With this, we end up with `Vitals and biometrics dashboard` instead of `Vitalsandbiometrics Dashboard` in the current case.

## Screenshots

> Before
image

> After

> Dashboard names consisting of more than one word 
Screenshot 2022-02-16 at 23 35 45
Screenshot 2022-02-16 at 23 35 29
Screenshot 2022-02-16 at 23 35 23

> One word dashboard names (these remain as before)
Screenshot 2022-02-16 at 23 35 16
Screenshot 2022-02-16 at 23 35 39

### DIFF
--- a/packages/esm-patient-biometrics-app/src/dashboard.meta.tsx
+++ b/packages/esm-patient-biometrics-app/src/dashboard.meta.tsx
@@ -1,5 +1,5 @@
 export const dashboardMeta = {
-  name: 'vitalsAndBiometrics',
+  name: 'vitals_and_biometrics',
   slot: 'patient-chart-vitals-biometrics-dashboard-slot',
   config: { columns: 1 },
   title: 'Vitals & Biometrics',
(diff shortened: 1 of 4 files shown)


brandones:
Why don't we just use e.g. `name: "Vitals and Biometrics"`?

brandones:
Ping, @denniskigen ?

denniskigen:
> Why don't we just use e.g. `name: "Vitals and biometrics"`?

Because in addition to being used as breadcrumb dashboard titles, the value of the `name` property is what gets interpolated in the URL as well when you navigate to a dashboard. So in that sense, I think `.../chart/vitals_and_biometrics` reads better than `../chart/vitalsAndBiometrics` (what we have currently).

brandones:
Yeah I think slugifying names ("Vitals and Biometrics" --> "vitals_and_biometrics") is a much more reliable transformation than the reverse. Un-slugifying names leaves open the question of which words should be capitalized. (You can think of this in information theory terms: the slug form has less information than the title form (namely, it does not have capitalization), so the transformation from title to slug is natural while the transformation the other way requires the computer to do some inference/invention.)

denniskigen:
Many thanks, @brandones!

PR #564, opened 2022-02-18
(feat) O3-1038 Workspace system should allow notifying with unsaved changes

brandones:
## Requirements

## Summary

Adds a mechanism for workspaces to toggle whether or not to prompt before opening another workspace.

Default behavior is the same as what happens currently.

Usage is as demonstrated in the Allergies form in this PR:

```typescript
 useEffect(() => {
    promptBeforeClosing(() => Boolean(input1 || input2));
  }, [promptBeforeClosing, input1, input2]);
```

### Note

I chose to implement this like this to allow for the theoretical possibility of performance optimizations via uncontrolled state, e.g.

```tsx
const inputText = useRef('');

useEffect(() => {
  promptBeforeClosing(() => Boolean(inputText.current));
}, [promptBeforeClosing]);
```

So in this example the `inputText` would only be checked when `promptBeforeClosing` is called.

This might never be used this way, and so might never matter. But anyway it's not really any more complicated than the alternative API I considered:

```tsx
const [text, setText] = useState();

useEffect(() => {
  setPromptBeforeClosing(Boolean(text));
}, [setPromptBeforeClosing, text]);
```

## Screenshots

Peek 2022-02-17 19-05

## Issue

https://issues.openmrs.org/browse/O3-1038

### DIFF
--- a/packages/esm-patient-allergies-app/src/allergies/allergies-form/allergy-form.component.tsx
+++ b/packages/esm-patient-allergies-app/src/allergies/allergies-form/allergy-form.component.tsx
@@ -37,7 +37,7 @@ enum AllergenTypes {
   ENVIRONMENT = 'ENVIRONMENT',
 }
 
-const AllergyForm: React.FC<DefaultWorkspaceProps> = ({ closeWorkspace, patientUuid }) => {
+const AllergyForm: React.FC<DefaultWorkspaceProps> = ({ closeWorkspace, promptBeforeClosing, patientUuid }) => {
   const { t } = useTranslation();
   const isTablet = useLayoutType() === 'tablet';
   const { concepts } = useConfig();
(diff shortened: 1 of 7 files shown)


packages/esm-patient-allergies-app/src/allergies/allergies-form/allergy-form.component.tsx
```diff
@@ -66,6 +66,22 @@ const AllergyForm: React.FC<DefaultWorkspaceProps> = ({ closeWorkspace, patientU
...
   const severityLevels = [t('mild', 'Mild'), t('moderate', 'Moderate'), t('severe', 'Severe')];
 
+  useEffect(() => {
+    promptBeforeClosing(() => {
+      return Boolean(
+        nonCodedAllergenType || nonCodedAllergicReaction || onsetDate || selectedAllergen || severityOfWorstReaction,
+      );
+    });
+  }, [
+    promptBeforeClosing,
```
jonathandick, line 76: is there an easier way of observing the form has values other than for watching on every question? This makes me concerned for the more complex forms which have dozens of questions.

brandones: Yeah I mean ideally our forms would use some kind of abstraction that would make this easy. Like we should definitely handle this in a generic way in the actual form engine.

ibacher (approved):
LGTM

PR #571, opened 2022-02-21
O3-1116: Update test-result observation request count

Piumal1999:
## Requirements

## Summary
Currently, the esm-test-results loadPatientData makes 6 Observation requests with the size of 100 even if the total is less than 100.  
I updated the count and the number of chunks to prefetch the first 300 objects with a single request. If the total is bigger than 300, it will request the remaining objects in chunks of 300.

## Screenshots

image

## Issue

https://issues.openmrs.org/browse/O3-1116

## Other

*None.*


jonathandick:
@ZacButko would you be able to review this?

ZacButko (approved):
This seems to work locally to reduce /Observation calls for John Wilson down form 6 calls to 1 as required by OS-1116.

PR #578, opened 2022-02-25
(fix) Improve use of extensions API; simplify dashboard system; remove cruft

brandones:
## Requirements

## Summary

Moves away from deprecated parts of the extensions API. Gets rid of unused parts of the dashboard system "tabs" and "custom."

**Must be merged together with https://github.com/openmrs/openmrs-esm-core/pull/337**


packages/esm-patient-common-lib/src/types/index.ts
```diff
@@ -1,13 +1,8 @@
 export * from './test-results';
 export * from './workspace';
 
-export interface DashbardGridConfig {
+export interface DashbardConfig {
```
denniskigen, line 4: `export interface DashboardConfig`

packages/esm-patient-chart-app/src/view-components/chart-review.component.tsx
```diff
@@ -46,32 +68,20 @@ const ChartReview: React.FC<ChartReviewProps> = ({ patientUuid, patient, view, s
...
-          patient={patient}
-          patientUuid={patientUuid}
-        />
+        <ExtensionSlot
+          key={dashboard.slot}
+          extensionSlotName={dashboard.slot}
+          className={styles.dashboard}
+          style={{ gridTemplateColumns }}
+        >
+          <Extension state={state} wrap={wrapItem} />
```
ZacButko, line 81: This seems weird to be passing patient and patientUuid through as props to ChartReview when ChartReview could just be using usePatient(), then even weirder still that we memoize it and pass it through to Extension. Why is this not something that Extension handles natively? And if it's not something that should be in Extension then the end components should be more responsible for calling usePatient() rather than passing props so deep?

ZacButko (approved):
Definitely in favor of simplifying and removing unused components. I just have the one question about why we memoize and pass patient and patientUuid as props instead of just loading it later in descendants with the usePatient hook

packages/esm-patient-chart-app/src/view-components/chart-review.component.tsx
brandones: Yeah, I think it is bad that we pass `patient` as a param. You are right that that should be obtained with `usePatient`. The reason we pass `patientUuid` as a param is so that widgets can be used in contexts where there is information for multiple patients displayed in the same page.

PR #583, opened 2022-02-28
(refactor) Refactor chart-review component

brandones:
## Requirements

- [x] My work includes tests, or is validated by existing tests.

## Summary

The revision of ChartReview in https://github.com/openmrs/openmrs-esm-patient-chart/pull/578 caused it to be able to enter an invalid state when only some extension metas are loading (not sure why that happens but it does). Here I move the dashboard rendering to its own component, which is only rendered when everything needed has been loaded.

This also gets rid of the `subview` parameter, which is cruft.


ZacButko (approved):
OK

PR #595, opened 2022-03-08
(feat) O3-1108  Left nav should support groupings

brandones:
## Requirements

## Summary
Adds a mechanism for adding groups to the left nav, either using code or config.

## Screenshots

https://www.loom.com/share/b093200976d2433681abdf1f10a0bcf7

## Related Issue

https://issues.openmrs.org/browse/O3-1108


packages/esm-patient-chart-app/src/config-schema.ts
```diff
@@ -22,6 +22,19 @@ export const esmPatientChartSchema = {
     _description: 'The UUID of the visit type to be used for the automatically created offline visits.',
     _default: 'a22733fa-3501-4020-a520-da024eeff088',
   },
+  navGroup: {
```
brandones, line 25: I think this makes it clear that we need extension-specific schemas. They can inherit from module schemas (which is desirable for things like concepts). But this stuff just doesn't make sense in the module schema.

packages/esm-patient-chart-app/src/side-nav/side-menu.component.tsx
```diff
@@ -1,20 +1,19 @@
...
 
 const SideMenuPanel: React.FC<SideMenuPanelProps> = () => {
   const layout = useLayoutType();
-  useNavMenu();
 
   return (
     isDesktop(layout) && (
       <SideNav expanded aria-label="Menu" className={styles.link}>
-        <ExtensionSlot extensionSlotName="nav-menu-slot" />
+        <PatientChartNavMenu />
```
brandones, line 16: Simplified this. The idea was to make the left nav change contents based on what page you're on. However at present the left nav only exists on the patient chart. Once we have left nav on other pages, we can re-introduce the abstraction, but using unistore rather than the extension system.

jonathandick (commented):
Brandon, is there anyway to simplify this so that one does not have to create the additional extensions but could could simply pass the parameters being used in createDashboardLink for each of the links within the group? Furthermore, it seems more in line with the pattern to create a proper named function to create the AccordionExtension component rather than the () => .... That function could then take a parameter of options much like seen in createDashboardLink. So might be: createDashboardGroup(title, sloteName, [{name:"drug-allergies-dashboard", title:"Drug Allergies"}])

packages/esm-patient-chart-app/src/config-schema.ts
samuelmale: @brandones I'm curious to know whether this supports multiple groups. In other words is possible to define multiple groups using the configuration approach?

brandones: It does! You can add the same extension multiple times and configure it using the `#id` pattern:

```json
"extensions": {
  "add": ["nav-group#hiv", "nav-group#heart"],
  "configure": {
    "nav-group#hiv": {
     "navGroup": { "title": "HIV", slotName: "hiv-nav-group-slot" }
    },
    "nav-group#heart": {
      "navGroup": { "title": "Cardiology", "slotName": "heart-nav-group-slot" }
    }
  }
}
```

brandones:
@jonathandick Sure, that was actually the way I initially implemented it, but then `AccordionExtension` ended up being useful for the configurable version. I think you're right that that's more natural; I'll add that back.

ZacButko (approved):
OK

brandones:
Turns out that this is terribly incomplete. Right now there's no way that adding a dashboard (link) to a group can work, because if the link is being added to `some-custom-slot`, then it's not being added to `patient-chard-dashboard-slot`, and it is from the latter that patient chart derives the list of existing dashboards.

I need to split dashboards from their links.

brandones:
Just kidding, I figured out a good way to make it work.

jonathandick (approved):
Great work!

brandones:
Jest for some reason wasn't liking the `WorkspaceWindowState` enum, so I replaced it with a disjoint union. Many people argue this is [preferable](https://fettblog.eu/tidy-typescript-avoid-enums/) anyway.

PR #598, opened 2022-03-09
OHRI-452 Form Patient Banner - Optional hideActionsOverflow, optional param

larslemos:
## Requirements

## Screenshots

This is the current Patient Banner on the Form:
Screen Shot 2022-03-09 at 13 58 13

The objective is to make this an option Overflow Button, since for this scenario, the actions don't apply for our context. 
Screen Shot 2022-03-09 at 14 02 12

## Related Issue

https://issues.openmrs.org/browse/O3-1154?atlLinkOrigin=c2xhY2staW50ZWdyYXRpb258aXNzdWU%3D

### DIFF
--- a/packages/esm-patient-banner-app/src/banner/patient-banner.component.tsx
+++ b/packages/esm-patient-banner-app/src/banner/patient-banner.component.tsx
@@ -15,9 +15,16 @@ interface PatientBannerProps {
   patientUuid: string;
   onClick?: (patientUuid: string) => void;
   onTransition?: () => void;
+  hideActionsOverflow?: boolean;
 }
 
-const PatientBanner: React.FC<PatientBannerProps> = ({ patient, patientUuid, onClick, onTransition }) => {
+const PatientBanner: React.FC<PatientBannerProps> = ({
+  patient,
+  patientUuid,
+  onClick,
+  onTransition,
+  hideActionsOverflow,
+}) => {
   const { t } = useTranslation();
   const overFlowMenuRef = React.useRef(null);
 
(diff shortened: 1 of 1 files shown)


packages/esm-patient-banner-app/src/banner/patient-banner.component.tsx
```diff
@@ -15,9 +15,16 @@ interface PatientBannerProps {
...
+  hideActionsOverflow?: boolean;
 }
 
-const PatientBanner: React.FC<PatientBannerProps> = ({ patient, patientUuid, onClick, onTransition }) => {
+const PatientBanner: React.FC<PatientBannerProps> = ({
+  patient,
+  patientUuid,
+  onClick,
+  onTransition,
+  hideActionsOverflow,
```
jonathandick, line 26: It would be preferable, I think, to do this via configuration, not through code.

brandones (approved):

packages/esm-patient-banner-app/src/banner/patient-banner.component.tsx
larslemos: Well, this solution took a lot of turns. 
1st Initially the plan was to make a stylesheet appracoach
2nd Tried using hooks, but then after an event, the re-render was paiting again the Action.
3rd Is the current commit.

The alternative way would be below. 

Have to refactor the module and:

1. In patient-banner-component.tsx: Add an extension slot to act as a container for the Actions Overflow (actions-overflow-container)

2. In /src, create a new folder (actions) with a component that returns the code for CustomOverflowMenuComponent 

3. In /src/index.ts, attach the actions component to the actions-overflow-container extension

4. In OHRI, add config to remove actions-overflow-container extension when rendering OHRIForm


This does mean it will, work cause every time I try something, there is a new side-effect I learn from the esm behaviour.

brandones: We are going to completely refactor this component, @larslemos . This solution is temporary.

PR #599, opened 2022-03-09
(chore) Use cross-env to pass TZ=UTC environment variable to jest

brandones:
## Requirements

## Summary

`yarn test` was not working on Powershell because Powershell doesn't understand POSIX syntax for environment variables. [cross-env](https://www.npmjs.com/package/cross-env) makes it work in any shell.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,8 +13,8 @@
     "verify": "lerna run lint && lerna run test && lerna run typescript",
     "prettier": "prettier --config prettier.config.js --write \"packages/**/*.{ts,tsx}\"",
     "prepare": "husky install",
-    "test": "TZ=UTC jest --config jest.config.json --verbose false --passWithNoTests",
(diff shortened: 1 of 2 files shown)


vasharma05:
Hey @brandones!
I visited the https://www.npmjs.com/package/cross-env, and found a warning there:
> NOTICE: cross-env still works well, but is in maintenance mode. No new features will be added, only serious and common-case bugs will be fixed, and it will only be kept up-to-date with Node.js over time.

Will this be an issue in the future? 
Other than this, I read about the cross-env and it's a viable solution for the same.

brandones:
Yep, it would be bad if it was unmaintained, but it is "in maintenance mode." It's an extremely simple utility and I don't imagine we'll be needing new features from it :)

vasharma05 (approved):

PR #602, opened 2022-03-11
03-1166: Add bottom navigation on tablet view

donaldkibet:
## Requirements

## Summary

This PR updates the patient chart to include bottom-navigation on tablet view as seen here in the [design](https://app.zeplin.io/project/60d59321e8100b0324762e05/screen/60d5983191ddc216af40749c) and also removes `Care Activities` button

## Screenshots
Kapture 2022-03-11 at 14 16 36

## Related Issue
https://issues.openmrs.org/projects/O3/issues/O3-1166?filter=allissues

## Other
TODO: Side rail for desktop view


packages/esm-patient-chart-app/src/patient-chart/action-menu/action-menu.component.tsx
```diff
@@ -49,13 +50,7 @@ export const ActionMenu: React.FC<ActionMenuInterface> = ({ open }) => {
...
     </aside>
   ) : (
-    <Button className={styles.actionBtn}>
-      <div>
-        <Edit20 />
-        {windowSize.size === 'hidden' && <WarningFilled16 className={styles.warningButton} />}
-      </div>
-      <span>{t('careActivities', 'Care Activities')}</span>
-    </Button>
+    <ExtensionSlot state={state} className={styles.extensionStyles} extensionSlotName={'action-menu-items-slot'} />
```
brandones, line 53: The fact that you're passing state in tablet mode but not in desktop mode is a red flag to me. But here, we don't need the state at all! Just use the `useWorkspaces` hook directly in components that need the list of workspaces.

Again: avoid state when possible.

brandones:
Looking at the desktop designs, it looks like the exact same set of buttons should appear whether in Desktop or Tablet mode:



This is very nice because it means we can get rid of a bunch of the conditional "ifTablet" logic. You can just use the same extensions in either mode. Only the styling needs to change, and the extension slot should be able to control that.

We should also make the "Workspace items" button an extension at some point (doesn't need to be this PR).

Please let me know if I'm misunderstanding something!

donaldkibet:
> Looking at the desktop designs, it looks like the exact same set of buttons should appear whether in Desktop or Tablet mode:
> 
> 
> 
> This is very nice because it means we can get rid of a bunch of the conditional "ifTablet" logic. You can just use the same extensions in either mode. Only the styling needs to change, and the extension slot should be able to control that.
> 
> We should also make the "Workspace items" button an extension at some point (doesn't need to be this PR).
> 
> Please let me know if I'm misunderstanding something!

@brandones  updated with the feedback comments. I have made the menu items be buttons instead of `div` and cleaned up the `action-menu` to include extensions and styles only

My intention was to implement the tablet view navigation, I have expanded the scope to also include desktop as seen in this gif
Kapture 2022-03-11 at 23 36 52

packages/esm-patient-forms-app/src/clinical-form-action-menu.component.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { useLayoutType } from '@openmrs/esm-framework';
+import { Button } from 'carbon-components-react';
+
+const ClinicalFormActionMenu: React.FC = () => {
+  const { t } = useTranslation();
+  const layout = useLayoutType();
+
+  if (layout === 'tablet')
+    return (
+      <Button kind="ghost" className={styles.container} onClick={() => {}}>
```
brandones, line 14: What's with the empty `onClick`?

packages/esm-patient-forms-app/src/clinical-form-action-menu.component.tsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import Document20 from '@carbon/icons-react/es/document/20';
+import styles from './clinical-form-action-menu.scss';
+import { useTranslation } from 'react-i18next';
+import { useLayoutType } from '@openmrs/esm-framework';
+import { Button } from 'carbon-components-react';
+
+const ClinicalFormActionMenu: React.FC = () => {
```
brandones, line 8: Also this looks like just a button. Why is it called `Menu`?

packages/esm-patient-chart-app/src/patient-chart/action-menu/action-menu.component.tsx
```diff
@@ -1,66 +1,18 @@
...
-      <span>{t('careActivities', 'Care Activities')}</span>
-    </Button>
-  );
-
   return (
     <>
-      {menu}
+      <aside className={styles.sideRail}>
+        <ExtensionSlot className={styles.extensionStyles} extensionSlotName={'action-menu-items-slot'} />
+      </aside>
```
brandones, line 15: Nice! Can we get rid of the `<HeaderPanel>` `drawer-slot` stuff below?

brandones:
Thanks for doing that @donaldkibet . It's looking great. Just a few smaller comments.

packages/esm-patient-forms-app/src/clinical-form-action-menu.component.tsx
donaldkibet: The clinical form workspace hasn't been implemented yet as seen here in the [designs](https://app.zeplin.io/project/60d59321e8100b0324762e05/screen/61f040dbd4e9d1a258f4e157) should this Action button be removed?

packages/esm-patient-forms-app/src/clinical-form-action-menu.component.tsx
donaldkibet: Renamed them to button

jonathandick:
This is great work @donaldkibet . Thanks for the thoughtful review @brandones . I'm going to merge this for now.

PR #611, opened 2022-03-15
(feat) Offline Form Improvements

manuelroemer:
## Requirements

## Summary
The PR contains major improvements towards how offline forms are handled. As part of this work, I had to make some (rather large) adjustments in various locations of `esm-form-entry-app` in order to support the new offline form submission flow. Among these, I tried my best to improve the code and make it easier to work with (introduce better typings, extract certain tasks into services, removed unused code, ...). 

With that being said, I focused on not making any *behavioral* changes in the code whenever possible. The services (and the whole module, actually) should still do the same thing as before, except, of course, for the changes with regard to offline form handling.

### Changes Regarding Offline Forms
**Previously**, offline forms were handled like this: `esm-form-entry-app` intercepted `POST` requests submitted by the MF with the help of the service worker. This was functional for the requirements back then, but by now, we needed offline forms to support more complex flows (like editing offline forms from the offline actions).
To support such tasks, I refactored `esm-form-entry-app` so that it is aware of offline mode and handles offline synchronization items directly, i.e. without any network interception magic. This allows us, and other MFs, to better interact with the queued data. It should further make the entire process of handling offline forms *much much much more stable* because the way data is submitted during create/edit flows is now handled differently.

With these changes in place, editing offline forms should now also be possible (though I will have to test this more extensively in the coming days, ideally with more forms). If there are follow-up issues, they can be addressed in another PR.

### Other Changes / Refactorings
The task described above required some deep changes in `esm-form-entry-app`. I had to touch a lot of files and took the chance to make some improvements (this is always subjective, I know) to the code:
* Slimmed down large files like `fe-wrapper.component.tsx` by extracting functions into separate services.
* Adding more types in the locations I touched.
* Use appropriate RX operators in the locations I touched.
* Removed dead code.
* ...

## Screenshots
_No UI changes._

## Related Issue
https://issues.openmrs.org/browse/O3-1032
https://issues.openmrs.org/browse/O3-949 (foundational work for this issue)

## Other

If this PR is merged the way it is, there are some follow-up tasks that should be done:
* The app's service worker can now be slimmed down because the network request interception logic is no longer needed (it was only ever introduced for the flow described above).
* There are some TODOs in the code. A major one is that esm-offline should be expanded with a new API for retrieving *full* sync items (and not just their content).


manuelroemer:
@donaldkibet @denniskigen Tagging you guys as reviewers since you probably know most about how forms work. It would be great if you guys could have a look at these changes and check if I missed something!

manuelroemer:
@ibacher Do you know if it is possible/a huge amount of effort to add more forms to Dev3? I used the legacy openmrs-spa env a lot since there are a lot more forms available, but that was not a good experience, really. Ideally, we'd have more forms available in our new dev environment. Maybe there is even a way to move the old forms from openmrs-spa over to the new dev3 version?

donaldkibet (approved):
LGTM 🍻 Thanks @manuelroemer

PR #615, opened 2022-03-17
(fix): only try to process dashboard extensions once the patient-chart-dashboard-slot has been mounted

ibacher:
## Requirements

## Summary

Without this in tablet mode I was getting the error: `TypeError: extensionStore.slots['patient-chart-dashboard-slot'] is undefined`.


brandones (approved):
Thanks @ibacher

brandones:
Because of https://github.com/openmrs/openmrs-esm-core/pull/359, the dashboard meta will be accessible even when the `patient-chart-dashboard-slot` has not been mounted—it only needs to have been registered. So I'm not sure why this is still a problem. Maybe out of date core? But anyway I suppose it can't hurt to make this component not crash everything if it happens to mount before the slot is registered (which is the other possibility, since the load/mount order is not deterministic).

brandones:
If you're using `yarn start`/`openmrs develop`, the core that gets used is the one corresponding to your version of `openmrs`.

PR #625, opened 2022-03-25
(fix) O3-1160: Fix Offline Visit Not Being Used Correctly | Offline Form Fixes

manuelroemer:
## Requirements

## Summary
Mainly fixes the issue described by O3-1160 where the offline visit was no longer being used correctly ever since the `useVisit` hook was refactored.
Also provides small fixes for offline forms. For instance, it was not possible to edit a submitted offline form from the offline tools (which should now be possible).

## Related Issue
https://issues.openmrs.org/browse/O3-1160

## Other
We should update the importmap of the dev tooling once this has been deployed to avoid any esm-patient-common-lib version discrepancies.


FlorianRappl (approved):
Great!

packages/esm-patient-common-lib/src/offline/patient.ts
```diff
@@ -0,0 +1,60 @@
+import { usePatient, getSynchronizationItems } from '@openmrs/esm-framework';
+import useSWR from 'swr';
+
+export const patientRegistrationSyncType = 'patient-registration';
+
+/**
+ * A hook similar to {@link usePatient} which also considers patients which have been newly registered
+ * in offline mode by the patient registration module.
+ */
+export function usePatientOrOfflineRegisteredPatient(patientUuid: string): ReturnType<typeof usePatient> {
```
brandones, line 10: @manuelroemer Should we just make this what `usePatient` does?

manuelroemer: @brandones No, I don’t think so. That would only work if all usePatient users could correctly handle offline registered patients. Giving a module which doesn’t expect offline patients one would certainly break things, e.g. when a follow up API request is made with the patient‘s UUID. That would always fail because the patient doesn’t exist in the BE. But there are also other failure points.

brandones: Makes sense, thanks @manuelroemer

PR #635, opened 2022-04-04
(fix) O3-1220: Vitals header does not update upon form submission

denniskigen:
## Requirements

## Summary

This PR fixes an issue where the values displayed in the vitals header do not update after submitting a new batch of values using the vitals and biometrics form. This is because the revalidation key passed to `mutate` does not match the URL that the form POSTs to.

This PR introduces a regular expression match to obtain matching keys and then invoke the `mutate` function on their pre-bound hooks. Because the key bound to the `useVitals` hook changes based on whether biometrics concepts are included or not, using a regex to perform a partial match is a more robust approach than trying to use a fully formed URL.

## Video

https://user-images.githubusercontent.com/8509731/161515582-cc4a4494-713a-427a-b8f1-d37655e95fdf.mp4

### DIFF
--- a/packages/esm-patient-vitals-app/src/vitals/vitals-biometrics-form/vitals-biometrics-form.component.tsx
+++ b/packages/esm-patient-vitals-app/src/vitals/vitals-biometrics-form/vitals-biometrics-form.component.tsx
@@ -1,22 +1,22 @@
 import React, { useEffect, useState } from 'react';
-import VitalsBiometricInput from './vitals-biometrics-input.component';
-import styles from './vitals-biometrics-form.component.scss';
 import { useTranslation } from 'react-i18next';
 import { useSWRConfig } from 'swr';
 import {
-  useConfig,
   createErrorHandler,
-  useSessionUser,
+  fhirBaseUrl,
   showToast,
   showNotification,
-  fhirBaseUrl,
+  useConfig,
   useLayoutType,
+  useSessionUser,
 } from '@openmrs/esm-framework';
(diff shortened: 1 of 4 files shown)


packages/esm-patient-vitals-app/src/vitals/vitals-biometrics-form/vitals-biometrics-form.component.tsx
```diff
@@ -36,24 +36,13 @@ const VitalsAndBiometricForms: React.FC<DefaultWorkspaceProps> = ({ patientUuid,
   const isTablet = useLayoutType() === 'tablet';
   const session = useSessionUser();
   const config = useConfig() as ConfigObject;
-  const { mutate } = useSWRConfig();
+  const { cache, mutate }: { cache: any; mutate: Function } = useSWRConfig();
```
brandones, line 39: Weird that this isn't typed. Are we missing a `@types` dependency?

brandones (approved):
This is an elegant solution. Thanks @denniskigen . If this comes up a bunch we should think of a way to move this into esm-api.

packages/esm-patient-vitals-app/src/vitals/vitals-biometrics-form/vitals-biometrics-form.component.tsx
denniskigen: No, we do have @types. 

`cache` here gets typed to `Cache<Data>`, which I presume is a wrapper around `Map`. I couldn't get `Map.prototype.keys()` to work with it, however.

PR #639, opened 2022-04-06
O3-1226: Filter form list according to user privileges

icrc-psousa:
**Ticket:** [O3-1226](https://issues.openmrs.org/browse/O3-1226)
**Description:**
- Fetch encounter type viewPrivilege and editPrivilege when fetching from data
- On esm-patient-forms-app, when populating the list of forms to display, filter out the ones of encounter types that the user has no edit privileges


ibacher (approved):
@icrc-psousa This LGTM, but I'm holding off on merging so I can fix `useSessionUser` to return a `LoggedInUser` rather than a `SessionUser`.

PR #649, opened 2022-04-09
(chore) Updates patient chart to latest esm-core

ibacher:
## Requirements

## Summary

This updates the patient chart to use the latest version of the framework, including the new public / internal API boundaries and SWC as transpiler.

## Other

There are parts of the patient chart that continue to depend on framework internals. The imports I had to add here have been marked with a `// FIXME` comment and should be addressed at some point.

I had to add [a Yarn resolution](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/) for `@types/react` since apparently Carbon depends on the `*` version which currently resolves to 18.0.0. There appears to be either things that will need to be addressed when we upgrade to React 18 or just simple typing incompatibilities caused by having multiple versions of `@types/react`.


brandones (approved):

ibacher:
Oops! This should've been closed!

PR #658, opened 2022-04-15
(fix) Eliminate the need for dashboards to have names

brandones:
## Requirements

## Summary

This removes the "name" attribute of dashboards. The title is used for everything.

## Screenshots

Peek 2022-04-14 17-24

### DIFF
--- a/packages/esm-patient-allergies-app/src/dashboard.meta.tsx
+++ b/packages/esm-patient-allergies-app/src/dashboard.meta.tsx
@@ -1,5 +1,4 @@
 export const dashboardMeta = {
-  name: 'allergies',
   slot: 'patient-chart-allergies-dashboard-slot',
   config: { columns: 1 },
   title: 'Allergies',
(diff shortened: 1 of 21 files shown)


brandones:
Doing something more radical.

brandones:
Just kidding, just renamed "title" to "name," nothing radical for now.

ZacButko (approved):

ibacher:
@brandones I'm curious about a point on this: how will this change interact with i18n? For instance, if I replace "Allergy" with "Alergia" does this also change the URL? (Can this be done?)

brandones:
Also not sure what happened to this change:

> Just kidding, just renamed "title" to "name," nothing radical for now.

Maybe I forgot to push...

ibacher:
> We could, but it would be pretty ugly and increase the chance of weird errors loading dashboards. I would prefer just to do i18n where we are displaying the title.

Oh, I absolutely agree that we should only do i18n where we are displaying the title and leave the URLs stable across versions. My concern is mostly just how we enable overriding the title in the left nav and breadcrumb without also overriding the title in the url.

brandones:
Oh yeah. Well right now we actually don't support overriding dashboard titles. I mean, the patient chart in the abstract [supports it](https://github.com/openmrs/openmrs-esm-patient-chart/blob/master/packages/esm-patient-chart-app/src/patient-chart/chart-review/chart-review.component.tsx#L41), but actually existing dashboards don't have configuration schemas associated with them, so they are effectively not configurable.

It seems like we should probably not make the titles overridable via config, but rather using i18n (and therefore i18n overrides in config).

PR #660, opened 2022-04-18
(fix) Forms: only try to load concepts for questions with a concept

icrc-psousa:
When building the unlabeled question list, traverse function was picking up nodes with no label but, also with no concept. No label should be loaded for those nodes (e.g.: AMPATH custom component).
Fixed the traverse function not to consider those nodes.


icrc-psousa:
@ibacher to review when possible

ibacher (approved):

PR #665, opened 2022-04-20
Making ampath forms work on Vitals section

MiguelAHPpih:
## Summary

The vitals section has been modified to include the usage of ampath forms as an option(implementer tools).
This was tested using "POC Sample Form 1".

## Related Issue
When the form is being loaded and there isn't an active visit. It will crash with no logs in the console.
This is adressed in this ticket:
https://issues.openmrs.org/browse/O3-1243


brandones:
Hi @MiguelAHPpih , can you please attach a screenshot, gif, or video?

MiguelAHPpih:
@brandones 
Vitals-AmpathForms

packages/esm-patient-vitals-app/src/index.ts
```diff
@@ -1,76 +1,76 @@
...
-            default: 'Record Vitals and Biometrics',
-          },
-        },
-      },
-    ],
-  };
-}
-
-export { backendDependencies, importTranslation, setupOpenMRS };
+import { defineConfigSchema, fhirBaseUrl, getAsyncLifecycle, messageOmrsServiceWorker } from '@openmrs/esm-framework';
```
brandones, line 1: What happened to this file? Seems like some strange whitespace change happened. Maybe `yarn prettier` or just `git checkout origin/master packages/esm-patient-vitals-app/src/index.ts`?

MiguelAHPpih: I remember doing some tests with console.log, but only that. It could be prettier

brandones: Please run `git checkout origin/master packages/esm-patient-vitals-app/src/index.ts` to revert these changes, and then `yarn prettier`. Commit whatever results.

MiguelAHPpih: Done

brandones (approved):

brandones:
@denniskigen Looks like the build step is OOMing. This also recently happened on master: https://github.com/openmrs/openmrs-esm-patient-chart/actions/runs/2238756863 . Maybe we need to limit the number of things that can run in parallel?

denniskigen:
> @denniskigen Looks like the build step is OOMing. This also recently happened on master: https://github.com/openmrs/openmrs-esm-patient-chart/actions/runs/2238756863 . Maybe we need to limit the number of things that can run in parallel?

Perhaps. I wonder though whether the build step has to be rerun inside of this `pre-release` job? Is it possible to reuse the pre-built artefacts from the `build` job?

brandones:
@denniskigen That would be great. I'm inclined to merge this for now since I don't think the problem has anything to do with @MiguelAHPpih 's code.

PR #667, opened 2022-04-21
refactor: renamed patient-banner props and exported PatientBanner component

N0-man:
## Requirements

## Summary
As part of [this PR](https://github.com/openmrs/openmrs-esm-patient-management/pull/152), we agreed to refactor the props name to make it more meaningful. In addition export the PatientBanner component to be consumed in a normal way i.e. without using parcels or ExtentionSlots

## Related Issue
https://bahmni.atlassian.net/browse/BAH-1492


brandones (approved):
I have some hesitation about this since we have not up to this point imported components across microfrontends. However I think it might be fine, and is probably preferable to the hacky thing we are doing with extension slots presently. Interested in opinions from @ibacher @ZacButko

N0-man:
require.context is webpack magic or sugar to import individual modules in a given directory. There are challenges when the code runs outside of webpack - such as jest or Node. Since we are now exporting the PatientBanner - we would have the issues in jest while testing other MF where PatientBanner is used. One way to solve it is using babel transform-require-context plugin - but that requires dependency on Babel

`require-context-polyfill.ts` is just a polyfill to do the same thing that react.context does - this would specifically help when react.context is unavailable (such as running in jest). Another approach could be to not use react.context and add the imports for translations individually OR externalize importTranslation construction using react.context outside of index.ts
 
A good place for the pollyfil or externalizing importTranslation would be inside `esm-utils`

vasharma05:
@N0-man, is this PR ready to be merged?

packages/esm-patient-banner-app/src/require-context-polyfill.ts
```diff
@@ -0,0 +1,49 @@
...
+ * require.context is webpack magic or sugar to import individual modules in a given directory
+ * There are challenges when the code runs outside of webpack - such as jest or Node.
+ * One way to solve it is using bable transform-require-context plugin - but that requires depdnency on Babel
+ * Below code is just a pollyfill to do the same thing that react.context does - when react.context is unavailable
+ * Another approach could be to not use react.context and add the imports for traslations individually
+ *
+ * A good place for this pollyfil would be inside esm-utils
+ */
+
+export const requireContext = (base = '../translations', scanSubDirectories = false, regularExpression = /.json$/) => {
```
ibacher, line 11: If we're going to use a polyfill here, it really should be written:

```ts
export const requireContext = request.context !== 'undefined' ? require.context : (base, scanSubdirectories, regularExpression) =>
```

packages/esm-patient-medications-app/src/order-basket/medication-order-form.component.tsx
```diff
@@ -134,7 +134,6 @@ export default function MedicationOrderForm({
                       id: orderBasketItem.dosage.dosage,
                       text: orderBasketItem.dosage.dosage,
                     }}
-                    placeholder={t('editDoseComboBoxPlaceholder', 'Dose')}
```
ibacher, line 137: Why are these removed as part of this PR?

ibacher (commented):
Exporting a component from a microfrontend seems to break the idea that this is a microfrontend at all. Importing components across microfrontends is going to cause us pain down the line and should be avoided.

Based on [the new designs](https://app.zeplin.io/project/60d59321e8100b0324762e05/screen/62443dcf03b20709ca8af9d7), I think we should move away from using the patient banner in search results at all and leave it for what it was designed for: the header on the patient chart.

@ZacButko I think you are working on implementing the new designs to support fast form entry (where, like @N0-man, we want to embed patient search into another widget). Do you have any update on that work? Is there a way we can get @N0-man or others from Bahmni to help contribute to it to help out with the LabLite work?

packages/esm-patient-banner-app/src/require-context-polyfill.ts
N0-man: > If we're going to use a polyfill here, it really should be written:
> 
> ```ts
> export const requireContext = request.context !== 'undefined' ? require.context : (base, scanSubdirectories, regularExpression) =>
> ```

yes that is the point, we should externalize require.context and not let it penetrate in all of the MF's index.js... I agree a polyfill is overly-done since outside of web-pack we wont need the functionality anyways...

packages/esm-patient-medications-app/src/order-basket/medication-order-form.component.tsx
ZacButko: placeholder is still valid in carbon. The typing library is currently incorrect. Instead for the time being we can use // @ts-ignore [like Vineet has done](https://github.com/openmrs/openmrs-esm-patient-chart/pull/675/files#diff-35064635b8bebb6e593e6d8dd084fee5577c4427ed31f2974f03632440c285b3)

N0-man: @ZacButko this is an outdated change... I believe you have already seen and responded to my message on Slack to Vineet? So basically I have locked the version to 7.49.0 with some coverage (no need to use @ts-ignore and be on infected version with missing coverage) - here is [my commit](https://github.com/openmrs/openmrs-esm-patient-chart/pull/667/commits/87f615553b401308150d86020270193ac42e74d4). I have also raised [this PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/60074) to fix `@types/carbon-component-react`

ZacButko: @N0-man yes, sorry, awesome. I attempted to start reviewing your PR and then was taken away again. Thank you for both of those things

ZacButko (approved):
I'm in favor of having reusable components, and even having components exported from modules. However I do hear what Ian is saying about this creating unwanted coupling between apps which could cause problems.

For my part I'm ok spiking on this for now and seeing how it goes, unless @ibacher you are saying strongly that leaving it coupled like it is now will cause problems

N0-man:
> The above is a long way of saying that I don't think we should merge this PR yet.



> I'm in favor of having reusable components, and even having components exported from modules. However I do hear what Ian is saying about this creating unwanted coupling between apps which could cause problems.
> 
> For my part I'm ok spiking on this for now and seeing how it goes, unless @ibacher you are saying strongly that leaving it coupled like it is now will cause problems

Appreciate your feedback @ZacButko. I was able to connect with @ibacher and also had few challenges to myself - merely for ensuring that we maintain the architecture purity and yet keep it code as maintainable and testable ...

I would like to meet you, @brandones and @ibacher sometime early next week to go over few options and opinions... perhaps if its better to even discuss in the MF Squad meet with broader community? Until then we can keep this PR open...

hadijahkyampeire:
@denniskigen is this PR still needed, I see it is approved it just needs resolving conflicts I could take it on just in case @N0-man got busy.

donaldkibet:
This has been stale for quite some time.  Closing this

PR #679, opened 2022-04-26
O3-1269: Form names should try use display property

donaldkibet:
## Requirements

## Summary
- Update the names of the form to default to the `display` property and fall back to `name` if the `display` property is not available. This is to support localization through the user's locale.


ibacher (approved):
LGTM! Thanks @donaldkibet!

PR #688, opened 2022-05-05
(chore) Limit workflow task parallelization

denniskigen:
## Requirements

## Summary

This PR limits the number of tasks that run in parallel in our CI workflow. This is to reduce the likelihood of the CI workflow execution running out of memory.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -27,8 +27,12 @@ jobs:
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           server-token: ${{ secrets.TURBO_SERVER_TOKEN }}
-      - run: yarn run test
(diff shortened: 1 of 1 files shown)


brandones:
Why not use the [`--concurrency`](https://turborepo.org/docs/reference/command-line-reference#--concurrency) flag?

denniskigen:
> Why not use the [`--concurrency`](https://turborepo.org/docs/reference/command-line-reference#--concurrency) flag?

Sure. I still reckon it's sensible to separate the formatting and type checking jobs and the build step. It'd be easier to determine what resources to allocate to the two jobs when separate. The build job would likely need a lower concurrency value to minimise resource use, whereas the formatting jobs could go full turbo.

brandones (approved):

PR #690, opened 2022-05-06
03-837 : Display `unsaved-changes` modal when start-visit-form changed

donaldkibet:
## Requirements

## Summary

Prevent users from closing `start-visit-form` after changes are made on the form, this adds the confirm-modal for appropriate action to be performed

## Screenshots
Kapture 2022-05-06 at 16 24 06

## Related Issue

https://issues.openmrs.org/projects/O3/issues/O3-837


brandones:
The workspace system [has generic support](https://github.com/openmrs/openmrs-esm-patient-chart/blob/6a01287330a294ce8764d1dbdfce157597b94a21/packages/esm-patient-common-lib/src/workspaces/workspaces.test.ts#L104) for "prompt before closing." See the [Allergy Form](https://github.com/openmrs/openmrs-esm-patient-chart/blob/bcfd84996851f68bc3262b53a9b024e9a8d3980c/packages/esm-patient-allergies-app/src/allergies/allergies-form/allergy-form.component.tsx#L40) for an example.

brandones (changes requested):
See above

donaldkibet:
> The workspace system [has generic support](https://github.com/openmrs/openmrs-esm-patient-chart/blob/6a01287330a294ce8764d1dbdfce157597b94a21/packages/esm-patient-common-lib/src/workspaces/workspaces.test.ts#L104) for "prompt before closing." See the [Allergy Form](https://github.com/openmrs/openmrs-esm-patient-chart/blob/bcfd84996851f68bc3262b53a9b024e9a8d3980c/packages/esm-patient-allergies-app/src/allergies/allergies-form/allergy-form.component.tsx#L40) for an example.

This function works when we want to open another form, while we have another form of the same type opened. It doesn't warn on closing a form that the user has made changes to it. 


I have added logic to `closeWorkspace` function on `esm-common-lib` to add the ability to check if the form has been touched, and to warn the user before closing the form.

brandones:
Oh that's right, I never did that part with [O3-1038](https://issues.openmrs.org/browse/O3-1038)! That's a wonderful addition and your code looks great. Could you please add a test in the [workspaces.test](https://github.com/openmrs/openmrs-esm-patient-chart/blob/3f518328e795f99d5c08b6c788cdabf285de2c15/packages/esm-patient-common-lib/src/workspaces/workspaces.test.ts#L120) file, and rename the existing "respects promptBeforeClosing function" test to something like "respects promptBeforeClosing function when opening a new workspace"?

donaldkibet:
Just realized that my code had a bug in this scenario.

1. `closeWorkspace` function is always called when we save the form, this is because while inputting data to the form, we have changed the form status to touched. This will cause the `unsaved changes` modal to be shown whenever we save data and close the workspace.

I have updated the closeWorkspace function to take a parameter, that determines whether we display the unsaved changes modal or straight away close the workspace without warming

brandones:
Nice, good catch @donaldkibet . I'm thinking that it will keep the client code simpler if the API is instead

```ts
closeWorkspace(ignoreChanges: boolean)
```

and then the line to check whether to prompt is

```ts
if (!ignoreChanges && promptCheckFcn && promptCheckFcn()) {
```

PR #694, opened 2022-05-13
O3-1253: Highlight abnormal vitals

denniskigen:
## Requirements

## Summary

Abnormal values for vital signs should be flagged in both the vitals header and the vitals widget.

This PR highlights abnormal vitals in the vitals widget as described in the ticket https://issues.openmrs.org/browse/O3-1253.

## Screenshots

Screenshot 2022-05-13 at 17 46 58

## Related Issue

https://issues.openmrs.org/browse/O3-1253


packages/esm-patient-vitals-app/src/vitals/vitals-biometrics-form/vitals-biometrics-form.utils.ts
```diff
@@ -3,6 +3,7 @@ import isNumber from 'lodash/isNumber';
 import { ConceptMetadata } from '@openmrs/esm-patient-common-lib';
 
 export function calculateBMI(weight: number, height: number): number {
+  if (!weight && !height) return;
```
ZacButko, line 6: Should this be `||` instead? 

Also should this function have a default return value?

packages/esm-patient-vitals-app/src/vitals/vitals.resource.tsx
```diff
@@ -168,3 +208,82 @@ export function editPatientVitals(
     },
   });
 }
+
+export function exists(...args: any[]): boolean {
```
ZacButko, line 212: Should this function be called `allExist`?  The difference of 'some' exist vs 'every' exist.

ZacButko (approved):
I'm curious about a few things but otherwise looks great.

packages/esm-patient-vitals-app/src/vitals/vitals.resource.tsx
denniskigen: Turns out this function is not strictly necessary.

packages/esm-patient-vitals-app/src/vitals/vitals-biometrics-form/vitals-biometrics-form.utils.ts
denniskigen: Yeah, it should. I don't think it should have a default return value.

PR #710, opened 2022-06-03
(feat) Dynamic Offline Data for Forms | Offline Support for Concept Labels

manuelroemer:
## Requirements

## Summary
This builds upon the changes in [this PR](https://github.com/openmrs/openmrs-esm-core/pull/447). It migrates the patient chart towards the new dynamic data API. Notably, these areas have been updated:

* Existing patient synchronization handlers using the previous API (from the legacy patient-list way).
* Offline forms have been updated to leverage the dynamic data API for storing which forms are made available offline.

In addition, since it builds upon the new dynamic data API, I integrated offline support for the Ampath form concept labels which have been introduced in [this commit](https://github.com/openmrs/openmrs-esm-patient-chart/commit/9e445be3d2784f967b7d53e7bd0cc495be963890) a while ago. This should allow forms using concept labels to properly display offline.

:warning: This will not build until the PR in esm-core is approved and on NPM. See the "Other" section in the esm-core PR for details about when the PRs should ideally be merged.

## Related Issue
https://issues.openmrs.org/browse/O3-1110
https://issues.openmrs.org/browse/O3-1293


packages/esm-patient-forms-app/src/offline.ts
```diff
@@ -73,3 +86,51 @@ async function syncPersonUpdate(personUuid?: string, personUpdate?: any) {
...
+  setupDynamicOfflineDataHandler({
+    id: 'esm-patient-forms-app:form',
+    type: 'form',
+    displayName: 'Patient forms',
+    async isSynced(identifier) {
+      const expectedUrls = await getCacheableFormUrls(identifier);
+      const absoluteExpectedUrls = expectedUrls.map((url) => window.origin + makeUrl(url));
+      const cache = await caches.open('omrs-spa-cache-v1');
+      const keys = (await cache.keys()).map((key) => key.url);
+      return absoluteExpectedUrls.every((url) => keys.includes(url));
```
manuelroemer, line 100: I am honestly torn about whether we should extract this logic here into a utility function in core. It is more or less the same code in every instance where it occurs (at the moment - it doesn't necessarily have to stay that way). I think it is duplicated around 5 times right now.
The reason for _not_ doing it at the moment is that _not_ moving it into core means that we don't have to commit to a new API that needs to be supported in the long run. Maybe it makes sense to wait and see how this develops and then, at some point, we can move it to core? Any thoughts?

packages/esm-form-entry-app/src/app/offline/caching.ts
```diff
@@ -18,3 +29,83 @@ export function setupOfflineDataSourcePrecaching() {
...
+    {
+      ...form,
+      pages: clobdata.pages ?? [],
+      referencedForms: clobdata.referencedForms ?? [],
+      processor: clobdata.processor,
+    },
+    [],
+  ) as FormSchema;
+
+  const conceptLang = (window as any).i18next?.language?.substring(0, 2).toLowerCase() || 'en';
```
manuelroemer, line 91: I am not sure if caching only the current language is sufficient. I think there are pros and cons.

Pro:
* Less network requests.
* Less data in the cache (potentially helps with storage quotas).
* I am assuming that in 99.99% of the cases users will only ever use one language. I assume that this language will also be selected _before_ downloading the form. It should probably work out fine most of the time.

Con:
Really only one flow that can fail: The user downloads a form, changes the language and then immediately goes offline without reloading the page. The concept labels would then not be precached.

I guess the potential remedies would either be:
* To listen to a language change and then redownload the concepts with the right language. I think that makes the most sense, but I am currently not sure how to ideally listen to a language change. Open for any help here!
* To download concepts in all (known) languages. That can obviously make the cache explode. Also, I don't know if we have a list of all known languages.

One other thing that is related to the language, but not to offline specifically: Does anybody know what happens (in online mode) when a concept in a specific language does not exist in the backend? Does the network request fail? If so, that would probably lead to the form not displaying anything in the current version.

ibacher: > Does anybody know what happens (in online mode) when a concept in a specific language does not exist in the backend?

We return the concept in the default language for the system.

> To download concepts in all (known) languages. That can obviously make the cache explode. Also, I don't know if we have a list of all known languages.

We do have a list of supported languages; however, as a practical matter, users will generally at most switch their language once to their preferred language and probably never look at the toggle again. If the issue comes up in demos and will go to bat to say that switching languages is generally not something users do, so while this is a potential problem, I don't think its one worth solving.

(The main use-case I can think of where there might be a valid need to do something like this is if we explicitly allowed multiple user to use the same device offline, but since this is not something we allow, it's not worth the complexity overhead to implement, IMO).

manuelroemer: > We return the concept in the default language for the system.
That's great, thanks!

> [...] so while this is a potential problem, I don't think its one worth solving.
Awesome, that was also my thought. If bad comes to worse, we could potentially also implement a fallback approach while offline (though that might become a bit finnicky on the code side).

packages/esm-patient-forms-app/src/offline.ts
manuelroemer: > [...] 3 times, its time to move it into a proper function [...]

That sounds like a good approach. I think in this specific case, I'd like to delay it a little bit to move things forward, but I'll keep it on my todo list (and perhaps even think about a proper way to do this - functions for cache interaction would generally be a nice addition to core, maybe even without an `@internal` modifier).

ibacher (approved):
Approved, with suggestion to change one function name

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -193,53 +190,36 @@ export class FormSchemaService {
...
-                if (question.extras.questionOptions.answers) {
-                  question.extras.questionOptions.answers.forEach((answer) => {
-                    if (!answer.label) {
-                      concepts.push(answer.concept);
-                    }
-                  });
-                }
-                break;
-            }
+  public static getUnlabeledConceptUuidsFromSchema(form: FormSchema): Array<string> {
```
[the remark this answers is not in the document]
manuelroemer: That sounds good, I will change it. Thanks!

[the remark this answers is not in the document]
dkayiwa: > but there is a plan underway to create an endpoint which would allow us to refer to concepts by something other than UUIDs, specifically concept maps

@ibacher doesn't the [/concept/CIEL:1065](https://demo.openmrs.org/openmrs/ws/rest/v1/concept/) REST API already support this?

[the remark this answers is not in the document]
ibacher: @dkayiwa Good point! Really it's less dependent on an endpoint and more dependent on refactoring the frontend code to be able to use those references correctly.

PR #716, opened 2022-06-13
(fix) Fixed label loading for concepts on repeating sections on AMPATH

icrc-psousa:
## Requirements

## Summary
When an AMPATH form is loaded, it's schema is traversed for concepts without specified labels so that they can be dynamically loaded, according to user lang, from the concept dictionary.

Currently, there's an issue that causes the concept data not to be loaded for questions that are inside AMPATH repeating sections, causing rendered labels to be blank on ngx-openmrs-formentry.

This PR fixes the above issue so that all concepts specified on AMPATH schema without labels are loaded. If concept data is not found within the concept dictionary, the concept id, as specified in the AMPATH schema, is used as label so that ngx-openmrs-formentry has always has a label to render.

### DIFF
--- a/packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
+++ b/packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
@@ -193,6 +193,21 @@ export class FormSchemaService {
     return formUuids;
   }
 
+  private processQuestionForUnlabeledConcepts(question) {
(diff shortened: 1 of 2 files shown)


icrc-psousa:
Abandoning pull request as the issue it fixes was already fixed with https://github.com/openmrs/openmrs-esm-patient-chart/pull/710

PR #717, opened 2022-06-13
(feat) Add a build script for patient-common-lib

ZacButko:
## Requirements

- [x] This PR has a title that briefly describes the work done, including the ticket number if there is a ticket.
- [n/a] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide).
- [n/a] My work includes tests or is validated by existing tests.

## Summary
Adds a build script for esm-patient-common-lib. Hopefully this will result in a `/dist/openmrs-esm-patient-common-lib.js` file after all github actions. This built file is required for outside modules who do not transpile TypeScript.

## Screenshots
none

## Related Issue
none

## Other
none

PR #722, opened 2022-06-17
Revert "(feat) Add a build script for patient-common-lib (#717)"

ibacher:
## Requirements

## Summary

With apologies to @ZacButko, this reverts commit 419aec97b7afb9f3248ef2d5eb2baf1a76b78c41 which may or may not be causing the problem on dev3.

### DIFF
--- a/packages/esm-patient-common-lib/package.json
+++ b/packages/esm-patient-common-lib/package.json
@@ -7,7 +7,6 @@
   "main": "src/index.ts",
   "source": true,
   "scripts": {
-    "build": "webpack --mode production",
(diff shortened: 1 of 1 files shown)


manuelroemer (approved):
I think that this is a good idea to at least try out right now since the patient chart is basically unusable at the moment (in the online version). If that doesn't fix the issue we can revert the revert afterwards.

If it _does_ help, we need to figure out how to build the common lib as a separate file without causing the chart to crash. I'm honestly not quite sure at the moment how to do that/why it's a problem in the first place. 

Thanks for creating the PR @ibacher!

PR #727, opened 2022-06-22
O3-1332:Results Viewer: Resizable Columns

jwnasambu:
## Requirements

## Summary
I added a drag element in-between the columns to enable a user drag the  column width to their desired width in a 'split' mode. While respecting some reasonable minWidth for each column.
## Screenshots

https://user-images.githubusercontent.com/33891016/175060443-a92d48d2-2c0b-459f-a5e7-01f5579bcc7c.mp4

### DIFF
--- a/packages/esm-patient-test-results-app/src/results-viewer/results-viewer.tsx
+++ b/packages/esm-patient-test-results-app/src/results-viewer/results-viewer.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useState } from 'react';
+import React, { useContext, useEffect, useRef, useState } from 'react';
 import {
   AccordionSkeleton,
   Button,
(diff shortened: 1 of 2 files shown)


ZacButko:
Cool work @jwnasambu ! Overall it looks great, I would be in favor of having the draggable separator visible even when at rest. What do you think @Ciaran ? 

Juliet, sorry I'm not able to verify this at the moment as it seems the dev concept dictionary has been reset.

packages/esm-patient-test-results-app/src/results-viewer/results-viewer.tsx
```diff
@@ -122,15 +163,36 @@ const ResultsViewer: React.FC<ResultsViewerProps> = ({ patientUuid, basePath, ty
...
+                ) : (
+                  <AccordionSkeleton open count={4} align="start" />
+                ))}
               {leftContent === 'panel' && <DesktopView />}
             </Column>
           )}
-          <Column sm={16} lg={tablet || expanded ? 12 : 7} className={`${styles.columnPanel}`}>
+          <Column
+            sm={16}
+            lg={tablet || expanded ? 12 : 12 - adjustWidth}
```
ZacButko, line 193: I think keeping this as an integer is what is what's making the UI feel jerky. In order to have a more fluid drag experience I think we would need to specify `width` in pixels instead

cduffy2:
Oh, this is interesting. A few UX points from my side:

There should be a vertical line in colour `grey-30` dividing the whole page, including the results viewer header, which is this thing:

Screenshot 2022-06-23 at 10 34 05

It would be nice if that mouse turned into the horizontal drag mode when it's 24px left or right of the horizontal line on the screen. This way it's easier to discover this feature. It would also be important for the results viewer header above to move left and right with the results. This way, the controls in the header that affect the things below move with the window. 

Hope that makes sense?

jwnasambu:
> Oh, this is interesting. A few UX points from my side:
> 
> There should be a vertical line in colour `grey-30` dividing the whole page, including the results viewer header, which is this thing:
> 
> Screenshot 2022-06-23 at 10 34 05
> 
> It would be nice if that mouse turned into the horizontal drag mode when it's 24px left or right of the horizontal line on the screen. This way it's easier to discover this feature. It would also be important for the results viewer header above to move left and right with the results. This way, the controls in the header that affect the things below move with the window.
> 
> Hope that makes sense?

Yes it does though am a bit confused and I kindly request for your clarification. should I include the `newest first` on the results viewer header? If yes, what should  be display when its clicked on or expanded.

packages/esm-patient-test-results-app/src/results-viewer/results-viewer.tsx
jwnasambu: @ZacButko Thanks for the review and fixing the UI am happily working now.

PR #736, opened 2022-06-29
Fix the fix for generic dashboards

ibacher:
## Requirements

## Summary

Add fallback to `window.location` if `useLocation()` returns undefined. Should fix the fix from f19d297.

### DIFF
--- a/packages/esm-patient-common-lib/src/dashboards/DashboardExtension.tsx
+++ b/packages/esm-patient-common-lib/src/dashboards/DashboardExtension.tsx
@@ -10,7 +10,7 @@ export interface DashboardExtensionProps {
 }
 
 export const DashboardExtension = ({ title, basePath }: DashboardExtensionProps) => {
-  const location = useLocation<Location>();
+  const location = useLocation<Location>() || window.location;
   const navLink = useMemo(() => decodeURIComponent(last(location.pathname.split('/'))), [location.pathname]);
 
   const activeClassName = title === navLink ? 'active-left-nav-link' : 'non-active';


packages/esm-patient-common-lib/src/dashboards/DashboardExtension.tsx
```diff
@@ -10,7 +10,7 @@ export interface DashboardExtensionProps {
 }
 
 export const DashboardExtension = ({ title, basePath }: DashboardExtensionProps) => {
-  const location = useLocation<Location>();
+  const location = useLocation<Location>() || window.location;
```
ZacButko, line 13: Do you want to throw in `?? {pathname: ""}` just to be super extra safe?

brandones (approved):

ZacButko (approved):
Do we know why this fixes it? Is there some time when useLocation is unavailable? Or what is causing this to fail?

packages/esm-patient-common-lib/src/dashboards/DashboardExtension.tsx
ZacButko: Or just put optional chaining everywhere

ibacher:
> Do we know why this fixes it? Is there some time when useLocation is unavailable? 

I haven't really tracked down the circumstances, but the error was that it's undefined... I'd suspect it's possible that one of the dashboards get mounted in a React tree that doesn't have a React-Router as a DOM. `window.location` should always be defined, though or else the browser is broken...

PR #740, opened 2022-06-30
(fix) Results Viewer Default Schema - Point to extant concepts

ZacButko:
## Requirements

- [n/a] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide).
- [n/a] My work includes tests or is validated by existing tests.

## Summary
Points to valid concepts so this will return data now, instead of 404

## Screenshots
none

## Related Issue
none

## Other
none

### DIFF
--- a/packages/esm-patient-test-results-app/src/config-schema.ts
+++ b/packages/esm-patient-test-results-app/src/config-schema.ts
@@ -15,19 +15,15 @@ export const configSchema = {
     },
     _default: [
       {
-        conceptUuid: '5035a431-51de-40f0-8f25-4a98762eb796',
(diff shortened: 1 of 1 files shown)


denniskigen:
Hey @ZacButko, it appears these concepts don't exist anymore. Mind having a look?

Screenshot 2022-07-29 at 12 27 03

ZacButko:
@denniskigen [here you go](https://github.com/openmrs/openmrs-esm-patient-chart/pull/756). Updating the uuid's is the easy part. The hard part is knowing what they will be the next time dev3 server gets reset 😄 . It would be good to make a more permanent solution on the initializer module for these uuid's to remain consistent.

PR #742, opened 2022-07-05
#O3-1329 implementation of round numbers by default and configuration…

icrc-agomes:
… in  obs-by-encounter widget

## Requirements

## Summary
Currently, the obs-by-encounter widget does not round displayed numbers. By default, it seems like the obs-by-encounter widget should round decimal points to, 2 numbers, but allow the rounding needs to be configured via the configuration.

## Screenshots
image

## Related Issue
https://issues.openmrs.org/browse/O3-1329

### DIFF
--- a/packages/esm-generic-patient-widgets-app/src/config-schema.ts
+++ b/packages/esm-generic-patient-widgets-app/src/config-schema.ts
@@ -23,6 +23,9 @@ export const configSchema = {
       color: {
         _type: Type.String,
       },
+      decimalPlaces: {
+        _type: Type.Number,
+      },
     },
     _default: [
       {
--- a/packages/esm-generic-patient-widgets-app/src/obs-table/obs-table.component.tsx
+++ b/packages/esm-generic-patient-widgets-app/src/obs-table/obs-table.component.tsx
@@ -48,7 +48,19 @@ const ObsTable: React.FC<ObsTableProps> = ({ patientUuid }) => {
               break;
 
             case 'Number':
-              rowData[obs.conceptUuid] = obs.valueQuantity?.value;
+              let decimalPlaces: number = config.data.find(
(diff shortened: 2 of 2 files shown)


packages/esm-generic-patient-widgets-app/src/config-schema.ts
```diff
@@ -23,6 +23,9 @@ export const configSchema = {
       color: {
         _type: Type.String,
       },
+      decimalPlaces: {
+        _type: Type.Number,
```
ibacher, line 27: It would be nice to have a validator here, e.g.:

```ts
_validator: [validators.inRange(0, 10)]
```

(`validators` above should be importable from `@openmrs/esm-framework`).

packages/esm-generic-patient-widgets-app/src/obs-table/obs-table.component.tsx
```diff
@@ -48,7 +48,19 @@ const ObsTable: React.FC<ObsTableProps> = ({ patientUuid }) => {
               break;
 
             case 'Number':
-              rowData[obs.conceptUuid] = obs.valueQuantity?.value;
+              let decimalPlaces: number = config.data.find(
```
ibacher, line 51:
```suggestion
              const decimalPlaces: number | undefined = config.data.find(
```

ibacher: Since we're already parsing the data to add the `dataType` is there any real reason to get the decimal places here instead of just filling it in when we determine the `dataType` is number?

ibacher (commented):
@icrc-agomes Thanks for this. Just a couple of touch-ups. Nice work!

ibacher (approved):

PR #754, opened 2022-07-25
O3-614- Support I18n for AMPATH forms

icrc-jofrancisco:
## Requirements

- [X] This PR has a title that briefly describes the work done, including the ticket number if there is a ticket.
- [X] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide).

## Summary
The patient chart will be responsible for receiving the translation resources specified in the AMPATH form schema and loading it into the ngxtranslate module, also used in AMPATH forms.
When loaded for the first time, the resource will be saved in the browser storage, similar to what happens with the form schema. On the next load, the patient chart will fetch the cached resource, or, if it's no longer available, fetch it from the server.

## Screenshots
--

## Related Issue
[Issue O3-614](https://issues.openmrs.org/browse/O3-614)

## Other
--


packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -6,18 +6,31 @@ import { FormResourceService } from '../openmrs-api/form-resource.service';
...
   constructor(
     private formsResourceService: FormResourceService,
     private localStorage: LocalStorageService,
     private formSchemaCompiler: FormSchemaCompiler,
+    private translate: ngxTranslate.TranslateService,
   ) {}
 
   public getFormSchemaByUuid(formUuid: string, cached: boolean = true): Observable<FormSchema> {
     const cachedCompiledSchema = this.getCachedCompiledSchemaByUuid(formUuid);
+    const currentLang = this.localStorage.getItem('i18nextLng') || 'en';
```
ibacher, line 22: This seems a bit hacky to me, especially when we can just use `window.i18next.language`...

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -6,18 +6,31 @@ import { FormResourceService } from '../openmrs-api/form-resource.service';
...
+    this.translate.setDefaultLang(currentLang);
+
     if (cachedCompiledSchema && cached) {
+      if (cachedCompiledSchema?.translations) {
+        cachedCompiledSchema.translations.forEach((translationResourceUuid: string) => {
+          let translationResource = this.getCachedTranslationsByUuid(translationResourceUuid);
```
ibacher, line 28:
```suggestion
          const translationResource = this.getCachedTranslationsByUuid(translationResourceUuid);
```

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -6,18 +6,31 @@ import { FormResourceService } from '../openmrs-api/form-resource.service';
...
   ) {}
 
   public getFormSchemaByUuid(formUuid: string, cached: boolean = true): Observable<FormSchema> {
     const cachedCompiledSchema = this.getCachedCompiledSchemaByUuid(formUuid);
+    const currentLang = this.localStorage.getItem('i18nextLng') || 'en';
+    this.translate.setDefaultLang(currentLang);
+
     if (cachedCompiledSchema && cached) {
+      if (cachedCompiledSchema?.translations) {
+        cachedCompiledSchema.translations.forEach((translationResourceUuid: string) => {
```
ibacher, line 27: Couldn't we compile the translations as part of this process and skip even reloading them from `localStorage`? That way we avoid issues like "what if local storage is out of date with the schema".

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -27,11 +40,26 @@ export class FormSchemaService {
...
+
+        if (formSchema?.translations) {
+          formSchema.translations.forEach((translationUuid: string) => {
+            this.getFormTranslationResourceByUuidFromServer(translationUuid)
+              .then((result) => {
+                this.cacheTranslationsByUuid(translationUuid, result);
+                this.translate.setTranslation(result.language, result.translations, true);
+              })
+              .catch((error) => {
+                console.error(`Error getting AMPATH translation resource ${JSON.stringify(error)}.`);
```
ibacher, line 52: It's good to have a `catch()` here. It would be better if we logged the actual error rather than calling `JSON.stringify()` on it. Also, should this display a message to the user?

packages/esm-form-entry-app/src/app/app.module.ts
```diff
@@ -18,6 +18,7 @@ import { ConfigResourceService } from './services/config-resource.service';
 import { LoaderComponent } from './loader/loader.component';
 import { SingleSpaPropsService } from './single-spa-props/single-spa-props.service';
 import { FormCreationService } from './form-creation/form-creation.service';
+import * as ngTranslate from '@ngx-translate/core';
```
ibacher, line 21: Why are we doing this rather than `import { TranslateService } from '@ngx-translate/core'`?

ibacher: In general, `*` imports should be avoided where at all possible.

hadijahkyampeire (approved):
This looks good to me, thanks @icrc-jofrancisco. PS: Please resolve the conflicts.

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -207,4 +239,12 @@ export class FormSchemaService {
 
     return [...results];
   }
+
+  private getCachedTranslationsByUuid(translationUuid: string): any {
+    return this.localStorage.getObject(translationUuid);
```
ibacher, line 244: Could we cache this in a store rather than localStorage? LocalStorage has some nice properties, but it's also going to create a caching headache here (what happens when the translations are updated on the backend but the user's local storage has a cached copy?)

hadijahkyampeire:
I am going to close this PR because it has a duplicate PR here which is the active one https://github.com/openmrs/openmrs-esm-patient-chart/pull/925 cc @icrc-jofrancisco

PR #769, opened 2022-08-15
(chore) upgrade to Yarn v3

ibacher:
## Requirements

## Summary

Pretty simple: gets us to Yarn v3.


package.json
```diff
@@ -72,7 +72,7 @@
     "webpack-cli": "^4.7.0"
   },
   "resolutions": {
-    "**/@types/react": "^18.0.14",
```
ZacButko, line 75: Do we no longer need to resolve React? This has always been painful in the past...

ZacButko (approved):

package.json
ibacher: Well, we might need to re-add it back depending on what happens with React 19, but for now I'd prefer to remove it 😁. (Yarn 3's syntax for this is also somewhat different so the `"**/@types/react"` is actually no longer valid).

PR #770, opened 2022-08-15
(fix)  Fixes concept url

ZacButko:
## Requirements

## Summary
In esm-form-entry-app concept.service.ts was getting concepts from `/rest/v1/conceptconcept?references...` and now it gets them from `/rest/v1/concept?references...`


ibacher:
I think this also needs the inverse tweak over [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/c71f9f53a381ed52e9bec9bd9814b8e9e0dd2f61/packages/esm-form-entry-app/src/app/offline/caching.ts#L94) so that this continues to work offline.

ibacher (approved):

PR #772, opened 2022-08-17
(feat) add the ability to load custom data sources into the AMPATH form engine

ibacher:
## Requirements

## Summary

This a proposal for how to support loading arbitrary custom data sources into the AMPATH form engine. Essentially, this works by having the user specify what custom data sources they want to load through the configuration of the esm-form-entry-app. I've also built a quick example [available here](https://github.com/openmrs/openmrs-esm-custom-data-source-example) of what a data source might look like that also has an example configuration.

This implementation depends on the 4.0 app shell, assumes that all custom data sources are available on the import map, and are loaded as part of the app shell. Among other things, this code _does not_  call `init()` on the module which is necessary for a federated module to be loaded properly.[^1] Custom data sources are assumed to be federated modules and this does not support loading any other type of module packaging. In the 4.0 app shell, this should be a non-issue since all modules are federated.

Limitations: For now, a custom data source is loaded only once, so it is best if data sources do not statically store possibly changable data (e.g. state of the current user, patient or visit—it's fine to have functions that look these things up though).

@denniskigen @enyachoke I'd particularly welcome your feedback on this as I'm not 100% sure I've completely understood data sources here, i.e., among other things, I'm not entirely sure I understand the point of the `unWrap` parameter and if I should add that to the configuration or not. I'm also assuming that implementations can somehow take advantage of these data sources, but it's a bit obscure to me how to do so.

Future work: If someone starts building a large number of these custom data sources there may need to be more optimisations done, i.e., to figure out which data sources are actually required by each form. But for right now, that didn't feel necessary.

PS While there's a few things going on here, the core of the implementation is just this:

```ts
this.dataSources.registerDataSource(name,
	(await window[slugify(moduleName)].get('./start'))();
);
```

[^1]: The technical reason for this is that the global object for tracking module scopes, `__webpack_share_scopes__`, is only made available in Webpack 5, but esm-form-entry-app is built using Webpack 4. By loading things through the app shell, the module will be properly initialized, and once that's done the module itself can be accessed by calling `get()` with no dependencies on Webpack-injected code. Once we can upgrade AMPATH forms and esm-form-entry-app to Angular 12+ this caveat can go away.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -11,20 +11,19 @@
         "paths": [
           {
             "name": "lodash",
-            "message": "Import specific methods from `lodash-es`. e.g. `import map from 'lodash-es/map'`"
+            "message": "Import specific methods from `lodash`. e.g. `import map from 'lodash/map'`"
           },
           {
             "name": "lodash-es",
-            "message": "Import specific methods from `lodash-es`. e.g. `import map from 'lodash-es/map'`"
-          }
-        ],
-        "patterns": [
+            "importNames": ["default"],
+            "message": "Import specific methods from `lodash-es`. e.g. `import { map } from 'lodash-es'`"
+          },
           {
-            "group": ["carbon-components-react"],
+            "name": "carbon-components-react",
             "message": "Import from `@carbon/react` directly. e.g. `import { Toggle } from '@carbon/react'`"
           },
           {
-            "group": ["@carbon/icons-react"],
+            "name": "@carbon/icons-react",
             "message": "Import from `@carbon/react/icons`. e.g. `import { ChevronUp } from '@carbon/react/icons'`"
           }
         ]
--- a/packages/esm-form-entry-app/src/app/fe-wrapper/fe-wrapper.component.ts
+++ b/packages/esm-form-entry-app/src/app/fe-wrapper/fe-wrapper.component.ts
@@ -1,7 +1,7 @@
 import { Component, HostListener, OnDestroy, OnInit } from '@angular/core';
 import { Form } from '@ampath-kenya/ngx-formentry';
 import { Observable, forkJoin, from, throwError, of, Subscription } from 'rxjs';
-import { catchError, map, mergeMap, take } from 'rxjs/operators';
+import { catchError, concatAll, map, mergeMap, take } from 'rxjs/operators';
 import { OpenmrsEsmApiService } from '../openmrs-api/openmrs-esm-api.service';
 import { FormSchemaService } from '../form-schema/form-schema.service';
 import { FormSubmissionService } from '../form-submission/form-submission.service';
@@ -70,7 +70,8 @@ export class FeWrapperComponent implements OnInit, OnDestroy {
     this.launchFormSubscription = this.loadAllFormDependencies()
       .pipe(
         take(1),
-        map((createFormParams) => this.formCreationService.initAndCreateForm(createFormParams)),
+        map((createFormParams) => from(this.formCreationService.initAndCreateForm(createFormParams))),
+        concatAll(),
         mergeMap((form) => {
           const unlabeledConcepts = FormSchemaService.getUnlabeledConceptIdentifiersFromSchema(form.schema);
           return this.conceptService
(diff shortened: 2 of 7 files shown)


ibacher:
@ZacButko Any chance you can take a look at this one?

ZacButko (commented):
Code wise of course this is fine. And I think it will help extend the functionality of custom data sources...

However I'm wondering if this is the right path to go down for custom data sources. A few small things jump out as concerns
- Unprotected plain-name resolution. Different modules can register `custom-patient` and one will be ignored which might be hard to debug. Some name-spacing to origin module could help mitigate this.
- Increases the number of modules that need to be included in a distro (very superficial I know) , and no clear connection that in order for this app to work custom data source module should also be included (more prone to user error). Let's say fast-data-entry-app depends on its own custom-data-srources module... I believe due to disallowing `setupOpenMRS` this would need to be its own package?
- These custom data sources will accumulate at the "top" of the application, always being loaded by form-entry-app and registered to it even if they are never used. I don't know how to make this type of module loading "lazy". In general for OpenMRS I'm concerned that we load way more dependencies at a time than we need to.
- I'm not quite clear how a function is being used after import? For your `custom-patient` example you use a patient search from with uuid as a parameter. What is telling the form engine to use a patient-uuid as the parameter to pass to your `custom-patient` data source?

In broader strokes I think the goal here is to allow a way to pre-fill data into forms, and maybe even make this accessible to BA, or at least not need to make platform / form-entry-app changes to support pre-filling a form field. I'm not sure this method abstracts away enough of the process to make it a significant improvement over say, passing a data object into a form extension-slot directly. In the end you will still need a developer involved to create a custom data source module, not 'link' it to an existing package, but 'situate them so they are always together', and I'm worried that the use case will be uncoupled from the machinery. 

What are your thoughts @ibacher ? Am I misunderstanding this approach?

ZacButko:
(Also, thanks, by the way @ibacher for taking such a thorough look at this, and even creating an example module to go with it)

ibacher:
Thanks @ZacButko for the very thorough review!

It's probably worth saying, up-front, that I think I'm solving a problem that's orthogonal to the problem you're thinking about. In particular, while this may offer a way for custom frontend modules to provide custom data sources, that's actually not the main problem I'm trying to solve with this. The main problem I'm trying to solve is when implementations need implementation-specific logic to make relevant data available to forms loaded by the form engine.

A good example of what I'm trying to support is [this monthly-scheduled-resource](https://github.com/openmrs/openmrs-esm-patient-chart/blob/c71f9f53a381ed52e9bec9bd9814b8e9e0dd2f61/packages/esm-form-entry-app/src/app/services/monthly-scheduled-resource.service.ts) data source. This feature is dependant on an API provided by the AMPATH [ETL Server](https://github.com/AMPATH/etl-rest-server/blob/77074a3e3141ad050148fe11a8c5c322e66210e8/etl-routes.js#L348), so an API that only exists at AMPATH—which is perfectly fine! We have several use-cases like this where implementers need to rely on implementation-specific logic in forms (e.g., the ICRC's logic around what counts as the "previous encounter"). I say that up front because that drives a lot of the responses here.

> Different modules can register custom-patient and one will be ignored which might be hard to debug.

Admittedly, one thing I would like to figure out how to solve is how to notify a user that this happens without also displaying a message every time a custom module gets loaded. That said, I think there are some improvements I can make here.

> Some name-spacing to origin module could help mitigate this.

The reason I'm a little hesitant about automated name-spacing is that the data source name isn't just random [it's part of the form schema](https://github.com/openmrs/openmrs-ngx-formentry/blob/ecec4b55b238a743ce2ec74a1150cfdf22246a5a/projects/ngx-formentry/src/form-entry/question-models/question-base.ts#L29). Although, I suppose there's not much harm in requiring, e.g., `@icrc/custom-data-source:encounter-history` as part of the form schema. But one use-case I'd like to be able to support is actually to clobber the default patient data source with a custom one.

> no clear connection that in order for this app to work custom data source module should also be included

I think this goes away when you realise that the form developer needs to be aware that the custom data source is defined.

> I believe due to disallowing setupOpenMRS this would need to be its own package?

Technically, there's no reason that you can't have `setupOpenMRS

ZacButko (approved):
Thanks for clarifying

PR #774, opened 2022-08-19
(feat) Refurbish Order Basket

samuelmale:
## Requirements

## Summary
This PR is focused on adding comprehensive support for capturing drug orders within the Order Basket 🛒 
- Incorporates support of ordertemplates
- The Drug Order component is aligned with the backend "order-config" 
- Other miscellaneous UX improvements

## Attachements

https://user-images.githubusercontent.com/26084581/185637284-bf078d4e-f389-43a7-8f42-30f73d8d4c1b.mp4

## Related Issue

- Epic: https://issues.openmrs.org/browse/O3-1185


packages/esm-patient-medications-app/src/api/drug-order-template.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export interface DosingInstructions {
+  dose: Array<MedicationDosage>;
+  units: Array<DosingUnit>;
+  route: Array<MedicationRoute>;
+  frequency: Array<MedicationFrequency>;
+  instructions?: Array<MedicationInstructions>;
+  asNeeded?: boolean;
+  asNeededCondtion?: string;
```
denniskigen, line 22:
```suggestion
  asNeededCondition?: string;
```

__mocks__/medication.mock.ts
```diff
@@ -1227,3 +1227,62 @@ export const mockDrugOrders = {
     ],
   },
 };
+
+export const mockOrderTemplates = [
+  {
+    uuid: 'cb43930e-3ed2-4099-ba0c-e6c841e22c4d',
+    display: 'Aspirin',
+    name: 'Aspirin',
+    description: 'Apsirin demo order template',
```
denniskigen, line 1236:
```suggestion
    description: 'Aspirin demo order template',
```

samuelmale: Nice catch!

packages/esm-patient-medications-app/src/order-basket/medication-order-form.component.tsx
```diff
@@ -54,19 +128,23 @@ export default function MedicationOrderForm({
...
               <strong className={styles.dosageInfo}>
-                {capitalize(orderBasketItem.commonMedicationName)} ({orderBasketItem.dosage.dosage})
+                {capitalize(orderBasketItem.commonMedicationName)} {doseWithUnitsLabel && `(${doseWithUnitsLabel})`}
               </strong>{' '}
-              <span className={styles.bodyShort01}>
-                &mdash; {orderBasketItem.route.name} &mdash; {orderBasketItem.dosageUnit.name} &mdash;{' '}
-              </span>
-              <span className={styles.caption01}>{t('dose', 'Dose').toUpperCase()}</span>{' '}
-              <strong className={styles.dosageInfo}>{orderBasketItem.dosage.dosage}</strong>
+              {template && (
```
denniskigen, line 133: Nit: `template` is not a boolean, so there might be occasions where this check leads to a `0` getting rendered. A ternary operator would work better in this circumstance.

```tsx
  {template ? () : null}
```

packages/esm-patient-medications-app/src/order-basket/order-basket-search-results.component.tsx
```diff
@@ -32,12 +33,20 @@ export default function OrderBasketSearchResults({
 
   useEffect(() => {
     const abortController = new AbortController();
-    searchMedications(searchTerm, encounterUuid, abortController, config.daysDurationUnit).then(
-      setSearchResults,
-      createErrorHandler,
-    );
+    if (searchTerm) {
+      if (searchTerm.length == 1) {
```
denniskigen, line 37: If this is intended to minimize the number of times we hit the server, we should consider debouncing the search instead (possibly in a follow-up PR).

denniskigen (approved):
Nice work, @samuelmale. Thanks for getting this done.

packages/esm-patient-medications-app/src/order-basket/order-basket-search-results.component.tsx
samuelmale: Nope, @denniskigen the minimum requirement in terms of the length of a `searchTerm` by the server is 2. Other words to expect non empty result(s) from the server, the length of the search should be greater that 1. But that doesn't nullify your suggestion, we definitely need to debounce..

packages/esm-patient-medications-app/src/order-basket/medication-order-form.component.tsx
samuelmale: So template is type "object", it's very unlikely for an order template's datatype to be otherwise. It's either an object if in a given scenario on was found or null.

denniskigen:
Testing this out on dev3, I don't get back any results when searching for a drug using the search box. Should this be the case, @samuelmale?

samuelmale:
@denniskigen yes that is happening because dev3 has no Concept Drugs and actual Drugs. I discussed with @gracepotma about this and she promised adding some Drugs

denniskigen:
Thanks for clarifying.

denniskigen:
@samuelmale would you mind reminding me why we got rid of the `Prescription Duration` and `Dispensing Information` sections of the form? I've forgotten the context?

samuelmale:
@denniskigen We removed some of those fields as an after-action for one of the many TAC calls that happened about Drug Order. Experts (BAs) in the program area reckoned that most of those fields were a dispensing detail and didn't make sense having them in the Order form. Looping in @ebambo

PR #776, opened 2022-08-22
(fix) reduce number of form re-renders

ibacher:
## Requirements

## Summary

**THIS DOES NOT SOLVE THE FORM RE-RENDERING ISSUE**

However, it does substantially _reduce_ the number of re-renders. While many of these changes really only affect the form entry part, the changes to `usePatientOrOfflineRegisteredPatient()` and `useVisitOrOfflineVisit()` should be a little more generally useful.

### DIFF
--- a/packages/esm-patient-common-lib/src/offline/patient.ts
+++ b/packages/esm-patient-common-lib/src/offline/patient.ts
@@ -1,5 +1,5 @@
-import { usePatient, getSynchronizationItems } from '@openmrs/esm-framework';
-import useSWR from 'swr';
+import { useEffect, useMemo, useState } from 'react';
+import { usePatient, getSynchronizationItems, useConnectivity } from '@openmrs/esm-framework';
 
 export const patientRegistrationSyncType = 'patient-registration';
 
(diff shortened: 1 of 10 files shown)


packages/esm-patient-common-lib/src/offline/patient.ts
```diff
@@ -8,53 +8,66 @@ export const patientRegistrationSyncType = 'patient-registration';
...
-      isLoading: false,
-      patient: onlinePatientState.patient,
-      patientUuid,
-      error: null,
-    };
-  }
+    if (offlinePatientState.data) {
+      return {
+        isLoading: false,
+        patient: offlinePatientState.data,
```
ZacButko, line 54: If `offlinePatientState.data` is supposed to contain multiple things I would change this to `offlinePatientState.data.patient`. Otherwise I would clean up as `offlinePatientState.patient`

ZacButko (approved):
Thanks Ian

packages/esm-patient-common-lib/src/offline/patient.ts
ibacher: Yeah... let me fix that... this is an artifact left over from replacing `useSWR()` with an initially identical API so I didn't screw up the rest of the logic.

PR #781, opened 2022-08-24
feat(form-entry-app): Add a hook to handle the encounter before it's created

ibacher:
## Requirements

## Summary

@ZacButko I think this is the hook you mentioned needing?

Always a wrapping component to add a handler to handle the form-engine's `EncounterCreate` object. Basically, allows a wrapper to modify the encounter that will be created before it's created.


ZacButko (approved):
Thanks Ian.

I think the name "handleEncounterCreate" is a little misleading. Maybe "modifyEncounterCreate" scopes the function responsibility better?

ibacher:
@ZacButko I went with `handle...()` mostly because you can basically do whatever you want with it... it's not dissimilar to a event handler.

PR #789, opened 2022-08-29
(fix) Fix vitals header flagging logic

denniskigen:
## Requirements

## Summary

This PR fixes the logic used to determine whether to show a warning background in the vitals header. Presently, that logic fails because of a bug in the `assessAllValues` function where the key-value pairs that get iterated over do not get correctly destructured from the source `Object.entries` array. I've replaced this function entirely with a function that only looks at the interpretation results for each vital sign and returns a boolean depending on whether any vital signs with abnormal interpretations exist.

## Screenshots

> Before
Screenshot 2022-08-29 at 14 10 12

> After
Screenshot 2022-08-29 at 14 09 23

### DIFF
--- a/packages/esm-patient-vitals-app/src/vitals/vitals-header/vitals-header.test.tsx
+++ b/packages/esm-patient-vitals-app/src/vitals/vitals-header/vitals-header.test.tsx
@@ -74,6 +74,7 @@ describe('VitalsHeader: ', () => {
     expect(screen.getByText(/Last recorded/i)).toBeInTheDocument();
     expect(screen.getByText(/19 — May — 2021/i)).toBeInTheDocument();
     expect(screen.getByText(/Record vitals/i)).toBeInTheDocument();
+    expect(screen.getByRole('img', { name: /warning/i })).toBeInTheDocument();
 
     const expandButton = screen.getByTitle(/ChevronDown/);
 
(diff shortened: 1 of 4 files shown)


packages/esm-patient-vitals-app/src/vitals/vitals-header/vitals-header.test.tsx
```diff
@@ -74,6 +74,7 @@ describe('VitalsHeader: ', () => {
     expect(screen.getByText(/Last recorded/i)).toBeInTheDocument();
     expect(screen.getByText(/19 — May — 2021/i)).toBeInTheDocument();
     expect(screen.getByText(/Record vitals/i)).toBeInTheDocument();
+    expect(screen.getByRole('img', { name: /warning/i })).toBeInTheDocument();
```
ZacButko, line 77: I'm curious what is role='img' here.. is that the icon?

ZacButko (approved):

packages/esm-patient-vitals-app/src/vitals/vitals-header/vitals-header.test.tsx
denniskigen: Yep, this one https://github.com/openmrs/openmrs-esm-patient-chart/blob/b98af1ecd254a97c37bd84d73ee4f6ed03565326/packages/esm-patient-vitals-app/src/vitals/vitals-header/vitals-header.component.tsx#L65/

PR #790, opened 2022-08-29
03-1429: Appointment Save and Close button.

hadijahkyampeire:
## Requirements

## Summary
Remove the condition for a selected service type as it was the one returning and stopping the submitting

## Screenshots
This is to confirm that the submitting is now making an API call as expected.
Screenshot 2022-08-30 at 16 19 38

## Related Issue

[Appointment Save and Close button](https://issues.openmrs.org/projects/O3/issues/O3-1429)

### DIFF
--- a/packages/esm-patient-appointments-app/src/appointments/appointments-form.component.tsx
+++ b/packages/esm-patient-appointments-app/src/appointments/appointments-form.component.tsx
@@ -52,8 +52,6 @@ const AppointmentsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, closeW
   }
 
   const handleSubmit = () => {
-    if (!selectedServiceType) return;
-
     const serviceType = serviceTypes.find((service) => service.name === selectedServiceType);
 
     const serviceUuid = services.find((service) => service.name === selectedService)?.uuid;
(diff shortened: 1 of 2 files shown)


packages/esm-patient-appointments-app/src/appointments/appointments-form.component.tsx
```diff
@@ -52,8 +52,6 @@ const AppointmentsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, closeW
   }
 
   const handleSubmit = () => {
-    if (!selectedServiceType) return;
```
ZacButko, line 55: Isn't `selectedServiceType` necessary for submitting?

hadijahkyampeire: It is not a required field by the backend but we were requiring it since we were getting the duration to make the `endTime` from the serviceType, but now that we mostly have no service types and we can access the durations on the service top level, it's better we don't make it a condition for submitting.

hadijahkyampeire:
@donaldkibet I have added a screenshot.

packages/esm-patient-appointments-app/src/appointments/appointments-form.component.tsx
ZacButko: @hadijahkyampeire ok, makes sense. What do you think @CynthiaKamau ?

hadijahkyampeire: @ZacButko do you know if it is possible to make appointment endpoints point to this backend because it is the one that works? `backend=[https://](http://loc/)data.kenyahmis.org:8500 --api-url=/openmrs`

ZacButko: @hadijahkyampeire  I'm not sure what you're asking.. for dev3 It would be preferable to get appointments working on that setup. If it's not pointing to a working instance right now we should make a talk post to get backend team to look at it.

if the question is just about syntax the better syntax would be `--backend "https://data.kenyahmis.org --port 8500 ...`

ZacButko (approved):

PR #796, opened 2022-09-02
(feat) Automatically revalidate order data after form submission

denniskigen:
## Requirements

## Summary

This PR is a follow-up to https://github.com/openmrs/openmrs-esm-patient-chart/pull/774. It adds a call to SWR's `mutate` function to the orders form submission handler so that when the form gets submitted successfully, a revalidation request gets broadcasted to SWR hooks using the same key. This means the user does not need to reload the patient chart to see newly created orders (as @samuelmale showed in the video demo in #774).

This PR also wraps an empty state `Tile` in the OrderBasketItemList component with the `Layer` component. This harmonizes its look and feel with that of the other empty states in the app.

### DIFF
--- a/packages/esm-patient-medications-app/src/order-basket/order-basket.component.tsx
+++ b/packages/esm-patient-medications-app/src/order-basket/order-basket.component.tsx
@@ -26,7 +26,7 @@ const OrderBasket = connect<OrderBasketProps, OrderBasketStoreActions, OrderBask
   orderBasketStoreActions,
 )(({ patientUuid, items, closeWorkspace, setItems }: OrderBasketProps & OrderBasketStore & OrderBasketStoreActions) => {
   const { t } = useTranslation();
-  const { mutate } = useSWRConfig();
+  const { cache, mutate }: { cache: any; mutate: Function } = useSWRConfig();
   const displayText = t('activeMedicationsDisplayText', 'Active medications');
   const headerTitle = t('activeMedicationsHeaderTitle', 'active medications');
   const isTablet = useLayoutType() === 'tablet';
@@ -109,7 +109,14 @@ const OrderBasket = connect<OrderBasketProps, OrderBasketStoreActions, OrderBask
           ),
         });
 
-        mutate(`/ws/rest/v1/order`);
+        const apiUrlPattern = new RegExp(
+          '\\/ws\\/rest\\/v1\\/order\\?patient\\=' + patientUuid + '\\&careSetting=' + config.careSettingUuid,
+        );
+
+        // Find matching keys from SWR's cache and broadcast a revalidation message to their pre-bound SWR hooks
+        Array.from(cache.keys())
+          .filter((url: string) => apiUrlPattern.test(url))
+          .forEach((url: string) => mutate(url));
       }
     });
     return () => abortController.abort();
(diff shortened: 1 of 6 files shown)


packages/esm-patient-medications-app/src/order-basket/order-basket.component.tsx
```diff
@@ -131,64 +135,60 @@ const OrderBasket = connect<OrderBasketProps, OrderBasketStoreActions, OrderBask
...
+            onItemClicked={(order) => openMedicationOrderFormForUpdatingExistingOrder(items.indexOf(order))}
+            onItemRemoveClicked={(order) => {
+              const newOrders = [...items];
+              newOrders.splice(items.indexOf(order), 1);
+              setItems(newOrders);
+            }}
+          />
+          {(() => {
+            if (isLoadingOrders) return <DataTableSkeleton role="progressbar" />;
+            if (isError) return <ErrorState error={isError} headerTitle={headerTitle} />;
```
ZacButko, line 166: `ErrorState` prefers if the actual error response object is passed, but this error is named like it should be a boolean. Would it be appropriate here to pass the error response object instead? Or label the variable so we don't think its going to be a boolean?

ZacButko (approved):
This looks it will be useful. I'm curious if you can think of a way to add a test to cover this functionality?

packages/esm-patient-medications-app/src/order-basket/order-basket.component.tsx
denniskigen: Yes, that variable is wrongly named. I've amended that in the latest commit.

denniskigen:
> This looks like it will be useful. I'm curious if you can think of a way to add a test to cover this functionality.

We use this pattern in multiple places, and those usages are backed up with tests that assert that a success notification gets rendered when the form gets submitted successfully. Adding a test for this shouldn't be difficult, though I'd prefer to do it in a separate PR.

ZacButko:
It's good that tests check for the notification popup on endpoint success. I was specifically wondering about the `mutate` affecting a separate piece of the UI which are using the same endpoint

denniskigen:
That's a valid concern. There's a pattern we could use to guard against the wrong keys being inadvertently revalidated https://github.com/openmrs/openmrs-esm-patient-chart/blob/441052b8f919d16918f51043f7cddb277434fbe7/packages/esm-patient-vitals-app/src/vitals/vitals-biometrics-form/vitals-biometrics-form.component.tsx#L76. 

For this specific key, we could use the following RegExp to match it exactly:

```tsx
const apiUrlPattern = new RegExp(
  '\\/ws\\/rest\\/v1\\/order\\?patient=' + patientUuid + '\\&careSetting=' + config.careSettingUuid,
)

// Find matching keys from SWR's cache and broadcast a revalidation message to their pre-bound SWR hooks
Array.from(cache.keys())
  .filter((url: string) => apiUrlPattern.test(url))
  .forEach((url: string) => mutate(url));
```

samuelmale:
@denniskigen I think adding the regex based filter is our based option here.

samuelmale (approved):
Nice work @denniskigen

ZacButko (approved):

PR #797, opened 2022-09-02
(chore) Bump `@carbon/react`, `@carbon/charts-react` and `d3`

denniskigen:
## Requirements

## Summary

Just some good ol' housekeeping. No breaking changes.

### DIFF
--- a/packages/esm-generic-patient-widgets-app/package.json
+++ b/packages/esm-generic-patient-widgets-app/package.json
@@ -35,9 +35,9 @@
   },
   "dependencies": {
     "@carbon/charts-react": "^1.5.2",
-    "@carbon/react": "^1.9.0",
+    "@carbon/react": "^1.12.0",
     "@openmrs/esm-patient-common-lib": "^4.0.0",
-    "d3": "^7.0.3",
+    "d3": "^7.6.1",
     "lodash-es": "^4.17.15"
   },
   "peerDependencies": {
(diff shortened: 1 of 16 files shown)


ZacButko (commented):
Hey Dennis. I think in general it's great to upgrade packages. looking at these `package.lock` files for me raises more questions than it answers.
1. I thought we had fully moved to listing `@carbon/react` as a `devDependency` and `peerDependency` for all of these? If not we totally should to avoid @carbon/react being loaded multiple times. 
2. I think d3 is hardly if ever used. It's been on my to-do list to clean out the unnecessary dependence on it. Is this something you want to do?

Sorting out all of these awkward peer dependencies is something I'm really looking forward to and one of the reasons I'm excited to move to [one monorepo](https://talk.openmrs.org/t/o3-merging-monorepos/37438/2).

denniskigen:
Re: `d3`, it's needed for carbon charts to work correctly. See https://charts.carbondesignsystem.com/react/?path=/story/docs-getting-started--instructions.

About `@carbon/react`, I vaguely recall us doing that. Though I think carbon/react is installed like so in all our monorepos.

ZacButko:
I know d3 is used by `@carbon/charts-react`. But for example `esm-patient-allergies-app` does not use `@carbon/charts-react` and I can see no other reason why d3 should be included as a dependency for that package.

denniskigen:
Correct. `@carbon/charts-react` shouldn't be used in the allergies app. I'll fix that.

ZacButko (approved):
The PR is fine by the way. Thanks again.

PR #805, opened 2022-09-12
(chore) run pre-release before deploy

ibacher:
Also patches some unnecessary dependencies

## Requirements

## Summary

This should correctly version preleases before they are deployed

### DIFF
--- a/packages/esm-patient-immunizations-app/package.json
+++ b/packages/esm-patient-immunizations-app/package.json
@@ -41,7 +41,7 @@
     "@openmrs/esm-framework": "4.x",
     "@openmrs/esm-patient-common-lib": "4.x",
     "dayjs": "1.x",
-    "react": "4.x",
+    "react": "18.x",
     "react-i18next": "11.x",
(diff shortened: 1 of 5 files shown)


ibacher:
So, we're a little stuck with Node 16 ATM for this repo. The culprit here seems to be Angular which is using Webpack 4. There's [this workaround](https://github.com/webpack/webpack/issues/14532#issuecomment-947012063), but I'd rather not apply that if we can avoid it.

denniskigen (approved):

packages/esm-patient-immunizations-app/package.json
```diff
@@ -41,7 +41,7 @@
     "@openmrs/esm-framework": "4.x",
     "@openmrs/esm-patient-common-lib": "4.x",
     "dayjs": "1.x",
-    "react": "4.x",
```
ZacButko, line 44: 😅

ZacButko (approved):

PR #807, opened 2022-09-14
enhancement: Add identifier type name alongside value on patient-banner

donaldkibet:
## Requirements

## Summary
- Enhancement add identifier type name alongside the value to enable users know which identifier they are looking at. Screenshot for reference
- Remove grey underscore on active visit tag 

## Screenshots
Screenshot from 2022-09-14 21-59-22


makombe:
@donaldkibet  thanks for this, I think you can rearrange them display key then value. where key is the identifiertype eg `Openmrs Id: MGJAUL`

CynthiaKamau (approved):
Thanks Donald

packages/esm-patient-banner-app/src/banner/patient-banner.scss
```diff
@@ -1,6 +1,8 @@
+@use '@carbon/styles/scss/spacing';
 @use '@carbon/styles/scss/type';
 @import '~@openmrs/esm-styleguide/src/vars';
 
+
 .container {
   border-top: 0.063rem solid $ui-03;
```
CynthiaKamau, line 7: Why not use carbon spacing

denniskigen: Counterpoint: I actually reckon we shouldn't be using the Carbon spacing tokens - they're not parseable at a glance in my opinion and I reckon we should prefer using em/rem-based spacing directly.

PR #813, opened 2022-09-21
Support for rendering values from older to newer on generic widgets

icrc-jofrancisco:
## Requirements

## Summary
As a developer, I want it to be possible to use a configuration to invert the order of the values. By inverting the data, it will be possible to render the values from the oldest to the most recent.

## Screenshots

Before:
image

After (with dateAxisReverse config set to true):
image

## Related Issue
None
## Other

None

### DIFF
--- a/packages/esm-generic-patient-widgets-app/src/config-schema.ts
+++ b/packages/esm-generic-patient-widgets-app/src/config-schema.ts
@@ -11,7 +11,7 @@ export const configSchema = {
     _description: 'Displayed in messages about this data',
     _default: 'results',
   },
-  dateAxisReverse: {
+  graphOldestFirst: {
     _type: Type.Boolean,
     _description: 'Show graph values from most oldest to recent',
     _default: false,
@@ -57,7 +57,7 @@ export const configSchema = {
 export interface ConfigObject {
   title: string;
   resultsName: string;
-  dateAxisReverse: boolean;
+  graphOldestFirst: boolean;
   data: Array<{
     concept: string;
     label: string;
--- a/packages/esm-generic-patient-widgets-app/src/config-schema.ts
+++ b/packages/esm-generic-patient-widgets-app/src/config-schema.ts
@@ -11,6 +11,11 @@ export const configSchema = {
     _description: 'Displayed in messages about this data',
     _default: 'results',
   },
+  dateAxisReverse: {
+    _type: Type.Boolean,
+    _description: 'Show graph values from most oldest to recent',
+    _default: false,
+  },
   data: {
     _type: Type.Array,
     _elements: {
@@ -52,6 +57,7 @@ export const configSchema = {
 export interface ConfigObject {
   title: string;
   resultsName: string;
+  dateAxisReverse: boolean;
   data: Array<{
     concept: string;
     label: string;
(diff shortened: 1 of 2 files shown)


packages/esm-generic-patient-widgets-app/src/config-schema.ts
```diff
@@ -11,6 +11,11 @@ export const configSchema = {
     _description: 'Displayed in messages about this data',
     _default: 'results',
   },
+  dateAxisReverse: {
```
ibacher, line 14: Maybe call it: `graphOldestFirst`

packages/esm-generic-patient-widgets-app/src/obs-graph/obs-graph.component.tsx
```diff
@@ -33,17 +33,20 @@ const ObsGraph: React.FC<ObsGraphProps> = ({ patientUuid }) => {
...
-        .filter((obs) => obs.conceptUuid === selectedConcept.uuid && obs.dataType === 'Number')
-        .map((obs) => ({
-          group: selectedConcept.label,
-          key: formatDate(new Date(obs.issued), { year: false, time: false }),
-          value: obs.valueQuantity.value,
-        })),
-    [obss, selectedConcept.uuid, selectedConcept.label],
-  );
+  const chartData = useMemo(() => {
+    let chartRecords = obss
```
ibacher, line 37:
```suggestion
    const chartRecords = obss
```

packages/esm-generic-patient-widgets-app/src/obs-graph/obs-graph.component.tsx
```diff
@@ -33,17 +33,20 @@ const ObsGraph: React.FC<ObsGraphProps> = ({ patientUuid }) => {
...
+      .filter((obs) => obs.conceptUuid === selectedConcept.uuid && obs.dataType === 'Number')
+      .map((obs) => ({
+        group: selectedConcept.label,
+        key: formatDate(new Date(obs.issued), { year: false, time: false }),
+        value: obs.valueQuantity.value,
+      }));
+
+    if (config?.graphOldestFirst) {
```
ibacher, line 45:
```suggestion
    if (config.graphOldestFirst) {
```

ibacher (approved):

gracepotma:
@ibacher I see you approved this PR already - were you waiting on anything in particular to Squash & Merge?

Quick question as well - what was the use case for this feature? Reason I ask is, we were really concerned about keeping these charts consistent, because of the potential for users to just glance at the trendline and assume a Left-to-Right timeline, without double-checking the X-axis. So we had tried to keep L2R chart layout consistent. 
Is the main use case to support R2L languages?

icrc-jofrancisco:
The idea here is to be able to see the data from the oldest to the most recent values. If this PR is accepted, this option is only available if the configuration is set up for it. In theory, if you don't change the configuration (as it is now), the user keeps with the same view. however, if the "graphOldestFirst" property is set to "true", the graph will be sorted chronologically (old to new). 
This implementation was requested for the ICRC. 
I hope I have answered the question.

Thanks,

icrc-jofrancisco:
Thanks for the great clarification @ibacher !
Quick answer to your question: No.
However, the code that was made is just inverting the order of the values, I don't think it's a problem, The performance should be the same.
Thanks,

PR #814, opened 2022-09-22
(feat) Add a forms dashboard section that can be configured

icrc-loliveira:
## Requirements

## Summary
Add a forms dashboard that with different sections that can be configured. This dashboard will be shown when pressing the form button on the left side of the app.

## Screenshots
image
Note: second button is because of test env.

## Related Issue
None

## Other
None


ibacher:
@icrc-loliveira Could you resolve the conflict? I think you need to rebase on `main` and re-run `yarn install`.

packages/esm-forms-dashboard-app/src/hooks/use-forms.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    const dynamicFormData = await getDynamicOfflineDataEntries('form');
+    return forms.filter((form) => dynamicFormData.some((entry) => entry.identifier === form.uuid));
+  });
+}
+
+export function useEncounters(patientUuid: string) {
+  const { useCurrentVisitDates } = useConfig() as ConfigObject;
+  const visits = useVisit(patientUuid);
+  let startDate = dayjs(new Date()).startOf('year').subtract(2, 'years').toDate().toISOString();
```
ibacher, line 35: Why two years? Isn't the right thing here to have `startDate` be `null` and just not add the parameter if has a value? Maybe instead of just doing string concatenation here, it would be clearer to use something like [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) to construct the query? That would allow you to programmatically construct the object and then just have:

```ts
return useSWR(`${encounterUrl}?${urlParams.toString()}` /* ... */)
```

packages/esm-forms-dashboard-app/src/hooks/use-forms.ts
```diff
@@ -0,0 +1,141 @@
...
+  let endDate = dayjs(new Date()).endOf('day').toDate().toISOString();
+
+  if (useCurrentVisitDates) {
+    startDate = visits?.currentVisit
+      ? new Date(visits.currentVisit.startDatetime).toISOString() // Date visit
+      : dayjs(new Date()).startOf('year').add(1, 'year').toDate().toISOString(); // Date in the future
+  }
+
+  const url = `/ws/rest/v1/encounter?v=${customEncounterRepresentation}&patient=${patientUuid}&fromdate=${startDate}&todate=${endDate}`;
+  return useSWR(url, (url) => openmrsFetch<ListResponse<EncounterWithFormRef>>(url));
```
ibacher, line 45:
```suggestion
  return useSWR<ListResponse<EncounterWithFormRef>>(url, openmrsFetch);
```

packages/esm-forms-dashboard-app/src/config-schema.ts
```diff
@@ -0,0 +1,81 @@
...
+  showConfigurableForms: {
+    _type: Type.Boolean,
+    _description: 'Whether to use customURL to fetch forms, works with `customFormsUrl`',
+    _default: false,
+  },
+  customFormsUrl: {
+    _type: Type.String,
+    _description: 'Custom forms endpoint to fetch forms using a custom url need `showConfigurableForms`',
+    _default: '',
+  },
```
ibacher, line 51: Can't we just combine these into one setting? I.e., if you have a `customFormsUrl` then we use that and if not we use the default?

packages/esm-forms-dashboard-app/src/hooks/use-forms.ts
```diff
@@ -0,0 +1,141 @@
...
+import useSWR from 'swr';
+import { getDynamicOfflineDataEntries, openmrsFetch, useConfig, useVisit } from '@openmrs/esm-framework';
+import { ListResponse, Form, EncounterWithFormRef, CompletedFormInfo, FormsSection } from '../types';
+import { customEncounterRepresentation, formEncounterUrl, formEncounterUrlPoc } from '../constants';
+import { ConfigObject, FormsSectionConfig } from '../config-schema';
+
+export function useForms(cachedOfflineFormsOnly = false, patientUuid: string = '') {
+  const { showConfigurableForms, customFormsUrl, showHtmlFormEntryForms } = useConfig() as ConfigObject;
+  const url = showConfigurableForms
+    ? customFormsUrl.concat(`?patientUuid=${patientUuid}`)
```
ibacher, line 11: Instead of this `concat()`, it would be better to use the [`interpolateUrl()`](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-config/src/navigation/interpolate-string.ts#L36) function from the framework. The patientUuid could be supplied as an additional property, but that way we're giving people more freedom over the URL they use...

packages/esm-forms-dashboard-app/src/forms-dashboard/forms-dashboard.component.tsx
```diff
@@ -0,0 +1,61 @@
+import React, { useState } from 'react';
+import { getConfig } from '@openmrs/esm-framework';
+import { useTranslation } from 'react-i18next';
+import styles from './forms-dashboard.scss';
+import { EmptyDataIllustration, useVisitOrOfflineVisit } from '@openmrs/esm-patient-common-lib';
+import FormsList from './form-list/forms-list.component';
+import { useFormsToDisplay } from '../hooks/use-forms';
+import { DataTableSkeleton, Layer, Tile } from '@carbon/react';
+import { HtmlFormEntryForm } from '@openmrs/esm-patient-forms-app/src/config-schema';
```
ibacher, line 9: This is actually exactly the kind of thing we want to avoid as we now have a strong dependency between `esm-froms-dashboard-app` and `esm-patient-forms-app`, i.e., they are no longer self-contained. We should either just duplicate the interface here or define it in, e.g. `@openmrs/esm-patient-common-lib` (which is intended to be shared).

packages/esm-forms-dashboard-app/src/forms-dashboard/form-list/forms-list.component.tsx
```diff
@@ -0,0 +1,182 @@
...
+  TableRow,
+} from '@carbon/react';
+import { Edit } from '@carbon/react/icons';
+import { CompletedFormInfo, FormsSection } from '../../types';
+import { ConfigObject } from '../../config-schema';
+import { defaultPaginationSize } from '../../constants';
+import { launchFormEntryOrHtmlForms } from '@openmrs/esm-patient-forms-app/src/form-entry-interop';
+import { HtmlFormEntryForm } from '@openmrs/esm-patient-forms-app/src/config-schema';
+
+export interface FormsListPros {
```
ibacher, line 28:
```suggestion
export interface FormsListProps {
```

packages/esm-forms-dashboard-app/src/forms-dashboard/form-list/forms-list.component.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+const FormsList: React.FC<FormsListPros> = ({
+  patientUuid,
+  patient,
+  visit,
+  formsSection,
+  searchTerm,
+  pageSize = defaultPaginationSize,
+  htmlFormEntryForms,
+}: FormsListPros) => {
```
ibacher, line 46:
```suggestion
}) => {
```

packages/esm-forms-dashboard-app/src/forms-dashboard/forms-dashboard.component.tsx
```diff
@@ -0,0 +1,61 @@
...
+  }, []);
+
+  if (data) {
+    return (
+      <div className={styles.widgetCard}>
+        {data?.map((formsSection, i) => {
+          let pageSize = undefined;
+          return (
+            <FormsList
+              {...{ patientUuid, patient, visit: currentVisit, formsSection, searchTerm, pageSize, htmlFormEntryForms }}
```
ibacher, line 38: I think that passing `htmlFormEntryForms` here to the formList is a mistake. There's really no good reason to be loading the `htmlFormEntryForms` setting here rather than in the form list.

packages/esm-forms-dashboard-app/src/forms-dashboard/form-list/forms-list.component.tsx
```diff
@@ -0,0 +1,182 @@
...
+  TableContainer,
+  TableHead,
+  TableHeader,
+  TableRow,
+} from '@carbon/react';
+import { Edit } from '@carbon/react/icons';
+import { CompletedFormInfo, FormsSection } from '../../types';
+import { ConfigObject } from '../../config-schema';
+import { defaultPaginationSize } from '../../constants';
+import { launchFormEntryOrHtmlForms } from '@openmrs/esm-patient-forms-app/src/form-entry-interop';
```
ibacher, line 25: This, also, absolutely should not be imported from `esm-patient-forms-app`.

packages/esm-forms-dashboard-app/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@openmrs/esm-forms-dashboard-app",
+  "version": "1.0.0",
```
ibacher, line 3: Version number here should match the other ESMs in this repo for now.

packages/esm-forms-dashboard-app/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "license": "MPL-2.0",
+  "description": "Forms dashboard",
```
ibacher, line 5: So, dashboard has a specific meaning inside the patient chart, specifically, a configuration of bits on the left menu. I would suggest we rename this `esm-forms-workspace-app` or something similar and change the necessary places. It would also be nice if this description told me anything about hat this does, e.g., "Widget to open forms from the workspace on the patient chart" or something.

packages/esm-forms-dashboard-app/src/index.ts
```diff
@@ -0,0 +1,47 @@
...
+            default: 'Forms Dashboard',
+          },
+          type: 'forms-dashboard',
+        },
+      },
+    ],
+  };
+}
+
+export { backendDependencies, importTranslation, setupOpenMRS };
```
ibacher, line 47: This should also have the `version` export like all the other MFs.

packages/esm-forms-dashboard-app/src/index.ts
```diff
@@ -0,0 +1,47 @@
...
+        name: 'form-action-menu-item',
+        slot: 'action-menu-items-slot',
+        load: getAsyncLifecycle(
+          () => import('./clinical-form-action-button/clinical-form-action-button.component'),
+          options,
+        ),
+        order: 2,
+      },
+      {
+        name: 'forms-dashboard-workspace',
```
ibacher, line 33: It would actually be better to use the `registerWorkspace` function from patient-common-lib. See, e.g., [esm-patient-forms-app](https://github.com/openmrs/openmrs-esm-patient-chart/blob/9e3c013770b9fdd641e9a7efdc1a5ab63193fb07/packages/esm-patient-forms-app/src/index.ts)

ibacher (dismissed):
Thanks @icrc-loliveira! I haven't gone through everything, but here are some initial comments.

I think some screenshots or screen caps would be helpful to evaluate this.

Also, why did you move the clinical-form-action-button here?

packages/esm-forms-dashboard-app/src/hooks/use-forms.ts
icrc-loliveira: I try this but I have following error:
```
No overload matches this call.
```

packages/esm-forms-dashboard-app/src/forms-dashboard/forms-dashboard.component.tsx
icrc-loliveira: We have discussed this specif case in one of our meetings.
In this case I am using the `getConfig` to use the `esm-patient-forms-app` config so we don't have to replicate the configuration to open the O2 forms. 
So I think it make sence also use the interfaces for that config that already exist.

packages/esm-forms-dashboard-app/src/hooks/use-forms.ts
icrc-loliveira: I have done it to keep a similar beavior of `esm-patient-forms-app`
```
startDate: Date = dayjs(new Date()).startOf('day').subtract(500, 'day').toDate(),
```

I just use 2 years instead of 500 days, it make more logic for me. I can try to remove it or try to use the 500 days. How should I proceed?

packages/esm-forms-dashboard-app/src/forms-dashboard/forms-dashboard.component.tsx
icrc-loliveira: Actually the `htmlFormEntryForms` has a configuration for all sections, That means if I do it  in the `formList` I will have to do the same call multiple times while in the `FormsDashboard` I only have to do ir onece.

gracepotma:
@denniskigen can I humbly request that you take over for @ibacher on this PR and help @icrc-loliveira see this through to completion? (Normally I'd ask Vineet to review feature-related tickets and reserve your time for O3-framework-related issues, but this seems like a situation where your deeper knowledge of forms is relevant - LMK if you this this would be better for Vineet to take on though.)

I'm so excited to see this PR - it will be **great** to have this feature included!!

icrc-loliveira:
Hello @gracepotma,
I have added the new info in the README.

vasharma05:
Hi @ibacher, can you please re-review the PR changes after your previous review.
Thanks a lot!

packages/esm-forms-workspace-app/src/config-schema.ts
```diff
@@ -0,0 +1,81 @@
...
+        _type: Type.String,
+        _description: 'Name of the section.',
+      },
+      labelCode: {
+        _type: Type.String,
+        _description: 'Label to be used to translate the section.',
+      },
+      forms: {
+        _type: Type.Array,
+        _description: 'Configuration of the forms.',
```
ibacher, line 17:
```suggestion
        _description: 'List of forms that should be associated with this section',
```

packages/esm-forms-workspace-app/src/config-schema.ts
```diff
@@ -0,0 +1,81 @@
...
+        _elements: {
+          name: {
+            _type: Type.String,
+            _description: 'Name of form.',
+          },
+          uuid: {
+            _type: Type.String,
+            _description: 'Uuid of the form.',
```
ibacher, line 25: Do we need both of these? Isn't there a UUID type? Rather than having an element named UUID, I'd think this could just be a list of form identifiers (i.e., things we can use to uniquely identify a form in the /forms REST endpoint)...

packages/esm-forms-workspace-app/src/config-schema.ts
```diff
@@ -0,0 +1,81 @@
...
+    _default: undefined,
+  },
+  showHtmlFormEntryForms: {
+    _type: Type.Boolean,
+    _default: true,
+    _description: 'Whether HTML Form Entry forms should be included in lists of forms.',
+  },
+  showRecommendedFormsTab: {
+    _type: Type.Boolean,
+    _description: 'Whether to display recommended forms tab. Requires AMPATH ETL.',
```
ibacher, line 39:
```suggestion
    _description: 'Whether to display recommended forms tab.',
```

packages/esm-forms-workspace-app/src/constants.ts
```diff
@@ -0,0 +1,8 @@
+export const customFormRepresentation =
+  '(uuid,name,display,encounterType:(uuid,name,viewPrivilege,editPrivilege),version,published,retired,resources:(uuid,name,dataType,valueReference))';
+export const customEncounterRepresentation = `custom:(uuid,encounterDatetime,encounterType:(uuid,name,viewPrivilege,editPrivilege),form:${customFormRepresentation}`;
+
+export const formEncounterUrl = `/ws/rest/v1/form?v=custom:${customFormRepresentation}`;
+export const formEncounterUrlPoc = `/ws/rest/v1/form?v=custom:${customFormRepresentation}&q=poc`;
+
+export const defaultPaginationSize = 10000;
```
ibacher, line 8: Shouldn't these just go in the place that we're making the API calls? I don't really understand why we need them here...

packages/esm-forms-workspace-app/src/declarations.d.tsx
```diff
@@ -0,0 +1,3 @@
+declare module '*.css';
+declare module '*.scss';
+declare module '*.png';
```
ibacher, line 3:
```suggestion
```

packages/esm-forms-workspace-app/src/config-schema.ts
```diff
@@ -0,0 +1,81 @@
...
+    _type: Type.String,
+    _description: 'Custom forms endpoint to fetch forms using a custom url need `showConfigurableForms`',
+    _default: '',
+  },
+  useCurrentVisitDates: {
+    _type: Type.Boolean,
+    _description: 'Only load encounters inside of the current visits dated.',
+    _default: false,
+  },
+  orderFormsByName: {
```
ibacher, line 57: Rather than having endless numbers of boolean properties, we should try to have as few properties as possible. I'd suggest we just have an "orderBy" setting that takes a string value that can be, e.g., "name" or "most recent". This allows us to expand this option down the road...

packages/esm-forms-workspace-app/src/forms-workspace/form-list/forms-list.component.tsx
```diff
@@ -0,0 +1,182 @@
...
+  visit,
+  formsSection,
+  searchTerm,
+  pageSize = defaultPaginationSize,
+  htmlFormEntryForms,
+}) => {
+  const { t } = useTranslation();
+  const { orderFormsByName } = useConfig() as ConfigObject;
+  const layout = useLayoutType();
+  const [formsInfo, setFormsInfo] = useState<Array<CompletedFormInfo>>(formsSection.completedFromsInfo);
```
ibacher, line 50: Shouldn't the type here be inferrable?

packages/esm-forms-workspace-app/src/forms-workspace/form-list/forms-list.component.tsx
```diff
@@ -0,0 +1,182 @@
...
+                    </TableRow>
+                  </TableHead>
+                  <TableBody>
+                    {rows.map((row, i) => (
+                      <TableRow {...getRowProps({ row })} key={row.id}>
+                        <TableCell key={row.cells[0].id}>
+                          <Link
+                            style={{ cursor: 'pointer' }}
+                            onClick={() => {
+                              handleFormOpen(row.id, '', tableRows[i].formName);
```
ibacher, line 131: Better to use `null` here...

```suggestion
                              handleFormOpen(row.id, null, tableRows[i].formName);
```

packages/esm-forms-workspace-app/src/forms-workspace/form-list/forms-list.scss
```diff
@@ -0,0 +1,72 @@
...
+  & > label {
+    cursor: pointer;
+  }
+
+  &:global(.cds--data-table tr) {
+    height: spacing.$spacing-09;
+  }
+}
+
+#wide-table.bx--data-table-container {
```
ibacher, line 59:
```suggestion
#wide-table.cds--data-table-container {
```

ibacher (commented):
At a high-level, I don't understand why this isn't just included as part of the form-app... It's better if we don't unnecessarily spawn multiple applications with overlapping concerns.

gracepotma:
Update per convo w/ @denniskigen today: We just want to address the issue where the workspace is thinking that opening this list of forms means the user has "a new form open" (it interprets this list as a form), so when the user clicks on one of the options, the user sees a warning "Are you sure you want to leave this unsaved form?". @denniskigen is going to review and update accordingly here.

gracepotma:
Update: Discussed today w/ Dennis & Ian; plan - Dennis to investigate a different approach for implementing the dashboard within the next few business days (akin to how we're doing the Order Basket dashboard).

packages/esm-forms-workspace-app/src/config-schema.ts
```diff
@@ -0,0 +1,81 @@
...
+    _type: Type.Array,
+    _elements: {
+      name: {
+        _type: Type.String,
+        _description: 'Name of the section.',
+      },
+      labelCode: {
+        _type: Type.String,
+        _description: 'Label to be used to translate the section.',
+      },
```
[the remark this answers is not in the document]
icrc-loliveira: This is in order that the section name can handle translations. 
Please let me know if there are any other away to handle translations for labels that come from a config.
```
eg. {t(labelCode, name)}
```

packages/esm-forms-workspace-app/src/config-schema.ts
icrc-loliveira: "forms":{
   "_type": "Type.Array",
   "_description": "Configuration of the forms, it can be used the `Form UUID` or the `Encounter Type UUID` if you have both the `Form UUID` will be used.",
   "_elements": {
      "formUuid": {
         "_type":"Type.String",
         "_description":"Uuid of the form."
      },
      "encounterTypeUuid": {
         "_type":"Type.String",
         "_description":"Uuid of the encounter type."
      }
   }
}

packages/esm-forms-workspace-app/src/config-schema.ts
icrc-loliveira: done

packages/esm-forms-workspace-app/src/forms-workspace/form-list/forms-list.component.tsx
icrc-loliveira: done

denniskigen:
Hey, @icrc-loliveira, I see you're pushing commits to this PR. Might it be worthwhile to consider building on top of the existing clinical forms dashboard? https://github.com/openmrs/openmrs-esm-patient-chart/pull/977

packages/esm-forms-workspace-app/src/config-schema.ts
```diff
@@ -0,0 +1,81 @@
...
+    _type: Type.Boolean,
+    _default: true,
+    _description: 'Whether HTML Form Entry forms should be included in lists of forms.',
+  },
+  showRecommendedFormsTab: {
+    _type: Type.Boolean,
+    _description: 'Whether to display recommended forms tab. Requires AMPATH ETL.',
+    _default: false,
+  },
+  showConfigurableForms: {
```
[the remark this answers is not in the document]
icrc-loliveira: Removed

packages/esm-forms-workspace-app/src/config-schema.ts
icrc-loliveira: I added support for the encounter type as described above.

delphinepas:
@ibacher could you please provide an update on this PR review? thank you cc @icrc-loliveira

delphinepas:
@ibacher can you please let us know what needs to be done to move forward on this ticket? thank you cc @icrc-loliveira @icrc-fdeniger

denniskigen:
Hi @delphinepas, I posted this [comment](https://github.com/openmrs/openmrs-esm-patient-chart/pull/814#issuecomment-1472039672) a while back. A forms dashboard [already exists](https://github.com/openmrs/openmrs-esm-patient-chart/pull/977) in O3. My suggestion was that we find what's missing in the current implementation and Luis's work and introduce those in a separate PR. How does that sound?

packages/esm-patient-forms-app/src/config-schema.ts
```diff
@@ -72,20 +101,21 @@ export const configSchema = {
...
   },
-  orderFormsByName: {
+  useCurrentVisitDates: {
     _type: Type.Boolean,
-    _description: 'Order forms alphabetically.',
-    _default: true,
+    _description: 'Only load encounters inside of the current visits dated.',
+    _default: false,
+  },
+  orderBy: {
```
ibacher, line 114: Would be good to have a validator attached here that ensures the value is one of the two accepted values

packages/esm-patient-forms-app/src/config-schema.ts
```diff
@@ -1,6 +1,35 @@
...
+            _description: 'Uuid of the form.',
+          },
+          encounterTypeUuid: {
+            _type: Type.String,
+            _description: 'Uuid of the encounter type.',
+          },
+        },
+      },
+    },
+    _default: undefined,
```
ibacher, line 31: I don't think the config system works really well with default `undefined`. Better to just use an empty array (which is still falsy in JS).

```suggestion
    _default:[],
```

packages/esm-patient-forms-app/src/config-schema.ts
```diff
@@ -72,20 +101,21 @@ export const configSchema = {
...
     _type: Type.String,
     _description: 'Custom forms endpoint to fetch forms using a custom url need `showConfigurableForms`',
     _default: '',
   },
-  orderFormsByName: {
+  useCurrentVisitDates: {
     _type: Type.Boolean,
-    _description: 'Order forms alphabetically.',
-    _default: true,
+    _description: 'Only load encounters inside of the current visits dated.',
```
ibacher, line 111:
```suggestion
    _description: 'Only load encounters inside of the current visit.',
```

packages/esm-patient-forms-app/src/forms-workspace/form-list/forms-list.component.tsx
```diff
@@ -0,0 +1,210 @@
...
+  htmlFormEntryForms,
+}) => {
+  const { t } = useTranslation();
+  const { orderBy } = useConfig() as ConfigObject;
+  const isTablet = useLayoutType() === 'tablet';
+  const [formsToShow, setFormsToShow] = useState(formsSection.completedFromsInfo);
+  const orderForms = (orderBy: string, forms: Array<CompletedFormInfo>): Array<CompletedFormInfo> => {
+    switch (orderBy) {
+      case OrderBy.Name:
+        return forms.sort((a, b) => (a.form.display < b.form.display ? -1 : 1));
```
ibacher, line 55: Might be good to make this case-insensitive...

packages/esm-patient-forms-app/src/hooks/use-form-encounters.ts
```diff
@@ -0,0 +1,39 @@
+import { getDynamicOfflineDataEntries, openmrsFetch, useConfig, useVisit } from '@openmrs/esm-framework';
+import { ConfigObject } from '../config-schema';
+import { customEncounterRepresentation, formEncounterUrl, formEncounterUrlPoc } from '../constants';
+import useSWR from 'swr';
+import { EncounterWithFormRef, Form, ListResponse } from '../types';
+import dayjs from 'dayjs';
+
+export function useFormEncounters(cachedOfflineFormsOnly = false, patientUuid: string = '') {
+  const { customFormsUrl, showHtmlFormEntryForms } = useConfig() as ConfigObject;
```
ibacher, line 9:
```suggestion
  const { customFormsUrl, showHtmlFormEntryForms } = useConfig<ConfigObject>();
```

ibacher (changes requested):
Some quick initial comments. A few UI things:

1. I don't think we need the Forms header inside the workspace; it's inconsistent with the rest of our designs:

Forms Header

2. On small desktop, each table-row should be the same width:

This PR:
Table of forms with mixed spacing

Current State:

Table of forms with consistent spacing

3. Also note the `Never`s above should still be rendered.

I still need to check through some of the more complex logic here, but there were the things that immediately lept out.

packages/esm-patient-forms-app/src/forms-workspace/form-list/forms-list.component.tsx
```diff
@@ -0,0 +1,210 @@
...
+  patientUuid: string;
+  patient: fhir.Patient;
+  visit: Visit;
+  formsSection: FormsSection;
+  searchTerm: string;
+  pageSize: number;
+  htmlFormEntryForms: Array<HtmlFormEntryForm>;
+}
+
+const FormsList: React.FC<FormsListProps> = ({
```
ibacher, line 39: It would probably be a good idea to refactor things so that the `FormsList` was less directly tied to the workspace component. That way, we can use the same view in other places...

packages/esm-patient-forms-app/translations/fr.json
```diff
@@ -1,6 +1,6 @@
 {
   "all": "Tout",
-  "clinicalForm": "Formulaire clinique",
+  "clinicalForms": "Formulaire clinique",
```
ibacher, line 3: I see a change to `clinicalForms` here and in the Cambodian translations, but not in the English. These two should probably be undone... Otherwise, we need to have the translations re-checked (the French for "Clinical forms" should be something like "Formulaires clinique"; I have no idea what change or even if a change is needed in Khmer)

packages/esm-patient-forms-app/src/config-schema.ts
icrc-loliveira: Hello @ibacher,
Can you be more specific on this one ?

packages/esm-patient-forms-app/src/forms-workspace/form-list/forms-list.component.tsx
icrc-loliveira: What is our suggestion?
The way I see it only has what needs to work.

-   patientUuid: string;
-   patient: fhir.Patient;
-   visit: Visit;
-   formsSection: FormsSection;
-   searchTerm: string;
-   pageSize: number;
-   htmlFormEntryForms: Array<HtmlFormEntryForm>;

I could simplify the `formsSection`, but I would still need the information that is inside it to work.
Maybe add an option to not show the title?

packages/esm-patient-forms-app/src/config-schema.ts
ibacher: Goes with `_type` and `_description` below.

packages/esm-patient-forms-app/src/forms-workspace/form-list/forms-list.component.tsx
icrc-loliveira: After a talk with @ibacher, we decided to ignore this.

packages/esm-forms-workspace-app/src/config-schema.ts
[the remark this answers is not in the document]
brandones: `t(name, name)` is fine

brandones:
This would now require a lot lot of merging, and it seems like it might be redundant with the existing forms dashboard?

Closing as stale.

Obviously feel free to update and re-open if relevant, @icrc-loliveira @ibacher @delphinepas .

PR #816, opened 2022-09-23
(feat) Show All Encounters Table

ZacButko:
## Requirements

## Summary
This adds another tab and table on patient chart -> visits page which is using the 'encounters' endpoint to pull all encounters, vs the already existing visits table (previously called encounters-list) which is getting data from the 'visits' endpoint. This allows encounters not attached to a visit to be found.

## Screenshots
In this video you can see data being entered through fast data entry for patient 100019A. Through the form submission we get the notice that "form has saved". Now if we navigate to the patient's chart we can see indeed that this data has been recorded as an Encounter, even if it is not associated with a visit.
https://user-images.githubusercontent.com/5445264/191969573-1e9017fc-f7ef-4258-866a-2ce09d018d39.mov


vasharma05 (approved):
LGTM 🚀 

Thanks for the work!

ZacButko:
I think a key paradigm of the OpenMRS is that most everything really only makes sense inside of a 'visit' context. Therefore the existing All Visits tab (coupled with Test Results Viewer) should be sufficient. Showing data outside of a visit context is "dangerous" because it will make people think that that is the way we are organizing data, and we should as much as possible continue to align with the visit-frist paradigm. 

However this is the first real feature to fetch and show all encounters, and could be very useful for, if nothing else, debugging. I selfishly want this feature for that reason, and can't really think of a better way to get or display it in such an easy to use way. That, and the organizational changes this PR makes make the PR worth merging. In order to hide this behind a debug mask I think what we should do is make a configuration to toggle the tab on and off. Let's merge this for now and make hiding via configuration a subsequent PR.

PR #821, opened 2022-09-28
Fix error upon saving vitals

nanfuka:
## Requirements

## Summary
1. When a vital entered is not valid, an error message is displayed at the bottom of the entry in red indicating that the value is not valid
2. the save button is disabled in case the user has not entered the weight


vasharma05:
Please add a proper name and description to the PR.
Thanks!

PR #837, opened 2022-10-07
(feat) Configure Show All Encounters

ZacButko:
## Requirements

## Summary
This makes it so that the all encounters tab of patient-chart visits section is turned off by default, and can be set in configuration to be on.

Additionally, it separates out the All Visits tab from the All Encounters tab, so that patients with encounters but not visits will be able to find their data

## Screenshots
Default configuration is off: 

Screen Shot 2022-10-07 at 5 16 46 AM

Configuring on shows the All Encounters tab: 

Screen Shot 2022-10-07 at 5 16 27 AM

Data shown on all encounters
Screen Shot 2022-10-07 at 4 38 34 AM


hadijahkyampeire (approved):
Thanks @ZacButko LGTM

PR #841, opened 2022-10-13
(chore)Husky fix: Replaced prepare with postinstall

vasharma05:
## Requirements

## Summary
According to the statement on the Husky website [here](https://typicode.github.io/husky/#/?id=install), it states that
```
Yarn 2+ doesn't support prepare lifecycle script, so husky needs to be installed differently (this doesn't apply to Yarn 1 though). See [Yarn 2+ install](https://typicode.github.io/husky/#/?id=yarn-2).
```
Hence I have incorporated the changes according to the instructions [here](https://typicode.github.io/husky/#/?id=yarn-2).


hadijahkyampeire (approved):
Nice catch @vasharma05

PR #842, opened 2022-10-13
O3-698:Introduce select feature for adding immunizations to table

jwnasambu:
## Requirements

## Summary
I introduced a select element to enable a user select the appropriate vaccine from the dropdown.

## Screenshots

https://user-images.githubusercontent.com/33891016/199489961-1e2e8836-05ce-4e0f-8d07-83a380ee921c.mp4

## Related Issue

https://issues.openmrs.org/browse/O3-693

### DIFF
--- a/packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
+++ b/packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
@@ -1,4 +1,4 @@
-import React, { SyntheticEvent, useEffect, useState } from 'react';
+import React, { SyntheticEvent, useEffect, useState, useCallback, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import {
   Button,
@@ -51,15 +51,20 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
     formChanged: false,
   };
   const [formState, setFormState] = useState(initialState);
-  const updateSingle = <T extends keyof ImmunizationFormState>(name: T, value: typeof formState[T]) =>
+  const updateSingle = useCallback(<T extends keyof ImmunizationFormState>(name: T, value: typeof formState[T]) => {
     setFormState((state) => ({ ...state, [name]: value }));
-  const onChangeComboBox = (event) => {
-    if (event.selectedItem) {
-      updateSingle('vaccineUuid', event.selectedItem.uuid);
-    } else {
-      updateSingle('vaccineUuid', '');
-    }
-  };
+  }, []);
+
+  const onChangeComboBox = useCallback(
+    (event) => {
+      if (event.selectedItem) {
+        updateSingle('vaccineUuid', event.selectedItem.uuid);
+      } else {
+        updateSingle('vaccineUuid', null);
+      }
+    },
+    [updateSingle],
+  );
 
   const { t } = useTranslation();
   const currentUser = useSession();
@@ -164,9 +169,7 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
               titleText={t('vaccineComboBoxTitle', 'Vaccines')}
               items={vaccineState}
               itemToString={(item) => item?.display ?? ''}
-              onChange={(event) => {
-                onChangeComboBox(event);
-              }}
+              onChange={onChangeComboBox}
             />{' '}
           </section>
         ) : null}
@@ -267,3 +270,6 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
 };
 
 export default ImmunizationsForm;
+function setFormState(arg0: (state: any) => any) {
+  throw new Error('Function not implemented.');
+}
--- a/packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
+++ b/packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
@@ -56,6 +56,8 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
   const onChangeComboBox = (event) => {
     if (event.selectedItem) {
       updateSingle('vaccineUuid', event.selectedItem.uuid);
+    } else {
+      updateSingle('vaccineUuid', '');
     }
   };
 
@@ -149,8 +151,9 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
     updateSingle('currentDose', currentDose);
   };
 
-  if (isLoading || vaccineState === undefined)
+  if (isLoading || vaccineState === undefined) {
     return <span className={styles.vaccineLoader}>{`${t('loadingVaccine', 'Loading Vaccine')} ...`} </span>;
+  }
 
   return (
     <Form className={styles.form} onSubmit={handleFormSubmit}>
@@ -158,7 +161,7 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
         {vaccines?.length ? (
           <section className={styles.immunizationSequenceSelect}>
             <ComboBox
-              titleText="Vaccines"
+              titleText={t('vaccineComboBoxTitle', 'Vaccines')}
               items={vaccineState}
               itemToString={(item) => item?.display ?? ''}
               onChange={(event) => {
@@ -169,7 +172,7 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
         ) : null}
 
         {hasSequences(formState.sequences) && (
-          <section className={styles.immunizationSequenceSelect}>
+          <div className={styles.immunizationSequenceSelect}>
             <Select
               id="sequence"
               name="sequence"
@@ -189,9 +192,9 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
                 );
               })}
             </Select>
-          </section>
+          </div>
         )}
-        <section className={styles.immunizationSequenceSelect}>
+        <div className={styles.immunizationSequenceSelect}>
           <DatePicker
             id="vaccinationDate"
             className="vaccinationDate"
@@ -208,8 +211,8 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
               type="text"
             />
           </DatePicker>
-        </section>
-        <section className={styles.immunizationSequenceSelect}>
+        </div>
+        <div className={styles.immunizationSequenceSelect}>
           <DatePicker
             id="vaccinationExpiration"
             className="vaccinationExpiration"
@@ -226,8 +229,8 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
               type="text"
             />
           </DatePicker>
-        </section>
-        <section className={styles.immunizationSequenceSelect}>
+        </div>
+        <div className={styles.immunizationSequenceSelect}>
           <TextInput
             type="text"
             id="lotNumber"
@@ -235,8 +238,8 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
             value={formState.lotNumber}
             onChange={(evt) => updateSingle('lotNumber', evt.target.value)}
           />
-        </section>
-        <section className={styles.immunizationSequenceSelect}>
+        </div>
+        <div className={styles.immunizationSequenceSelect}>
           <TextInput
             type="text"
             id="manufacturer"
@@ -244,7 +247,7 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
             value={formState.manufacturer}
             onChange={(evt) => updateSingle('manufacturer', evt.target.value)}
           />
-        </section>
+        </div>
       </Stack>
       <ButtonSet className={isTablet ? styles.tablet : styles.desktop}>
         <Button className={styles.button} kind="secondary" onClick={() => closeWorkspace()}>
--- a/packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
+++ b/packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
@@ -1,4 +1,4 @@
-import React, { SyntheticEvent, useEffect, useState, useCallback, useMemo } from 'react';
+import React, { SyntheticEvent, useEffect, useState, useCallback } from 'react';
 import { useTranslation } from 'react-i18next';
 import {
   Button,
--- a/packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
+++ b/packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
@@ -6,6 +6,7 @@ import {
   DatePicker,
   DatePickerInput,
   Form,
+  InlineLoading,
   Select,
   SelectItem,
   TextInput,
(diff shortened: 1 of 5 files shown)


packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
```diff
@@ -119,14 +132,40 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
...
   return (
-    <Form className={styles.form} onSubmit={handleFormSubmit} data-testid="immunization-form">
-      <div>
-        <h4 className={styles.immunizationSequenceSelect}>
-          {`${t('vaccine', 'Vaccine')} : ${formState?.vaccineName}`}{' '}
-        </h4>
+    <Form className={styles.form} onSubmit={handleFormSubmit}>
+      <Stack gap={4}>
+        <section className={styles.immunizationSequenceSelect}>
+          <span>{t('vaccine', 'vaccine')}</span>
```
ibacher, line 158:
```suggestion
          <span>{t('vaccine', 'Vaccine')}</span>
```

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
```diff
@@ -119,14 +132,40 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
...
+  const vaccines = mockPaginatedImmunizations;
+  const VaccineSelect = () => (
+    <Select
+      id="vaccine"
+      invalidText="Required"
+      labelText=""
+      onChange={(event) => updateSingle('vaccineName', event.target.value)}
+      value={formState.vaccineName}
+    >
+      {!formState.vaccineName ? <SelectItem text={t('chooseVaccine', 'Choose a vaccine')} value="" /> : null}
```
ibacher, line 144: IIRC, there's a disabled property that should be set for a default but invalid drop-down option.

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
```diff
@@ -119,14 +132,40 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
     updateSingle('currentDose', currentDose);
   };
 
+  const vaccines = mockPaginatedImmunizations;
+  const VaccineSelect = () => (
+    <Select
+      id="vaccine"
+      invalidText="Required"
```
ibacher, line 139: Presumably, this should be translatable

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
```diff
@@ -119,14 +132,40 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
...
+      labelText=""
```
ibacher, line 140: I think it would be preferable to put the "Vaccine" string here rather than having it as part of a separate `span` tag. The label text is semantically connected to the field it labels in a way that the span is not, which makes it better for most use-cases.

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
```diff
@@ -1,13 +1,26 @@
...
+  Layer,
+} from '@carbon/react';
 import { showNotification, showToast, useSession, useVisit, useLayoutType } from '@openmrs/esm-framework';
 import { DefaultWorkspaceProps } from '@openmrs/esm-patient-common-lib';
 import { savePatientImmunization } from './immunizations.resource';
 import { mapToFHIRImmunizationResource } from './immunization-mapper';
 import { ImmunizationFormData, ImmunizationSequence } from './immunization-domain';
 import { immunizationFormSub } from './immunization-utils';
 import styles from './immunizations-form.scss';
+import { mockPaginatedImmunizations } from '../../../../__mocks__/immunizations.mock';
```
ibacher, line 23: I'm not sure this mock mocks anything that is reasonably provided by the backend. We should update this mock to look like, e.g., the result of listing a convenience set, since presumably the list of valid vaccinations comes from a convenience set of concepts.

ibacher (changes requested):
Thanks @jwnasambu! A few things that should be adjusted, but overall looks good.

packages/esm-patient-immunizations-app/src/immunizations/immunizations.resource.tsx
```diff
@@ -43,13 +43,13 @@ export async function getImmunizationsConceptSet(
   return result;
 }
 
-export function useImmunizationsConceptSet() {
+export function useImmunizationsConceptSet(customRepresentation = 'full') {
```
ibacher, line 46: Maybe instead of taking this as an argument, you can just hard-code the custom representation here? After all, this hook is only used in the code you're using (nice thing about MFs is that they are small, so you can pretty much modify the parts without worrying that someone else is using it).

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
```diff
@@ -119,14 +149,27 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
...
+              itemToString={(item) => item?.display ?? ''}
+              onChange={(event) => {
+                onChangeComboBox(event);
+              }}
+            />{' '}
+          </section>
+        ) : null}
+
         {hasSequences(formState.sequences) && (
-          <div className={styles.immunizationSequenceSelect}>
```
ibacher, line 129: `div` is probably better than `section` for most of these. A `section` is supposed to be part of a document, e.g.,

<section>
  <h2>This is a section</h2>
  <p>And here is some section content</p>
</section>

`div` are just for anything to group a bunch of stuff that need to be grouped like you're doing here.

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
```diff
@@ -119,14 +149,27 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
     updateSingle('currentDose', currentDose);
   };
 
+  if (isLoading || vaccineState === undefined)
```
ibacher, line 152: Always have a `{` block after an `if`.

```suggestion
  if (isLoading || vaccineState === undefined) {
```

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
```diff
@@ -119,14 +149,27 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
...
-      <div>
-        <h4 className={styles.immunizationSequenceSelect}>
-          {`${t('vaccine', 'Vaccine')} : ${formState?.vaccineName}`}{' '}
-        </h4>
+    <Form className={styles.form} onSubmit={handleFormSubmit}>
+      <Stack gap={4}>
+        {vaccines?.length ? (
+          <section className={styles.immunizationSequenceSelect}>
+            <ComboBox
+              titleText="Vaccines"
```
ibacher, line 161: Any user-visible text should be localisable, e.g.,

```suggestion
              titleText={t('vaccineComboBoxTitle', 'Vaccines')}
```

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
```diff
@@ -183,36 +226,37 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
...
         </Button>
         <Button
           className={styles.button}
           kind="primary"
-          disabled={isViewEditMode ? !enableEditButtons : !enableCreateButtons}
+          disabled={isViewEditMode ? !enableEditButton : !enableCreateButton}
+          type="submit"
         >
-          {t('save', 'Save')}
+          {t('saveAndClose', 'Save and close')}
```
ibacher, line 259: If you call this "Save and close" then the `handleFormSubmit` function should actually close the workspace, i.e., it needs to call `closeWorkspace()` at the appropriate point.

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
```diff
@@ -42,15 +53,35 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
   const [formState, setFormState] = useState(initialState);
   const updateSingle = <T extends keyof ImmunizationFormState>(name: T, value: typeof formState[T]) =>
     setFormState((state) => ({ ...state, [name]: value }));
+  const onChangeComboBox = (event) => {
```
ibacher, line 56: A couple of things here:

1. Wrap this function in [`useCallback`](https://reactjs.org/docs/hooks-reference.html#usecallback), as it's being passed to a stateful component.
2. When using this function, it can just be written: `onChange={onChangeComboBox}`
3. I notice in most places, we just directly call `updateSingle()` in the event hander. Why do we need this special handling here? (If there's no selected item, surely `vaccineUuid` should be set to, e.g., `null`)

ibacher (changes requested):
Thanks @jwnasambu! Lot's of good stuff here. A few more things I think need to be addressed before this is mergeable.

jwnasambu:
@ibacher Kindly feel free to review my PR again. I have captured the proposed changes. Thanks

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
```diff
@@ -40,17 +51,44 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
     formChanged: false,
   };
   const [formState, setFormState] = useState(initialState);
-  const updateSingle = <T extends keyof ImmunizationFormState>(name: T, value: typeof formState[T]) =>
+  const updateSingle = useCallback(<T extends keyof ImmunizationFormState>(name: T, value: typeof formState[T]) => {
```
vasharma05, line 54: We don't require to give the type for `useCallbacks`. We can directly give the type to the expected arguments.

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
```diff
@@ -40,17 +51,44 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
...
     setFormState((state) => ({ ...state, [name]: value }));
+  }, []);
+
+  const onChangeComboBox = useCallback(
+    (event) => {
+      if (event.selectedItem) {
+        updateSingle('vaccineUuid', event.selectedItem.uuid);
+      } else {
+        updateSingle('vaccineUuid', null);
+      }
```
vasharma05, line 64:
```suggestion
      updateSingle('vaccineUuid', event?.selectedItem?.uuid);
```

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
```diff
@@ -40,17 +51,44 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
...
+    },
+    [updateSingle],
+  );
 
   const { t } = useTranslation();
   const currentUser = useSession();
   const { currentVisit } = useVisit(patientUuid);
   const isTablet = useLayoutType() === 'tablet';
+  const { data, isLoading } = useImmunizationsConceptSet('custom:(display,answers:(uuid,display))');
```
vasharma05, line 73: Since the string is a constant, we don't need to pass it as an argument

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
```diff
@@ -40,17 +51,44 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
...
+  const vaccines = data?.answers;
+  const [vaccineState, setVaccineState] = useState(vaccines || []);
```
vasharma05, line 75:
```suggestion
  const [vaccineState, setVaccineState] = useState(data?.answers ?? []);
```

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
```diff
@@ -40,17 +51,44 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
...
-  const enableEditButtons = isViewEditMode && formState.formChanged;
+  const enableEditButton = isViewEditMode && formState.formChanged;
+  const enableCreateButton =
+    !isViewEditMode &&
+    !!formState.vaccinationDate &&
+    !!formState.vaccineUuid &&
+    !!formState.expirationDate &&
+    !!formState.vaccinationDate &&
+    !!formState.lotNumber &&
+    !!formState.manufacturer;
```
vasharma05, line 91: Why are we using this?

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
```diff
@@ -40,17 +51,44 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
...
   const { currentVisit } = useVisit(patientUuid);
   const isTablet = useLayoutType() === 'tablet';
+  const { data, isLoading } = useImmunizationsConceptSet('custom:(display,answers:(uuid,display))');
+  const vaccines = data?.answers;
+  const [vaccineState, setVaccineState] = useState(vaccines || []);
+  useEffect(() => {
+    if (!isLoading && vaccines?.length) {
+      setVaccineState(vaccines);
+    }
+  }, [vaccines, isLoading]);
```
vasharma05, line 80: Why are you creating state for a value coming from a hook. This is not the right way.

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
```diff
@@ -119,12 +156,24 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
     updateSingle('currentDose', currentDose);
   };
 
+  if (isLoading || vaccineState === undefined) {
+    return <span className={styles.vaccineLoader}>{`${t('loadingVaccine', 'Loading Vaccine')} ...`} </span>;
```
vasharma05, line 160: Use the [Skeleton](https://react.carbondesignsystem.com/?path=/story/components-skeleton-skeletontext--default) or [Loader](https://react.carbondesignsystem.com/?path=/story/components-loading--default) from carbon react

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.component.tsx
```diff
@@ -202,21 +251,25 @@ const ImmunizationsForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, close
...
         </Button>
       </ButtonSet>
     </Form>
   );
 };
 
 export default ImmunizationsForm;
+function setFormState(arg0: (state: any) => any) {
+  throw new Error('Function not implemented.');
+}
```
vasharma05, line 275: Where are we using this function?

packages/esm-patient-immunizations-app/src/immunizations/immunizations.resource.tsx
vasharma05: I agree with @ibacher, we don't need to pass the custom string as an argument, until and unless different payload is required at 2 or more places.,

vasharma05 (changes requested):

denniskigen:
Closing for now until we revisit work on immunizations. Sorry, you've done some great work here, and I hope this ends up getting merged soon @jwnasambu.

PR #680, opened 2022-04-26
O3-852:Ability to Mark Patient Deceased

jnsereko:
## Requirements

- [x] This PR has a title that briefly describes the work done, including the ticket number if there is a ticket.
- [x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide).
- [X] My work includes tests or is validated by existing tests.

## Summary

Ability to Mark Patient Deceased

## Screenshots

Screenshot 2022-04-26 at 16 01 14

## Related Issue

https://issues.openmrs.org/browse/O3-852

PR #844, opened 2022-10-13
03-852: Ability to Mark Patient Deceased v2

hadijahkyampeire:
## Requirements

## Summary
Adds functionality to mark patients as deceased and also mark them as alive in case you made a mistake

## Screenshots

https://user-images.githubusercontent.com/30952856/195926956-61f29839-82c1-4090-a54e-84a0af06bd55.mov

## Related Issue
[Jira Issue](https://issues.openmrs.org/browse/O3-852)

## Other
[Built from this PR](https://github.com/openmrs/openmrs-esm-patient-chart/pull/680)
Built from Version one, credit is to be shared with @jnsereko for the success of this work


gracepotma:
@hadijahkyampeire do you have screenshots or a screenrecording I could look at? Hard for me to approve as a product manager by looking at the code 😅

vasharma05:
Hi @hadijahkyampeire, great work, just that I find the notifications to be different than it should be saying.

image

image

packages/esm-patient-chart-app/src/deceased/deceased-form.scss
```diff
@@ -0,0 +1,115 @@
+@import "~@openmrs/esm-styleguide/src/vars";
+@import "~carbon-components/src/globals/scss/vars";
+@import "~carbon-components/src/globals/scss/mixins";
```
vasharma05, line 3: We no longer use `import` for carbon components. According to the new release by carbon, it was recommended by carbon to use `use` instead of `import`.

You can refer this file: https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-patient-search-app/src/compact-patient-search/compact-patient-banner.scss

packages/esm-patient-chart-app/src/deceased/deceased-form.scss
```diff
@@ -0,0 +1,115 @@
...
+}
+
+.heading {
+  @include carbon--type-style("productive-heading-03");
+  color: $text-01;
+  margin: $spacing-05;
+}
+
+.sectionTitle {
+  @include carbon--type-style("productive-heading-02");
```
vasharma05, line 20: `carbon--type-style` was changed with `type.type-style('productive-heading-02');`

packages/esm-patient-chart-app/src/deceased/deceased.resource.ts
```diff
@@ -0,0 +1,137 @@
...
+};
+
+export function markPatientDeceased(
+  deceasedDate: Date | null,
+  personUuid: string,
+  selectedCauseOfDeathValue: string | null,
+  abortController: AbortController,
+) {
+  const payload = {
+    deathDate: toDateObjectStrict(
```
vasharma05, line 77: We are using 2 Date objects, omrsISOString and dayjs, which is not efficient by the looks. What do you trying to do here?

packages/esm-patient-chart-app/src/deceased/deceased.resource.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+
+export function useConceptAnswers(conceptUuid: string) {
+  const { data, error, isValidating } = useSWR<{ data: { answers: ConcentAnswersResponse } }, Error>(
+    `/ws/rest/v1/concept/${conceptUuid}`,
+    openmrsFetch,
+  );
+  const result = useMemo(() => {
+    return {
+      conceptAnswers: data ? data?.data?.answers : null,
```
vasharma05, line 97: data?.data?.answers alone will do the work. It'll return undefined which is nearly same as null.

packages/esm-patient-chart-app/src/deceased/deceased.resource.ts
```diff
@@ -0,0 +1,137 @@
...
+      isConceptLoading: !data && !error,
+      conceptError: error,
+      isConceptAnswerValidating: isValidating,
+    };
+  }, [data, error, isValidating]);
+  return result;
+}
+
+export function usePersonfromPatient(patientUuid: string) {
```
vasharma05, line 106: Also, the personUuid and the patientUuid are always same, incase you are using the person for only the UUID

packages/esm-patient-chart-app/src/deceased/mark-alive-modal.component.tsx
```diff
@@ -0,0 +1,58 @@
...
+    e.preventDefault();
+    closeDialog();
+    markPatientAlive(patientUuid, new AbortController())
+      .then((response) => {
+        if (response.ok) {
+          closeDialog();
+          showToast({
+            critical: true,
+            kind: 'success',
+            title: t('markAsAlive', 'Mark As Alive'),
```
vasharma05, line 28: This is different from what I saw in the video. Seems like the same key is given different values.

Also, it should be "Marked as Alive"

packages/esm-patient-chart-app/src/index.ts
```diff
@@ -83,6 +83,14 @@ function setupOpenMRS() {
           moduleName,
         }),
       },
+      {
+        name: 'mark-alive-button',
+        slot: 'patient-actions-slot',
+        load: getAsyncLifecycle(() => import('./actions-buttons/mark-patient-alive.component'), {
+          featureName: 'patient-actions-slot',
```
vasharma05, line 90: featureName is actually related to the component, not the slotName. It's good this way too.

packages/esm-patient-chart-app/translations/en.json
```diff
@@ -45,11 +49,16 @@
   "loadVisitInfo": "Load Visit Info",
   "location": "Location",
   "male": "Male",
+  "markAlive": "Mark Alive",
+  "markAsAlive": "Are you sure, you want to mark patient as alive?",
```
vasharma05, line 53: This same key was having value `Mark as Alive`. Please make sure every key has different values.

When extract-translations runs, it gives warnings when it finds same key having different values.

vasharma05 (changes requested):

dkayiwa:
> do we need an option to mark a patient alive back, seems weird to me

Don't you believe in resurrection? 😊

packages/esm-patient-chart-app/src/deceased/mark-alive-modal.component.tsx
hadijahkyampeire: I updated that after recording and I didn't do another recording but it is fixed, I was using the same keys.

hadijahkyampeire:
> > do we need an option to mark a patient alive back, seems weird to me
> 
> Don't you believe in resurrection? 😊

hehehe, @dkayiwa you are making me laugh, so @vasharma05 , it is in the ticket requirements just in case someone made a mistake, and also for developer testing purposes.

hadijahkyampeire:
Thanks for the review @vasharma05 I will push the fixes ASAP.

packages/esm-patient-chart-app/src/deceased/base-concept-answer.component.tsx
```diff
@@ -0,0 +1,83 @@
...
+    if (!isEmpty(searchTerm)) {
+      return conceptAnswers.filter(
+        (conceptAnswer) => conceptAnswer.display.toLowerCase().search(searchTerm.toLowerCase()) !== -1,
+      );
+    } else {
+      return conceptAnswers;
+    }
+  }, [searchTerm, conceptAnswers]);
+
+  const handleSearch = React.useMemo(() => debounce((searchTerm) => setSearchTerm(searchTerm), 300), []);
```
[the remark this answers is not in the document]
hadijahkyampeire: This is not really expecting serSearchTerm as a dependency and it won't be changing.

packages/esm-patient-chart-app/src/deceased/deceased.resource.ts
ibacher: Is there some reason we're loading the `patient` endpoint if all we care about is the `person`? There's also a `person` endpoint....

packages/esm-patient-chart-app/src/deceased/base-concept-answer.component.tsx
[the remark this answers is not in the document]
ibacher: Sure; however, I think including dependencies in the dependency array even when it is, strictly speaking, not necessary is a good practice. The reason the lint rule is written that way is not _just_ that it's easier to write. It's that getting into the habit of adding everything to the React dependency array is a good practice unless there's an explicit reason not to.

The concern is this: If someone comes along and sees this code and doesn't understand _why_ those things don't need to be in the dependency array they can easily copy-and-paste this snippet, tweak it, and wind up with a broken component. So, really, the burden should be to satisfy the lint rule and if you're not satisfying the lint rule, explicitly document why this case is exceptional.

packages/esm-patient-chart-app/src/deceased/deceased.resource.ts
hadijahkyampeire: So it's true this looks off but it looks like the `toDateObjectStrict` requires a date in that format I have looked in many places that is the way we have been sending the date. check this https://github.com/openmrs/openmrs-esm-patient-chart/blob/39a9dd30fc9ced3d06df75e8f0f8e0c9786853c4/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx#L94

packages/esm-patient-chart-app/src/deceased/deceased.resource.ts
hadijahkyampeire: which endpoint is that I have tried to look around but I only see patient endpoints not specifically person.

hadijahkyampeire: Oh, I have seen it. Thanks.

packages/esm-patient-banner-app/src/banner-tags/deceased-patient-tag.component.tsx
```diff
@@ -2,26 +2,34 @@ import React from 'react';
...
-        <div role="tooltip" className={styles.tooltipPadding}>
-          <h6 style={{ marginBottom: '0.5rem' }}>{t('deceased', 'Deceased')}</h6>
-          <span>{formatDatetime(parseDate(patient.deceasedDateTime))}</span>
-        </div>
-      }
-    >
-      <Tag type="red">{t('deceased', 'Deceased')}</Tag>
-    </DefinitionTooltip>
-  ) : null;
+  const { person } = useOmrRestPatient(patientUuid);
```
vasharma05, line 14:
```suggestion
  const { person } = useOmrsRestPatient(patientUuid);
```

packages/esm-patient-banner-app/src/hooks/usePatientAttributes.tsx
```diff
@@ -41,3 +45,21 @@ export const usePatientContactAttributes = (patientUuid: string) => {
     isLoading,
   };
 };
+
+export function useOmrRestPatient(patientUuid: string) {
```
vasharma05, line 49:
```suggestion
export function useOmrsRestPatient(patientUuid: string) {
```

vasharma05 (approved):
Great work!

Just a small suggestion.
Thanks!

PR #846, opened 2022-10-18
Added units in the test results viewer and different value types to correctly display coded and text test results 

vasharma05:
## Requirements

## Summary
I've added units in the test results viewer and different value types to correctly display coded and text test results. 

## Screenshots
image

## Related Issue

https://issues.openmrs.org/browse/O3-1562

### DIFF
--- a/packages/esm-patient-test-results-app/src/panel-view/panel.component.tsx
+++ b/packages/esm-patient-test-results-app/src/panel-view/panel.component.tsx
@@ -76,7 +76,7 @@ const LabSetPanel: React.FC<LabSetPanelProps> = ({ panel, observations, activePa
               content: <span>{`${test.value} ${test.meta?.units}`}</span>,
             },
             interpretation: test.interpretation,
-            range: `${test.meta?.range} ${test.meta?.units}` ?? '--',
+            range: test.meta.range ? `${test.meta?.range} ${test.meta?.units}` : '--',
           }))
         : observations.map((test) => ({
             id: test.id,
--- a/packages/esm-patient-test-results-app/src/panel-view/panel.component.tsx
+++ b/packages/esm-patient-test-results-app/src/panel-view/panel.component.tsx
@@ -73,24 +73,16 @@ const LabSetPanel: React.FC<LabSetPanelProps> = ({ panel, observations, activePa
             id: test.id,
             testName: test.name,
             value: {
-              content: (
-                // <span className={getClass(test.obs?.[0]?.interpretation)}>
-                <span>{test.value}</span>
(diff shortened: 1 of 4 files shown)


packages/esm-patient-test-results-app/src/panel-view/panel.component.tsx
```diff
@@ -73,24 +73,16 @@ const LabSetPanel: React.FC<LabSetPanelProps> = ({ panel, observations, activePa
...
-              content: (
-                // <span className={getClass(test.obs?.[0]?.interpretation)}>
-                <span>{test.value}</span>
-                // </span>
-              ),
+              content: <span>{`${test.value} ${test.meta?.units}`}</span>,
             },
             interpretation: test.interpretation,
-            range: test.meta?.range ?? '--',
+            range: `${test.meta?.range} ${test.meta?.units}` ?? '--',
```
hadijahkyampeire, line 79: Isn't it weird to condition on strings, even if it is undefined it will be 'undefined' which looks like a value, you should probably condition on the prop and then display a value else you display the `emDash`

packages/esm-patient-test-results-app/src/panel-view/types.ts
```diff
@@ -90,15 +100,15 @@ export interface ConceptMeta {
   lowAbsolute: number;
   lowCritical: number;
   units: string;
-  getInterpretation: (value: number) => OBSERVATION_INTERPRETATION;
+  getInterpretation: (value: string) => OBSERVATION_INTERPRETATION;
```
hadijahkyampeire, line 103: Are you sure this is not a breaking change elsewhere? maybe you should expect number or string so that it doesn't break when it comes as a number.

hadijahkyampeire (commented):
Nice work @vasharma05 just minor suggestions

packages/esm-patient-test-results-app/src/panel-view/types.ts
vasharma05: No, I have verified it thoroughly and this is not breaking. If you see the `getObservationValue`, that is returning a string altogether, and hence I changed the type of the value with string. 

If you scroll down a few lines, I have a `isNaN` check in place so that it doesn't break.
Thanks!

packages/esm-patient-test-results-app/src/panel-view/panel.component.tsx
vasharma05: Thanks for the catch!
I'll fix this asap.

hadijahkyampeire (approved):
Good work!

PR #861, opened 2022-11-09
O3-1589:Refactor code on the visit form

jwnasambu:
## Requirements

## Summary
I created a little mini-component for the location, datePicker  and  timePicker instead of having a ternary with two branches with the same text in each then wrapped it in a layer.

## Issue worked on:
https://issues.openmrs.org/browse/O3-1589


hadijahkyampeire (approved):
Thanks @jwnasambu for this refactor

PR #863, opened 2022-11-10
O3-1552: Drug Orders multiply dosage dangerously and other small fixes

vasharma05:
## Requirements

## Summary
This PR addresses and solves the following existing issues in the medications and order basket.

1. Medications should show Drug orders, i.e. orders having Order Type = Drug. Since a patient named "Joshua Johnson" was also having Test Orders (orderType = testorder) the medications widget was throwing 400 error, since the custom representation passed into `/ws/rest/v1/order` was not matching with that of test orders. Hence I pass the drug order type to the order API.
image

3. Drug orders were multiplying dangerously.
Say for a drug Aspirin having strength 81mg, was prescribed to a patient with dose of 81mg (which means 1 tablet), then the medications row was showing the dosage of 81*81mg. Now, we just show the dose prescribed and the unit of dose, instead of multiplying the dose with drug strength.
image

4. The order item in the order basket was not properly showing the drug strength and was showing the dosage ordered.
#### Before
image

#### Now
image

5. The empty state of past medications had the "Record medication", which is now shifted to the active medication's Empty state.
#### Before
image

#### Now
image

## Screenshots
Attached above

## Related Issue
https://issues.openmrs.org/browse/O3-1552

### DIFF
--- a/packages/esm-patient-medications-app/src/components/medications-details-table.component.tsx
+++ b/packages/esm-patient-medications-app/src/components/medications-details-table.component.tsx
@@ -84,13 +84,13 @@ const MedicationsDetailsTable = connect<
           <div className={styles.medicationRecord}>
             <div>
               <p className={styles.bodyLong01}>
-                <strong>{capitalize(medication.drug?.name)}</strong> &mdash; {medication.drug?.strength.toLowerCase()}{' '}
-                &mdash; {medication.doseUnits?.display.toLowerCase()}
+                <strong>{capitalize(medication.drug?.concept?.display)}</strong> &mdash;{' '}
+                {medication.drug?.strength.toLowerCase()} &mdash; {medication.drug.dosageForm.display.toLowerCase()}
               </p>
               <p className={styles.bodyLong01}>
                 <span className={styles.label01}>{t('dose', 'Dose').toUpperCase()}</span>{' '}
                 <span className={styles.dosage}>
-                  {getDosage(medication.drug?.strength, medication.dose).toLowerCase()}
+                  {medication.dose} {medication.doseUnits.display.toLowerCase()}
                 </span>{' '}
                 &mdash; {medication.route?.display.toLowerCase()} &mdash; {medication.frequency?.display.toLowerCase()}{' '}
                 &mdash;{' '}
--- a/packages/esm-patient-medications-app/src/api/api.ts
+++ b/packages/esm-patient-medications-app/src/api/api.ts
@@ -1,6 +1,7 @@
 import useSWR from 'swr';
-import { ConfigObject, openmrsFetch, Session } from '@openmrs/esm-framework';
+import { openmrsFetch, Session, useConfig } from '@openmrs/esm-framework';
 import { OrderPost, PatientMedicationFetchResponse } from '../types/order';
+import { ConfigObject } from '../config-schema';
 
 /**
  * Fast, lighweight, reusable data fetcher with built-in cache invalidation that
@@ -9,6 +10,8 @@ import { OrderPost, PatientMedicationFetchResponse } from '../types/order';
  * @param status The status/the kind of orders to be fetched.
  */
 export function usePatientOrders(patientUuid: string, status: 'ACTIVE' | 'any', careSettingUuid: string) {
+  const config = useConfig() as ConfigObject;
+  const { drugOrderTypeUUID } = config;
   const customRepresentation =
     'custom:(uuid,dosingType,orderNumber,accessionNumber,' +
     'patient:ref,action,careSetting:ref,previousOrder:ref,dateActivated,scheduledDate,dateStopped,autoExpireDate,' +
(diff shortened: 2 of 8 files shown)


denniskigen (approved):
CI builds are broken, I suspect due to your changes to the lockfile, so have a look at those. Looks generally good to me, thanks @vasharma05. Any feedback from the design review?

samuelmale:
@vasharma05 Nice work. One thing that got me confused is I'm not sure I see the logic related to fix "Drug orders were multiplying dangerously." in this PR?

packages/esm-patient-medications-app/src/components/medications-details-table.component.tsx
```diff
@@ -84,13 +84,13 @@ const MedicationsDetailsTable = connect<
...
-                <strong>{capitalize(medication.drug?.name)}</strong> &mdash; {medication.drug?.strength.toLowerCase()}{' '}
-                &mdash; {medication.doseUnits?.display.toLowerCase()}
+                <strong>{capitalize(medication.drug?.concept?.display)}</strong> &mdash;{' '}
+                {medication.drug?.strength.toLowerCase()} &mdash; {medication.drug.dosageForm.display.toLowerCase()}
               </p>
               <p className={styles.bodyLong01}>
                 <span className={styles.label01}>{t('dose', 'Dose').toUpperCase()}</span>{' '}
                 <span className={styles.dosage}>
-                  {getDosage(medication.drug?.strength, medication.dose).toLowerCase()}
+                  {medication.dose} {medication.doseUnits.display.toLowerCase()}
```
gracepotma, line 93: @samuelmale this is where the logic is replaced - the problematic logic was introduced by the getDosage function, which is the one causing the multiplier. Do you think we need an additional/deeper fix as well?

dkayiwa:
> One thing that got me confused is I'm not sure I see the logic related to fix "Drug orders were multiplying dangerously." in this PR?

That is exactly the problem with pull requests that do more than one thing. They are harder to review. For a bug as terrible as this one, i would not mix its fix with anything else at all. 😊 

Just in case: https://opensource.com/article/18/6/anatomy-perfect-pull-request

dkayiwa:
Does this also take care of the rest of the places mentioned here? https://issues.openmrs.org/browse/O3-1552?focusedCommentId=285539&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-285539

packages/esm-patient-medications-app/src/components/medications-details-table.component.tsx
samuelmale: Oohhh I see thanks Grace. (It wouldn't be sarcastic if someone requested you to review a PR with chunks of commits  think you should start writing some code)

@vasharma05 so what's gonna happen to malicious `getDosage(...)` function? Is left as dead code? Is it being consumed in a different place?

vasharma05:
> That is exactly the problem with pull requests that do more than one thing. They are harder to review. For a bug as terrible as this one, i would not mix its fix with anything else, at all. 

Hi @dkayiwa !
I always try to mention everything that is going to be changed in my PR. In this PR as well, as you can see, the major changes was regarding the removal of `getDosage` function itself, other changes pt.3, 4, 5 are changes related to that itself.

Regarding pt. 1, this was something that I just came because this issue of Drug order was totally new to me

dkayiwa:
@vasharma05 i like the way you have simplified https://github.com/openmrs/openmrs-esm-patient-chart/pull/868. It is so easy to review even by a non frontend expert like me. 😊

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/medications-summary.component.tsx
```diff
@@ -29,7 +29,7 @@ const MedicationSummary: React.FC<MedicationSummaryProps> = ({ medications }) =>
                       <span className={styles.label01}> {t('dose', 'Dose').toUpperCase()} </span>{' '}
                     </span>
                     <span className={styles.dosage}>
-                      {getDosage(medication.order.drug?.strength, medication.order?.dose)?.toLowerCase()}
+                      {medication.order.dose} {medication.order.doseUnits.display.toLowerCase()}
```
FlorianRappl, line 32: Are these things guaranteed to be there? Otherwise, if these depend on user settings I'd recommend using the optional chaining operator (`?.`), e.g., `medication.order?.dose`. The removed line was by the way inconsistent, as `medication.order.drug?.strength` assumed `order` is always given while `medication.order?.dose` did not. So either the first one would crash or the second was unnecessary.

packages/esm-patient-medications-app/src/components/medications-details-table.component.tsx
vasharma05: Removed the dead import. Thanks @samuelmale !

vasharma05:
Thanks @dkayiwa !

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/medications-summary.component.tsx
vasharma05: Thanks @FlorianRappl for the catch, I'll just fix it.

PR #873, opened 2022-11-15
(fix) Explicitly process change detection and its side-effects on form action

icrc-psousa:
## Requirements

## Summary
When triggering an action on code running outside Angular, the changes weren't properly triggering an UI update to reflect ocurred changes. 

This is experienced, for example, when running [openmrs-esm-fast-data-entry-app](https://github.com/openmrs/openmrs-esm-fast-data-entry-app) that makes use of esm-form-entry-app and triggers it's onSubmit action. After triggering the esm-form-entry-app onSubmit action, the form is validated and messages related to the validations should appear but, as the action is triggered by React on openmrs-esm-fast-data-entry-app, they won't be shown unless some event triggers a re-render on Angular.

This PR fixes the issue described above by explicitly processing change detection and its side-effects when a form action is triggered.

## Screenshots
image


icrc-psousa:
@ibacher It seems I'm not able to assign reviewers to the PR. Would it be possible for you to assign some reviewers in order to get feedback and push this forward.

denniskigen:
Nothing comes to mind immediately @icrc-psousa. I'll revert tomorrow after some research.

FlorianRappl:
It all depends *how* Angular knows that there is a change. For instance, usually this is / should be done by zone.js, which patches all kinds of browser integrated functionality to notify Angular when / what changed.

Another way is to use the `ChangeDetectionRef`, which allows you to actually notify Angular on your terms. This can also be partially used implicitly (there is a strategy that can be set on component level, see, e.g., https://blog.angular-university.io/onpush-change-detection-how-it-works/) or explicitly (inject `ChangeDetectionRef` into the constructor, use it later to `detectChanges`).

Not sure if this helps you or if I only tell you things that you know already @icrc-psousa .

icrc-psousa:
Thanks for your input @FlorianRappl! Actually, this is all new information for me. I'll have a look.
Also, as briefly discussed with @denniskigen, maybe we can also have a look at esm-form-entry-app on openmrs-esm-patient-chart as it also renders the form on the left pane but the above problems are not present there. Perhaps there's something on openmrs-esm-patient-chart to cope with this that I've missed.

icrc-psousa:
It seems the 1st proposed fix, using `this.applicationRef.tick()` to manually trigger change detection outside Angular zone, was breaking Angular change handlers meaning that, an external action would indeed update the view because `this.applicationRef.tick()` was called but, after that, no other changes would be automatically detected unless Angular was explicitly notified.

Alternatively, returning the code execution that would update the view to Angular zone by using `ngZone.run()` seems to fix the issue.
https://github.com/openmrs/openmrs-esm-patient-chart/blob/a61fa7bae8310d6b28ef667048c744e4c3f9109a/packages/esm-form-entry-app/src/app/fe-wrapper/fe-wrapper.component.ts#L265

Now, triggering an action on code running outside Angular, properly triggers an UI update to reflect occurred changes and Angular change handlers keep working as intended.

@denniskigen @FlorianRappl I've updated the PR, can you review it?

FlorianRappl (approved):
LGTM - works!

icrc-psousa:
Hi @denniskigen, @gracepotma,
Would it be possible to validate this and push it forward? This issue is preventing [openmrs-esm-fast-data-entry-app](https://github.com/openmrs/openmrs-esm-fast-data-entry-app) from working properly. Thanks!

ibacher:
Thanks @icrc-psousa and @FlorianRappl!

PR #877, opened 2022-11-16
03-1084: No warning message if you leave patient chart with open/unsaved form

hadijahkyampeire:
## Requirements

## Summary
- Added a warning dialog before closing the visit header when any form is open.

## Caveat
- Still need to handle the aspect of whether or not a form has been changed.

## Screen recording

https://user-images.githubusercontent.com/30952856/202297133-f4dd6cbf-3d73-4c38-8e34-8a0bf18175b0.mov

## Screenshots showing Visit Notes and Order Basket cases
Screenshot 2022-11-17 at 22 50 13
Screenshot 2022-11-17 at 22 30 41

## Related Issue
[No warning message if you leave patient chart with open/unsaved form](https://issues.openmrs.org/projects/O3/issues/O3-1084)
[Order basket case](https://issues.openmrs.org/browse/O3-1618)


packages/esm-patient-chart-app/translations/en.json
```diff
@@ -71,6 +72,11 @@
   "noObservationsFound": "No observations found",
   "notes": "Notes",
   "openAnyway": "Open anyway",
+  "openFormSpecificWarningText": "The {workspaceTitle} form will not be saved if you exit the patient chart now.",
```
dkayiwa, line 75: Will the form be saved if i exit the patient chart `later` instead of `now`? 😊

packages/esm-patient-chart-app/src/visit-header/confirmation-modal.component.tsx
```diff
@@ -0,0 +1,58 @@
...
+  leavePatientChart: (e: any) => void;
+  props: any;
+}
+
+const ConfirmLeavingPatientChart: React.FC<ConfirmationDialogProps> = ({ closeDialog, leavePatientChart, props }) => {
+  const { t } = useTranslation();
+  const { type = '', title = '', additionalProps = {} } = props;
+  const workspaceTitle = additionalProps && additionalProps?.workspaceTitle;
+
+  const unifiedTitle = title.includes('Form') ? title : `${title} Form`;
```
ibacher, line 17: This needs to be localizable if we're using it. It's not clear to me why we need this though?

packages/esm-patient-chart-app/src/visit-header/confirmation-modal.component.tsx
```diff
@@ -0,0 +1,58 @@
...
+                'openFormSpecificWarningText',
+                'The {workspaceTitle} form will not be saved if you exit the patient chart now.',
+                { workspaceTitle },
+              )
+            : t('openFormWarningText', 'The form will not be saved if you exit the patient chart now.')
+          : t('openWorkspaceWarningText', 'The workspace will not be saved if you exit the patient chart now.')}
+      </ModalBody>
+      <ModalFooter>
+        <Button kind="secondary" onClick={closeDialog}>
+          {t('no', 'No')}
```
ibacher, line 42: This should probably be something like "Return to patient chart" or at least "Cancel". "No" doesn't clearly communicate what action is taken by clicking this button.

packages/esm-patient-chart-app/translations/en.json
hadijahkyampeire: If later comes after the form is saved, I can change that but the [zeplin mocks](https://app.zeplin.io/project/60d59321e8100b0324762e05/screen/61bb8ad37c91686a70eec2d2) have the wording that way. You might suggest a better word @dkayiwa

packages/esm-patient-chart-app/src/visit-header/confirmation-modal.component.tsx
```diff
@@ -0,0 +1,58 @@
...
+        </Button>
+        <Button
+          kind="danger"
+          onClick={(e) => {
+            closeDialog();
+            leavePatientChart(e);
+          }}
+        >
+          {t('delete', 'Delete')}
```
ibacher, line 51: This should probably be either "Close patient chart" or "Continue". Again "Delete" is very ambiguous...

packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
```diff
@@ -21,8 +21,10 @@ import {
   useVisit,
   navigate,
   useConfig,
+  showModal,
+  // getGlobalStore,
```
ibacher, line 25: Please don't leave commented-out code handing around. Either uncomment it or remove it...

packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
```diff
@@ -113,6 +115,17 @@ const VisitHeader: React.FC = () => {
   const navMenuItems = useAssignedExtensions('patient-chart-dashboard-slot').map((extension) => extension.id);
   const { startVisitLabel } = useConfig();
 
+  // I thought this would tell when a form is updated
+  // const globalFormStore = getGlobalStore('ampath-form-state', {}).getState();
```
ibacher, line 119: So, if we need to be informed by the form engine whether there are changes, that change needs to be part of the form engine or the wrapper around it.

packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
```diff
@@ -113,6 +115,17 @@ const VisitHeader: React.FC = () => {
...
+
+  // const hasFormChanges = Object.values(globalFormStore).includes('ready');
+  // console.log(hasFormChanges, 'gloable-----', globalFormStore)
+
+  const store = getWorkspaceStore();
+  const state = store.getState();
+  const hasOpenForm = state.openWorkspaces?.[0]?.type === 'form';
+  const workSpaceProps = state.openWorkspaces?.[0];
```
ibacher, line 127: It's not clear to me why things are ordered this way? Also, isn't it possible to have multiple "open" workspaces? Shouldn't this code work with that situation?

ibacher: I.e., a more appropriate implementation would be something like:

```
const formWorkspace = state.openWorkspaces?.find((workspace) => workspace.type === "form");
```

packages/esm-patient-chart-app/src/visit-header/confirmation-modal.component.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+
+import { Button, ModalBody, ModalFooter, ModalHeader } from '@carbon/react';
+import { useTranslation } from 'react-i18next';
+
+interface ConfirmationDialogProps {
+  closeDialog: () => void;
+  leavePatientChart: (e: any) => void;
+  props: any;
```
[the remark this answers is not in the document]
ibacher: Realistically, however, you can probably even borrow the type definition from the place where they're originally defined.

packages/esm-patient-chart-app/src/visit-header/confirmation-modal.component.tsx
```diff
@@ -0,0 +1,58 @@
...
+        closeModal={closeDialog}
+        title={
+          type === 'form'
+            ? t('openFormWarningHeader', "You haven't saved the {formTitle}", { formTitle: unifiedTitle })
+            : t('openWorkspaceWarningHeader', "You haven't saved this workspace")
+        }
+      />
+      <ModalBody>
+        {type === 'form' && workspaceTitle
+          ? workspaceTitle
```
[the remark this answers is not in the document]
hadijahkyampeire: Yah, I forgot to remove it, I will do that.

packages/esm-patient-chart-app/translations/en.json
dkayiwa: I would also tell the user what they need to do, to have the form saved. 😊

packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
hadijahkyampeire: Actually, this will not only apply to the Ampath forms but also all the other forms like `startVisit Form`, `markDeceased`, `Allergies`, `vitals`, and all the other forms we have. That's my most challenge right now.

hadijahkyampeire:
Thanks, @ibacher for the review, just a note, I am going to also add conditions for the order basket, I just realized it is also a requirement for this ticket.

packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
hadijahkyampeire: I was trying to see if I can detect form updates from the global store but that didn't seem to work as expected, I will remove this implementation now.

packages/esm-patient-chart-app/src/visit-header/confirmation-modal.component.tsx
hadijahkyampeire: Just for uniformity because some form titles don't have the word form and the look off and inconsistent.

packages/esm-patient-chart-app/translations/en.json
hadijahkyampeire: That sounds much better, thanks, Daniel.

gracepotma:
So happy to see this coming together ❤️ This has been long needed.
Quick question that I can't tell from the screencap: Does this also work for the Visit Note form (the one where you click the pencil in the right siderail)?

hadijahkyampeire:
> So happy to see this coming together ❤️ This has been long needed. Quick question that I can't tell from the screencap: Does this also work for the Visit Note form (the one where you click the pencil in the right siderail)?

Yes @gracepotma it works for the Visit Note and Order Basket as well but I didn't capture those in the video, I will add screenshots for them.

packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
ibacher: It should be the responsibility of the component mounted into the workspace to tell you that... This code shouldn't be trying to deduce that at all.

hadijahkyampeire: That's true, I have talked to some members of the team about including the caveat in a follow-up PR since it might need more discussions and critical thinking, so I will create a ticket for that work to capture updates from all the possible components that we mount in our workspaces. How does that sound @ibacher

packages/esm-patient-chart-app/src/visit-header/confirmation-modal.component.tsx
```diff
@@ -0,0 +1,69 @@
...
+    additionalProps?: {
+      workspaceTitle?: string;
+    };
+  };
+}
+
+const ConfirmLeavingPatientChart: React.FC<ConfirmationDialogProps> = ({ closeDialog, leavePatientChart, props }) => {
+  const { t } = useTranslation();
+  const { type = '', title = '', additionalProps = {} } = props;
+  const workspaceTitle = additionalProps && additionalProps?.workspaceTitle;
```
ibacher, line 21:
```suggestion
  const workspaceTitle = additionalProps?.workspaceTitle;
```

packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
ibacher: There was some discussion of this in the past. I think no one just got around to actually implementing it, so ticketing it sounds fine to me.

gracepotma:
> > So happy to see this coming together ❤️ This has been long needed. Quick question that I can't tell from the screencap: Does this also work for the Visit Note form (the one where you click the pencil in the right siderail)?
> 
> Yes @gracepotma it works for the Visit Note and Order Basket as well but I didn't capture those in the video, I will add screenshots for them.

Oh it's okay @hadijahkyampeire no need to add another screenshot, I trust you; if you say it's done, it's done :)

hadijahkyampeire:
@ibacher I have fixed all the issues mentioned in the review, please take another look when you get some time.

ibacher (commented):
Thanks @hadijahkyampeire! I do wonder if there's a simpler solution here, though. Looking through the workspace code, it looks like we already have a `promptBeforeClosing` as part of the `closeWorkspace` function. Maybe instead of a new dialog here we can just leverage that as part of the visit header? This, ultimately, seems too imperative, i.e., for every new thing that's added to the workspace, we need to add yet another thing to the `hasOpenForm` and since implementation can create as many workspaces as they like, this isn't going to be very scalable.

hadijahkyampeire:
I will try that and see if it works with the visit header, thanks @ibacher

hadijahkyampeire:
@ibacher I have managed to use the `promptBeforeClosing` function and it works pretty well, the only issue I am seeing is that my titles and bodyText translations want to disappear when I extract translations, why would they behave like that?

hadijahkyampeire:
@ibacher I think this PR is ready for approval, please take another look when you get some time.

packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
```diff
@@ -113,6 +123,15 @@ const VisitHeader: React.FC = () => {
...
   const { startVisitLabel } = useConfig();
 
+  const store = getWorkspaceStore();
+  const state = store.getState();
+  const workSpaceProps: workspaceProps = state.openWorkspaces?.[0] || {};
+
+  const formWorkspace = state.openWorkspaces?.find((workspace) => workspace.type === 'form');
+  const orderWorkspace = state.openWorkspaces?.find((workspace) => workspace.type === 'order');
+  const visitNoteWorkspace = state.openWorkspaces?.find((workspace) => workspace.type === 'visit-note');
+  const hasOpenForm = formWorkspace || orderWorkspace || visitNoteWorkspace;
```
ibacher, line 133: So, my suggestion was really about generifying things here. Keep in mind that the workspace if pluggable. We do not really know before hand the list of possible workspaces. My suggestion was that we should be able to iterate through the list of workspaces and just call the `closeWorkspace()` on each.

hadijahkyampeire: I am sorry @ibacher , I do not fully understand the above, do you mean call the `promptBeforeClosing()` or `closeWorskpace()`

PR #880, opened 2022-11-17
O3-1617: location_id is not saved in encounter while saving a form

icrc-agomes:
## Requirements

## Summary
In this development we are implementing, when saving a form that doesn't have a field to select a location, the use of the location of the current session.

## Related Issue
[O3-1617](https://issues.openmrs.org/browse/O3-1617)


packages/esm-form-entry-app/src/app/form-submission/form-submission.service.ts
```diff
@@ -30,6 +31,7 @@ interface FormSubmissionResult {
 @Injectable()
 export class FormSubmissionService {
   constructor(
+    private readonly openmrsApi: OpenmrsEsmApiService,
```
denniskigen, line 34: How are we using this service?

denniskigen (commented):
Thank you, @icrc-agomes. This is very close! Just a couple of concerns to address.

PR #882, opened 2022-11-18
Add patient clinical identifiers on patient banner

gabriel090:
## Requirements

## Summary

Add NUPI and CCC number, and Patient Clinical number as the primary identifier on Patient banner and chart banner
identifier1
identifier2

### DIFF
--- a/packages/esm-patient-chart-app/src/banner/patient-banner.component.tsx
+++ b/packages/esm-patient-chart-app/src/banner/patient-banner.component.tsx
@@ -2,10 +2,11 @@ import React, { MouseEvent } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Button, Tag } from '@carbon/react';
 import { ChevronDown, ChevronUp, OverflowMenuVertical } from '@carbon/react/icons';
-import { ExtensionSlot, age, formatDate, parseDate } from '@openmrs/esm-framework';
+import { ExtensionSlot, age, formatDate, parseDate, useConfig } from '@openmrs/esm-framework';
 import ContactDetails from './contact-details/contact-details.component';
 import CustomOverflowMenuComponent from './ui-components/overflow-menu.component';
 import styles from './patient-banner.scss';
+import { ConfigObject } from '../config-schema';
 
 interface PatientBannerProps {
   patient: fhir.Patient;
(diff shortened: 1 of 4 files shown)


vasharma05:
Please correct me if I'm wrong, but do we show the labels for every identifier?
I remember the label was only for the primary identifier.

CC: @denniskigen

vasharma05:
Please resolve the conflicts.
Thanks!

packages/esm-patient-chart-app/src/banner/patient-banner.component.tsx
```diff
@@ -118,8 +122,8 @@ const PatientBanner: React.FC<PatientBannerProps> = ({
           </div>
           <div className={styles.row}>
             <div className={styles.identifiers}>
-              {patient?.identifier?.length
-                ? patient?.identifier.map(({ value, type }) => (
+              {filterPatientIdentifier?.length
+                ? filterPatientIdentifier.map(({ value, type }) => (
```
FlorianRappl, line 126: I'd change this to `filterPatientIdentifier?.map()` - no need for ternary operator as `null` or `undefined` will not render. If the length is zero you get an empty array, which also will not render. So you wrote less code that is actually also better for React (right now we still have one text node in case nothing should be rendered).

packages/esm-patient-chart-app/src/config-schema.ts
```diff
@@ -42,6 +42,11 @@ export const esmPatientChartSchema = {
     _description: 'Custom label to use on the start visit actions',
     _default: '',
   },
+  defaultPatientIdentifier: {
```
FlorianRappl, line 45: General question: What is the reason to duplicate all this and have it in patient-chart and patient-banner? Should we maybe have only patient-banner? There might be some reason behind it, but I fail to see / remember it.

cc @denniskigen

denniskigen: Me neither @FlorianRappl, I was very surprised when I first saw this duplication. I intend to clean it up soon.

denniskigen:
@gabriel090, Ciaran is working on a design mockup for how multiple identifiers should be displayed on the banner. Could you mind holding tight until we get that? Shouldn't take long.

denniskigen:
So here's Ciaran's design feedback https://www.loom.com/share/f8d094da2cae4aacaccee36c5e6d461c. Let me know what you think about it @gabriel090 and all.

PR #887, opened 2022-11-24
O3-1641: Ability to display configurable visit attributes as tags on patient header

vasharma05:
## Requirements

## Summary
This PR supports showing visit attributes, which are recorded when starting a visit, on the patient banner as tags.
The user can configure whether to show a visit attribute on the patient banner or not using the below mentioned configuration:

```
"@openmrs/esm-patient-chart-app": {
  "visitAttributeTypes": Array<{
    "uuid": (string) UUID of the visit attribute type whose field is to be shown in the start visit form,
    "required": (boolean) Whether filling the visit attribute type field is required or not,
    "displayInThePatientBanner": (boolean) Whether the visit attribute's value should be shown in the patient banner or not
  }>
}
```

## Screenshots
image
image

## Related Issue
https://issues.openmrs.org/browse/O3-1641

## Other
Dependent on https://github.com/openmrs/openmrs-esm-core/pull/575

### DIFF
--- a/packages/esm-patient-chart-app/src/patient-banner-tags/visit-attribute-tags.component.tsx
+++ b/packages/esm-patient-chart-app/src/patient-banner-tags/visit-attribute-tags.component.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Tag } from '@carbon/react';
-import { formatDate } from '@openmrs/esm-framework';
+import { formatDate, useConfig } from '@openmrs/esm-framework';
+import { ChartConfig } from '../config-schema';
 import { useVisitOrOfflineVisit } from '@openmrs/esm-patient-common-lib';
 
 interface VisitAttributeTagsProps {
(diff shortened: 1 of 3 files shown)


packages/esm-patient-chart-app/src/patient-banner-tags/visit-attribute-tags.component.tsx
```diff
@@ -0,0 +1,53 @@
...
+    case 'org.openmrs.customdatatype.datatype.ConceptDatatype':
+      return value?.display;
+    case 'org.openmrs.customdatatype.datatype.FloatDatatype':
+      return value;
+    case 'org.openmrs.customdatatype.datatype.FreeTextDatatype':
+      return value;
+    case 'org.openmrs.customdatatype.datatype.LongFreeTextDatatype':
+      return value;
+    case 'org.openmrs.customdatatype.datatype.BooleanDatatype':
+      return value;
```
denniskigen, line 22: It looks as though all these `case` clauses return the same value as the `default` clause. Is this intentional? Should there maybe be an empty state in the `default` clause (something like `--`)? If so, that bit of code could be shortened to:

```suggestion
    case 'org.openmrs.customdatatype.datatype.FloatDatatype':
    case 'org.openmrs.customdatatype.datatype.FreeTextDatatype':
    case 'org.openmrs.customdatatype.datatype.LongFreeTextDatatype':
    case 'org.openmrs.customdatatype.datatype.BooleanDatatype':
      return value;
```

denniskigen (approved):
LGTM, thank you @vasharma05!

PR #888, opened 2022-11-28
Add Khmer translations

rbuisson:
## Requirements

- [X] This PR has a title that briefly describes the work done, including the ticket number if there is a ticket.

## Summary
Adding Khmer (km) translations from Transifex


reagan-meant (approved):

rbuisson:
Thanks for the approval @reagan-meant .
Though I still can't merge, as it appears one GA is failing.

@ibacher do you know what might cause this?

ibacher:
I see this:

> The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled.

Which kind of suggests an internal GitHub thing. In any case, we shouldn't explicitly be requiring the bundle size check to pass... it's informational rather than something strictly enforced.

packages/esm-form-entry-app/translations/km.json
```diff
@@ -0,0 +1 @@
+{}
```
ibacher, line 1:
```suggestion
{}

```

ibacher (approved):

PR #889, opened 2022-11-29
03-1652: Order basket not reporting backend errors properly.

hadijahkyampeire:
## Requirements

## Summary
- Caught the error message returned by the backend and show it in a notification

## Screenshots
Screenshot 2022-11-29 at 02 20 18

## Related Issue
- https://issues.openmrs.org/browse/O3-1652


packages/esm-patient-medications-app/src/order-basket/order-basket.component.tsx
```diff
@@ -46,6 +53,29 @@ const OrderBasket = connect<OrderBasketProps, OrderBasketStoreActions, OrderBask
...
+          setIsLoading(false);
+        }
+      },
+      (error) => {
+        const apiError = error.responseBody.error.message;
+        const noProviderError = "[could not execute statement => Column 'provider_id' cannot be null]";
+        setIsLoading(false);
+        showNotification({
+          title: t('noProviderError', 'Error creating an encounter'),
+          kind: 'error',
```
hadijahkyampeire, line 71: I was wondering whether this would look better as a warning just.

hadijahkyampeire: It would look like this.
Screenshot 2022-11-30 at 02 02 58

ibacher (changes requested):
A couple of small points in the comments, but I think there are some bigger issues that need to be discussed.

The problem you're originally reporting arises because [in this block](https://github.com/openmrs/openmrs-esm-patient-chart/blob/05a390c549974a98f20aba408b8fa0b33810e752/packages/esm-patient-medications-app/src/api/api.ts#L91), it's plausible for the frontend to send a bit of JSON like this:

```json
"encounterProviders": [{
	"provider": null,
    "encounterRole": <some uuid>
}]
```

hadijahkyampeire:
I will then fix the JSON and see if it works even without a provider

packages/esm-patient-medications-app/src/order-basket/order-basket.component.tsx
vasharma05: No, it should be an error.

hadijahkyampeire:
Closing this after refining requirements for the order basket fix, I will push another PR for the appropriate fix.

PR #895, opened 2022-12-08
(feat)Replaced dose's combo input with a number input

vasharma05:
## Requirements

## Summary
Replacing the Dose's combo input with number input in the order basket form

## Screenshots
None

## Related Issue
None


packages/esm-patient-medications-app/src/order-basket/medication-order-form.component.tsx
```diff
@@ -221,55 +210,19 @@ export default function MedicationOrderForm({ initialOrderBasketItem, onSign, on
...
-                    selectedItem={
-                      dosageOptions?.length
-                        ? {
-                            id: orderBasketItem.dosage ? `${orderBasketItem.dosage?.value}` : null,
-                            text: orderBasketItem.dosage ? `${orderBasketItem.dosage?.value}` : null,
-                          }
-                        : null
-                    }
-                    // @ts-ignore
                     placeholder={t('editDoseComboBoxPlaceholder', 'Dose')}
```
denniskigen, line 216: Reckon we could omit the placeholder and just show a [default value](https://carbondesignsystem.com/components/number-input/usage/#default-values) instead (which we're already doing with the value prop). Additionally, I've noticed something subtle about how the NumberInput gets rendered:

Screenshot 2022-12-14 at 15 25 11

It doesn't quite fill the available vertical space. The culprit here is its `display: flex` property. You can apply this style override in the `gridRow` class to fix that:

```scss
.gridRow {
  @extend .row;
  padding: 0px;

  :global(.cds--number) {
    display: inline-block;
  }
}
```

packages/esm-patient-medications-app/src/order-basket/medication-order-form.component.tsx
```diff
@@ -221,55 +210,19 @@ export default function MedicationOrderForm({ initialOrderBasketItem, onSign, on
...
-                      }
-                      if (value && !valueExists && draftIndex == -1 && !isNaN(Number(value))) {
-                        dosageOptions.push({ id: 'draft', text: value });
-                      }
+                    label={t('editDoseComboBoxTitle', 'Enter Dose')}
+                    value={orderBasketItem?.dosage ?? 0}
+                    onChange={(e, { value }) => {
+                      setOrderBasketItem({
+                        ...orderBasketItem,
+                        dosage: value ? parseFloat(value) : 0,
```
denniskigen, line 222: Why not `parseInt`?

packages/esm-patient-medications-app/src/types/order-basket-item.ts
```diff
@@ -1,10 +1,4 @@
-import {
-  DosingUnit,
-  MedicationDosage,
-  MedicationFrequency,
-  MedicationRoute,
-  OrderTemplate,
-} from '../api/drug-order-template';
+import { DosingUnit, MedicationFrequency, MedicationRoute, OrderTemplate } from '../api/drug-order-template';
```
denniskigen, line 1: We could get rid of the `MedicationDosage` type if it's no longer in use.

denniskigen (commented):
LGTM, bar a few issues I've flagged.

denniskigen:
Also, the proper semantic commit prefix for this should be (feat), not (chore).

packages/esm-patient-medications-app/src/order-basket/medication-order-form.component.tsx
ibacher: Technically, the dosage could be a float, unfortunately.

denniskigen: Ok

packages/esm-patient-medications-app/src/types/order-basket-item.ts
vasharma05: Sure!

hadijahkyampeire (changes requested):
Thanks, @vasharma05 this is almost done, just a tinny visual bug that needs to be fixed. The `dose` input and `unit` are joined, you probably wanna give a margin-right to the does input.
Screenshot 2022-12-22 at 13 24 47

Minor: Also I don't know if you had to do all these changes in this ticket since it is just changing the input. We need to be sure that changing the dose interface/type does not break other places and we won't need it as an object sometime in the future.

vasharma05:
Let me have a look at the breaking UI.

For the changes, as I discussed yesterday, all these changes are related to the dose input itself. I am more specific to the correct data that is stored and the way it is stored, and hence the changes.

vasharma05:
Detailed description:

The `DrugOrderTemplate` is the type of data that is fetched from the drug's template. The Drug template keeps it as an array of object containing the dosage count in the `value` key. From these drug templates, we fetch the default value of the dosage count, as is done with frequency count and dosage form. These values is then stored in the `OrderBasketItem` interface, for which the type of OrderBasketItem's dosage value is changed from an `object` (not an array of objects) to the type `number` and all the places where the changes are required is also taken care of.

packages/esm-patient-medications-app/src/order-basket/medication-order-form.component.tsx
```diff
@@ -232,55 +220,19 @@ export default function MedicationOrderForm({ initialOrderBasketItem, onSign, on
...
-                      const draftIndex = dosageOptions.findIndex((opt) => opt.id == 'draft');
-                      // validate and clean up
-                      if (draftIndex >= 0 && value) {
-                        dosageOptions[draftIndex].text = value;
-                      } else if (draftIndex >= 0) {
-                        dosageOptions.pop();
-                      }
-                      if (value && !valueExists && draftIndex == -1 && !isNaN(Number(value))) {
-                        dosageOptions.push({ id: 'draft', text: value });
-                      }
```
vasharma05, line 281: @hadijahkyampeire the whole above approach of getting the dosage input and then saving the value in the object is replaced with a single line of code as is mentioned below.
```
dosage: value ? parseFloat(value) : 0
```

Also, there's no point of storing it an object of type 
```
dosage: {
    "value": <value>,
}
```

This not only minimise the bytes but also keeps the developer more into sync to the real value type (number), so that any unintended mistake can be avoided in the future.

hadijahkyampeire: Great to know if it doesn't affect other places, thanks.

vasharma05:
Thank you @hadijahkyampeire

PR #897, opened 2022-12-09
O3-1594: Closing patient chart always returns to home page (part)

vasharma05:
P.S. This ticket was assigned to @SharleenAwinja and she has done the work. I did help with distributing the work into different repositories, and hence this ticket is assigned to @SharleenAwinja and hence she is the rightful owner.

## Requirements

## Summary
This PR adds the event listener to record the origin page for the patient-chart, so whenever we exit from the patient chart, we should be directed back to the origin page from where we jumped into the patient chart.

## Screenshots
None

## Related Issue
https://issues.openmrs.org/browse/O3-1594

## Other
Dependent on https://github.com/openmrs/openmrs-esm-core/pull/579

### DIFF
--- a/packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
@@ -126,13 +126,12 @@ const VisitHeader: React.FC = () => {
 
   const noActiveVisit = !isLoading && visitNotLoaded;
 
-  const originPage = localStorage.getItem('fromPage');
-
   const onClosePatientChart = useCallback(() => {
-    originPage ? navigate({ to: `${window.spaBase}/${originPage}` }) : navigate({ to: `${window.spaBase}/home` });
+    const originPage = window.localStorage.getItem('patient-chart-origin-page');
+    originPage ? navigate({ to: `${window.spaBase}${originPage}` }) : navigate({ to: `${window.spaBase}/home` });
     setShowVisitHeader((prevState) => !prevState);
     localStorage.removeItem('fromPage');
-  }, [originPage]);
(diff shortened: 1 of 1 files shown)


packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
```diff
@@ -126,13 +126,12 @@ const VisitHeader: React.FC = () => {
 
   const noActiveVisit = !isLoading && visitNotLoaded;
 
-  const originPage = localStorage.getItem('fromPage');
-
   const onClosePatientChart = useCallback(() => {
-    originPage ? navigate({ to: `${window.spaBase}/${originPage}` }) : navigate({ to: `${window.spaBase}/home` });
+    const originPage = window.localStorage.getItem('patient-chart-origin-page');
```
ibacher, line 130: This may not be the point of this PR, but caching this data in local storage is not the right thing to do. Session storage would be slightly more appropriate, but really, I'd just store it in memory (i.e., as a variable or piece of state.

vasharma05: Thanks for pointing the scenario if the user opens multiple tabs.
I'll try to test it then.
Thanks!

vasharma05:
Hi @SharleenAwinja ! Can you please look into the suggestions?
Thanks!

hadijahkyampeire (approved):
This should work for now but as Ian suggested if it is easy to save it in session storage, that would be much reliable.

hadijahkyampeire:
@vasharma05 is this still needed or we should close it.

denniskigen:
Closing as stale.

PR #903, opened 2022-12-12
feat:(form-app) mostRecentObsValueBefore data source in form-entry (O3-1661)

hadijahkyampeire:
## Requirements

## Summary
- Add services to fetch all concepts and the observation values for all concepts.
- Filter through concepts to return values before a given date.

## Screenshots

- Showing the datasource being loaded when a form is opened
Screenshot 2022-12-15 at 01 49 35

## Related Issue
https://issues.openmrs.org/browse/O3-1661


packages/esm-form-entry-app/src/app/form-creation/form-creation.service.ts
```diff
@@ -90,6 +90,8 @@ export class FormCreationService {
   private wireDataSources(createFormParams: CreateFormParams) {
     const visitTypeUuid = this.singleSpaPropsService.getPropOrThrow('visitTypeUuid');
     const patient = this.singleSpaPropsService.getPropOrThrow('patient');
+    const patientUuid = this.singleSpaPropsService.getPropOrThrow('patientUuid');
```
ibacher, line 93: This is redundant... You should be able to get the patient UUID from the patient object already (it's the id field).

packages/esm-form-entry-app/src/app/form-creation/form-creation.service.ts
```diff
@@ -90,6 +90,8 @@ export class FormCreationService {
...
+    const date = new Date().toISOString();
```
ibacher, line 94: We can just leave out the date stuff for now

packages/esm-form-entry-app/src/app/form-creation/form-creation.service.ts
```diff
@@ -111,6 +113,10 @@ export class FormCreationService {
     this.dataSources.registerDataSource('patient', this.formDataSourceService.getPatientObject(patient), true);
     this.dataSources.registerDataSource('rawPrevEnc', createFormParams.previousEncounter, false);
     this.dataSources.registerDataSource('userLocation', createFormParams.user.sessionLocation);
+    this.dataSources.registerDataSource(
```
ibacher, line 116: So, the data source, in this case, should be an object containing this function, not the result of the function call

packages/esm-form-entry-app/src/app/form-creation/form-creation.service.ts
```diff
@@ -111,6 +113,10 @@ export class FormCreationService {
...
+      'mostRecentObsValue',
+      this.formDataSourceService.getMostRecentObsValueBefore(date, patientUuid),
```
ibacher, line 118: Also, note that the patient UUID should already be available inside the form... The second parameter here was supposed to be a list of concepts.

packages/esm-form-entry-app/src/app/form-data-source/form-data-source.service.ts
```diff
@@ -114,6 +117,56 @@ export class FormDataSourceService {
     );
   }
 
+  public getMostRecentObsValueBefore(date: string, patientUuid: string) {
+    type ObsResult = FHIRResource['resource'] & {
+      conceptUuid: string;
+      dataType?: string;
+    };
+
+    function isUuid(input: string) {
```
ibacher, line 126: It's a bad idea to declare a function inside a function, especially as in this case where there's to dependency on data in the containing function. The reason is that functions declared inside a function are re-created on every function execution, i.e., we're asking the interpreter to do unnecessary work.

packages/esm-form-entry-app/src/app/form-data-source/form-data-source.service.ts
```diff
@@ -114,6 +117,56 @@ export class FormDataSourceService {
...
+    const concepts = this.conceptResourceService.getAllConcepts();
+    concepts.subscribe((concepts) => {
+      const uuids = concepts.map((c) => c.uuid).join(',');
+
+      return this.observationResourceService.getMostRecentObsValues(date, uuids, patientUuid).subscribe((data) => {
+        const observations =
+          data?.entry?.map((entry) => {
+            const observation: ObsResult = {
+              ...entry.resource,
+              conceptUuid: entry.resource.code.coding.filter((c) => isUuid(c.code))[0]?.code,
```
ibacher, line 143: It would probably be more robust to just check for the first code element that doesn't have a `system` property, if this needs to be here... I'm a little confused as to why we need this at all, though...

packages/esm-form-entry-app/src/app/form-data-source/form-data-source.service.ts
```diff
@@ -114,6 +117,56 @@ export class FormDataSourceService {
     );
   }
 
+  public getMostRecentObsValueBefore(date: string, patientUuid: string) {
+    type ObsResult = FHIRResource['resource'] & {
```
ibacher, line 121: While this isn't per-se harmful, type declarations should generally not be declared in functions. There's already a `types.ts` file that we're importing from... Why not declare this type there?

packages/esm-form-entry-app/src/app/form-data-source/form-data-source.service.ts
```diff
@@ -114,6 +117,56 @@ export class FormDataSourceService {
...
+      return this.observationResourceService.getMostRecentObsValues(date, uuids, patientUuid).subscribe((data) => {
+        const observations =
+          data?.entry?.map((entry) => {
+            const observation: ObsResult = {
+              ...entry.resource,
+              conceptUuid: entry.resource.code.coding.filter((c) => isUuid(c.code))[0]?.code,
+            };
+
+            if (entry.resource.hasOwnProperty('valueString')) {
+              observation.dataType = 'Text';
```
ibacher, line 147: I'm not exactly sure why this is necessary?

packages/esm-form-entry-app/src/app/openmrs-api/obs-resource.service.ts
```diff
@@ -0,0 +1,25 @@
...
+import { map } from 'rxjs/operators';
+import { WindowRef } from '../window-ref';
+import { fhirBaseUrl } from '@openmrs/esm-framework';
+
+@Injectable()
+export class ObservationResourceService {
+  constructor(protected http: HttpClient, protected windoRef: WindowRef) {}
+
+  public getMostRecentObsValues(date, conceptUuids, patientUuid) {
+    const url = `/openmrs${fhirBaseUrl}/Observation/$lastn`;
```
ibacher, line 13: Don't hard-code `/openmrs` We actually support this being changed. Just use the `makeUrl` function exported from esm-framework, i.e., 

```
makeUrl(`${fhirBaseUrl}/Observation/$lastn`);
```

ibacher: Actually, though, is there a reason we can't use `openmrsFetch` here? I'd always prefer using that unless it's impossible to do so... (`openmrsFetch` handles a number of concerns...).

packages/esm-form-entry-app/src/app/form-data-source/form-data-source.service.ts
```diff
@@ -3,9 +3,11 @@ import { Injectable } from '@angular/core';
 import { take, map, tap } from 'rxjs/operators';
 import { Observable, Subject } from 'rxjs';
 
+import { FHIRResource } from '@openmrs/esm-framework';
```
ibacher, line 6: Always use `import type` if we're just importing types from a library... It allows the declaration to be erased.

```suggestion
import type { FHIRResource } from '@openmrs/esm-framework';
```

packages/esm-form-entry-app/src/app/form-data-source/form-data-source.service.ts
```diff
@@ -114,6 +117,56 @@ export class FormDataSourceService {
...
+            if (entry.resource.hasOwnProperty('valueCodeableConcept')) {
+              observation.dataType = 'Coded';
+            }
+
+            return observation;
+          }) ?? [];
+
+        const mostRecentObsBefore = observations.filter(
+          (observation) => new Date(observation.effectiveDateTime).getTime() > new Date(date).getTime(),
+        );
```
ibacher, line 163: I think doing this filtering here is a mistake. We need to do it on the backend. Since the backend will only return you a single result per concept and patient (that's what the `max=1` specifies) filtering here just means that including a data will make data unavailable.

ibacher: Also, you could simplify this to `new Date(observation.effectiveDateTime) > new Date(date)`...

packages/esm-form-entry-app/src/app/form-data-source/form-data-source.service.ts
```diff
@@ -114,6 +117,56 @@ export class FormDataSourceService {
     );
   }
 
+  public getMostRecentObsValueBefore(date: string, patientUuid: string) {
```
ibacher, line 120: See earlier comments on the parameters here (and check the model signature from the ticket, since this doesn't implement the requirements. Also, why would we pass the date as a string here? Just take a JS date object...

ibacher (changes requested):
Thanks Hadijah! Some good stuff in here. I think you're right that we need the patient UUID as a parameter, so the function signature should be something like:

```ts
getMostRecentObsValues(patient: string, concepts: string[])
```

But I'd probably do something like this:

```ts
getMostRecentObsValues(patient: string, date: Date, concepts: string[]);
```

Or:

```ts
getMostRecentObsValues(patient: string, { date: Date, concepts: string[] });
```

packages/esm-form-entry-app/src/app/form-creation/form-creation.service.ts
hadijahkyampeire: So we won't even need to do any filtering?

packages/esm-form-entry-app/src/app/form-data-source/form-data-source.service.ts
hadijahkyampeire: oh, I get it, I will move it out. Thanks for that tip.

packages/esm-form-entry-app/src/app/form-data-source/form-data-source.service.ts
hadijahkyampeire: I saw in some resources how they were formatting observations and I used the same approach but I will look closely and see if it is indeed necessary.

packages/esm-form-entry-app/src/app/form-data-source/form-data-source.service.ts
hadijahkyampeire: The returned obs data had many things and was an object so I wanted to return an array with formatted values, @ibacher please advise whether we need to format the data anyway, if we do, what is the expected JSON structure?

packages/esm-form-entry-app/src/app/openmrs-api/obs-resource.service.ts
hadijahkyampeire: I am not sure on how to use `openmrsFetch` on the angular side, from what I have seen in the existing services, they mostly use the approach I used, but I would also prefer using openmrsFetch, I will try it and see.

packages/esm-form-entry-app/src/app/form-creation/form-creation.service.ts
ibacher: For now, let's focus on the core thing: get the most recent Obs usable in a form. We can add the date filtering after.

packages/esm-form-entry-app/src/app/form-data-source/form-data-source.service.ts
ibacher: Yeah, we use that in some formatting code, but since this is designed to be called in an expression, I don't think we have the same set of concerns... (In some cases we need to work out whether the Obs should be displayed as a textbox, a number input, etc).

packages/esm-form-entry-app/src/app/openmrs-api/obs-resource.service.ts
ibacher: You should just be able to import it from `@openmrs/esm-framework`... I guess there's a conundrum about how the form will access the results... `openmrsFetch` is Promise-based; Angular's HTTP framework is built on rxjs... Neither of those return straight-forward values.

ibacher: We might have to use XMLHttpRequest for this...

packages/esm-form-entry-app/src/app/form-data-source/form-data-source.service.ts
FlorianRappl: Yeah *usually* the declaration gets erased anyway, but (1) it makes sure that it definitely *will* be erased (even though TS might think there are side-effects etc.) and (2) it's also a good hint for the reader that the imported object it really just a type.

For things like `interface` and `type` the story is anyway simple, but if we'd import a `class` then such a `type` import would also guard us from actually using it in actual code; which can be a good thing for a couple of reasons.

Anyway - just wanted to emphasize that this suggestion is great and should be followed.

packages/esm-form-entry-app/src/app/openmrs-api/concept-resource.service.ts
```diff
@@ -29,6 +29,15 @@ export class ConceptResourceService {
       );
   }
 
+  public getAllConcepts(): Observable<any[]> {
```
FlorianRappl, line 32: Can we give this a better type than `any`? Reason is quite simple - here we say "hey, that's any", but below, e.g., we say "it has an `uuid` field". So I wonder - what else do we know / assume? We should put in code here.

(same for the existing code, e.g., `getConceptByUuid` below - everything we know we should already have there)

packages/esm-form-entry-app/src/app/openmrs-api/obs-resource.service.ts
```diff
@@ -0,0 +1,25 @@
+import { Injectable } from '@angular/core';
+import { HttpClient, HttpParams } from '@angular/common/http';
+
+import { map } from 'rxjs/operators';
+import { WindowRef } from '../window-ref';
+import { fhirBaseUrl } from '@openmrs/esm-framework';
+
+@Injectable()
+export class ObservationResourceService {
+  constructor(protected http: HttpClient, protected windoRef: WindowRef) {}
```
FlorianRappl, line 10: Any reason for `protected` instead of `private`?

FlorianRappl: (I'm not seeing this being a parent class right now / in the future)

packages/esm-form-entry-app/src/app/openmrs-api/obs-resource.service.ts
FlorianRappl: If all that we need is an observable we can just use rxjs for this:

```js
import { from } from 'rxjs';
const observable = from(promise);
```

but I don't know why the given approach was selected in favor of `openmrsFetch`.

packages/esm-form-entry-app/src/app/openmrs-api/concept-resource.service.ts
hadijahkyampeire: From the discussion I had with @rbuisson today, I will be removing this, I don't need to fetch concepts because the function will be expecting selected concept uuids, so I am going to remove this.

packages/esm-form-entry-app/src/app/openmrs-api/obs-resource.service.ts
hadijahkyampeire: No reason really, I have very limited knowledge of angular, so I was just following other files. I will change that to private.

hadijahkyampeire:
@ibacher @FlorianRappl I have implemented the requested changes, please take another look.

packages/esm-form-entry-app/src/app/openmrs-api/obs-resource.service.ts
```diff
@@ -0,0 +1,19 @@
+import { Injectable } from '@angular/core';
+import { HttpClient, HttpParams } from '@angular/common/http';
+import { Observable } from 'rxjs';
+
+import { map } from 'rxjs/operators';
+import { fhirBaseUrl, openmrsObservableFetch } from '@openmrs/esm-framework';
+
+@Injectable()
+export class ObservationResourceService {
+  public openmrsFetch(url): Observable<any> {
```
denniskigen, line 10:
```suggestion
  public openmrsFetch<T>(url): Observable<FetchResponse<T>> {
```

hadijahkyampeire:
Thanks @ibacher for the commits

FlorianRappl (approved):
LGTM!

hadijahkyampeire:
Thanks @FlorianRappl for the approval, I guess we shall merge it after @ibacher and I test the data source  in a sample form

hadijahkyampeire:
Thanks, @FlorianRappl for the approval, I guess we shall merge it after @ibacher and I test the data source  in a sample form

packages/esm-form-entry-app/src/app/form-data-source/form-data-source.service.ts
```diff
@@ -86,6 +89,40 @@ export class FormDataSourceService {
...
+
+    return (patient: string) => {
+      return this.fetchMostRecentObsValue(patient, [...conceptIdentifiers]);
+    };
+  }
+
+  private extractMostRecentObsConceptIds(questions: Array<Questions>, concepts: Set<string>) {
+    for (const question of questions) {
+      const useMostRecentValue = question.questionOptions?.useMostRecentValue ?? false;
+      if (useMostRecentValue === 'true' || (typeof useMostRecentValue === 'boolean' && useMostRecentValue)) {
```
hadijahkyampeire, line 108: @ibacher is this going to be a string named `"true"` or a boolean `true`

hadijahkyampeire: Or this is an angular syntax 🤔 ?

ibacher: The idea is to support both, since that's what is used elsewhere in the form engine.

hadijahkyampeire: okay, thanks for the polish, it looks much better and makes a lot of sense now 👍 .

ibacher (approved):

PR #910, opened 2022-12-15
OZ-52: (fix) Fixed the failing test results viewer when leaf.obs is undefined

vasharma05:
## Requirements

## Summary
This PR fixes the failing test results viewer when leaf.obs is undefined

## Screenshots
None

## Related Issue
None

### DIFF
--- a/packages/esm-patient-test-results-app/src/filter/filter-reducer.ts
+++ b/packages/esm-patient-test-results-app/src/filter/filter-reducer.ts
@@ -1,4 +1,4 @@
-import { ReducerAction, ReducerState, ReducerActionType, TreeParents, TreeNode } from './filter-types';
+import { ReducerAction, ReducerState, ReducerActionType, TreeParents, TreeNode, LowestNode } from './filter-types';
 
 export const getName = (prefix, name) => {
   return prefix ? `${prefix}-${name}` : name;
@@ -11,12 +11,12 @@ const computeParents = (
   parents: TreeParents;
   leaves: Array<string>;
   tests: Array<[string, TreeNode]>;
-  lowestParents: Array<TreeNode>;
+  lowestParents: Array<LowestNode>;
 } => {
   let parents: TreeParents = {};
   const leaves: Array<string> = [];
   const tests: Array<[string, TreeNode]> = [];
-  const lowestParents: Array<TreeNode> = [];
+  const lowestParents: Array<LowestNode> = [];
   if (node?.subSets?.length && node.subSets[0].obs) {
     // lowest parent
     const activeLeaves: Array<string> = [];
@@ -33,7 +33,7 @@ const computeParents = (
     });
     leaves.push(...activeLeaves);
     tests.push(...activeTests);
-    lowestParents.push({ flatName: node.flatName, display: node.display, ...node });
+    lowestParents.push({ flatName: node.flatName, display: node.display });
   } else if (node?.subSets?.length) {
     node.subSets.forEach((subNode) => {
       const {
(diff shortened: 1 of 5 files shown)


packages/esm-patient-test-results-app/src/filter/filter-reducer.ts
```diff
@@ -19,7 +19,7 @@ const computeParents = (prefix, node) => {
     });
     let activeTests = [];
     node.subSets.forEach((leaf) => {
-      if (leaf.obs.length) {
+      if (leaf.obs?.length) {
```
FlorianRappl, line 22: I'd do `Array.isArray(leaf?.obs) && leaf.obs.length > 0` - this way you not only mitigate `leaf` for some reason being undefined, but you also make sure that the object is indeed an `Array`; and you explicitly state that its length should be greater 0 (same interpretation as just `leaf.obs.length`, but just more explicit and easier to read.

On further notice: This function should be refactored anyway. It uses `var` (bad, prefer `let`, but always try to use `const` if applicable), does not introduce any types (we have a lot of implicit `Array<any>`, and even a strange dictionary (`parents`) etc.). Furthermore nothing in this file is typed for some reason... It should definitely be improved to ensure proper usage.

FlorianRappl: Also make `activateTests` a const, same with `activeLeaves`, `lowestParents` etc.

vasharma05: Sure @FlorianRappl, I can go through for a revamp, but since I'm taken into some other work currently, @mks-d , please confirm the urgency on this, can this wait till the next week's end?

Thanks!

vasharma05: Sure @FlorianRappl, I can go through for a revamp, but since I'm taken into some other work currently, @mks-d , please confirm the urgency on this, can this wait till the next week's end?

Thanks!

FlorianRappl: There is no need to spend much time on this; just that you change the variable declarations for now. Adding a few types would also be beneficial and shouldn't take too long.

In any case - it's something we should do sooner or later.

vasharma05: Thanks @FlorianRappl !

denniskigen (approved):
Sweet. Thanks, @vasharma05!

packages/esm-patient-test-results-app/src/filter/filter-types.ts
```diff
@@ -22,12 +31,29 @@ export interface FilterLeafProps {
...
+
+export interface TreeCheckboxes {
+  [key: string]: boolean;
+}
+
+export interface TreeTests {
+  [key: string]: TreeNode;
+}
+
+export interface TreeNode {
```
denniskigen, line 46: One of these `TreeNode` definitions ought to be deleted.

denniskigen, line 46: One of these TreeNode definitions ought to be deleted.

packages/esm-patient-test-results-app/src/filter/filter-reducer.ts
```diff
@@ -1,31 +1,39 @@
...
     node.subSets.forEach((leaf) => {
-      if (leaf.obs.length) {
+      if (Array.isArray(leaf?.obs) && leaf.obs.length > 0) {
         activeTests.push([leaf.flatName, leaf]);
       }
     });
     leaves.push(...activeLeaves);
     tests.push(...activeTests);
-    lowestParents.push({ flatName: node.flatName, display: node.display });
+    lowestParents.push({ flatName: node.flatName, display: node.display, ...node });
```
denniskigen, line 36: Is this change where you're spreading extra object properties from the `node` into `lowestParents` intentional?

packages/esm-patient-test-results-app/src/filter/filter-types.ts
vasharma05: Done

packages/esm-patient-test-results-app/src/filter/filter-reducer.ts
vasharma05: Made a new Type `LowestNode` and removed `...node`

PR #925, opened 2022-12-22
(feat): I18n translations in patient chart

hadijahkyampeire:
## Requirements

## Summary
- Support form translations in the patient chart

## Screenshots
- Logged Translations
Screenshot 2023-02-10 at 16 25 38

## Related Issue
- https://issues.openmrs.org/browse/O3-1657

## Caveat
- Still have some weird errors so the forms aren't displaying but at least there is no compilation error.

This PR is branched off @icrc-jofrancisco PR: https://github.com/openmrs/openmrs-esm-patient-chart/pull/754. Just with a few fixes.


packages/esm-form-entry-app/package.json
```diff
@@ -87,7 +89,7 @@
     "karma-coverage-istanbul-reporter": "~3.0.2",
     "karma-jasmine": "~4.0.0",
     "karma-jasmine-html-reporter": "^1.5.0",
-    "openmrs": "next",
+    "openmrs": "4.0.2-pre.336",
```
FlorianRappl, line 92: Why not keep the generic `next`? Right now it would be [4.0.3-pre.441](https://www.npmjs.com/package/openmrs/v/4.0.3-pre.441).

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -211,4 +243,12 @@ export class FormSchemaService {
 
     return [...results];
   }
+
+  private getCachedTranslationsByUuid(translationUuid: string): any {
+    return this.localStorage.getObject(translationUuid);
+  }
+
+  private cacheTranslationsByUuid(translationUuid, schema): void {
```
FlorianRappl, line 251: What are the types of these parameters?

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -6,18 +6,31 @@ import { FormResourceService } from '../openmrs-api/form-resource.service';
...
   constructor(
     private formsResourceService: FormResourceService,
     private localStorage: LocalStorageService,
     private formSchemaCompiler: FormSchemaCompiler,
+    private translate: TranslateService,
   ) {}
 
   public getFormSchemaByUuid(formUuid: string, cached: boolean = true): Observable<FormSchema> {
     const cachedCompiledSchema = this.getCachedCompiledSchemaByUuid(formUuid);
+    const currentLang: string = (window as any).i18next.language.substring(0, 2).toLowerCase() || 'en';
```
FlorianRappl, line 22: Do we need to be so explicit regarding the type (`string`) here? Wouldn't it be better to type `window.i18next` (and make it `i18n | undefined`, such that `window.i18next?.language...` would be a thing).

hadijahkyampeire:
@ibacher I saw the suggested changes for implementing the cache in a store as opposed to localStorage in this [PR](https://github.com/openmrs/openmrs-esm-patient-chart/pull/754), could please elaborate more about that so that I can fix it in this PR?

FlorianRappl (commented):
Good work

packages/esm-form-entry-app/package.json
hadijahkyampeire: oh I forgot that I will fix it ASAP.

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
hadijahkyampeire: Thanks @FlorianRappl I will change that.

hadijahkyampeire: @FlorianRappl I don't understand the suggested change here, is there a type name `i18n`?

packages/esm-form-entry-app/package.json
```diff
@@ -47,6 +47,8 @@
     "@angular/router": "~12.2.16",
     "@carbon/styles": "~1.14.0",
     "@ng-select/ng-select": "^6.1.0",
+    "@ngx-translate/core": "^13.0.0",
+    "@ngx-translate/http-loader": "^6.0.0",
```
ibacher, line 51: I don't think we need this at all...

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -27,11 +40,26 @@ export class FormSchemaService {
     }).pipe(
       map(({ unCompiledSchema, formMetadataObject }) => {
         const formSchema: any = unCompiledSchema.form;
+
+        if (formSchema?.translations) {
```
ibacher, line 44: This needs to be adapted to the new translations-lookup mechanism. Essentially, when loading the form, the form should have a `resources` section with a list of form resources. Those named "<form_name>_translations_<lang>" are the resource we need to fetch (there should be a link in the result). This is similar to how we pick the schema out of the form object, so hopefully it's straight-forward.

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -6,18 +6,31 @@ import { FormResourceService } from '../openmrs-api/form-resource.service';
...
+    private translate: TranslateService,
   ) {}
 
   public getFormSchemaByUuid(formUuid: string, cached: boolean = true): Observable<FormSchema> {
     const cachedCompiledSchema = this.getCachedCompiledSchemaByUuid(formUuid);
+    const currentLang = window.i18next?.language.substring(0, 2).toLowerCase() || 'en';
+    this.translate.setDefaultLang(currentLang);
+
     if (cachedCompiledSchema && cached) {
+      if (cachedCompiledSchema?.translations) {
```
ibacher, line 26: The caching mechanism probably needs to be re-written. In my view, we should just add the route for the link to the translation resource to the things cached by the offline tools and work with that, but if not, we need to extract the translations and store them in local storage. See my previous comment for how this is to be done.

packages/esm-form-entry-app/package.json
hadijahkyampeire: @ibacher do you mean only the `http-loader` or even the `core` ?

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -6,18 +6,31 @@ import { FormResourceService } from '../openmrs-api/form-resource.service';
...
   constructor(
     private formsResourceService: FormResourceService,
     private localStorage: LocalStorageService,
     private formSchemaCompiler: FormSchemaCompiler,
+    private translate: TranslateService,
   ) {}
 
   public getFormSchemaByUuid(formUuid: string, cached: boolean = true): Observable<FormSchema> {
     const cachedCompiledSchema = this.getCachedCompiledSchemaByUuid(formUuid);
+    const currentLang = (window as any).i18next?.language.substring(0, 2).toLowerCase() || 'en';
```
FlorianRappl, line 22: Instead of treating `window` as `any` I recommend to extend the Window interface with `i18next`:

```ts
declare global {
  interface Window {
    i18next?: {
      language: string;
    };
  }
}
```

(Ideally the definition of the `i18next` property / instance is taken from `i18next` directly - I think its called something like `i18n`.)

vasharma05:
Hi @hadijahkyampeire , please check the comments by @ibacher and @FlorianRappl !
Thanks!

hadijahkyampeire:
@FlorianRappl @ibacher @denniskigen I have been blocked by this error, I have tried searching about it and it looks that the issue is unused service but I don't know how to go about it `openmrs-esm-form-entry-app.js:1 ERROR NullInjectorError: R3InjectorError(r)[Ut -> St -> wt -> wt -> wt]: 
  NullInjectorError: No provider for wt!` . Please advise.

packages/esm-form-entry-app/src/app/app.module.ts
```diff
@@ -31,6 +32,7 @@ import { FormCreationService } from './form-creation/form-creation.service';
     MonthlyScheduleResourceService,
     ConfigResourceService,
     SingleSpaPropsService,
+    TranslateService,
```
FlorianRappl, line 35: I think we might need some additional `imports` for the injection to work.

Definitely use `HttpClientModule` here.

Also Angular is at 12, but the ngx translate is at 13. Not sure if they aligned, but sometimes that is a hint..

hadijahkyampeire: I have added more imports and the error is now gone, thanks @FlorianRappl

gracepotma:
@denniskigen can you please advise Hadijah here? Up to your discretion to pull in Florian.

FlorianRappl (approved):

hadijahkyampeire:
@FlorianRappl @gracepotma this still needs to fetch the translation files from the right place, I made some local changes to do that following @ibacher 's guidance but I wasn't able to see any `translation` file. Ian mentioned today that it might be because we have not reset dev3, once that happens, `Test Form 1` will have `fr` translations.

So I would say this is not yet done but I will request to sync with @ibacher or @FlorianRappl or @denniskigen  so that we get it done quickly.

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -23,58 +23,25 @@ export class FormSchemaService {
...
-              .then((result) => {
-                this.cacheTranslationsByUuid(translationUuid, result);
-                this.translate.setTranslation(result.language, result.translations, true);
-              })
-              .catch((error) => {
-                console.error(`Error getting AMPATH translation resource: ${error.display}.`);
-              });
-          });
-        }
+    const formMetadataObject: any = this.formsResourceService.getFormMetaDataByUuid(formUuid).pipe(take(1));
```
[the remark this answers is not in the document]
ibacher: Also, it would be much better if we could just use the form metadata we're already loading [over here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/7bc9d5f6f86b62822f5ba4f4bbd41c635719ed85/packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts#L24). There's no constraint that says we need to use this function with this signature and we should only be loading the form object once from the server if at all possible.

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -6,18 +6,56 @@ import { FormResourceService } from '../openmrs-api/form-resource.service';
...
+    private translate: TranslateService,
   ) {}
 
+  public getFormTranslationsByUuid(formUuid: string) {
+    const currentLang = (window as Window).i18next?.language.substring(0, 2).toLowerCase() || 'en';
+    this.translate.setDefaultLang(currentLang);
+
+    const formMetadataObject: any = this.formsResourceService.getFormMetaDataByUuid(formUuid).pipe(take(1));
+
+    return this.getFormTranslationsByUuidFromServer(formUuid).subscribe(({ form, referencedComponents }) => {
```
ibacher, line 32: Why do we need to pass the `formUuid` instead of the `formMetadataObject` we already got? How are we using the `formUuid` differently here?

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -6,18 +6,56 @@ import { FormResourceService } from '../openmrs-api/form-resource.service';
...
+      if (form?.translations) {
+        this.translate.setTranslation(form?.language, form?.translations, true);
+      }
+
+      formMetadataObject.translations = form?.translations;
```
ibacher, line 37: Is this line necessary? What is this doing?

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -6,18 +6,56 @@ import { FormResourceService } from '../openmrs-api/form-resource.service';
...
+      return this.formSchemaCompiler.compileFormSchema(
+        formMetadataObject,
+        referencedComponents,
+      ) as unknown as FormSchema;
+    });
+  }
+
   public getFormSchemaByUuid(formUuid: string, cached: boolean = true): Observable<FormSchema> {
     const cachedCompiledSchema = this.getCachedCompiledSchemaByUuid(formUuid);
+    const currentLang = (window as Window).i18next?.language.substring(0, 2).toLowerCase() || 'en';
```
ibacher, line 47: Why do we need to do this parsing more than once?

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -6,18 +6,56 @@ import { FormResourceService } from '../openmrs-api/form-resource.service';
...
+    this.translate.setDefaultLang(currentLang);
```
ibacher, line 48: Ditto

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
ibacher: As per my above comment, this is no longer how we're storing translations (that is, as part of the form schema) so why is this code still necessary?

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -80,6 +133,72 @@ export class FormSchemaService {
     return formSchema;
   }
 
+  private getFormTranslationsByUuidFromServer(formUuid: string) {
+    const formTranslations: ReplaySubject<any> = new ReplaySubject(1);
+    this.fetchFormTranslationUsingFormMetadata(formUuid).subscribe(
+      (translations: any) => {
```
ibacher, line 139: Type here should **not** be `any`.

ibacher: Explicit types help us catch programming errors, especially when other people need to use your code. Using `any` where it's not called for (i.e., the object can literally be anything) makes maintenance harder and pretty well guarantees that we will add bugs. If you don't know the type, set it to `unknown`, but in this case, we seem to know the type...

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -80,6 +133,72 @@ export class FormSchemaService {
     return formSchema;
   }
 
+  private getFormTranslationsByUuidFromServer(formUuid: string) {
+    const formTranslations: ReplaySubject<any> = new ReplaySubject(1);
```
ibacher, line 137: This looks like it duplicates a lot of code from `getFormSchemaWithReferences()`. Why do we need two functions doing the same things?

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
```diff
@@ -80,6 +133,72 @@ export class FormSchemaService {
...
+        console.error(err);
+        formTranslations.error(err);
+      },
+    );
+    return formTranslations;
+  }
+
+  private getFormTranslationResourceByUuidFromServer(translationResourceUuid: string) {
+    return this.formsResourceService.getFormClobDataByUuid(translationResourceUuid).toPromise();
+  }
```
ibacher, line 168: Does this function need to be here? It looks like it exists to turn an RXJS object into a Promise, but it's not clear what the advantage of doing that is...

ibacher (commented):
Thanks @hadijahkyampeire. Lot's of little things in here still needing to be addressed... The real test of whether or not this works though is whether when running this code locally you can see the French translations when your locale is set to French in the app, so a screenshot of that would be extremely valuable...

packages/esm-form-entry-app/src/app/form-schema/form-schema.service.ts
hadijahkyampeire: I was trying to add translations to the formMetaData object.

hadijahkyampeire:
Thanks @ibacher for the review but honestly this who translation implementation is confusing and I could really use your expertise 🙂 🙏 . I don't even understand the flow very well.

hadijahkyampeire:
Closing this as @ibacher has made a rework of this feature. Thanks @ibacher

PR #932, opened 2023-01-09
(fix) Error handling when failed to fetch a visit attribute type

vasharma05:
## Requirements

## Summary
This PR brings following fixes:
1. Fix to the UI break: If the visit attribute type UUID mentioned in the config is not present in the backend server, the start visit form is breaking down.
2. Error Notification: If an error occurs when fetching visit attribute type from the backend, the following notification will appear. 
As per conversation with Paul Adams [here](https://openmrs.slack.com/archives/CKS32D55G/p1673346520563139?thread_ts=1672935762.284029&cid=CKS32D55G), 

a. **If the field is required to start a visit**
Show an error at the start of the form blocking it from saving the form, giving an action to reload the form, and reporting the error to the administrator.

b. **If the field is not required to start a visit**
Show a warning but not to block the form, only report an error to administrator.

3. Selecting an option by default: If the visit attribute type value is required, the default option "Select an option" will be disabled.

## Screenshots

2.a
image

2.b
image

## Related Issue
None

## Other
None


vasharma05:
Converting to draft, needs confirmation on the Design changes from Paul Adams!

vasharma05:
As per my conversation with Paul Adams, summary of the call is as follows:

If the field is required to start a visit
Show an error at the start of the form, giving an action to reload the form, and reporting the error to the administrator.

If the field is not required to start a visit
Do not show any error, only report an error to administrator.

packages/esm-patient-chart-app/src/visit/hooks/useVisitAttributeType.tsx
```diff
@@ -62,21 +52,11 @@ export function useVisitAttributeType(uuid) {
 }
 
 export function useConceptAnswersForVisitAttributeType(conceptUuid) {
-  const { data, error } = useSWRImmutable<FetchResponse<Concept>, Error>(
+  const { data, error } = useSWRImmutable<FetchResponse<Concept>, any>(
```
ibacher, line 55: Not sure why you've changed the typing here. However, if you just want to indicate "we don't care what this is", `unknown` is more correct than `any`...

ibacher (commented):
At some point, we need to ensure that the error gets logged to the console. I'm fine with the idea of not displaying a technical error to users, but we do need to ensure that someone can hunt down the error...

packages/esm-patient-chart-app/src/visit/hooks/useVisitAttributeType.tsx
vasharma05: We are using the `error` variable, it's just that I'm not type casting it

packages/esm-patient-chart-app/src/visit/hooks/useVisitAttributeType.tsx
```diff
@@ -35,21 +35,11 @@ const visitAttributeTypeCustomRepresentation =
   'custom:(uuid,display,name,description,datatypeClassname,datatypeConfig)';
 
 export function useVisitAttributeType(uuid) {
-  const { data, error } = useSWRImmutable<FetchResponse<VisitAttributeType>, Error>(
+  const { data, error } = useSWRImmutable<FetchResponse<VisitAttributeType>, any>(
```
ibacher, line 38: Why the change to `any`?

vasharma05: Resolved, thanks @ibacher for the catch!

vasharma05:
Hi @ibacher, the requested changes are done, please re-review the PR.
Thanks!

ibacher (approved):

PR #947, opened 2023-01-16
(feat) O3-1738: Show more than just ten items in conditions widget 

hadijahkyampeire:
## Requirements

## Summary
- Request for `500` results from the Backend so as to increase the returned conditions.

## Screenshots
Screenshot 2023-01-17 at 08 47 29

## Video recording

https://user-images.githubusercontent.com/30952856/212733598-70829170-d912-4175-a401-6a0cd1099111.mov

## Related Issue
- https://issues.openmrs.org/browse/O3-1738

## Other
- Ignore the blue links, it looks like I have some cache on my local.
- I removed the pagination, so ignore the one in the video.

### DIFF
--- a/packages/esm-patient-conditions-app/src/conditions/conditions.resource.tsx
+++ b/packages/esm-patient-conditions-app/src/conditions/conditions.resource.tsx
@@ -26,7 +26,7 @@ export type CodedCondition = {
 
 export function useConditions(patientUuid: string) {
   const { data, error, isValidating } = useSWR<{ data: FHIRConditionResponse }, Error>(
-    `${fhirBaseUrl}/Condition?patient=${patientUuid}`,
+    `${fhirBaseUrl}/Condition?patient=${patientUuid}&_count=500`,
     openmrsFetch,
   );
 
--- a/packages/esm-patient-conditions-app/src/conditions/conditions.resource.tsx
+++ b/packages/esm-patient-conditions-app/src/conditions/conditions.resource.tsx
@@ -26,7 +26,7 @@ export type CodedCondition = {
 
 export function useConditions(patientUuid: string) {
   const { data, error, isValidating } = useSWR<{ data: FHIRConditionResponse }, Error>(
-    `${fhirBaseUrl}/Condition?patient=${patientUuid}&_count=500`,
+    `${fhirBaseUrl}/Condition?patient=${patientUuid}&_count=100`,
     openmrsFetch,
   );
 
--- a/packages/esm-patient-conditions-app/src/conditions/conditions-detailed-summary.component.tsx
+++ b/packages/esm-patient-conditions-app/src/conditions/conditions-detailed-summary.component.tsx
@@ -15,14 +15,8 @@ import {
   TableRow,
 } from '@carbon/react';
 import { Add } from '@carbon/react/icons';
-import { formatDate, parseDate, usePagination } from '@openmrs/esm-framework';
-import {
-  CardHeader,
-  EmptyState,
-  ErrorState,
-  launchPatientWorkspace,
-  PatientChartPagination,
-} from '@openmrs/esm-patient-common-lib';
+import { formatDate, parseDate } from '@openmrs/esm-framework';
+import { CardHeader, EmptyState, ErrorState, launchPatientWorkspace } from '@openmrs/esm-patient-common-lib';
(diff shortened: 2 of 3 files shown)


kdaud:
Thanks @hadijahkyampeire I am going to test the changes to night and get the feedback asap.

gracepotma:
Thanks so much for your very quick work on this @hadijahkyampeire! This is indeed an improvement and I support merging it since it fixes the underlying issue. **However** many, many patients have >10 conditions, and we wouldn't want serious ones to be hidden behind pagination. So there will need to be a second PR (or we should change this one I guess?) to instead apply the same logic that we use on the Orders and Visits pages: Showing the **full list**. Sorry this requirement was kind of buried at the end of the ticket description: 
image

Alternatively, if it's easier, could we just set the pagination length for this widget to default to 100 conditions?

hadijahkyampeire:
Thanks, @gracepotma I can remove the pagination and we just make the table scrollable if the conditions become very many.

packages/esm-patient-conditions-app/src/conditions/conditions.resource.tsx
```diff
@@ -26,7 +26,7 @@ export type CodedCondition = {
 
 export function useConditions(patientUuid: string) {
   const { data, error, isValidating } = useSWR<{ data: FHIRConditionResponse }, Error>(
-    `${fhirBaseUrl}/Condition?patient=${patientUuid}`,
+    `${fhirBaseUrl}/Condition?patient=${patientUuid}&_count=500`,
```
vasharma05, line 29: Hey @hadijahkyampeire , shouldn't we have pagination in place instead of calling count of 500.

By default, it should only fetch 10 conditions, and when user clicks on the next arrow, it should fetch another set of results. Also, in the fhir module, we get a `total` variable giving us a total results, and hence we can use pagination more effectively.


Consider this, if there are around 100s of conditions, they'll be fetched all together everytime, even if the user doesn't look at the conditions widget at all. This take up memory and processing, where it is not even needed.

vasharma05 (changes requested):

packages/esm-patient-conditions-app/src/conditions/conditions.resource.tsx
dkayiwa: And bandwidth too. :)

hadijahkyampeire: @vasharma05 @dkayiwa as you see I have a commit for removing pagination, so no we don't want pagination we just want to request for more results from the backend, please read the ticket requirements and the proposed solution.

hadijahkyampeire:
Hello @dkayiwa, do you know if there is a place where I can find FHIR endpoints for conditions well documented with expected payload, queryParams, response etc. That would be helpful in identifying whether pagination can be done on the backend (if yes what params do I pass), Also I am adding `edit` and `delete` functionality but I am not sure what the endpoints are. cc @ibacher

packages/esm-patient-conditions-app/src/conditions/conditions.resource.tsx
hadijahkyampeire: The allowance for pagination according to @gracepotma  should be maybe showing 100 results per page, I can maybe put back the pagination with 100 per page but I would say we just leave it for now because the table can scroll and we don't want to hide important conditions behind pagination.

vasharma05:
Hey @gracepotma !

Instead of removing the pagination to pull out important conditions, shouldn't there be a priority among the conditions so that we can order the conditions, instead of filling the page with around 100 conditions?

vasharma05 (approved):
Approving now, as a quick fix, but this needs to be implemented in a proper way in the future.

gracepotma:
@vasharma05 fair question re. the ordering by severity; however, I don't believe we should implement something like that for 2 reasons: (1) different types of clinicians can have very different opinions about what is higher priority (e.g. a cardiologist will have a different opinion than an oncologist, etc), and (2) it would become something of a metadata management nightmare. Mayyyybe in the future we could have something like a move up/move down UI that clinicians could use to reorder the list but again for reasons #1 and #2 I don't think it would work out long-term. I'm happy with how it looks for now with the long-page view :) Although of course if we find there are performance concerns due to this implementation, I'm happy to reevaluate.

PR #961, opened 2023-01-21
(Fixes):Pagination, multiple submissions and sorting patient chart appointments.(03-1765 & 03-1795)

hadijahkyampeire:
## Requirements

## Summary
- Fix the pagination behavior when one switches to another table view before the pagination was keeping the previous table's page e.g 3, and when you switch to view maybe past Appointments, it would take you to page 3 which is most likely empty and it looks weird.
- Handling sorting differently depending on the kind of appointments.(Refer to the ticket to see the detailed convo)
- Fix the multiple saving of an appointment by disabling the save button after clicking it, this prevents other clicks.
- Change `pageSize` to `10` instead of 5.

## Screenshots
Screenshot 2023-01-21 at 23 44 13

## Recording

https://user-images.githubusercontent.com/30952856/213887699-2829d286-7cb3-412f-b6a3-9a0abd1168d1.mov

## Related Issue
- https://issues.openmrs.org/browse/O3-1765
- https://issues.openmrs.org/browse/O3-1795


vasharma05 (approved):
The use of `switchedView` looks flaky to me, if you are able to get something else working, it might be better.

Otherwise, the changes looks good.

hadijahkyampeire:
Thanks @vasharma05 for the review, do you have something better at the top of your mind? that would really help, I was thinking of giving each tab it's own table but again that is also not good and if we want pagination to reset when we change the tabs, then we need a value that tells us when the tab has been changed. We can always come back to it if we figure out a better approach and I think all the tables we have which have tabs behave the same way when paginated. So we might need a global fix.

PR #964, opened 2023-01-24
(refactor) Use the bound mutate API instead of global mutate

denniskigen:
## Requirements

## Summary

SWR offers two approaches for using the `mutate` API to mutate data: 

- The global mutate API which can mutate any key, 
- And the bound mutate API only mutates the data of its corresponding SWR hook.

This PR refactors usages of `mutate` to use the bound mutate API so that developers don't have to pass `keys` to SWR hooks.

### DIFF
--- a/packages/esm-patient-notes-app/src/notes/visit-notes-form.component.tsx
+++ b/packages/esm-patient-notes-app/src/notes/visit-notes-form.component.tsx
@@ -1,6 +1,5 @@
-import React, { SyntheticEvent, useMemo } from 'react';
+import React, { SyntheticEvent, useCallback, useMemo, useState } from 'react';
 import dayjs from 'dayjs';
-import { useSWRConfig } from 'swr';
 import { useTranslation } from 'react-i18next';
 import debounce from 'lodash-es/debounce';
 import {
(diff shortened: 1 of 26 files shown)


ibacher (approved):

packages/esm-patient-notes-app/src/notes/visit-notes-form.component.tsx
```diff
@@ -32,68 +31,42 @@ import {
...
-        setProviderUuid(data.uuid);
-      });
-    }
-    if (currentSessionLocationUuid) {
-      fetchLocationByUuid(ac, currentSessionLocationUuid).then(({ data }) => setLocationUuid(data.uuid));
-    }
-  }, [currentSessionLocationUuid, currentSessionProviderUuid]);
+  const { mutateVisitNotes } = useVisitNotes(patientUuid);
+  const { locationUuid } = useLocationUuid(session?.sessionLocation?.uuid);
+  const { providerUuid } = useProviderUuid(session?.currentProvider?.uuid);
```
ibacher, line 69: I don't know that these hooks are useful... I mean, I don't really see the point of the code they're replacing either... If we load these UUIDs from the session endpoint, those UUIDs willl always be valid so the re-checking with the backend is a a couple of useless calls. I'd just go with:

```
const locationUuid = session?.sessionLocation?.uuid;
const providerUuuid = session?.currentProvider?.uuid;
```

denniskigen: Great catch. Felt icky writing this.

PR #965, opened 2023-01-24
(feat) O3-1410: Conditions widget should filter conditions by status

denniskigen:
## Requirements

## Summary

The Conditions widget should filter conditions by their clinical status. This means the user should be able to see the following states:

- All conditions
- Active conditions
- Inactive conditions

This PR adds a dropdown that the user can click on to select the clinical status they'd like to filter the conditions list by.

## Screenshots

https://user-images.githubusercontent.com/8509731/214384282-ead9e34e-17a1-48cb-9650-b90f4161542e.mp4

## Related issue

https://issues.openmrs.org/browse/O3-1410


packages/esm-patient-conditions-app/src/conditions/conditions-detailed-summary.component.tsx
```diff
@@ -24,39 +25,57 @@ function ConditionsDetailedSummary({ patient }) {
   const { t } = useTranslation();
   const displayText = t('conditions', 'Conditions');
   const headerTitle = t('conditions', 'Conditions');
+  const [filter, setFilter] = useState('');
```
ibacher, line 28: Three things here:

1. I think we should default the condition list to "Active" conditions instead of "All". In this context, "Active" implies "is part of the patient's care", so it's much more important that those be shown.
2. We know what the default is, so we may as well use that.
3. I think stricter typing is better here to prevent future errors.

```suggestion
  const [filter, setFilter] = useState<'All' | 'Active' | 'Inactive'>('Active');
```

ibacher (approved):
Thanks @denniskigen! This basically LGTM.

PR #968, opened 2023-01-25
(feat) O3-1797: Re-order Summary page widgets to match designs (and remove some)

vasharma05:
## Requirements

## Summary
This PR fixes the ordering of the widgets in the Patient Summary sections.

The order goes as follows:
1. Vitals
2. Biometrics
3. Conditions
4. Active Medications

Reference: [Design](https://app.zeplin.io/project/60d59321e8100b0324762e05/screen/60d59c8c1ff239bbe989d7d5) and Ciaran's comment on O3-1797.
 
The remaining widgets are not yet created and hence we have the above 4 widgets left.

## Screenshots
image

image

## Related Issue
https://issues.openmrs.org/browse/O3-1797

## Other
https://github.com/openmrs/openmrs-esm-core/pull/596

### DIFF
--- a/packages/esm-patient-conditions-app/src/index.ts
+++ b/packages/esm-patient-conditions-app/src/index.ts
@@ -1,5 +1,5 @@
 import { defineConfigSchema, getAsyncLifecycle, getSyncLifecycle } from '@openmrs/esm-framework';
-import { createDashboardLink } from '@openmrs/esm-patient-common-lib';
+import { createDashboardLink, getPatientSummaryOrder } from '@openmrs/esm-patient-common-lib';
 import { configSchema } from './config-schema';
 import { dashboardMeta } from './dashboard.meta';
 
@@ -29,7 +29,7 @@ function setupOpenMRS() {
       {
         name: 'conditions-overview-widget',
         slot: 'patient-chart-summary-dashboard-slot',
-        order: 3,
+        order: getPatientSummaryOrder('Conditions'),
         load: getAsyncLifecycle(() => import('./conditions/conditions-overview.component'), options),
         meta: {
           columnSpan: 4,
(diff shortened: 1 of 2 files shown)


vasharma05:
Hey @ibacher @rbuisson, asking this question as per implementation view.

I have deleted the extensions from the root in this PR. Should I keep the extension intact, removing only the slot from extension, which can be added back to Patient Summary via configuration by implementers, or is it okay to remove it, keeping the design view same for all the distributions?

Thanks!

packages/esm-patient-conditions-app/src/index.ts
```diff
@@ -29,7 +29,7 @@ function setupOpenMRS() {
       {
         name: 'conditions-overview-widget',
         slot: 'patient-chart-summary-dashboard-slot',
-        order: 6,
+        order: 3,
```
ibacher, line 32: It would be nice if we had a distinct order per widget. Both this and medications are at order 3.

ibacher:
> I have deleted the extensions from the root in this PR. Should I keep the extension intact, removing only the slot from extension, which can be added back to Patient Summary via configuration by implementers, or is it okay to remove it, keeping the design view same for all the distributions?

Yeah, that's a good point. We probably don't really need the forms overview widget at all (since it will be replaced by the workspace). I guess it might make sense to keep the others around but not assigned to a slot, at least for now.

packages/esm-patient-conditions-app/src/index.ts
vasharma05: Thanks for the catch!

denniskigen:
@vasharma05  @ibacher how can we move this work forward?

ibacher (approved):
IMHO This should be marked as a feat, as it really does require at least a minor version bump to the patient chart (since it's a fairly large UI change). Otherwise, LGTM.

vasharma05:
Duly noted @ibacher !
Thanks!

PR #980, opened 2023-02-01
(fix) Fixed regression when fetching concept dictionary data using multiple requests

icrc-psousa:
When fetching concept dictionary data to render the form, requests will be spitted in batches of 100. Prior to this PR, the final result was only assuming the data from the last request. This PR fixes that regression and ensures that all requested concept data is concatenated and returned even if is spitted between multiple requests on concept service.

## Requirements


ibacher (approved):

denniskigen:
@icrc-psousa Thanks, Pedro!

PR #995, opened 2023-02-10
(Improvement: 03-1856): Appointments cancel modal improvements.

hadijahkyampeire:
## Requirements

## Summary
- Better naming of the appointments cancel modal action buttons texts.

## Screenshots
- 
Screenshot 2023-02-10 at 18 45 29

## Related Issue
- [Better naming of the Appointments cancel modal.](https://issues.openmrs.org/browse/O3-1856)


ibacher:
Can these be separated into a PR-per-ticket? I know they're small things, but we should be committing at the level of features, meaning each PR should (ideally) fix one thing.

ibacher:
PS for the condition thing: instead of messing with the CSS of disabled text (which is going to be hard to get exactly right), why not just render a fixed label if we're editing and skip the search component altogether?

hadijahkyampeire:
@ibacher I have removed the `conditions` changes I will do them in a seperate PR.

vasharma05 (approved):
LGTM! Thanks!

PR #997, opened 2023-02-14
(fix) 03-1857: Make the disabled condition a label instead of a disabled search.

hadijahkyampeire:
## Requirements

## Summary
- Display the condition as a label in edit mode.

## Screenshots

https://user-images.githubusercontent.com/30952856/218719218-ccec042e-aa1c-4373-a63a-b70ee162acac.mov

## Related Issue
- https://issues.openmrs.org/browse/O3-1857

### DIFF
--- a/packages/esm-patient-conditions-app/src/conditions/conditions-widget.component.tsx
+++ b/packages/esm-patient-conditions-app/src/conditions/conditions-widget.component.tsx
@@ -14,6 +14,7 @@ import {
   Search,
   Stack,
   Tile,
+  FormLabel,
 } from '@carbon/react';
 import { createErrorHandler, showNotification, showToast, useLayoutType, useSession } from '@openmrs/esm-framework';
 import {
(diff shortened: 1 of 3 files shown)


hadijahkyampeire:
@denniskigen I have implemented the feedback you shared during the coffee break.

packages/esm-patient-conditions-app/src/conditions/conditions-widget.component.tsx
```diff
@@ -222,59 +223,71 @@ const ConditionsWidget: React.FC<ConditionsWidgetProps> = ({
...
-                      <span>
-                        {t('noResultsFor', 'No results for')} <strong>"{conditionToLookup}"</strong>
-                      </span>
-                    </Tile>
-                  </Layer>
-                </>
-              );
-            })()}
-          </div>
+          {context === 'editing' ? (
```
denniskigen, line 226: When editing a condition, we should show the `display` value property from the `condition` prop. This value is immediately available, unlike the `selectedCondition` property. That means we shouldn't need to show a loading state either. This should be refactored to:

```tsx
const displayName = getFieldValue(condition?.cells, 'display');
```

Which could then get used as follows:

```tsx
{context === 'editing' ? (
  <FormLabel className={styles.conditionLabel}>. 
    {displayName}
  </FormLabel>
...
}
```

packages/esm-patient-conditions-app/translations/en.json
```diff
@@ -24,6 +24,7 @@
   "enterCondition": "Enter condition",
   "errorDeletingCondition": "Error deleting a condition",
   "inactive": "Inactive",
+  "loadingCondition": "Loading Condition",
```
denniskigen, line 27: This translation shouldn't be required based on the advice above.

denniskigen (commented):
This is very close @hadijahkyampeire. Just a few issues to address.

packages/esm-patient-conditions-app/src/conditions/conditions-widget.component.tsx
hadijahkyampeire: oh okay, I see that, thanks.

hadijahkyampeire:
@denniskigen I have made the tweaks and done some small refactoring since you have work for the most refactoring.

denniskigen (approved):
LGTM. Thanks, @hadijahkyampeire!

PR #1005, opened 2023-02-15
(refactor) Refactor conditions widget

denniskigen:
## Requirements

- [x]  My work includes tests or is validated by existing tests.

## Summary

Refactors the conditions widget as follows:

- Refactors API functions in the conditions resource to asynchronous functions that use promises instead of observables. Also moves the Create and Edit payloads into the resource instead of the Conditions Widget component.
- Amends the primary buttons in the Create, Edit and Delete forms to show a loading state when the form gets submitted. This also helps ensure the user doesn't inadvertently click the primary button more than once.
- Refactors the Conditions Form so that inline notifications are shown when errors occur when creating or editing a condition.
- Remove translations from the Conditions table filter's `items` list. Having translations breaks this functionality at the moment. I'll look for a different approach to adding translations back at some point in the future.


packages/esm-patient-conditions-app/src/conditions/conditions-detailed-summary.component.tsx
```diff
@@ -92,11 +92,11 @@ function ConditionsDetailedSummary({ patient }) {
             <div className={styles.filterContainer}>
               <Dropdown
                 id="conditionStatusFilter"
-                initialSelectedItem={t('active', 'Active')}
+                initialSelectedItem="Active"
```
ibacher, line 95: Doesn't this undo our ability to translate these strings?

packages/esm-patient-conditions-app/src/conditions/conditions-form.component.tsx
```diff
@@ -1,51 +1,103 @@
...
-  context: string;
-  closeWorkspace: () => void;
 }
 
-const ConditionsForm: React.FC<ConditionFormProps> = ({ closeWorkspace, patientUuid, context, condition }) => {
+const ConditionsForm: React.FC<ConditionFormProps> = ({ closeWorkspace, condition, formContext, patientUuid }) => {
   const { t } = useTranslation();
   const isTablet = useLayoutType() === 'tablet';
-  const [hasSubmissibleValue, setHasSubmissibleValue] = React.useState(false);
   const submissionNotifier = useMemo(() => new BehaviorSubject<{ isSubmitting: boolean }>({ isSubmitting: false }), []);
```
ibacher, line 21: Do we need the submissionNotifier at all? I.e., can't we just use `isSubmittingForm` / `setIsSubmittingForm` directly?

ibacher (approved):
Mostly LGTM. A couple of small points, though.

packages/esm-patient-conditions-app/src/conditions/conditions-form.component.tsx
samuelmale: @ibacher There was this idea of supporting "Forms Embeddable widgets". This was in effort to reuse existing RefApp widgets within Forms as extensions. This works on a concept of "self contained components" and stuff like the `submissionNotifier` helped facilitate that.  Checkout thread for some background info => https://talk.openmrs.org/t/o3-embedding-conditions-widget-in-forms/36642

ibacher: ... I don't love that, but ok.

packages/esm-patient-conditions-app/src/conditions/conditions-detailed-summary.component.tsx
ibacher: Ah... yeah, I see. So, theoretically, there would be a way for the backend to localise the display value, but that's not what's being operated on here.

In the [FHIR Condition](https://www.hl7.org/fhir/condition.html), `clinicalStatus` takes a concept from a required [ValueSet](https://www.hl7.org/fhir/valueset-condition-clinical.html), so, e.g., "active" is not intended to be a display string "active"; it's actually the concept ID for the "active" concept that might have different translations in different locales. From that perspective, [this](https://github.com/openmrs/openmrs-esm-patient-chart/blob/f12632e57e02fb77fc612778937bc3ad0a6dd04a/packages/esm-patient-conditions-app/src/conditions/conditions.resource.tsx#L91) is a bit odd.

Let me think about how we might translate this best... It would be nice if we didn't end up in a situation where we needed to synchronize both backend and frontend translation strings.

PR #1018, opened 2023-02-22
(chore) Specify max concurrency in build script

denniskigen:
## Requirements

## Summary

Specifies a max concurrency value for the build script in the `Report bundle size` workflow. This should lead to fewer failed builds.


denniskigen:
Re: the failing report bundle size workflow - the reason for this is it appears `ng build` doesn't play well with Node 18.x. Downgrading to Node 16 fixes the problem.

ibacher (approved):

PR #1020, opened 2023-02-24
O3-1896 [Enhancement]: add minimum and maximum values for the Record vitals form

Jexsie:
## Requirements

## Summary

As **Ian** suggested,  if a user is using the numeric dials, they should start at the absolute minimum and go only as far as the absolute maximum

## Screenshots

[Screencast from 2023-02-23 19-02-19.webm](https://user-images.githubusercontent.com/104269786/221114112-71ec8287-70cb-4916-9f9a-6ec5797a9e7f.webm)

## Related Issue

https://issues.openmrs.org/browse/O3-1896

## Other

Had no reference to the exact measures. I  just dialed until I see warnings and used that range. If there is any desired change, am willing to.

### DIFF
--- a/packages/esm-patient-vitals-app/src/vitals/vitals-biometrics-form/vitals-biometrics-form.component.tsx
+++ b/packages/esm-patient-vitals-app/src/vitals/vitals-biometrics-form/vitals-biometrics-form.component.tsx
@@ -47,6 +47,30 @@ const VitalsAndBiometricForms: React.FC<DefaultWorkspaceProps> = ({ patientUuid,
   const [isSubmitting, setIsSubmitting] = useState<boolean>(false);
   const encounterUuid = currentVisit?.encounters?.find((enc) => enc?.form?.uuid === config.vitals.formUuid)?.uuid;
 
+  const uuids = {
+    systolicBloodPressureUuid: config.concepts.systolicBloodPressureUuid,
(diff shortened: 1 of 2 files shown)


jona42-ui (commented):
Thanks @Jexsie  however am curious to know from folks like @ibacher @dkayiwa @bmamlin @gracepotma whether standards like SNOMED-CT, LOINC, and ICD-10, which can be used to represent vital sign concepts in a standardized way are not used?

gracepotma:
Completely agree with @jona42-ui's reaction. Let's *please* not hard-code ranges into the software (and if / when we really have to, we'd need to document that very thoroughly and obviously, so that this is not missed when there are changes to ranges in the future - folks won't magically remember all the places that hard-coded values would need to be changed). 
The Vitals Ranges app should be leveraging the max/min established _on the associated concepts_. That way if we change the ranges on the source concepts, the app auto-adjusts. 
Perhaps try looking at how the Vitals Header references the concepts and their values: https://github.com/openmrs/openmrs-esm-patient-chart/tree/main/packages/esm-patient-vitals-app/src/vitals/vitals-header

Jexsie:
thanks a lot @gracepotma @ibacher @jona42-ui  for the corrections. Let me fix that.

Jexsie:
Hey @jona42-ui  and @ibacher, I have updated my PR with a new commit. Kindly check it out.

PR #1022, opened 2023-02-26
(feat) O3-1896: Number input steppers in vitals and biometrics form shouldn't exceed ranges

Jexsie:
… and biometrics form

## Requirements

## Summary

- As Ian suggested, if a user is using the numeric dials, they should start at the absolute minimum and go only as far as the absolute maximum
- Updated unit test to handle the change 

## Screenshots

[Screencast from 2023-02-23 19-02-19.webm](https://user-images.githubusercontent.com/104269786/221401249-270d7523-abd2-457e-94bf-1cb3fa442754.webm)

## Related Issue

https://issues.openmrs.org/browse/O3-1896

### DIFF
--- a/packages/esm-patient-common-lib/src/useVitalsConceptMetadata.tsx
+++ b/packages/esm-patient-common-lib/src/useVitalsConceptMetadata.tsx
@@ -19,13 +19,16 @@ export function useVitalsConceptMetadata() {
     : new Map<string, string>([]);
 
   const conceptRanges = conceptMetadata?.length
-    ? new Map<string, { lowAbsolute: number | string | null; highAbsolute: number | string | null }>(
+    ? new Map<string, { lowAbsolute: number | null; highAbsolute: number | null }>(
         conceptMetadata.map((concept) => [
           concept.uuid,
-          { lowAbsolute: concept.lowAbsolute, highAbsolute: concept.hiAbsolute },
+          {
+            lowAbsolute: concept.lowAbsolute ?? null,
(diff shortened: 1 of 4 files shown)


Jexsie:
cc @ibacher, @denniskigen, @gracepotma

packages/esm-patient-common-lib/src/useVitalsConceptMetadata.tsx
```diff
@@ -17,11 +17,22 @@ export function useVitalsConceptMetadata() {
   const conceptUnits = conceptMetadata?.length
     ? new Map<string, string>(conceptMetadata.map((concept) => [concept.uuid, concept.units]))
     : new Map<string, string>([]);
+
+  const conceptRanges = conceptMetadata?.length
+    ? new Map<string, { lowAbsolute: number | string | null; highAbsolute: number | string | null }>(
```
ibacher, line 22: Do we need to include the "string" type here? These should basically always be numbers. Also, note that the values may be unset (i.e., undefined). If `undefined` isn't a valid value for `min` or `max` we should probably ensure that the value will be `null`, e.g., `{ lowAbsolute: concept.lowAbsolute ?? null }`.

ibacher (commented):
Thanks @Jexsie! Mostly looks good, just some TS questions.

Jexsie:
Hey @ibacher, pushed an update for this

ibacher (approved):

PR #1026, opened 2023-02-27
(fix) 03-1917: Forms icon in the side rails loses the active class.

hadijahkyampeire:
## Requirements

## Summary
- This PR fixes the bug where we were losing the active styles on the form icon when we click a specific form. This was because we were only checking for the `clinical-forms-workspace` and the form itself had another name `patient-form-entry-workspace` so I added that as a check to add the active class to the forms icons.

## Screenshots

https://user-images.githubusercontent.com/30952856/221686837-2e1c0e57-158c-4227-9b2f-345a529fd38f.mov

## Related Issue
- https://issues.openmrs.org/browse/O3-1917


jwnasambu:
Thanks @hadijahkyampeire for the great work. LGTM!

denniskigen (approved):
Thanks, @hadijahkyampeire!

PR #1028, opened 2023-02-28
(fix) Attempt to fix weird overflow menu UI

denniskigen:
## Requirements

## Summary

This PR is an absolute shot in the dark at fixing an issue where overflow menus in the visits table get rendered with a green background. Nothing in this PR directly addresses that issue, and the weird thing is that neither @hadijahkyampeire nor I could replicate it locally. Per the browser inspector tool, the offending `background-color` property appears to come from the `cds--btn-primary` class in the overflow menu wrapper. This is weird because this implementation matches the conditions widget overflow menu implementation almost to the letter, yet that issue doesn't occur in the conditions widget.

## Screenshots

> Devtools inspector

Screenshot 2023-03-01 at 00 07 49

> Issue on dev3 (note the green background color on the overflow menu buttons)

Screenshot 2023-03-01 at 00 14 24

> Same UI when run locally
 
Screenshot 2023-03-01 at 00 22 23

### DIFF
--- a/packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/visits-table/visits-table.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/visits-table/visits-table.component.tsx
@@ -132,7 +132,6 @@ const VisitTable: React.FC<VisitTableProps> = ({ showAllEncounters, visits }) =>
 
   return (
     <DataTable
-      data-floating-menu-container
       filterRows={handleFilter}
       headers={tableHeaders}
       rows={tableRows}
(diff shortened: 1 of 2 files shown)


ibacher (approved):
🙏

ibacher:
I think that is the right colour for `.cds—btn—primary` which means, maybe more bizarrely, that the same CSS class isn’t being applied when run locally?

Admittedly I have seen some weird corner-cases (the login app not working on first load on dev3, but working fine if run locally, comes to mind)…

hadijahkyampeire:
Thanks @denniskigen for the blind shot, let's see if it fixes it 🙏

PR #1029, opened 2023-02-28
(fix) 03-1932: Form Icon should be disabled if a specific form is open

hadijahkyampeire:
## Requirements

## Summary
- This PR adds the ability to disable the form icon in the side rail when a form is open so as to prevent switching forms without first discarding the open form.

## Screenshots

https://user-images.githubusercontent.com/30952856/222009469-4eb3e27f-fdef-485a-b6c9-c515122f360c.mov

## Related Issue
- https://issues.openmrs.org/browse/O3-1932

### DIFF
--- a/packages/esm-patient-forms-app/src/clinical-form-action-button.component.tsx
+++ b/packages/esm-patient-forms-app/src/clinical-form-action-button.component.tsx
@@ -17,12 +17,15 @@ const ClinicalFormActionButton: React.FC = () => {
     workspaces?.[0]?.name?.match(/clinical-forms-workspace/i) ||
     workspaces?.[0]?.name?.match(/patient-form-entry-workspace/i);
 
+  const isFormOpen = workspaces.filter((w) => w.name === 'patient-form-entry-workspace')?.length >= 1;
+
   if (layout === 'tablet') {
     return (
       <Button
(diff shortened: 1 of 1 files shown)


denniskigen:
LGTM. Any concerns, @rbuisson?

denniskigen (approved):
LGTM! Does this change look ok to you, @rbuisson?

donaldkibet:
@hadijahkyampeire @denniskigen i believe instead of disabling the form-icon, we should show a close button as shown on the designs [here](https://app.zeplin.io/project/61434fa756474d5545f65cf4/screen/614362c96abfb0557aa9e357), where when you hover on the forms icon, a close action is shown. I might be wrong though

rbuisson:
I wouldn't have done it this way @hadijahkyampeire . Does disabling the button when the form is open has the side effect that the user may not be able to re-open when the form is collapsed?

rbuisson:
Very good point @donaldkibet  and thanks for pointing to the designs. @hadijahkyampeire  why not implementing it that way? Did you receive some other design instructions?

hadijahkyampeire:
Thanks @rbuisson and @donaldkibet I will push a PR to fix this.

PR #1033, opened 2023-03-03
Adds configuration to hide patient identifiers

pirupius:
## Requirements

## Summary
Different implementations have different ways to uniquely identify their patients/customers. This PR adds a configuration to conditionally hide some system identifiers using with a given identifier type coding uuid

## Screenshots
This screenshot has an example with the `OpenMRS` identifier hidden

Screenshot 2023-03-03 at 03 44 38


packages/esm-patient-chart-app/src/patient-chart/patient-chart.component.tsx
```diff
@@ -47,7 +46,7 @@ const PatientChart: React.FC = () => {
           >
             <ExtensionSlot extensionSlotName="breadcrumbs-slot" />
             <aside>
-              <PatientBanner {...state} />
+              <ExtensionSlot extensionSlotName="patient-header-slot" state={state} />
```
pirupius, line 49: Deleted some duplicate code for the patient banner within the patient chart esm and replaced it with an extension to load the version from the patient banner esm

CynthiaKamau (commented):
Thank you @pirupius for this, is there a reason why you deleted the relationships and contact details component ?

pirupius:
@CynthiaKamau am not sure at what point it got in but there was some duplication of code with the exact components within the `esm-patient-banner-app` existing in the `esm-patient-chart-app` . So after confirmation from @denniskigen i harmonized any changes that might have been non-existent in the `esm-patient-banner-app` and deleted the files in the `esm-patient-chart-app`. With this the patient chart will use the ones in the patient banner.

packages/esm-patient-chart-app/src/patient-chart/patient-chart.component.tsx
vasharma05: Hi @pirupius !
Actually this was done intentionally because as an extension, the patient banner was taking a lot of time to load and we wanted it to be the first thing to be loaded on the patient chart page.

pirupius: @vasharma05 thanks for your feedback. I have however tested this with a backend server with over 350k patients and i haven't yet seen an issue. Could it be that this isn't a blocker anymore?

donaldkibet (approved):

PR #1052, opened 2023-03-16
Change on getCurrentUser to return the user data and sessionLocation

icrc-agomes:
## Requirements

## Summary
When we called the getCurrentUser function the property sessionLocation was always null.
This caused, when we tried to save a form, to be stored with no location.


njiddasalifu (commented):
Good change. I like the identation
one suggestion, I think you can add some comments in the code to just explain how the session is mapped.

nanfuka:
great work here, please indicate a link to the ticket you are fixing

ibacher:
Closing in favour of #1070, which solves the same problem.

PR #1070, opened 2023-03-24
(fix) fix setting provider and location values on form-entry launch

donaldkibet:
## Requirements

## Summary

This PR fixes a bug that occurs when a user launches the form engine and both provider and location aren't set automatically. This happens because the `getCurrentUser` function is intended to return a session object instead of `User` object. I have provided a fix for that. But we will need to rename [getCurrentUser](https://github.com/openmrs/openmrs-esm-core/blob/304fd2e7d7f20e85cf8151453e67baddf84794ab/packages/framework/esm-api/src/shared-api-objects/current-user.ts#L71) function to `getCurrentSession` or something close to what the function does.

## Screenshots
userLocation

### DIFF
--- a/packages/esm-form-entry-app/src/app/fe-wrapper/fe-wrapper.component.ts
+++ b/packages/esm-form-entry-app/src/app/fe-wrapper/fe-wrapper.component.ts
@@ -121,7 +121,7 @@ export class FeWrapperComponent implements OnInit, OnDestroy {
 
     return forkJoin({
       formSchema: this.fetchCompiledFormSchema(this.formUuid, language).pipe(take(1)),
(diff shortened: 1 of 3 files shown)


denniskigen:
Kind of what [this](https://github.com/openmrs/openmrs-esm-patient-chart/pull/1052) is solving, right?

donaldkibet:
> Kind of what [this](https://github.com/openmrs/openmrs-esm-patient-chart/pull/1052) is solving, right? (haven't gone through either this or your PR yet, mind)

Yes, i believe the `getCurrentUser` is a function hitting the session endpoint. Therefore returning the session Object which also contain user information. Secondly we also need to extract `currentProvider` information from session object. In my opinion providing the entire `session` object is the best approach. @denniskigen

PR #1073, opened 2023-03-26
O3-1998: Allergy Table shows wrong Onset date

gracepotma:
Issue: https://issues.openmrs.org/browse/O3-1998
1. Enter an allergy; enter a date of onset sometime in the past. Save the allergy.
2. View the Allergies summary table: See it shows "Today" under "Since", not the "Date of first onset". This is wrong. 

Blocker for Faimer.

### DIFF
--- a/packages/esm-patient-allergies-app/src/allergies/allergies-detailed-summary.component.tsx
+++ b/packages/esm-patient-allergies-app/src/allergies/allergies-detailed-summary.component.tsx
@@ -44,10 +44,6 @@ const AllergiesDetailedSummary: React.FC<AllergiesDetailedSummaryProps> = ({ pat
     },
     { key: 'reaction', header: t('reaction', 'Reaction') },
     {
-      key: 'onsetDate',
-      header: t('since', 'Since'),
-    },
-    {
       key: 'lastUpdated',
       header: t('lastUpdated', 'Last updated'),
     },
@@ -74,7 +70,6 @@ const AllergiesDetailedSummary: React.FC<AllergiesDetailedSummaryProps> = ({ pat
           </span>
         ),
       },
-      onsetDate: formatDate(parseDate(allergy.onsetDate), { day: false, time: false }) ?? '--',
       lastUpdated: allergy.lastUpdated ? formatDate(parseDate(allergy.lastUpdated), { time: false }) : '--',
       reaction: allergy.reactionManifestations?.join(', '),
       note: allergy?.note ?? '--',
--- a/packages/esm-patient-allergies-app/src/allergies/allergies-detailed-summary.component.tsx
+++ b/packages/esm-patient-allergies-app/src/allergies/allergies-detailed-summary.component.tsx
@@ -44,7 +44,7 @@ const AllergiesDetailedSummary: React.FC<AllergiesDetailedSummaryProps> = ({ pat
     },
     { key: 'reaction', header: t('reaction', 'Reaction') },
     {
-      key: 'recordedDate',
+      key: 'onsetDate',
       header: t('since', 'Since'),
     },
     {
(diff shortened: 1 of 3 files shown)


packages/esm-patient-allergies-app/src/allergies/allergies-detailed-summary.component.tsx
```diff
@@ -74,7 +74,7 @@ const AllergiesDetailedSummary: React.FC<AllergiesDetailedSummaryProps> = ({ pat
           </span>
         ),
       },
-      recordedDate: formatDate(parseDate(allergy.recordedDate), { day: false, time: false }) ?? '--',
+      onsetDate: formatDate(parseDate(allergy.onsetDate), { day: false, time: false }) ?? '--',
```
hadijahkyampeire, line 77: @gracepotma looking at the data being returned by the API, we do not have a field named `onsetDate` we have a `recordedDate` and `lastUpdated Date`. So if we want to show the `onSetDate` we will need the API to be modified so that it returns it. cc @ibacher @denniskigen

hadijahkyampeire: Screenshot 2023-03-27 at 01 15 36
I have also looked at the fetched resource, I have not seen the `onSetDate`.

gracepotma: Ah I see. Wow. I assumed onsetDate could be used as a reference because I saw it used elsewhere in the allergies app here: https://github.com/openmrs/openmrs-esm-patient-chart/blob/f5330c710fa4f6f6d75d0659e7d15a7d8887d683/packages/esm-patient-allergies-app/src/allergies/allergies-form/allergy-form.component.tsx#L55

gracepotma: If there is no possible way to represent this in the data model, then I propose we just remove this column completely for now, as it is misleading.

hadijahkyampeire: Okay I will push that change on this PR.

denniskigen:
@ibacher, could you please chime?

gracepotma:
> @ibacher, could you please chime in?

@ibacher could we actually just extend the allergy resource by adding an attribute via a concept, e.g. we could have a concept for "onsetDate" and attach that?

gracepotma:
> @ibacher, could you please chime in?

@ibacher could we actually just extend the allergy resource by adding an attribute via a concept, e.g. we could have a concept for "onsetDate" and attach that?

ibacher:
> could we actually just extend the allergy resource by adding an attribute via a concept, e.g. we could have a concept for "onsetDate" and attach that?

Short version: no. The Allergy base type doesn't support attributes or, really, any form of extension right now.

I'm for just removing it from the table for now. We do need a way of recording an onset date, but neither the date recorded, the encounter date, or the last updated date are the right date for that.

gracepotma:
OK - could we at least add text to the comment box to allow the user to capture the onset date as free text?

ibacher:
>  could we at least add prompt text to the comment box to encourage the user to capture the onset date as free text?

Yes!

dkayiwa:
@gracepotma do you see a need for us to implement allergy attributes as part of the platform 2.7 roadmap?

ibacher:
> do you see a need for us to implement allergy attributes as part of the platform 2.7 roadmap?

I do think we should have adding an onset date (and an onset date for AllergyReaction) on the roadmap.

ibacher (approved):

PR #1074, opened 2023-03-26
(fix) 03-1995: Fix patient conditions sorting on "onSetDate".

hadijahkyampeire:
## Requirements

## Summary
- This PR fixes the sorting bug we had when we sort on `onSetDate` on the patient chart conditions table. This was reported by the `faimer team`. The date string we show on the UI was not properly sorting since it is just alphabetical, so there was need to specify the sortKey which is the actual dateTime string.

## Screenshots
Screenshot 2023-03-27 at 00 58 50

- Loom recording
https://www.loom.com/share/10a1246a12d74fc8ba7a02e7a3d45270

## Related Issue
- https://issues.openmrs.org/browse/O3-1995

### DIFF
--- a/packages/esm-patient-conditions-app/src/conditions/utils.ts
+++ b/packages/esm-patient-conditions-app/src/conditions/utils.ts
@@ -1,0 +1,19 @@
+/**
+ * Enables a comparison of arbitrary values with support for undefined/null.
+ * Requires the `<` and `>` operators to return something reasonable for the provided values.
+ */
+export function compare<T>(x?: T, y?: T) {
+  if (x == undefined && y == undefined) {
+    return 0;
+  } else if (x == undefined) {
+    return -1;
+  } else if (y == undefined) {
+    return 1;
+  } else if (x < y) {
+    return -1;
+  } else if (x > y) {
+    return 1;
+  } else {
+    return 0;
+  }
+}
--- a/packages/esm-patient-conditions-app/src/conditions/utils.ts
+++ b/packages/esm-patient-conditions-app/src/conditions/utils.ts
@@ -1,19 +1,23 @@
+import { parseDate } from '@openmrs/esm-framework';
 /**
  * Enables a comparison of arbitrary values with support for undefined/null.
  * Requires the `<` and `>` operators to return something reasonable for the provided values.
  */
-export function compare<T>(x?: T, y?: T) {
-  if (x == undefined && y == undefined) {
+export function compare<T extends string>(x?: T, y?: T) {
+  if (x === y || (!x && !y)) {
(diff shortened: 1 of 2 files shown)


packages/esm-patient-conditions-app/src/conditions/utils.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Enables a comparison of arbitrary values with support for undefined/null.
+ * Requires the `<` and `>` operators to return something reasonable for the provided values.
+ */
+export function compare<T>(x?: T, y?: T) {
```
vasharma05, line 5: You didn't consider the case for `--`

hadijahkyampeire: Let me check that, but I was handling it as an empty string.

hadijahkyampeire: @vasharma05 actually this compare function does not accept the `emdash` values, but I am handling such cases as undefined. Do you have some pointers on how you would want the `emdash` to be handled?

hadijahkyampeire: The `parseDate` will return a date string and that fails the `typescript` flags it as not being a number.

ibacher: `parseDate()` returns a JS Date object...

ibacher: So:

```ts
  if (x === y || !x && !y) {
    return 0;
  } else if (!x) {
    return -1;
  } else if (!y) {
    return 1;
  } else {
    const xDate = parseDate(x);
    const yDate = parseDate(y);
    
    if (xDate === yDate) {
       return 0;
    }

    return xDate < yDate ? -1 : 1;
  }
```

hadijahkyampeire: Thanks @ibacher

ayush-AI (commented):
Hello @hadijahkyampeire, thanks you for this PR. In the PR title, mentioned issue should be O3 -1995 instead of 03 -1995. Please fix that.

packages/esm-patient-conditions-app/src/conditions/utils.ts
vasharma05: Ohh okay, I remembered `parseDate` returned millseconds, but forgot about the JS Date object.
Thanks Ian for the correction!

vasharma05 (approved):
Thanks for the fix @hadijahkyampeire !

PR #1060, opened 2023-03-20
(fix) O3-1973: Separate active orders and past orders in the medications  summary table

Jexsie:
## Requirements

- [x] This PR has a title that briefly describes the work done, including the ticket number if there is a ticket.
- [x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://zeroheight.com/23a080e38/p/880723-introduction).
- [x] My work includes tests or is validated by existing tests.

## Summary

 - Gets all orders and filters out **pastOrders** and **activeOrders** to solve the issue explained in the jira link below

### Before
Screenshot from 2023-03-20 11-02-34

### After
image

## Related Issue

https://issues.openmrs.org/browse/O3-1973

PR #1076, opened 2023-03-27
Revert "(fix) O3-1973: Separate active orders and past orders in the medications  summary table"

hadijahkyampeire:
Reverts openmrs/openmrs-esm-patient-chart#1060

### DIFF
--- a/packages/esm-patient-medications-app/src/medications-summary/medications-summary.component.tsx
+++ b/packages/esm-patient-medications-app/src/medications-summary/medications-summary.component.tsx
@@ -1,13 +1,12 @@
-import React, { useMemo } from 'react';
+import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { DataTableSkeleton } from '@carbon/react';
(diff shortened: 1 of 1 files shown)


hadijahkyampeire:
@ibacher I have created this PR to revert the merged work, I don't know if that is okay, we can have @Jexsie push the work together with the remaining work on a new PR.

ibacher:
We don't need to do this... we can just move forward with what's there... it's not broken just needs a quick fix commit

Jexsie:
Hey @ibacher, what changes should I commit now

ibacher:
@Jexsie Fixed things in [this commit](https://github.com/openmrs/openmrs-esm-patient-chart/commit/63204b7f746524c52f9c8921cdb7dd777a5ba540). The typings were wrong; the object we were getting from the backend had strings, rather than dates (which makes sense; JSON does not use JS's Date object).

Jexsie:
Thanks @ibacher.

PR #1077, opened 2023-03-28
(feat) add ability to use ohri-form engine on patient chart

donaldkibet:
## Requirements

## Summary
Add ability to launch and use the `OHRI` form engine. I have also made it configurable to retain the current implementation where we are using the `AMPATH` form engine.

Only to be merged once https://github.com/UCSF-IGHS/openmrs-ohri-form-engine-lib/pull/86 has been merged. @samuelmale @denniskigen @ibacher @gracepotma what are your thought on this?

## Screenshots
ohri-form

## TODO
- Better error handling
- Edit mode
- Fix error while saving


ibacher:
I like this for the most part; however:

1. I'd prefer that the workspace embedding the OHRI form engine be moved to it's own app, rather than integrated into the patient-forms-app
2. I'd prefer that the workspace embedding the OHRI form engine not use an extension. This feels like an abuse of the extension system since we're expecting one and only one type of thing to be rendered there.

Ideally, we'd have a single workspace, called `patient-form-entry-workspace` and an implementation could either chose to use the AMPATH form engine (which would then render in that workspace) or use the OHRI form engine (which would then render in that same workspace). This would basically remove the need for any configuration other than what's declared in the import map (and we can add an error if an implementation somehow ends up with two workspaces with the same registration, similar to what we do with extensions).

donaldkibet:
> I like this for the most part; however:
> 
> 1. I'd prefer that the workspace embedding the OHRI form engine be moved to it's own app, rather than integrated into the patient-forms-app
> 2. I'd prefer that the workspace embedding the OHRI form engine not use an extension. This feels like an abuse of the extension system since we're expecting one and only one type of thing to be rendered there.
> 
> Ideally, we'd have a single workspace, called `patient-form-entry-workspace` and an implementation could either chose to use the AMPATH form engine (which would then render in that workspace) or use the OHRI form engine (which would then render in that same workspace). This would basically remove the need for any configuration other than what's declared in the import map (and we can add an error if an implementation somehow ends up with two workspaces with the same registration, similar to what we do with extensions).

Alright make sense. In that regard am closing this PR.

samuelmale:
Thank you for starting this, Donald! There is a plan to develop an app that dispenses OHRI forms within the RefApp. I was hoping to start working on this next week. Donald, would you like to collaborate with me to make sure we don't duplicate efforts?

donaldkibet:
> Thank you for starting this, Donald! There is a plan to develop an app that dispenses OHRI forms within the RefApp. I was hoping to start working on this next week. Donald, would you like to collaborate with me to make sure we don't duplicate efforts?

Sure

PR #1083, opened 2023-03-30
(fix) Improve performance of patient chart

ibacher:
## Requirements

## Summary

This is a draft for the changes I'm trying to implement for the patient chart, mostly to discuss what's happening here. Commentary to follow.


packages/esm-patient-chart-app/src/patient-chart/patient-chart.component.tsx
```diff
@@ -31,32 +31,40 @@ const PatientChart: React.FC = () => {
...
-                <ChartReview {...state} view={view} />
-                <WorkspaceNotification />
+      <>
+        <div
+          className={`${styles.innerChartContainer} ${
+            windowSize.size === 'normal' && active ? styles.closeWorkspace : styles.activeWorkspace
+          }`}
+        >
+          <ExtensionSlot extensionSlotName="breadcrumbs-slot" />
+          {isLoadingPatient ? (
```
ibacher, line 49: Realistically, I want to get rid of this `<Loader />` component here at all. There's not really any state that we're passing to child objects aside from "`usePatient()` resolved". Currently whatever renders in the `patient-header-slot` and the conditions app on the chart seem to be broken without waiting for that, but that should be fixable at the level of those components.

packages/esm-patient-chart-app/src/workspace/workspace-window.component.tsx
```diff
@@ -13,7 +13,8 @@ interface ContextWorkspaceParams {
   patientUuid?: string;
 }
 
-const WorkspaceWindow: React.FC<ContextWorkspaceParams> = ({ patientUuid }) => {
+const WorkspaceWindow: React.FC<ContextWorkspaceParams> = () => {
```
ibacher, line 16: Changed from a parameter to using a hook. `usePatient()` itself should basically guarantee that we're only fetching as appropriate

packages/esm-patient-common-lib/src/offline/patient.ts
```diff
@@ -1,60 +0,0 @@
-import { usePatient, getSynchronizationItems } from '@openmrs/esm-framework';
-import useSWR from 'swr';
-
-export const patientRegistrationSyncType = 'patient-registration';
```
ibacher, line 4: Deleting this constant may have been a mistake, but we don't need the rest of this file...

packages/esm-patient-chart-app/src/workspace/workspace-window.component.tsx
ibacher: Empirically, not really for me at least relative to the delay it had before hand, but yeah, I was hoping `usePatient()` would deduplicate things, but it doesn't.

denniskigen (approved):
Thanks, Ian.

ibacher:
@denniskigen Are you happy for me to merge this in? 

@donaldkibet You raised a good point about multiple requests for patients that I do want to address in the future. However, currently on dev3, before this change, when I load the patient on the chart, I see 8 requests to the `https://dev3.openmrs.org/openmrs/ws/fhir2/R4/Patient/<patient uuid>` endpoint, so I don't think effectively moving one of those requests to a different point in the React tree is going to be a real performance hit (`usePatientOrOfflinePatient()` doesn't actually de-dedupe requests anyways).

denniskigen (approved):

denniskigen:
I guess I put Descartes before the proverbial horse, @ibacher, but yes, very happy to merge this in, and the corollary of this is quite palpable for me on dev3. Thanks for the great work!

The other incredible thing I hadn't noticed about this is this size diff: 😮 

Screenshot 2023-04-14 at 1 31 30 PM

PR #1139, opened 2023-04-28
03 1970 Fix Forms don’t open after having opened a visit form.

sainingo:
## Requirements

## Summary

[Walkthrough Video](https://www.loom.com/share/d895202dde9541f4a373083386738433)

## Screenshots

startVisitvsist

## Related Issue

https://issues.openmrs.org/projects/O3/issues/O3-1970

### DIFF
--- a/packages/esm-patient-chart-app/src/actions-buttons/start-visit.component.tsx
+++ b/packages/esm-patient-chart-app/src/actions-buttons/start-visit.component.tsx
@@ -26,7 +26,7 @@ const StartVisitOverflowMenuItem: React.FC<StartVisitOverflowMenuItemProps> = ({
           }}
         >
           <span className="cds--overflow-menu-options__option-content">
-            {!startVisitLabel ? <>{t('startVisit', 'Start visit')}</> : startVisitLabel}
+            {startVisitLabel ? <>{t('startVisit', 'Start visit')}</> : startVisitLabel}
           </span>
         </button>
       </li>
--- a/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
@@ -86,7 +86,7 @@ const StartVisitForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, closeWor
   const handleSubmit = useCallback(
     (event) => {
       event.preventDefault();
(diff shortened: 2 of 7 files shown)


packages/esm-patient-chart-app/src/actions-buttons/start-visit.component.tsx
```diff
@@ -26,7 +26,7 @@ const StartVisitOverflowMenuItem: React.FC<StartVisitOverflowMenuItemProps> = ({
           }}
         >
           <span className="cds--overflow-menu-options__option-content">
-            {!startVisitLabel ? <>{t('startVisit', 'Start visit')}</> : startVisitLabel}
+            {startVisitLabel ? <>{t('startVisit', 'Start visit')}</> : startVisitLabel}
```
ibacher, line 29: Doesn't this do things incorrectly, i.e., if a user set the `startVisitLabel` via config, they get the translated "Start visit" screen, whereas if they did not, they get the `startVisitLabel` (which will be some kind of falsy value).

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -121,10 +123,10 @@ const StartVisitForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, closeWor
         .pipe(first())
         .subscribe(
           (response) => {
+            getVisitFormData.currentVisit = response.data;
+            localStorage.setItem('visit-form', JSON.stringify(getVisitFormData));
```
ibacher, line 127: I've said this before, but please don't use `localStorage` for tracking temporary data. We really don't need any storage here and could likely just set a variable value, but if we _must_ use storage for this, use sessionStorage (local storage is for values that persist across sessions—not at all what we want here)

ibacher (commented):
I get that this works, but it's very form specific. We have an overarching probably that whenever a workspace opens that requires a visit and we show the Start a Visit form, the user is not redirected to the workspace once they've started the visit, so it would be nicer if there were a way to solve this that wasn't specific to each thing in the workspace, i.e., if we can add the hooks to the part of the code that is actually opening the "Start Visit" workspace so that once that's complete, the expected workspace is opened.

packages/esm-patient-chart-app/src/actions-buttons/start-visit.component.tsx
sainingo: Yes,.. I will revert that

PR #1154, opened 2023-05-11
Modify the visit date to accommodate the new encounter date

icrc-psousa:
## Requirements

### DIFF
--- a/packages/esm-form-entry-app/src/app/form-submission/form-submission.service.ts
+++ b/packages/esm-form-entry-app/src/app/form-submission/form-submission.service.ts
@@ -1,7 +1,7 @@
 import { Injectable } from '@angular/core';
 
 import { forkJoin, Observable, of, from } from 'rxjs';
-import { catchError, mergeMap } from 'rxjs/operators';
+import { catchError, flatMap, map, mapTo, mergeMap, switchMap, take } from 'rxjs/operators';
 import { EncounterAdapter, PersonAttribuAdapter, Form } from '@openmrs/ngx-formentry';
 import { NodeBase } from '@openmrs/ngx-formentry/form-entry/form-factory/form-node';
 import { EncounterResourceService } from '../openmrs-api/encounter-resource.service';
@@ -19,6 +19,7 @@ import cloneDeep from 'lodash-es/cloneDeep';
 import { mutateEncounterCreateToPartialEncounter } from '../offline/syncItemMutation';
 import { SingleSpaPropsService } from '../single-spa-props/single-spa-props.service';
 import { v4 } from 'uuid';
+import { VisitResourceService } from '../openmrs-api/visit-resource.service';
 
 /**
  * The result of submitting a form via the {@link FormSubmissionService.submitPayload} function.
@@ -36,6 +37,7 @@ export class FormSubmissionService {
     private readonly personResourceService: PersonResourceService,
     private readonly formDataSourceService: FormDataSourceService,
     private readonly singleSpaPropsService: SingleSpaPropsService,
+    private readonly visitResourceService: VisitResourceService,
   ) {}
 
   public submitPayload(form: Form): Observable<FormSubmissionResult> {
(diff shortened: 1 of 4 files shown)


packages/esm-form-entry-app/src/app/form-submission/form-submission.service.ts
```diff
@@ -131,6 +129,47 @@ export class FormSubmissionService {
...
+    );
+  }
+
+  private submitEncounter(encounterCreate: EncounterCreate): Observable<any | undefined> {
+    if (!encounterCreate) {
+      return of(undefined);
+    }
+
+    if (encounterCreate.uuid) {
+      return this.encounterResourceService.getEncounterByUuid(encounterCreate.uuid).pipe(
```
icrc-psousa, line 144: Unnecessary request to get encounter so that visit metadata is fetched?
Encounter data passed by openmrs-ngx-formentry contains no visit info. Is it worth looking into openmrs-ngx-formentry to try getting all visit data?

packages/esm-form-entry-app/src/app/form-submission/form-submission.service.ts
```diff
@@ -131,6 +129,47 @@ export class FormSubmissionService {
...
+        take(1),
+        mergeMap((encounter) => {
+          if (
```
icrc-psousa, line 147: Currently, this is always checking encounter date against visit date. Even though encounter date may not have been changed or the POST to update the encounter would return 200 meaning dates are OK. Would it make sense to let the POST go through to update visit 1st and, based on the returned error (400?), check the dates?

packages/esm-form-entry-app/src/app/form-submission/form-submission.service.ts
ibacher: In general, I think it's preferable to try to get the visit date. Since the user is required to start (or select an existing visit) before creating the form, we should already have that data without needing to fetch anything new. Maybe it can be passed in from the workspace as a property?

packages/esm-form-entry-app/src/app/form-submission/form-submission.service.ts
ibacher: I think it's preferable to do the check here, since that ensures that we have the same logic in offline mode. Otherwise, an offline form might end up being rejected (then again, I'm a little vague on whether or how we're tracking visits for offline mode)

packages/esm-form-entry-app/src/app/form-submission/form-submission.service.ts
```diff
@@ -131,6 +129,47 @@ export class FormSubmissionService {
     }
   }
 
+  private confirmVisitDateAdjustment() {
+    return confirm(
+      'The encounter date falls outside the designated visit date range. Would you like to modify the visit date to accommodate the new encounter date?',
```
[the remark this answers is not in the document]
ibacher: Tbf, we can always do the initial version with a confirm dialog and add the full Carbon UI later.

packages/esm-form-entry-app/src/app/openmrs-api/visit-resource.service.ts
```diff
@@ -0,0 +1,36 @@
...
+  public getUrl(): string {
+    return this.windowRef.openmrsRestBase + 'visit';
+  }
+
+  public getVisitByUuid(uuid: string, v: string = null): Observable<any> {
+    let url = this.getUrl();
+    url += '/' + uuid;
+
+    const params: HttpParams = new HttpParams().set('v', v && v.length > 0 ? v : this.v);
+    return this.http.get<Person>(url, { params });
```
ibacher, line 23: At least I assume we're not expecting this to return a `Person`

```suggestion
    return this.http.get<Visit>(url, { params });
```

packages/esm-form-entry-app/src/app/types/index.ts
```diff
@@ -260,7 +260,15 @@ export interface Concept {
...
 
+export interface Visit {
+  display?: string,
+  startDatetime: string,
+  stopDatetime?: string,
+  uuid: string,
+}
+
 export interface Encounter {
+  visit: Visit;
```
ibacher, line 271: Also, this should be further down or at least after the `uuid` property.

```suggestion
  visit?: Visit;
```

packages/esm-form-entry-app/src/app/form-submission/form-submission.service.ts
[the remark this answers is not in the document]
icrc-psousa: Agreed. Also, regarding the translations. Would it be ok to leave the hard-coded English msg until we have ngx-translate in place. I believe translations will be supported with this PR: https://github.com/openmrs/openmrs-ngx-formentry/pull/42

packages/esm-form-entry-app/src/app/types/index.ts
icrc-psousa: Corrected

delphinepas:
@ibacher would it be possible to review this PR? thank you cc @icrc-psousa @icrc-fdeniger

ibacher (approved):
LGTM. Thanks @icrc-psousa!

PR #1155, opened 2023-05-12
(feat) O3-2091:Support translations for breadcrumbs

jwnasambu:
## Requirements

## Summary
This PR allows translating the breadcrumbs' text from the place where the breadcrumbs are registered.

## Screenshots
image

## Related Issue
https://issues.openmrs.org/browse/O3-2091


packages/esm-patient-chart-app/src/index.ts
```diff
@@ -29,19 +30,20 @@ function setupOpenMRS() {
   defineExtensionConfigSchema('nav-group', genericNavGroupConfigSchema);
   defineExtensionConfigSchema('dashboard', genericDashboardConfigSchema);
 
+  const t = i18next.getFixedT(null, moduleName);
   registerBreadcrumbs([
     {
       path: spaBasePath,
-      title: 'Patient',
+      title: () => Promise.resolve(t('patientBreadcrumb', 'Patient')),
```
ibacher, line 37: We shouldn't need the `Promise.resolve()` here, right? It's generally better to avoid wrapping things in promises if we can. (Also applies below).

packages/esm-patient-chart-app/src/index.ts
```diff
@@ -29,19 +30,20 @@ function setupOpenMRS() {
...
       parent: `${window.spaBase}/home`,
     },
     {
```
ibacher, line 40: We probably need some comments here, at least to cover the standard dashboards, e.g.

```ts
/**
 * t('visitDashboard', 'Visit dashboard');
 */
```

(Not sure if that's a real one.

ibacher:
I'd also expect to, at the very least, see updates to the en.json file to add the `patientBreadcrumb` key. Please ensure that you are running the git commit hooks (i.e., don't commit or push with `--no-verify`).

jwnasambu:
@ibacher thanks for the review!

vasharma05:
Marking the PR as a draft PR, since it's not working locally, as discussed with @jwnasambu.

As soon as the solution works, we'll mark the PR open for review.

@ibacher we'll need your feedback on the PR.

jwnasambu:
@vasharma05 Thanks so much for the advice. Am so humbled!

packages/esm-patient-chart-app/src/index.ts
```diff
@@ -30,20 +29,25 @@ function setupOpenMRS() {
   defineExtensionConfigSchema('nav-group', genericNavGroupConfigSchema);
   defineExtensionConfigSchema('dashboard', genericDashboardConfigSchema);
 
-  const t = i18next.getFixedT(null, moduleName);
   registerBreadcrumbs([
     {
       path: spaBasePath,
-      title: () => Promise.resolve(t('patientBreadcrumb', 'Patient')),
+      title: () => Promise.resolve(window.i18next.t('patientBreadcrumb', { defaultValue: 'Patient', ns: moduleName })),
```
vasharma05, line 35: Hi @ibacher !
Is window.i18next working for you?
It's throwing error, since i18next is undefined

ibacher: Yes... I only committed this once it was working. `window.i18next` should be defined, especially by the point this gets called...

vasharma05: Ohh okay 🤔 

I got the catch

vasharma05: I was actually calling the `window.i18next` inside the `setupOpenMRS`, but you called it inside the breadcrumb function.

Yes, by the time breadcrumb function was called, it is setup.

ibacher: Hmmm... that's weird.

vasharma05: I went through the app shell too. [Here](https://github.com/openmrs/openmrs-esm-core/blob/1acdec571b0284d0c78c367b1377d544bc87485a/packages/shell/esm-app-shell/src/run.ts#L345) it showed that apps are loading first, before the i18 is setup. (Correct me if I'm wrong)

vasharma05: Thanks for the help @ibacher !
I tried a lot of ways but wasn't able to figure this out

vasharma05:
Hi @ibacher !
Please re-review the changes

packages/esm-patient-chart-app/src/index.ts
ibacher: > it showed that apps are loading first, before the i18 is setup

Yeah, that's why I had to be done inside a function that's called later. We could rework things, but we'd have to be a little careful about it; translations are loaded [from the module's `importTranslation()` export](https://github.com/openmrs/openmrs-esm-core/blob/1acdec571b0284d0c78c367b1377d544bc87485a/packages/shell/esm-app-shell/src/locale.ts#L91) function, but we're only able to load the module once `loadModules()` is called, so we'd have to add a step between `loadApps()` and `setupApps()` to initialize the i18next context if we really need this to work inside `setupOpenMRS()`.

I'm not sure of the value of doing that, though?

ibacher: Also, `i18next` probably needs to be shared so we can avoid loading directly from `window`.

packages/esm-patient-chart-app/src/index.ts
```diff
@@ -32,12 +32,17 @@ function setupOpenMRS() {
   registerBreadcrumbs([
     {
       path: spaBasePath,
-      title: 'Patient',
+      // @ts-ignore
```
ibacher, line 35: Well, I don't like the `// @ts-ignore`. I'd prefer the `Promise.resolve()` to that.

packages/esm-patient-chart-app/src/index.ts
```diff
@@ -32,12 +32,17 @@ function setupOpenMRS() {
...
+      title: () => window.i18next.getFixedT(null, moduleName)('patientBreadcrumb', 'Patient'),
       parent: `${window.spaBase}/home`,
     },
     {
       path: `${spaBasePath}/:view`,
-      title: ([_, key]) => `${decodeURIComponent(key)} dashboard`,
+      title: ([_, key]) =>
+        `${window.i18next.getFixedT(null, moduleName)(decodeURIComponent(key))} ${window.i18next.getFixedT(
```
ibacher, line 42: I don't think this is a correct translation of what was happening before... PS, there's no value in using `getFixedT()` over `t` in this case...

ibacher (changes requested):

packages/esm-patient-chart-app/src/index.ts
```diff
@@ -32,16 +32,17 @@ function setupOpenMRS() {
   registerBreadcrumbs([
     {
       path: spaBasePath,
-      title: 'Patient',
+      title: () => Promise.resolve(window.i18next.t('Patient_title', { defaultValue: 'Patient', ns: moduleName })),
```
vasharma05, line 35:
```suggestion
      title: () => Promise.resolve(window.i18next.t('patientBreadcrumb', { defaultValue: 'Patient', ns: moduleName })),
```

packages/esm-patient-chart-app/src/index.ts
```diff
@@ -32,16 +32,17 @@ function setupOpenMRS() {
...
       parent: `${window.spaBase}/home`,
     },
     {
       path: `${spaBasePath}/:view`,
-      title: ([_, key]) => `${decodeURIComponent(key)} dashboard`,
+      title: ([_, key]) =>
+        Promise.resolve(window.i18next.t(`${decodeURIComponent(key)} Dashboard`, { ns: moduleName, defaultValue: 'Default Value' })),
```
vasharma05, line 41:
```suggestion
        Promise.resolve(window.i18next.t(`${decodeURIComponent(key)} Dashboard`, { ns: moduleName, defaultValue: `${decodeURIComponent(key)} Dashboard` })),
```

vasharma05 (commented):
Some more changes

vasharma05 (commented):
Some more changes

vasharma05 (commented):
Some more changes

packages/esm-patient-chart-app/src/index.ts
vasharma05: Sorry for the same.
Thanks!

packages/esm-patient-chart-app/src/index.ts
vasharma05: I agree, thanks for the review!

vasharma05:
@ibacher @jwnasambu , I have updated the PR description

packages/esm-patient-chart-app/src/index.ts
```diff
@@ -32,16 +32,17 @@ function setupOpenMRS() {
...
       path: spaBasePath,
-      title: 'Patient',
+      title: () => Promise.resolve(window.i18next.t('patientBreadcrumb', { defaultValue: 'Patient', ns: moduleName })),
       parent: `${window.spaBase}/home`,
     },
     {
       path: `${spaBasePath}/:view`,
-      title: ([_, key]) => `${decodeURIComponent(key)} dashboard`,
+      title: ([_, key]) =>
+        Promise.resolve(window.i18next.t(`${decodeURIComponent(key)} Dashboard`, { ns: moduleName, defaultValue: `${decodeURIComponent(key)} Dashboard` })),
```
ibacher, line 41: If we're using the same string as the key and default value, we only need the key.

packages/esm-patient-chart-app/src/index.ts
```diff
@@ -32,16 +32,17 @@ function setupOpenMRS() {
   registerBreadcrumbs([
```
ibacher, line 32: It would probably be good to add some comments so that these keys get loaded, e.g.,

```
/**
 * t('patientBreadcrumb', 'Patient')
 * t'('Patient Summary Dashboard')
 */
```

ibacher (approved):

jwnasambu:
@ibacher, @vasharma05 kindly feel free to review the proposed changes at your convinient time please!

packages/esm-patient-chart-app/translations/fr.json
```diff
@@ -85,6 +87,10 @@
   "other": "Autre",
   "pastVisitErrorText": "Erreur de visite passée",
   "pastVisits": "Visites passées",
+  "patientSummary":"résumé du patient",
+  "patient":"patiente",
+  "patientBreadcrumb": "patiente",
+  "Patient Summary Dashboard": "tableau de bord récapitulatif des patients",
   "priority": "Priorité",
```
Jexsie, line 94: wouldn't it look better if these are sentence cased like **_Résumé du patient, Patiente_** or this can be specified in the config?

denniskigen: Correct, @Jexsie. We generally try to use [sentence case](https://writer.com/blog/sentence-case/) for translation strings.

jwnasambu: Thanks @Jexsie and @denniskigen for the review!

vasharma05 (approved):

PR #1157, opened 2023-05-13
(chore) Improve build efficiency of form-engine-app

ibacher:
## Requirements

## Summary

1. Removes @openmrs/esm-patient-common-lib from `dependencies`. This should shrink the size of the built application and it's unnecessary here as @openmrs/esm-patient-common-lib can be resolved from the workspace.
2. Fixes the `scriptRuleConfig.exclude` regex to actually do what it's intended to do instead of [this](https://regexper.com/#%28node_modules%5B%5E%5C%2F%40openmrs%5C%2Fesm%5C-patient%5C-common%5C-lib%5D%29). I think the current regex only works by mistake and likely causes additional libraries to be loaded by the SWC loader at build time.
3. Removes the alias for `@openmrs/esm-framework/src/internal` which doesn't seem to be necessary or useful. I left the other aliases in place because I am not familiar with those libraries.

While this will likely result in a larger code size reported by the automated tooling, it should actually result in a slightly smaller code size loaded for this app at runtime.

### DIFF
--- a/packages/esm-form-engine-app/package.json
+++ b/packages/esm-form-engine-app/package.json
@@ -35,7 +35,6 @@
   },
   "dependencies": {
     "@carbon/react": "^1.12.0",
-    "@openmrs/esm-patient-common-lib": "^4.4.0",
     "@openmrs/openmrs-form-engine-lib": "next",
     "lodash-es": "^4.17.15",
     "react-error-boundary": "^4.0.3"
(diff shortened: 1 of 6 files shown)


denniskigen (approved):
LGTM. Thanks, @ibacher.

donaldkibet (approved):

PR #1183, opened 2023-05-30
(feat) Add i18n for the left menu in patient chart

jwnasambu:
## Requirements

## Summary

This PR enables translating menu items in the patient chart left nav menu. It fixed the `encodeURIComponent(title.toString())`, ensuring that the title value is converted to a string before being passed to the `encodeURIComponent` function.


jwnasambu:
@ibacher and @denniskigen Kindly review my work at your convinient time please!

packages/esm-patient-common-lib/src/dashboards/DashboardExtension.tsx
```diff
@@ -1,24 +1,35 @@
...
+        setResolvedTitle(resolvedValue);
+      });
+    } else {
+      setResolvedTitle(title);
+    }
+  }, [title]);
 
   return (
-    <div key={title}>
+    <div key={resolvedTitle}>
```
ibacher, line 27: If `resolvedTitle` isn't set, then this should not render anything.

packages/esm-patient-common-lib/src/dashboards/DashboardExtension.tsx
```diff
@@ -1,24 +1,35 @@
...
       <ConfigurableLink
-        to={`${basePath}/${encodeURIComponent(title)}`}
-        className={`cds--side-nav__link ${title === navLink && 'active-left-nav-link'}`}
+        to={`${basePath}/${encodeURIComponent(resolvedTitle)}`}
```
ibacher, line 29: I'm a little nervous that this won't work properly because we're now assuming that the _translated_ title will resolve to a URL that serves the correct page / component and I don't think that's the case. I think we need to change the `DashboardExtension` to take 3 properties:

1. The base path
2. The translatable title
3. The URI fragment used for this component.

Basically, the thing used in `to` and `className` should be the URI fragment (so, non-translated) and the translated title should _only_ be used as the content of the link tag.

ibacher: Basically, the URL generated for a page in OpenMRS should not depend on the user's language.

ibacher (changes requested):

ibacher (approved):

packages/esm-patient-chart-app/src/side-nav/generic-dashboard.component.tsx
```diff
@@ -5,11 +5,15 @@ import { DashboardExtension } from '@openmrs/esm-patient-common-lib';
...
+    _description: 'The display string for this dashboard',
+    _type: Type.String,
+  },
+  path: {
+    _description: 'The URL fragment this link points to',
     _type: Type.String,
   },
   slot: {
-    _default: 'new-dashboard-slot',
+    _description: 'The slot that this dashbaord config renders',
```
denniskigen, line 16:
```suggestion
    _description: 'The slot that this dashboard config renders',
```

denniskigen:
It'd be great if we could also include the updated translations (`en` and maybe some other locale like `fr`) in this PR.

jwnasambu:
@denniskigen and @ibacher am fixing other  locales and will push changes shortly!

ibacher:
@denniskigen We should be managing translations through Transifex rather than PRs, so once we get this merged in, we can add additional languages as a follow-up.

denniskigen:
These changes ought to be made in `esm-core` as well. Clicking the `Offline Actions` menu item in the left nav causes the app to hang at the moment.

ibacher:
Yeah, it's definitely on the list. I wish that wasn't implemented in core... Some more thing to fix in the chart first. Sorry that this has been a really sloppy implementation job.

PR #1225, opened 2023-06-21
(fix) App bar should display workspace header not patient's name in tablet

Jexsie:
## Requirements

## Summary

- Updates the z-index to `8000` which is so far the highest, neglecting the notifications.
- Also fixes the small space between the lower navbar and the open workspace that appears in tablet.

### DIFF
--- a/packages/esm-patient-chart-app/src/workspace/workspace-window.scss
+++ b/packages/esm-patient-chart-app/src/workspace/workspace-window.scss
@@ -92,7 +92,7 @@
     top: 0;
     left: 0;
     right: 0;
-    z-index: 1000;
+    z-index: 8000;
   }
 
   .header {
@@ -115,7 +115,7 @@
 
   .fixed {
     top: 3rem;
-    bottom: spacing.$spacing-10;
+    bottom: 3.96rem;
     background-color: $ui-01;
   }
 


packages/esm-patient-chart-app/src/workspace/workspace-window.scss
```diff
@@ -115,7 +115,7 @@
 
   .fixed {
     top: 3rem;
-    bottom: spacing.$spacing-10;
+    bottom: 3.96rem;
```
ibacher, line 118: Is there a variable we could use here instead? Why `3.96rem`? Why not `4rem`?

packages/esm-patient-chart-app/src/workspace/workspace-window.scss
```diff
@@ -92,7 +92,7 @@
     top: 0;
     left: 0;
     right: 0;
-    z-index: 1000;
+    z-index: 8000;
```
ibacher, line 95: I think we should adjust a competing z-index down, rather than just pushing this number very high, especially in a global override...

Jexsie: According to what I've realized so far, the app bar was set to a z-index of `8000` allover. So this is what am trying to follow along with.  And for the case of this, the workspace header wasn't showing up 'cause it was blocked by the patient header

packages/esm-patient-chart-app/src/workspace/workspace-window.scss
denniskigen: It's probably better to adjust this value to `4rem` and tweak the nav height accordingly rather than use decimal values like `3.96rem` that are harder to reason about.

packages/esm-patient-chart-app/src/patient-chart/action-menu/action-menu.scss
```diff
@@ -62,6 +62,12 @@ $actionPanelOffset: 3rem;
     width: 100%;
     display: flex;
     justify-content: stretch;
+
+    & div {
+      & button {
+        height: spacing.$spacing-10;
```
vasharma05, line 68: This is for which component on the UI?

packages/esm-patient-chart-app/src/workspace/workspace-window.scss
vasharma05: Make it 8001 instead, I believe

packages/esm-patient-chart-app/src/workspace/workspace-window.scss
```diff
@@ -115,7 +115,7 @@
 
   .fixed {
     top: 3rem;
-    bottom: spacing.$spacing-10;
+    bottom: calc(spacing.$spacing-10 + 1px); // plus the top border
```
vasharma05, line 118: Does using the below property avoid the +1px?
Please try and let me know
Thanks!
```
box-sizing: border-box;
```

packages/esm-patient-chart-app/src/patient-chart/action-menu/action-menu.scss
Jexsie: the bottom navs

packages/esm-patient-chart-app/src/workspace/workspace-window.scss
Jexsie: tried this, but the top border is not shown when a workspace is opened

packages/esm-patient-chart-app/src/workspace/workspace-window.scss
Jexsie: looks good on `8000`
image

vasharma05 (approved):

PR #1227, opened 2023-06-21
(feat) Dashboard Extension uses the `moduleName` as the namespace for translation

vasharma05:
## Requirements

## Summary
This PR introduces improvement by using the moduleName as the namespace for `useTranslation` hook. This will also help in any re-renders by `useTranslation` hook inside the `DashboardExtension` which was not possible earlier  by changing the prop of an extension.

### DIFF
--- a/packages/esm-patient-allergies-app/src/dashboard.meta.ts
+++ b/packages/esm-patient-allergies-app/src/dashboard.meta.ts
@@ -2,4 +2,5 @@ export const dashboardMeta = {
   slot: 'patient-chart-allergies-dashboard-slot',
   columns: 1,
   path: 'Allergies',
+  title: 'Allergies',
 };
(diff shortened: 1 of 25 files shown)


ibacher:
A better option would actually be to wrap the component in the `openmrsComponentDecorator()`.

packages/esm-patient-allergies-app/src/dashboard.meta.ts
```diff
@@ -1,5 +1,5 @@
 export const dashboardMeta = {
   slot: 'patient-chart-allergies-dashboard-slot',
   columns: 1,
-  path: 'Allergies',
+  title: 'Allergies',
```
ibacher, line 4: So, I get that this will work, but I actually had a separate motivation in adding a `path` property, specifically, I actually think it would be more consistent to allow the URL to vary regardless of the title. `Vitals & Biometrics` is great display text, but gets a bit weird as soon as you start realising how much URL encoding and decoding we need to do and is not obviously more useful in that format than, e.g., `vitals-and-biometrics` as a URL slug.

vasharma05:
@denniskigen, please review the PR whenever you get some time
Thanks!

hadijahkyampeire (approved):
This looks good to me

PR #1269, opened 2023-07-10
(fix) O3-1895 Edit Past Visit button does not work

brandones:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.
- [ ] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://zeroheight.com/23a080e38/p/880723-introduction).
- [x] My work includes tests or is validated by existing tests.

## Summary

Fixes the "edit past visit" functionality. Modifies the active visit tag so you can tell. Fixes active visit tag tooltip styling.

It's split into two commits, the change and a refactor. Review separately for pleasant diffs.

Note that this is not quite right—the Retrospective Entry tag should appear also after clicking "Start visit" and entering a visit in the past. But this requirement maybe needs to be specified a bit better. What if it's a multi-day visit started the previous day, which has not ended yet? Getting the logic right for identifying a retrospective visit is left to [O3-2242](https://issues.openmrs.org/browse/O3-2242).

## Screenshots

Peek 2023-07-07 23-38

Old:

New:

## Related Issue

https://issues.openmrs.org/browse/O3-1895

## Other

Depends on https://github.com/openmrs/openmrs-esm-core/pull/724

PR #1271, opened 2023-07-10
(chore) Cache artefacts from lint checks more reliably

denniskigen:
## Requirements

- [x] This PR has a title that describes the work done, including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.

## Summary

Fixes an issue where the artefacts from linting weren't getting cached correctly by turbo, leading to some lint issues peeling not getting picked up, such as in https://github.com/openmrs/openmrs-esm-patient-chart/pull/1269/commits/0e752ed16e258b4871460413a6886487be81df76. Prefixing the lint script with TIMING=1 just adds [rule performance profiling](https://eslint.org/docs/latest/extend/custom-rules#profile-rule-performance) and should not affect how lint rules are run. Adding that flag to the command gives turbo artefacts to cache and compare future runs against. This means it can more reliably determine what changed.

This PR also removes some unused properties from the turbo config file.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,4 @@
 src/**/*.test.tsx
 src/**/*.spec.tsx
-**/*.d.tsx
+**/*.d.ts
 **/node_modules/**/*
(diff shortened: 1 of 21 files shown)


ibacher (approved):
It might be good to integrate `cross-env` so this also works on Windows, but LGTM.

PR #1283, opened 2023-07-19
(feat) Put `Patient Flags` feature behind a feature toggle

denniskigen:
## Requirements

## Summary

The Patient Flags feature is currently still a work in progress and isn't ready for production. It makes sense to put it behind a feature flag. This PR adds a feature flag called `patientFlags` and wraps the feature behind it.

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-chart/assets/8509731/a10c8385-d9d1-4a5b-97a9-033ed4ae99a6

## Related Issue

https://issues.openmrs.org/browse/O3-2279

## Other
*None*

### DIFF
--- a/packages/esm-patient-flags-app/src/index.ts
+++ b/packages/esm-patient-flags-app/src/index.ts
@@ -1,4 +1,4 @@
-import { defineConfigSchema, getAsyncLifecycle } from '@openmrs/esm-framework';
+import { defineConfigSchema, getAsyncLifecycle, registerFeatureFlag } from '@openmrs/esm-framework';
 import { configSchema } from './config-schema';
 
 const moduleName = '@openmrs/esm-patient-flags-app';
@@ -7,6 +7,12 @@ export const importTranslation = require.context('../translations', false, /.jso
 
 export function startupApp() {
   defineConfigSchema(moduleName, configSchema);
+
+  registerFeatureFlag(
+    'patientFlags',
+    'Patient Flags',
+    'Visual components that enable healthcare providers to see relevant patient information with a glance in the Patient chart. Flags are displayed in the Patient Summary, just below the patient banner, and can link users to other areas of the chart to perform relevant actions during a visit.',
+  );
 }
 
 export const flagTags = getAsyncLifecycle(() => import('./flags/flags-highlight-bar.component'), {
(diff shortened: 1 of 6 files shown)


packages/esm-patient-flags-app/src/index.ts
```diff
@@ -7,6 +7,12 @@ export const importTranslation = require.context('../translations', false, /.jso
 
 export function startupApp() {
   defineConfigSchema(moduleName, configSchema);
+
+  registerFeatureFlag(
+    'patientFlags',
+    'Patient Flags',
+    'Visual components that enable healthcare providers to see relevant patient information with a glance in the Patient chart. Flags are displayed in the Patient Summary, just below the patient banner, and can link users to other areas of the chart to perform relevant actions during a visit.',
```
brandones, line 14: Beautiful description, thank you :clap:

brandones:
Why are we making changes within `FlagsHighlightBar`, `FlagsList`, and `Flags`? Why not just keep those components out of the tree in the first place and call that good?

denniskigen:
No need to call `useFeatureFlag` at any point?

brandones:
No need to call useFeatureFlag at any point?

Yeah, right? Like we just prevent the extension from mounting at all and we're good.

> Maybe I'm missing something in my local setup?

What? Oh, do you mean that only modifying the extension definition doesn't change it locally for you? Is it because you need to coerce your dev server into reloading the routes somehow?

denniskigen:
> What? Oh, do you mean that only modifying the extension definition doesn't change it locally for you? Is it because you need to coerce your dev server into reloading the routes somehow?

That's it! I've just checked the routes registry file and noticed that the `featureFlag` property isn't getting picked up regardless of how many times I reload. For context, I'm running a dev server off esm-core (via `yarn run:shell`) and then adding an import map override for the flags app (via the import map overrides panel). The approach seems sound despite the fact I can't test locally. So I've updated the PR to remove the use `useFeatureFlag` hook. Rendering the extensions is now dependent on the `featureFlag` property in the routes registry.

brandones (approved):
Thanks Dennis. I'm good approving this as is, but heads up to @ibacher about the difficulties with the dev server and routes—I think this challenge is an already familiar one but just want to make sure.

denniskigen:
Thanks for the suggestions and the reviews!

PR #1286, opened 2023-07-21
(feat) Add an extension slot for additional visit summary items

ibacher:
NB All the other tabs will disable themselves if there is no data to display. These extensions currently don't support that functionality.

## Requirements

## Summary

Adds an extension slot to the Visit Summary component to allow additional panels to be rendered.

I didn't convert the existing tabs into panels because I didn't have a good strategy to support the `disabled` property for the existing panels (which hides them when they don't have any contents).

### DIFF
--- a/packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/visit-summary.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/visit-summary.component.tsx
@@ -1,14 +1,32 @@
 import React, { useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Tab, Tabs, TabList, TabPanel, TabPanels, Tag } from '@carbon/react';
-import { formatTime, OpenmrsResource, parseDate, useConfig, useLayoutType, Visit } from '@openmrs/esm-framework';
-import { Order, Encounter, Note, Observation, OrderItem, Diagnosis, mapEncounters } from '../visit.resource';
+import {
+  type AssignedExtension,
+  Extension,
+  ExtensionSlot,
+  formatTime,
+  parseDate,
+  useConfig,
+  useConnectedExtensions,
+  useLayoutType,
+  type Visit,
+} from '@openmrs/esm-framework';
+import {
+  type Order,
+  type Encounter,
+  type Note,
+  type Observation,
+  type OrderItem,
+  type Diagnosis,
+  mapEncounters,
+} from '../visit.resource';
 import VisitsTable from './visits-table/visits-table.component';
 import MedicationSummary from './medications-summary.component';
 import NotesSummary from './notes-summary.component';
 import TestsSummary from './tests-summary.component';
+import type { ExternalOverviewProps } from '@openmrs/esm-patient-common-lib';
 import styles from './visit-summary.scss';
-import { ExternalOverviewProps } from '@openmrs/esm-patient-common-lib';
 
 interface DiagnosisItem {
   diagnosis: string;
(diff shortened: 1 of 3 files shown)


packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/visit-summary.component.tsx
```diff
@@ -128,6 +149,14 @@ const VisitSummary: React.FC<VisitSummaryProps> = ({ visit, patientUuid }) => {
           >
             {t('encounters_title', 'Encounters')}
           </Tab>
+          {extensions.map((extension, index) => (
+            <Tab key={index} className={styles.tab} id={`${extension.meta.title || index}-tab`}>
```
ibacher, line 153: @brandones Is there some obvious way I can replicate the `disabled` property of the other tabs? (Obviously `config.disableEmptyTabs` is easy to use; but can the extension some how communicate "I have no content").

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/visit-summary.component.tsx
```diff
@@ -142,6 +171,11 @@ const VisitSummary: React.FC<VisitSummaryProps> = ({ visit, patientUuid }) => {
           <TabPanel>
             <VisitsTable visits={mapEncounters(visit)} showAllEncounters={false} patientUuid={patientUuid} />
           </TabPanel>
+          <ExtensionSlot name={visitSummaryPanelSlot}>
+            <TabPanel>
```
ibacher, line 175: I forgot that this worked... it's an extremely cool feature!

denniskigen (approved):
LGTM

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/visit-summary.component.tsx
brandones: Ok, here are three ways we might accomplish that. None particularly pleasant, but I think 1 or 3 are both reasonable. 3 is probably the best.

1. **Crazy meta API.** Let these extensions define a meta which is a string that gets `eval`ed by this component to determine whether the tab should render or not. So like `"meta": "visit?.encounters.length > 0"`. This would expand the contract between this component and these extensions to include what are the variables that form the context for this evaluation. This contract would have to be stable and documented well.

2. **Two extensions per tab.** You could make the tabs their own extensions. They can then have whatever complex render logic you want. The tabs go in a slot, and they refer to panels that are different extensions in a different slot. If you add the panel as an extension but not the corresponding tab, or vice-versa, tough luck. I dislike this option; it is a bad implementer experience. You could do something clever such that they detect each others' existence and complain loudly if they don't find each other, but still. Not awesome.

3. **Portal magic.** I'm not 100% sure this would work but I don't see why it wouldn't. The extensions essentially become the tabs, and the slot contains the tabs. But you get a reference to a DOM node inside `<TabPanels>` and pass that as state through the slot into the extensions. Each extension mounts its panel component to the DOM node using [createPortal](https://react.dev/reference/react-dom/createPortal). Maybe it would have to be multiple DOM nodes? And maybe you'd have to do something to ensure the order is guaranteed? Maybe it would be better to have the panel be the actual extension, since that's a more reasonable thing to just put somewhere, while the tab would go to a portal if the slot provides a DOM node to attach it to?

ibacher: Thanks! That's really insightful! I think I'm actually going to go a slightly simpler route and try to pass a callback to the extension through the extensions `state` property, though. That seems more straight-forward to me, but not an idea I would've hit on without thinking through these other options. (I like the crazy meta API; I've actually contemplated doing something similar with the "Display properties" configuration, to give implementations a way of providing custom logic, but figuring out what values should be in the evaluation context and how we get them is not straight-forward).

brandones: Very cool! Looking forward to seeing the code for that approach.

PR #1287, opened 2023-07-21
(fix) resolves valid route with grouped chart dashboards

pirupius:
## Requirements

## Summary
When the patient chart has only folders, the combined array of ungrouped and grouped dashboards has an empty path on index 0 which doesn't resolve. This PR adds a filter to remove dashboards without the `path` key to ensure the index 0 is valid

### DIFF
--- a/packages/esm-patient-chart-app/src/patient-chart/chart-review/chart-review.component.tsx
+++ b/packages/esm-patient-chart-app/src/patient-chart/chart-review/chart-review.component.tsx
@@ -46,8 +46,11 @@ const ChartReview: React.FC<ChartReviewProps> = ({ patientUuid, patient, view })
     )
     .flat();
   const dashboards = ungroupedDashboards.concat(groupedDashboards) as Array<DashboardConfig>;
+  const filteredDashboards = dashboards.filter((dashboard) => {
+    if ('path' in dashboard) return dashboard;
+  });
 
-  const defaultDashboard = dashboards[0];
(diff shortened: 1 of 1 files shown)


packages/esm-patient-chart-app/src/patient-chart/chart-review/chart-review.component.tsx
```diff
@@ -46,8 +46,11 @@ const ChartReview: React.FC<ChartReviewProps> = ({ patientUuid, patient, view })
     )
     .flat();
   const dashboards = ungroupedDashboards.concat(groupedDashboards) as Array<DashboardConfig>;
+  const filteredDashboards = dashboards.filter((dashboard) => {
+    if ('path' in dashboard) return dashboard;
+  });
```
samuelmale, line 51:
```suggestion
  const filteredDashboards = dashboards.filter(dashboard => 'path' in dashboard);

```

Question: @pirupius Shouldn't our filter only give a pass for dashboards with a valid path? Like a truthy one?

```js
 const filteredDashboards = dashboards.filter(dashboard => dashboard.path);
```

I think we should also log to console all excluded dashboards and flag them as missing a `path` property

pirupius: The dashboards without paths are mostly flags and other components that necessarily don't appear in the sidenav so there's no need to add a console log

samuelmale (approved):

PR #1298, opened 2023-07-26
O3-2242: 'Retrospective Entry' label on Chart Header when doing RDE

brandones:
## Requirements

## Summary
Puts a label in the top bar when in retrospective entry mode. Makes the patient banner visit tag only reflect whether there is an ongoing visit, so it has nothing to do with RDE.

## Screenshots
Screencast from 2023-07-25 20-09-28

## Related Issue
https://issues.openmrs.org/browse/O3-2242

## Other
Pending https://github.com/openmrs/openmrs-esm-core/pull/738


brandones:
@ibacher @denniskigen ?

packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
```diff
@@ -91,17 +82,21 @@ const PatientInfo: React.FC<PatientInfoProps> = ({ patient }) => {
...
+      {patientNameIsTooLong ? (
+        <Tooltip
+          align="bottom-left"
+          width={100}
+          label={
+            <>
+              <p className={styles.tooltipPatientName}>{name}</p>
+              <p className={styles.tooltipPatientInfo}>{`${parseInt(age(patient?.birthDate))}, ${getGender(
+                patient?.gender,
+              )}`}</p>
```
ibacher, line 96: I'm a bit unclear on why this extra information is included here but not, say, as a tooltip on the patient name if it's 25 characters or less.

ibacher (approved):
So, I think we'd still want a retrospective entry tag in addition to the header display. I know that kind of duplicates information on the screen, but it's quite important context (since a user might get used to just looking for the Active Visit tab).

packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
brandones: Yeah that is goofy. I'll fix it. Oh yeah sorry I forgot to mention this PR includes a bunch of random refactoring.

brandones:
> So, I think we'd still want a retrospective entry tag in addition to the header display. I know that kind of duplicates information on the screen, but it's quite important context (since a user might get used to just looking for the Active Visit tab).

A lot is in the air on the design side right now. For now I'm gonna merge this in, and then we'll see where the design lands.

packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
brandones: Actually looking at it now, I think it makes sense. My only concern is that the patient name by default looks kinda clickable, when it is actually only clickable if it is a very long name.

image

image

PR #1299, opened 2023-07-26
(fix) Restore configurable logo to print preview

denniskigen:
## Requirements

## Summary

This PR restores the ability to configure the logo shown when the user clicks the print button on the Vitals widget. 

## Screenshots
*None*

## Related Issue
*None*

## Other
*None*


ibacher:
Don't we already have the logo configured somewhere? Wouldn't it be best to just use that configuration?

packages/esm-patient-vitals-app/src/vitals/print/print.component.tsx
```diff
@@ -16,18 +16,28 @@ interface PrintComponentProps {
...
...3.407 7.196 0 12.096 4.906 12.096 13.354v1.362h-20.597c.325 4.413 3.704 8.173 8.335 8.173 3.65 0 6.105-1.307 8.12-3.868l3.219 2.508ZM227.854 59.356c0-2.346-.216-4.36-.216-6.05h4.031c0 1.363.11 2.777.11 4.195h.11c1.144-2.505 4.306-4.85 8.5-4.85 6.705 0 9.699 4.252 9.699 10.41v15.748h-4.248v-15.31c0-4.253-1.856-6.924-5.833-6.924-5.503 0-7.903 3.979-7.903 9.811V78.81h-4.25V59.356ZM259.211 41.008h6.708L278.8 70.791h.107l12.982-29.782h6.549v37.99h-4.506V47.124h-.106L280.192 79h-2.738l-13.629-31.875h-.107V79h-4.507V41.01ZM312.392 57.752h4.023c4.992 0 11.487 0 11.487-6.282 0-5.47-4.776-6.276-9.177-6.276h-6.333v12.558Zm-4.506-16.744h9.711c7.352 0 15.132 1.072 15.132 10.462 0 5.527-3.594 9.125-9.495 10.037L334.018 79h-5.525l-10.304-17.063h-5.797V79h-4.506V41.01ZM358.123 47.712c-1.506-2.413-4.187-3.486-6.926-3.486-3.973 0-8.1 1.88-8.1 6.385 0 3.49 1.931 5.047 7.994 6.98 5.903 1.878 11.377 3.809 11.377 11.267 0 7.567-6.495 11.11-13.36 11.11-4.402 0-9.125-1.45-11.7-5.262l3.862-3.165c1.61 2.794 4.83 4.24 8.105 4.24 3.862 0 8.263-2.253 8.263-6.601 0-4.669-3.165-5.474-9.928-7.728-5.366-1.771-9.442-4.134-9.442-10.463 0-7.298 6.277-10.945 12.929-10.945 4.241 0 7.836 1.178 10.625 4.45l-3.699 3.218Z"
```
denniskigen, line 37: It looks like `xlinkHref` / `href` doesn't work in the print context, otherwise, I'd have preferred to use something like [this](https://github.com/openmrs/openmrs-esm-core/blob/8c423dab6b2f3dabc07f7b2178f70f287769240f/packages/apps/esm-login-app/src/login/login.component.tsx#L159C7-L159C7).

Sidenote: `xlinkHref` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href) and ought be replaced with `href`.

ibacher: It looks like react-to-print works by creating an iframe and copying any styles or fonts into that iframe, so within that context, the `id` doesn't resolve. The iframe, however, has a fixed id (`printWindow`) and there is a `onBeforePrint()` function, so it's at least theoretically possible to use that hook to copy the SVG DOM node to the iframe, which might do the trick?

That said, that sounds a little finicky.

denniskigen: Hey @ibacher, should we still try to improve this as you suggested above?

ibacher: No, I think it's fine to defer both changes I requested for later (apparently we still don't have a version of `useConfig()` that takes a module name?).

ibacher: My real concern here is that we're going to have to define the same "logo" config object for every printable table, and we should probably just use whatever is configured in primary navigation.

PR #1300, opened 2023-07-31
(feat) Form entry app to use new API endpoint

ibacher:
## Requirements

## Summary

This PR replaces a huge chunk of form pre-processing done on the frontend with a call to a backend endpoint that supplies the same data.

Specifically:

* We no longer need a compile step with referenced forms; referenced pages, sections and questions will all be substituted before the schema is sent.
* We no longer need to separately load translations, as translations should be provided inside the form schema
* We no longer need to load unreferenced concepts; the backend will return a `conceptReferences` map that collects every concept used in the form and (by default) returns the UUID and localised name for it.
* This PR removes the localStorage caching mechanism altogether. At least in my local testing, even complex forms with multiple component forms, translations, and concepts are returned in < 1 second.

There are two goals for this work:
1. It should not be necessary to cache the form locally (except for offline usage purposes), ensuring that user's always see the latest version of a form.
2. It should reduce the number of requests involved in loading a form from (roughly) `4 + n + 4o` (where `n` is the number of unlabelled concepts and `o` is the number of referenced forms) to exactly `1`.

Leaving this as a draft until we have a "stable" version of the O3 module (coming soon).

### DIFF
--- a/packages/esm-form-entry-app/src/app/fe-wrapper/fe-wrapper.component.spec.ts
+++ b/packages/esm-form-entry-app/src/app/fe-wrapper/fe-wrapper.component.spec.ts
@@ -6,13 +6,11 @@ import { FeWrapperComponent } from './fe-wrapper.component';
 import { FormEntryModule } from '@openmrs/ngx-formentry';
 import { ReactiveFormsModule } from '@angular/forms';
 import { OpenmrsEsmApiService } from '../openmrs-api/openmrs-esm-api.service';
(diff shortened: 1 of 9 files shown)


ibacher:
@enyachoke Was the form metadata used for anything other than loading additional form resources?

vasharma05 (approved):
LGTM!

enyachoke (approved):

PR #1320, opened 2023-08-18
Added support for Dates in fr

icrc-loliveira:
## Requirements

### DIFF
--- a/packages/esm-form-entry-app/src/app/app.module.ts
+++ b/packages/esm-form-entry-app/src/app/app.module.ts
@@ -19,6 +19,9 @@ import { SingleSpaPropsService } from './single-spa-props/single-spa-props.servi
 import { FormCreationService } from './form-creation/form-creation.service';
 import { TranslateModule, TranslateService, TranslateStore } from '@ngx-translate/core';
 import { ProgramResourceService } from './openmrs-api/program-resource.service';
+import { registerLocaleData } from '@angular/common';
+import localeFr from '@angular/common/locales/fr';
+registerLocaleData(localeFr);
 
 @NgModule({
   declarations: [AppComponent, EmptyRouteComponent, FeWrapperComponent, LoaderComponent],


gracepotma:
❤️ bravo and thank you @icrc-loliveira!

delphinepas:
@ibacher could you please review this PR as it is a blocking issue for some users in the field. many thanks cc @icrc-loliveira @icrc-fdeniger

packages/esm-form-entry-app/src/app/app.module.ts
```diff
@@ -19,6 +19,9 @@ import { SingleSpaPropsService } from './single-spa-props/single-spa-props.servi
 import { FormCreationService } from './form-creation/form-creation.service';
 import { TranslateModule, TranslateService, TranslateStore } from '@ngx-translate/core';
 import { ProgramResourceService } from './openmrs-api/program-resource.service';
+import { registerLocaleData } from '@angular/common';
+import localeFr from '@angular/common/locales/fr';
+registerLocaleData(localeFr);
```
ibacher, line 24: So, this unconditionally loads the French locale as part of the application. I don't think we can accept this into the community version for that reason. Inter alia, for instances not using French and where French is not a display language, we wouldn't want to load this. Also, for locations which are not in French but want a different date format, this won't work.

Instead, we could support this by doing something similar around [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/02e38dbf3a6060e3ed795735b35917e0184b8970/packages/esm-form-entry-app/src/app/fe-wrapper/fe-wrapper.component.ts#L119C1-L119C1) (the point where we register the form language). Basically, we should be able to do something like:

```js
import(`@angular/common/locales/${language}`).then(registerLocaleData).catch(console.error);
```

The above might need some modifications to work with the RxJS chains.

packages/esm-form-entry-app/src/app/fe-wrapper/fe-wrapper.component.ts
```diff
@@ -116,12 +117,19 @@ export class FeWrapperComponent implements OnInit, OnDestroy {
...
     const identifiers = this.formDataSourceService.getPatientObject(patient)?.identifiers ?? [];
-    const language = window.i18next?.language?.substring(0, 2) ?? '';
-    this.translateService.addLangs([language]);
-    this.translateService.use(language);
+    const locale = window.i18next?.language?.substring(0, 2) ?? '';
+    this.translateService.addLangs([locale]);
+    this.translateService.use(locale);
+
+    import(
+      /* webpackInclude: /(en|fr)\.js$/ */
```
ibacher, line 125:
```suggestion
      /* webpackInclude: /\.js$/ */
```

ibacher (approved):

PR #1322, opened 2023-08-20
(fix) O3-2158: Clear order basket search input and trap focus when search gets reset.

NYESIGA:
fix)O3-2158:Clear order basket search input and trap focus when search gets reset. i Declare a ref in the Drug Search component that tracks the Search input using the useRef hook and passed it down in Drug Search Result component and Update the logic in it to to accept the ref as a prop. Create a boolean state variable in the OrderBasketSearchResults component that tracks whether the search input was cleared and set to false and create function onClick handler's logic so that it sets the value of the state variable i created above to true. and i added useEffect hook in OrderBasketSearchResults component that traps focus to the Search input if the boolean state variable is true. After trapping focus, it set the state variable back to false.

https://github.com/openmrs/openmrs-esm-patient-chart/assets/110772072/a3b63172-a453-4af6-8161-bfdb19edb750

https://issues.openmrs.org/browse/O3-2158

### DIFF
--- a/packages/esm-patient-medications-app/src/add-drug-order/drug-search/drug-search.component.tsx
+++ b/packages/esm-patient-medications-app/src/add-drug-order/drug-search/drug-search.component.tsx
@@ -1,4 +1,4 @@
-import React, { useRef,useState } from 'react';
+import React, { useRef, useState } from 'react';
 import debounce from 'lodash-es/debounce';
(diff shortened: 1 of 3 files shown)


NYESIGA:
https://issues.openmrs.org/browse/O3-2158

vasharma05 (approved):

PR #1325, opened 2023-08-23
(fix) O3-1676 & O3-1677: Workspace implementation improved as per design

vasharma05:
## Requirements

## Summary
This PR brings the following changes to the workspaces:
### Workspace Variants
The registered workspaces can be registered with a variant keyword, which can be of 3 types:
#### Independent workspace (variant = "independent")
This variant denotes to the set of workspaces which are not collapsible and has a `X` button on the top right corner.
image

image

#### Siderail workspace (variant = "siderail")
These workspaces are allowed to callapse and marked as active on the corresponding action button on the side rail.
image

image

#### Clinical forms (variant = "clinical-form")
These workspaces are also allowed to collapse, but also can be maximised to full width and have a total width of 520px, as compared to the above 2 variants, which are 420px in width.
image
image

### Workspace width
The workspaces are now confined to 420px/ 520px, depending on the variant. The workspace width will remain fixed as the desktop size changes.

Reference: https://zeroheight.com/23a080e38/p/483a22-workspace

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-chart/assets/51502471/c08bc38c-5204-42af-ab6a-0a1487bcf01a

## Related Issue
https://issues.openmrs.org/browse/O3-1676
https://issues.openmrs.org/browse/O3-1677

### DIFF
--- a/packages/esm-patient-chart-app/src/workspace/workspace-window.scss
+++ b/packages/esm-patient-chart-app/src/workspace/workspace-window.scss
@@ -4,6 +4,8 @@
 @import "../root.scss";
 
 .header {
+  box-sizing: content-box;
+  border-bottom: 1px solid $text-03;
   background-color: $ui-03;
   top: 3rem;
   z-index: 100;
@@ -45,9 +47,6 @@
   width: calc(100% - 3rem);
 }
 
-.dynamicWidth {
-  width: 32.5rem;
-}
 
 /* Desktop */
 :global(.omrs-breakpoint-gt-tablet) {
@@ -61,26 +60,32 @@
   }
 
   .container {
-    margin-right: 2.825rem;
-    width: 32.5rem;
+    margin-right: 3rem;
     height: 100vh;
   }
 
-  .headerGlobalBar {
-    &>button {
-      min-height: spacing.$spacing-08;
-      background-color: $ui-02;
-      border-color: $color-gray-30;
-      border-style: solid;
-      border-width: 0 1px 0 0;
+  .clinicalForm {
+    width: 32.25rem;
+    .dynamicWidth {
+      width: 32.25rem;
     }
   }
-}
 
-/* Small desktop */
-:global(.omrs-breakpoint-lt-large-desktop){
-  .dynamicWidth {
-    width: 32.5rem;
+  .independent {
+    width: 26.25rem;
+    .dynamicWidth {
+      width: 26.25rem;
+    }
+  }
+
+  .headerGlobalBar {
+    button {
+      background-color: $ui-02;
+      border-right: 1px solid $ui-03;
+    }
+    button:last-child {
+      border-right: none;
+    }
   }
 }
 
@@ -128,7 +133,7 @@
 html[dir='rtl'] {
   :global(.omrs-breakpoint-gt-tablet) {
     .container {
-      margin-left: 2.825rem;
+      margin-left: 3rem;
       margin-right: unset;
     }
   }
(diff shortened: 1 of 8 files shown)


hadijahkyampeire (approved):
LGTM, thanks @vasharma05

brandones:
`variant` is a very general name for something that has very specific behavior. Semantically it seems redundant with "type." And you did not document it anywhere (the type declaration would have been the place). I would much rather you have added two booleans, `canMaximize` and `canHide`, and then you only show the close button if `canHide` is `false`.

brandones:
@vasharma05 This cannot possibly be the correct window width.

image

vasharma05:
> variant is a very general name for something that has very specific behavior. Semantically it seems redundant with "type." And you did not document it anywhere (the type declaration would have been the place). I would much rather you have added two booleans, canMaximize and canHide, and then you only show the close button if canHide is false.

Actually I tried to go via the available documentation itself. 
Documentation available here: https://zeroheight.com/23a080e38/p/483a22-workspace

vasharma05:
If in the long run, the variant comes out to be an issue, I'll be happy to shift to your idea of 2 booleans.

vasharma05:
For the order basket form, I'm going to ask Paul and Ciaran for the same.
Thanks for pointing it out.
Thanks!

packages/esm-patient-chart-app/src/workspace/workspace-window.scss
```diff
@@ -61,26 +60,32 @@
...
+    .dynamicWidth {
+      width: 32.25rem;
     }
   }
-}
 
-/* Small desktop */
-:global(.omrs-breakpoint-lt-large-desktop){
-  .dynamicWidth {
-    width: calc(50% - 1.625rem);
```
pirupius, line 83: Hey @vasharma05 are we able to revert the width size? the workspace has become really small to display form content correctly. 

>On desktop screens, the workspace sits on the right side of the page. Its width is 420px in most cases, and 520px if it's a clinical form. This width still allows actions to be performed with full visibility of information in the chart.

In the case that we're to follow the documentation, let's change the 26.25rem that represents the lower point of 420px, in favor of the upper limit of 520px (~32.5rem) to give more space

brandones:
@vasharma05 Okay, let's look at what's happened. These are the designs:

image

This is what has been implemented:

image

It looks like you're right about the overall width, and the problem is that some fields are not sized correctly to display well. And the little "x" in the Dose Unit dropdown is probably superfluous and not helping either. Let's please fix the form fields so that the form is usable again.

brandones:
@pirupius What form are you having trouble with? Could you send a screenshot or something?

packages/esm-patient-chart-app/src/workspace/workspace-window.scss
vasharma05: Variant: `clinical-form` will provide you with the width of 520px as I mentioned above as well.

You can provide the variant in either the `meta` of the extension registration, directly in the `registerWorkspace`. You can see the changed files for reference.
Thanks!

vasharma05:
> It looks like you're right about the overall width, and the problem is that some fields are not sized correctly to display well. And the little "x" in the Dose Unit dropdown is probably superfluous and not helping either. Let's please fix the form fields so that the form is usable again.

Sure @brandones !
I'll create a ticket and fix this.
Thanks for the deep dive!

pirupius:
@brandones only noticed it with the medication forms.

vasharma05:
Hi @brandones @pirupius , please review the improvement to this PR: https://github.com/openmrs/openmrs-esm-patient-chart/pull/1333

PR #1327, opened 2023-08-24
(fix) Common component for siderail components and global state for managing workspace nav buttons 

vasharma05:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.
- [x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://zeroheight.com/23a080e38/p/880723-introduction).
- [x] My work includes tests or is validated by existing tests.

## Summary
1. Common component in the common-lib for generating side components
2. Global state for workspace nav buttons in the sidenav, leveraging upon the workspace global store

## Screenshots
Will be uploaded soon

PR #1335, opened 2023-08-28
(fix) Fixed the workspace type for siderail nav buttons

vasharma05:
## Requirements

## Summary
This PR fixes workspace type for the siderail nav buttons.

Comment from @brandones: https://github.com/openmrs/openmrs-esm-patient-chart/pull/1327#discussion_r1305702753

## Screenshots
None

## Related Issue
None


packages/esm-patient-common-lib/src/workspaces/workspaces.ts
```diff
@@ -33,7 +33,7 @@ export interface WorkspaceRegistration {
   /** Use `getLifecycle` or `getAsyncLifecycle` to get the value of `load` */
   load(): Promise<any>;
   /** Only one of each "type" of workspace is allowed to be open at a time. The default is "form" */
-  type?: string;
+  type?: 'clinical-form' | 'visit-note' | 'order' | string;
```
ibacher, line 36: Does having specific types help with anything here?

brandones: It should either have the specific constants or `string`, not both. I originally made it "string" so that if third-party modules wanted to add their own workspace types and buttons to the siderail they could do that. Limiting the allowed values of "type" might help us catch invalid type errors in our code a little bit easier, but would make it harder for third-party modules using Typescript to have their own workspace types. So this is a judgement call based on whether we want to support third party modules being able to do that. I suspect that we do.

vasharma05: I have reverted the type to just string, I agree, that might have bring in a lot of issues.
Having it string, and making it default to `form` is okay I believe, because in the community version, the main distinction is actually between siderail workspaces and other workspaces.

If we consider the third-party modules, yes it will also conflict there as well.
Thanks for the help @brandones @ibacher !

brandones (approved):
Thanks for this @vasharma05 !

vasharma05:
I understand,  I am on it. 
Thanks a lot for informing!

PR #1339, opened 2023-08-30
(fix) O3-2363: Workspace window state should be part of global workspace state

vasharma05:
## Requirements

## Summary
This PR brings the `workspaceWindowState` context's value in the global `workspace` store. This is done because the context's changes were not reflected in the siderail nav buttons, as the buttons render as extensions.

## Screenshots
No difference in the UI.

## Related Issue
https://issues.openmrs.org/projects/O3/issues/O3-2363


packages/esm-patient-common-lib/src/workspaces/useWorkspaces.tsx
```diff
@@ -30,10 +33,24 @@ export function useWorkspaces(): WorkspacesInfo {
...
-  };
+  useEffect(() => {
+    if (workspaces.length === 0) {
+      updateWorkspaceSizeState('hidden');
+    } else {
+      updateWorkspaceSizeState(workspaces[0].preferredWindowSize === 'maximized' ? 'maximized' : 'normal');
+    }
+  }, [workspaces]);
+
+  const memoisedResults = useMemo(
```
ibacher, line 44: I actually don't know that we need memoisation here. There's nothing very expensive about constructing an object.

packages/esm-patient-common-lib/src/workspaces/useWorkspaces.tsx
```diff
@@ -30,10 +33,24 @@ export function useWorkspaces(): WorkspacesInfo {
     }
   }, [workspaces]);
 
-  return {
-    active: workspaces.length > 0,
-    windowState,
-    workspaces,
-    prompt,
-  };
+  useEffect(() => {
```
ibacher, line 36: This looks like it's intended to replace the `windowState` memo, but it's still left in?

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
```diff
@@ -243,11 +244,22 @@ export function changeWorkspaceContext(patientUuid) {
...
+};
 export function getWorkspaceStore() {
   return getGlobalStore<WorkspaceStoreState>('workspace', initialState);
 }
 
+export function updateWorkspaceSizeState(value: WorkspaceWindowState) {
+  const store = getWorkspaceStore();
+  const state = store.getState();
+  store.setState({ ...state, workspaceWindowState: value });
+}
```
ibacher, line 261:
```suggestion
export function updateWorkspaceSizeState(value: WorkspaceWindowState) {
  getWorkspaceStore().setState((state) => { ...state, workspaceWindowState: value });
}
```

This might be a unistore vs zustand API difference.

ibacher: However, instead of having a separate exported function for this, it might make sense in `useWorkspaces()` to something like:

```ts
const { prompt, openWorkspaces: workspaces, workspaceWindowSize, updateWorkspaceSize  } = useStoreWithActions(getWorkspaceStore(), { updateWorkspaceSize: (state, value} => ({ ...state, workspaceWindowState: value) })})
```

ibacher (commented):
Thanks Vineet! Looks good, but a few suggestions.

packages/esm-patient-common-lib/src/workspaces/useWorkspaceWindowSize.tsx
```diff
@@ -1,65 +0,0 @@
...
-interface WorkspaceWindowSizeContext {
-  windowSize: WorkspaceWindowSize;
-  updateWindowSize?(value: WorkspaceWindowState): any;
-  active: boolean;
-}
-
-function reducer(state: WorkspaceWindowSize, action: WorkspaceWindowState): WorkspaceWindowSize {
-  switch (action) {
-    case 'minimized':
-    case 'reopened':
```
brandones, line 22: Can you remove `reopened` from the WorkspaceWindowState type?

packages/esm-patient-common-lib/src/workspaces/useWorkspaces.tsx
```diff
@@ -1,22 +1,25 @@
...
-import { Prompt, WorkspaceStoreState } from './workspaces';
+import { Prompt, WorkspaceStoreState, updateWorkspaceSizeState } from './workspaces';
 
 export interface WorkspacesInfo {
   active: boolean;
-  windowState: WorkspaceWindowState;
+  workspaceWindowState: WorkspaceWindowState;
   workspaces: Array<OpenWorkspace>;
   prompt: Prompt;
+  updateWorkspaceSizeState: (val: WorkspaceWindowState) => void;
```
brandones, line 10: Can we pick either `size` or `state` and just be consistent about it? I feel like either workspaceSize or workspaceState is fine.

packages/esm-patient-common-lib/src/workspaces/useWorkspaces.tsx
vasharma05: No, it isn't expensive, but what usually happens that the returned object gets created everytime on re-render of the hook. Hence the reference of the returned object changes on re-render. The values of the object are used as dependency in other hooks and hence can trigger corresponding hooks.

But now that I think, the inner values are memoised, hence the values' reference will not change 🤔 .

packages/esm-patient-common-lib/src/workspaces/useWorkspaces.tsx
vasharma05: Yes, actually I saw [this](https://github.com/openmrs/openmrs-esm-patient-chart/blob/4201b4aa390ead099b09aaebbc0d891b1fa0c112/packages/esm-patient-common-lib/src/workspaces/useWorkspaceWindowSize.tsx#L44) useEffect as well in the WorkspaceWindowSize Context.

brandones (approved):
A couple minor things but generally looks good!

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
brandones: The update function doesn't use anything from the hook, so I think it's fine to leave it as a function globally available. Contrast for example with `useOrderBasket`, where the update function actually uses the `useOrderBasket` arguments.

packages/esm-patient-common-lib/src/workspaces/useWorkspaces.tsx
brandones: Hard to know with these things. I think the usual guidance is to use `useMemo` when A) you are quite sure that not using it will cause a problem or B) after you've noticed a performance problem that would be fixed by `useMemo`.

In this case, since we always use `useWorkspaces` with destructuring, it doesn't matter whether the return object itself changes.

packages/esm-patient-common-lib/src/workspaces/useWorkspaces.tsx
```diff
@@ -30,10 +33,24 @@ export function useWorkspaces(): WorkspacesInfo {
...
+    }
+  }, [workspaces]);
+
+  const memoisedResults = useMemo(
+    () => ({
+      active: workspaces.length > 0,
+      workspaceWindowState,
+      workspaces,
+      prompt,
+      updateWorkspaceSizeState,
```
brandones, line 50: `updateWorkspaceSizeState` is available via import from `workspaces`. Why would you return it here?

brandones (changes requested):

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
brandones: This is also consistent with the rest of the workspace API, which follows this pattern, rather than using actions. I have no strong feelings between the two paradigms (actions vs global functions that act on the store), but I think consistency is good.

packages/esm-patient-common-lib/src/workspaces/useWorkspaces.tsx
ibacher: This is why I was suggesting making it an action. I’m basically happy as long as we have one clear way of doing this instead of two.

packages/esm-patient-common-lib/src/workspaces/useWorkspaces.tsx
vasharma05: Updated to `updateWorkspaceWindowState`, matches with it's value `workspaceWindowState`

packages/esm-patient-common-lib/src/workspaces/useWorkspaces.tsx
vasharma05: I used the `useStoreWithActions`, it helped out well!

vasharma05:
Hi @brandones @ibacher , please re-review.
Thanks!

packages/esm-patient-common-lib/src/workspaces/useWorkspaces.tsx
vasharma05: Actually, my point was to export all the functions (actions) with the state from the `useWorkspaces` only, but was quite doubtful.

@ibacher , your suggestion `useStoreWithActions` really panned out. One issue I found is that the returned action from the hook isn't memoized.

packages/esm-patient-chart-app/src/workspace/workspace-window.component.tsx
```diff
@@ -17,35 +17,16 @@ const WorkspaceWindow: React.FC<ContextWorkspaceParams> = () => {
...
-      setIsWorkspaceWindowOpen(false);
-    }
-  }, [workspaces.length, active, hidden, maximized, normal]);
-
-  useEffect(() => {
-    if (active && hidden) {
-      updateWindowSize('normal');
-    }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [workspaces, active]);
```
vasharma05, line 43: I shifted this block of useEffect in in the `useWorkspaces` only.

packages/esm-patient-common-lib/src/workspaces/useWorkspaceWindowSize.tsx
vasharma05: Done

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
vasharma05: I was thinking of returning the `updateWorkspaceWindowState` from `workspaces.ts` via `useWorkspaces`, which didn't quite fit right.

I implemented it with `useStoreWithActions` for now, so your comment late. Should I revert the changes, or we can go ahead with the same?
Thanks!

brandones: Yeah now we have this fragmented API where most workspace functions are imported from the library, except updateWorkspaceWindowSize which is returned from the hook for no good reason. Please change it back and remove it from the hook.

vasharma05: I understand, pushing the changes

vasharma05: Done!

ibacher (approved):

PR #1278, opened 2023-07-14
(feat) Migrating start visit form to use RHF and Zod

lumuchris256:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.
- [x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://zeroheight.com/23a080e38/p/880723-introduction).
- [x] My work includes tests or is validated by existing tests.

## Summary
Migrated the form to use RHF and Zod

## Screenshots
https://www.loom.com/share/eb8972ae8d1a4879942b7aa954913d82?sid=acfe6f65-34eb-4a48-92b3-5faeaaa580be

## Related Issue
https://issues.openmrs.org/browse/O3-2144

## Other
cc @hadijahkyampeire @denniskigen

PR #1346, opened 2023-09-06
(fix) Fix failing to save required visit attributes in the start visit form

vasharma05:
## Requirements

## Summary
As the PR #1278 was merged, a critical bug was introduced where the user was not able to save the required visit attributes in the start visit form.

This PR fixes the above issue with the following improvements/ fixes:
1. ZOD schema configured based on the config
2. Visit attribute fields now rendered with the RHF's `Controller` component
3. Error handling done by RHF and zod
4. Cleaning previous validation process

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-chart/assets/51502471/eac74670-cd1c-4659-bc7e-abf893ed69c1

## Related Thread
https://openmrs.slack.com/archives/C02UNMKFH8V/p1693986247424839


packages/esm-patient-chart-app/src/visit/visit-form/visit-attribute-type.component.tsx
```diff
@@ -71,13 +60,11 @@ const VisitAttributeTypeFields: React.FC<VisitAttributeTypeFieldsProps> = ({
 };
 
 interface AttributeTypeFieldProps {
+  fieldProps: any;
```
ibacher, line 63: Is there any way we can get a sensible type for this?

packages/esm-patient-chart-app/src/visit/visit-form/visit-attribute-type.component.tsx
```diff
@@ -131,13 +120,12 @@ const AttributeTypeField: React.FC<AttributeTypeFieldProps> = ({
...
+            {...fieldProps}
             labelText={labelText}
-            onChange={(e) => setVisitAttribute(e.target.value)}
-            required={required}
-            invalid={required && isMissingRequiredAttributes && !visitAttributes[uuid]}
+            invalid={!!errors.visitAttributes?.[uuid]}
             invalidText={t('fieldRequired', 'This field is required')}
           >
-            <SelectItem text={t('selectAnOption', 'Select an option')} value={null} />
+            <SelectItem text={t('selectAnOption', 'Select an option')} value={''} />
```
ibacher, line 128: Is this necessary? Seems better that the `selectAnOption` option isn't usable.

vasharma05: Actually, since the value is null, when "Select an option" is selected, instead of setting the value in the state as null, it saves "Select an option", which is truthy.

The user can select it again, if that is what you are afraid of

packages/esm-patient-chart-app/src/visit/visit-form/visit-attribute-type.component.tsx
vasharma05: The type for this wasn't exported by the RHF

packages/esm-patient-chart-app/src/visit/visit-form/visit-attribute-type.component.tsx
vasharma05: P.S. This is the Ozone distro cambodia environment, used for testing purposes.
https://github.com/openmrs/openmrs-esm-patient-chart/assets/51502471/9b41503c-b8f9-4833-91cd-384743ff45c7

packages/esm-patient-chart-app/src/visit/visit-form/visit-attribute-type.component.tsx
ibacher: Looks like it's `ControllerRenderProps<TFieldValues, TName>` (see [here](https://github.com/react-hook-form/react-hook-form/blob/ece261ac22762e12eea0e944eb52636e17e61e22/src/types/controller.ts#L88C12-L88C54)).

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -449,7 +483,7 @@ const StartVisitForm: React.FC<DefaultWorkspaceProps> = ({ patientUuid, closeWor
               <div className={styles.sectionTitle}>{isTablet && t('visitAttributes', 'Visit attributes')}</div>
               <div className={styles.sectionField}>
                 <VisitAttributeTypeFields
-                  isMissingRequiredAttributes={isMissingRequiredAttributes}
+                  // isMissingRequiredAttributes={isMissingRequiredAttributes}
```
denniskigen, line 486: Is this code required?

vasharma05: No

packages/esm-patient-chart-app/src/visit/visit-form/visit-attribute-type.component.tsx
```diff
@@ -71,13 +61,11 @@ const VisitAttributeTypeFields: React.FC<VisitAttributeTypeFieldsProps> = ({
 };
 
 interface AttributeTypeFieldProps {
+  fieldProps: ControllerRenderProps<VisitFormData, `visitAttributes.${string}`>;
```
ibacher, line 64: Shouldn't this be `FieldPath<VisitFormData>` or similar?

vasharma05: This is the type of the `field`.

image

vasharma05: image

denniskigen (approved):

packages/esm-patient-chart-app/src/visit/visit-form/visit-attribute-type.component.tsx
ibacher: Right, because we've told it that the property is `visitAttributes.${attributeType.uuid}`, which works at runtime, but not at the level of the type system, since the type system only exists when adding or editing code, so it infers `attributeType.uuid` is a type string and seems to substitute that.

vasharma05: image


I'm not able to understand this, with my understanding, the second arg `visitAttributes.${string}` will be the FieldPath, not the whole `field`, right?

PR #1357, opened 2023-09-14
(fix) Hiding the workspace re-opens the workspace window.

vasharma05:
## Requirements

## Summary
In addition to #1339, updating the workspace window state workflow was altered by me in wake of optimizing.
Before the handling was in the `useWorkspaces` hook and `workspace-window` component, which was meant to update the window state on the basis of workspaces.

Due to negligence, the useEffect written couldn't hide a workspace.

Solution: All the workspaces are opened by a single function `launchPatientWorkspace`, so it should be the only point of updating the workspace window state. Hence the workspace is hidden by `->` button and opened/ re-opened when calling the launchPatientWorkspace

### DIFF
--- a/packages/esm-patient-common-lib/src/workspaces/workspaces.ts
+++ b/packages/esm-patient-common-lib/src/workspaces/workspaces.ts
@@ -167,6 +167,11 @@ export function launchPatientWorkspace(name: string, additionalProps?: object) {
       });
     }
   }
+  if (store.getState().workspaceWindowState === 'hidden') {
+    updateWorkspaceWindowState(
+      store.getState().openWorkspaces[0].preferredWindowSize === 'maximized' ? 'maximized' : 'normal',
+    );
+  }
 }
 
 export function launchPatientChartWithWorkspaceOpen({
--- a/packages/esm-patient-common-lib/src/workspaces/useWorkspaces.tsx
+++ b/packages/esm-patient-common-lib/src/workspaces/useWorkspaces.tsx
@@ -24,18 +24,6 @@ export function useWorkspaces(): WorkspacesInfo {
     getWorkspaceStore().subscribe(update);
   }, []);
 
-  // This hook is meant to be triggered only when workspace changes
-  // Accordingly the workspaceWindowState will be updated
-  useEffect(() => {
-    if (workspaces.length === 0) {
-      updateWorkspaceWindowState('hidden');
-    } else if (workspaceWindowState === 'hidden') {
-      updateWorkspaceWindowState('normal');
-    } else {
-      updateWorkspaceWindowState(workspaces[0].preferredWindowSize === 'maximized' ? 'maximized' : 'normal');
-    }
-  }, [workspaces, workspaceWindowState]);
-
   const memoisedResults = useMemo(
     () => ({
       active: workspaces.length > 0,


brandones:
Wait wait wait. Sorry I missed this problem in the original code: we should not be mutating `preferredWindowSize` at all! That's a hard-coded property of each workspace registration. It tells the workspace system what window state it *would like* to open in. The thing that should be mutated is the actual window state, which is the `workspaceWindowState` property of the workspace store.

vasharma05:
Hi @brandones , I am not mutating the preferred window size, I was mutating the `workspaces` because the window state needs to be updated on change in the workspaces

Different scenarios are:
1. Launching a workspace will show the hidden workspaces
2. When all the workspaces are closed, the workspace window should be hidden
3. Whenever a workspace is opened/ closed, the latest workspace's preferred size should be the window size.

Before my #1339 , the implementation of showing/ hiding the workspaces was done at 2 places:
1. useWorkspaces (Link)
 `windowState` returned either the hidden state (if no workspace open), else the preferred window size.
2. workspace-window.component.tsx ([Link](https://github.com/openmrs/openmrs-esm-patient-chart/blob/2878fbbf15153f733f090d748e6c027881df4847/packages/esm-patient-chart-app/src/workspace/workspace-window.component.tsx#L26))
`isWorkspaceWindowOpen` is a state for handling whether to show/ hide the workspace. The useEffects below has a dependency for `workspaces`, and whenever workspaces in the global store used to change(opening a new workspace/ closing a workspace/ reopening the opened workspace), the useEffect runs and then updates the `isWorkspaceWindowOpen` to show/ hide the workspace.

What I did in #1339 was to merge the `windowState` and the `isWorkspaceWindowOpen` use Effect handlings into one. I did test my changes out, but somehow missed the bug. The bug was that whenever I hide the workspace, the `workspaceWindowState` updates, which re-runs the removed useEffect and hence in the 1st **else** case, updates the `workspaceWindowState` to normal, hence cannot hide the workspace window. This useEffect is actually meant to run on changing workspaces only.

Now, my approach here in this PR is that, before I reduced the handling from 2 places into 1, i.e. `useWorkspaces`, but the useEffect has multiple cases to handle as mentioned above. So instead of handling all the cases in a single useEffect, if we analyze the structure of the already implemented workspaces, the `launchPatientWorkspace` is the only place where the workspace is opened from. Hence, why not update the workspace window size (maximised or normal) from there only. Regarding the other handling for hiding the workspace, it is done in the current workspace-window.component.tsx. For closing the workspaces, the workspace window is shown only when `openedWorkspaces.length > 0` as implemented here.

Another approach I have for this is to improve the removed `useEffect` like below:

```
useEffect(() => {
    if (workspaces.length === 0) {
      updateWorkspaceWindowState('hidden');
    } else {
      updateWorkspaceWindowState(workspaces[0].preferredWindowSize === 'maximized' ? 'maximized' : 'normal');
    }
  }, [workspaces]);
```

Now only the update in the workspace will handle the windowState, fulfilling all the cases as:
1. Launching a workspace will show the hidden workspaces
When a new workspace is opened initially, it will update the openedWorkspaces' length from 0 to 1, hence opening the workspace with the top workspace's preferred window size.
2. When all the workspaces are closed, the workspace window should be hidden
When all the workspaces are closed, the `workspaceWindowState` will be `hidden`
3. Whenever a workspace is opened/ closed, the latest workspace's preferred size should be the window size.
reopening the workspace will call the `launchPatientWorkspace` which will update the workspace, updating the workspaceStore, which then updates the `useWorkspaces`' workspaces state, which then updates the workspace window.

But this is again something I'd say atomic, broken into simpler items, rather than handling multiple cases, hence my changes.

The basic approach here was to return the correct workspaceWindowState from the `useWorkspaces` and hence handling updating the workspaceWindowState from the only hook, rather than multiple places. (Implemented in #1339 ).

In this PR, the approach is that, the workspaceWindowState should be updated from where it is meant to be, rather than listening to the different variables to mutate and then depending the `useEffect` hook to handle the same.

If you have a better approach, please do let me know.
Thanks!
CC: @ibacher @denniskigen

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
```diff
@@ -167,6 +167,11 @@ export function launchPatientWorkspace(name: string, additionalProps?: object) {
       });
     }
   }
+  if (store.getState().workspaceWindowState === 'hidden') {
+    updateWorkspaceWindowState(
+      store.getState().openWorkspaces[0].preferredWindowSize === 'maximized' ? 'maximized' : 'normal',
```
brandones, line 172: It really looks very precisely like you are modifying the `preferredWindowSize`, which should be regarded as immutable. I appreciate your very thorough analysis of the requirements but this is not a good solution.

vasharma05: I might not be getting your point of mutating the preferred window size. I am just updating the workspace window size to either be maximized or normal. I am not changing the preferred window size of an opened workspace. I am knowing whether the workspace to opened in full screen ora normal size screen

brandones: Wow, I am an idiot. Sorry @vasharma05 . You're completely right.

brandones (approved):
Sorry for the hassle, nice work!

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
vasharma05: No worries @brandones !
Thanks for the review!

PR #1377, opened 2023-09-27
(enhc) Change default fallback values to empty string for visit attributes

donaldkibet:
## Requirements

## Summary
This pull request refines the default values for visit-related attributes to enhance system flexibility.

In earlier implementations, in the absence of data, the system would resort to null as a fallback for these attributes. This change transitions that fallback to an empty string (''). The primary motivation behind this shift is to offer greater adaptability, especially in scenarios where there's a need to launch forms without necessarily linking them to a specific visit.

## Related Issue

https://thepalladiumgroup.atlassian.net/browse/KHP3-3935


makombe:
@donaldkibet not so sure if I understand your use case. But I think 'Null' should just serve you well without necessary changing it.

vasharma05:
I agree with @makombe too, `null` represents no value, but `''` represents empty string.

PR #1406, opened 2023-10-10
(feat) Add support for configurable form sections

ibacher:
## Requirements

- [] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://zeroheight.com/23a080e38/p/880723-introduction).

## Summary

This PR does a few things, but the main one is that it adds the ability to configure "sections" of forms that will display in the form dashboard in places of the normal list of all forms. If this remains unconfigured, nothing is changed, but if configured, it looks like the below UI.

## Screenshots

Form Dashboard with Multiple Sections

## Other

Notably:

* Forms can be included in any number of sections
* Forms are only loaded once
* Forms per section can be specified either using the form UUID or the form name

Smaller things this does:

* `orderFormsByName` has been replaced by `orderBy` which can be set to either `name` (the default) or `most-recent`. Currently, this setting is global to the form list and not configurable per section.
* I changed the search from "partial match" to use the "fuzzy" library, though "fuzzy" may not be optimal for all languages, it should include more results than the "partial match".
* Each section is separately searchable.

There were two components involved in the form, the rather anemic `FormsDashboard`, which basically just rendered a `FormsList`, and the `FormsList` itself that basically implemented the feature. This refactors things a fair bit to change component responsibility. In particular:

* Adds a `FormsTable` component which just takes the headers, rows and a couple of helper elements and renders the actual table of forms and search box.
* Modifies `FormsList` to take a list of forms, an optional section name and simply render an appropriate `FormsTable`. it also handles searching for the list.
* The `FormsDashboard` is now responsible for all data loading and then renders a `FormsList` per section or just a `FormsList`, depending on configuration.

The goal of this refactoring was to move the slowest bits (fetching the list of forms) to the top of the tree (and also to keep `FormsList` from becoming a madness-inducing series of `if / else` conditions).


ibacher:
CC: @icrc-loliveira

packages/esm-patient-forms-app/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
   "extends": "../../tsconfig.json",
-  "include": ["src/**/*"],
+  "include": ["src/**/*", "../../tools/setupTests.ts"],
```
ibacher, line 3: @denniskigen I did this to allow TS to recognise the types from `@testing-library/jest-dom`. I think we talked about that once before. I think this is safe to do, as we don't currently use tsconfig to for actual compilation, just type-checking. If / when we do, we may need some scheme with a `tsconfig.test.json`, but that seems more complicated.

packages/esm-patient-forms-app/src/forms/forms-list.component.tsx
```diff
@@ -1,81 +1,69 @@
...
+import styles from './forms-list.scss';
+import type { CompletedFormInfo } from '../types';
 
 export type FormsListProps = {
-  currentVisit: Visit;
-  htmlFormEntryForms: Array<HtmlFormEntryForm>;
-  patient: fhir.Patient;
-  patientUuid: string;
+  completedForms?: Array<CompletedFormInfo>;
+  error?: any;
```
ibacher, line 14: I hate using `any`, but this is meant to be directly from SWR and that's the type it gives.

gracepotma:
Wow, beautiful work Ian!! Thanks for this PR 💐

packages/esm-patient-forms-app/src/forms/forms-dashboard.component.tsx
```diff
@@ -1,24 +1,61 @@
...
+
   return (
     <div className={styles.container}>
-      <FormsList
-        currentVisit={currentVisit}
-        htmlFormEntryForms={htmlFormEntryForms}
-        patient={patient}
-        patientUuid={patientUuid}
-      />
+      {!Boolean(sections) ? (
```
denniskigen, line 44: The boolean check doesn't distinguish between arrays that are empty and those that are not. When running a dev server with an empty `formSections` config value, I see an empty page. 

```suggestion
      {sections.length === 0) ? (
```

packages/esm-patient-forms-app/src/forms/forms-list.component.tsx
```diff
@@ -1,81 +1,69 @@
...
+      return completedForms;
     }
-    return data.filter((form) => {
-      const formName = form.form.display ?? form.form.name;
-      return formName.toLowerCase().includes(searchTerm.toLowerCase());
-    });
-  }, [data, searchTerm]);
 
-  const handleSearch = React.useMemo(() => debounce((searchTerm) => setSearchTerm(searchTerm), 300), []);
+    return fuzzy
```
denniskigen, line 49: I like this fuzzy search approach. Perhaps we should use it in most other places.

packages/esm-patient-forms-app/tsconfig.json
denniskigen: I agree. The existing workaround (needing to manually launch the `test-helpers.tsx`
file to get the IDE to load types) is too convoluted. I've not found a better approach.

packages/esm-patient-forms-app/src/forms/forms-list.component.tsx
denniskigen: `Error` usually works for me.

packages/esm-patient-forms-app/src/forms/forms-list.component.tsx
ibacher: I got the idea from @brandones... I like that library a lot; it's simple and to the point. I'm not entirely sure how well it will perform with different languages and scripts, but it should be no worse than what was there before.

denniskigen:
Sorry about the poorly formatted suggestion. Are you getting this too?


CleanShot 2023-10-19 at 10  21 51@2x

ibacher:
No, I'm not

denniskigen:
Do you have the sample config you used to get the screenshot somewhere?

ibacher:
I should've save it, but it was something like:

```json
{
  "@openmrs/esm-patient-forms-app": {
    "formSections": [
       {"name": "Section 1", "forms": ["HIV Green Card", "Two-Page Form", "Viral Load Lab Result"]},
       {"name": "Section 2", "forms": ["HIV Green Card", "Two-Page Form", "Viral Load Lab Result"]},
       {"name": "Section 3",  "forms": ["SOAP Note Template", "Test Form 1", "Covid 19", "Test Form", "Laboratory Test Orders"]},
     ]
  } 
}
```

denniskigen:
Thanks, will be useful when I write up the changelog for the next release.

ibacher:
This reminds me, I think there's an error logged with a config like the one above... It actually works, it's just something about the validation logic. I was going to ticket that up.

denniskigen:
Ok. Easy to get lost in the mire.

denniskigen (approved):
LGTM. Thanks, @ibacher.

PR #1430, opened 2023-10-24
(feat) Supporting O3-2510: Expose patient-common-lib on window

brandones:
## Requirements

## Summary
Makes esm-patient-common-lib available on the window object. This will allow ngx-formentry to access it despite not being a webpack module.

## Related Issue
https://issues.openmrs.org/browse/O3-2510

## Other
Copilot called me out for doing a hack
image

### DIFF
--- a/packages/esm-patient-chart-app/src/index.ts
+++ b/packages/esm-patient-chart-app/src/index.ts
@@ -7,12 +7,17 @@ import {
   registerFeatureFlag,
 } from '@openmrs/esm-framework';
 import { createDashboardLink } from '@openmrs/esm-patient-common-lib';
+import * as PatientCommonLib from '@openmrs/esm-patient-common-lib';
 import { esmPatientChartSchema } from './config-schema';
 import { moduleName, spaBasePath } from './constants';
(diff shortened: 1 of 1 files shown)


ibacher:
Question on the premise here: openmrs-ngx-formentry is, in fact, using Webpack (I realize the Angular team regards this as an "implementation detail", but it's one we rely on).

With that in mind, do you think this is still the right approach or should we use something like `ngx-build-plus` which allows us to modify the webpack config (and is what we use to turn the form-entry-app into a federated module)?

brandones:
Oh, interesting. I guess I am still inclined to do it this way. I don't think it poses major disadvantages, other than that there's no way to do explicit version dependency, and this can be substituted for checking expected function signatures and logging good error messages. I'm not sure it would be worth it to get the whole `ngx-build-plus` thing set up just for this. Above all, though, I'm not sure WMF supports conditional dependency loading in any way *other* than is more or less exactly this. The ngx-formentry code needs to be able to run without producing errors outside the context of the patient chart. It's only one specific question-type that depends on patient-chart.

ibacher (approved):

PR #1431, opened 2023-10-25
(fix:form-entry) update search location functionality

jecihjoy:
## Requirements

## Summary

Search location endpoint was returning all locations through endpoint `location?v=` and filtering was done in the service layer. This update help uterize seach location endpoint provided by the openmrs location resource `location?q=&v=`

Before 
Screenshot from 2023-10-25 09-38-22

After
Screenshot from 2023-10-25 09-35-07

### DIFF
--- a/packages/esm-form-entry-app/src/app/openmrs-api/location-resource.service.ts
+++ b/packages/esm-form-entry-app/src/app/openmrs-api/location-resource.service.ts
@@ -17,7 +17,7 @@ export class LocationResourceService {
   ) {}
 
   public getLocationByUuid(uuid: string): Observable<Location | undefined> {
-    const url = this.getUrl(uuid);
+    const url = this.getLcationByUuidUrl(uuid);
     return this.http.get<Location>(url).pipe(catchError(() => this.getLocationByUuidFallback(uuid)));
   }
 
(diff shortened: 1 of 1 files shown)


packages/esm-form-entry-app/src/app/openmrs-api/location-resource.service.ts
```diff
@@ -17,7 +17,7 @@ export class LocationResourceService {
   ) {}
 
   public getLocationByUuid(uuid: string): Observable<Location | undefined> {
-    const url = this.getUrl(uuid);
+    const url = this.getLcationByUuidUrl(uuid);
```
denniskigen, line 20:
```suggestion
    const url = this.getLocationByUuidUrl(uuid);
```

packages/esm-form-entry-app/src/app/openmrs-api/location-resource.service.ts
```diff
@@ -26,21 +26,22 @@ export class LocationResourceService {
...
+    let url = '';
+    searchText ? (url = this.getUrl(searchText)) : (url = this.getUrl());
     return this.http.get<ListResult<Location>>(url).pipe(map((r) => r.results));
   }
 
-  public getUrl(uuid?: string) {
-    return uuid
-      ? this.windowRef.openmrsRestBase + 'location/' + uuid + '?v=' + LocationResourceService.v
-      : this.windowRef.openmrsRestBase + 'location?q=&v=' + LocationResourceService.v;
+  public getLcationByUuidUrl(uuid?: string) {
```
denniskigen, line 38:
```suggestion
  public getLocationByUuidUrl(uuid?: string) {
```

denniskigen (dismissed):
Thanks, @jecihjoy.

ibacher:
Nice work! I just want to note that there's actually some value to having this load all locations. Specifically:

1. In offline mode, it's possible for us to pre-cache the all locations lookup easily and still do the same filtering on the frontend.
2. In a similar vein, a better designed version of the service here could load the locations once and continue to use that, which reduces the number of network calls, which will generally be a performance win.

I'm not opposed to merging this in, just making sure we're thinking of a technical design that will best serve our users.

jecihjoy:
@ibacher thank you for the thoughtful review. I took sometime to investigate why I was getting an empty list depicted in the first screenshot, that was the motivation for me to move searching functionality into the backend. I realized we had configure default rest pagination size to 150 and my search criteria was not part of the 150.
We have over 10,000 entries in the location table, which it's slow pulling all these at once, however with your suggested approach of caching I believe it'll improve perfomance.
So the short term, I suggest we proceed with the changes, but I'll raise a followup up ticket with your suggested enhancemnets.

PR #1458, opened 2023-11-08
(feat) The lab order form should have a reason for ordering field

makombe:
## Requirements

## Screenshots
lab-order-reason

## Related Issue

https://issues.openmrs.org/browse/O3-2361

## Other

Sample config schema

```ts
{
  "@openmrs/esm-patient-labs-app": {
    "labTestsWithOrderReasons": [
      {
        "labTestUuid": "854AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
        "orderReasons": [
          "161236AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
          "162080AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
          "160032AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
        ]
      }
    ]
  }
}
```

### DIFF
--- a/packages/esm-patient-labs-app/src/config-schema.ts
+++ b/packages/esm-patient-labs-app/src/config-schema.ts
@@ -52,8 +52,8 @@ export const configSchema = {
       orderReasons: {
         _type: Type.Array,
         _elements: {
-          uuid: {
-            _type: Type.UUID,
+          concept: {
+            _type: Type.ConceptUuid,
             _description: 'Array of coded concepts that represent reasons for ordering a lab test',
           },
           label: {
--- a/packages/esm-patient-labs-app/src/config-schema.ts
+++ b/packages/esm-patient-labs-app/src/config-schema.ts
@@ -52,15 +52,8 @@ export const configSchema = {
       orderReasons: {
         _type: Type.Array,
         _elements: {
-          concept: {
-            _type: Type.ConceptUuid,
-            _description: 'Array of coded concepts that represent reasons for ordering a lab test',
-          },
-          label: {
-            _type: Type.String,
-            _default: null,
-            _description: 'The label for the reason for ordering concept',
-          },
+          _type: Type.ConceptUuid,
+          _description: 'Array of coded concepts that represent reasons for ordering a lab test',
         },
         _default: [],
         _description: 'Coded Lab test order reason options',
@@ -81,7 +74,7 @@ export interface LabTestReason {
 }
 export interface OrderReason {
   labTestUuid: string;
-  orderReasons: Array<LabTestReason>;
+  orderReasons: Array<string>;
 }
 export interface ConfigObject {
   concepts: Array<ObsTreeEntry>;
--- a/packages/esm-patient-labs-app/src/config-schema.ts
+++ b/packages/esm-patient-labs-app/src/config-schema.ts
@@ -41,11 +41,47 @@ export const configSchema = {
       _default: '52a447d3-a64a-11e3-9aeb-50e549534c5e',
     },
   },
+  labTests: {
+    _type: Type.Array,
+    _elements: {
+      labTestUuid: {
+        _type: Type.UUID,
+        _description: "UUID for the 'Lab' Test that require order reason",
+        _default: '',
+      },
+      labTestOrderReasons: {
+        _type: Type.Array,
+        _elements: {
+          uuid: {
+            _type: Type.UUID,
+            _description: 'Lab test concept UUID',
+          },
+          label: {
+            _type: Type.String,
+            _default: null,
+            _description: 'The label for order reason',
+          },
+        },
+        _default: [],
+        _description: 'Coded Lab test order reason options',
+      },
+    },
+    _default: [],
+    _description: 'Whether to allow for provision of order reason',
+  },
 };
 
 export interface ObsTreeEntry {
   conceptUuid: string;
   defaultOpen: boolean;
+}
+export interface LabTestReason {
+  uuid: string;
+  label?: string;
+}
+export interface OrderReason {
+  labTestUuid: string;
+  labTestOrderReasons: Array<LabTestReason>;
 }
 export interface ConfigObject {
   concepts: Array<ObsTreeEntry>;
@@ -53,4 +89,5 @@ export interface ConfigObject {
   orders: {
     labOrderTypeUuid: string;
   };
+  labTests: Array<OrderReason>;
 }
--- a/packages/esm-patient-labs-app/src/config-schema.ts
+++ b/packages/esm-patient-labs-app/src/config-schema.ts
@@ -41,25 +41,25 @@ export const configSchema = {
       _default: '52a447d3-a64a-11e3-9aeb-50e549534c5e',
     },
   },
-  labTests: {
+  labTestsWithOrderReasons: {
     _type: Type.Array,
     _elements: {
       labTestUuid: {
         _type: Type.UUID,
-        _description: "UUID for the 'Lab' Test that require order reason",
+        _description: 'UUID of the lab test that requires a reason for ordering',
         _default: '',
       },
-      labTestOrderReasons: {
+      orderReasons: {
         _type: Type.Array,
         _elements: {
           uuid: {
             _type: Type.UUID,
-            _description: 'Concept UUID representing the order reason',
+            _description: 'Array of coded concepts that represent reasons for ordering a lab test',
           },
           label: {
             _type: Type.String,
             _default: null,
-            _description: 'The label for order reason',
+            _description: 'The label for the reason for ordering concept',
           },
         },
         _default: [],
@@ -81,7 +81,7 @@ export interface LabTestReason {
 }
 export interface OrderReason {
   labTestUuid: string;
-  labTestOrderReasons: Array<LabTestReason>;
+  orderReasons: Array<LabTestReason>;
 }
 export interface ConfigObject {
   concepts: Array<ObsTreeEntry>;
@@ -89,5 +89,5 @@ export interface ConfigObject {
   orders: {
     labOrderTypeUuid: string;
   };
-  labTests: Array<OrderReason>;
+  labTestsWithOrderReasons: Array<OrderReason>;
 }
--- a/packages/esm-patient-labs-app/src/config-schema.ts
+++ b/packages/esm-patient-labs-app/src/config-schema.ts
@@ -54,7 +54,7 @@ export const configSchema = {
         _elements: {
           uuid: {
             _type: Type.UUID,
-            _description: 'Lab test concept UUID',
+            _description: 'Concept UUID representing the order reason',
           },
           label: {
             _type: Type.String,
@@ -67,7 +67,7 @@ export const configSchema = {
       },
     },
     _default: [],
-    _description: 'Whether to allow for provision of order reason',
+    _description: 'Whether to allow for provision of coded order reason',
   },
 };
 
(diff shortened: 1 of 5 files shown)


packages/esm-patient-labs-app/src/lab-orders/add-lab-order/lab-order-form.tsx
```diff
@@ -24,6 +25,8 @@ export function LabOrderForm({ initialOrder, closeWorkspace }: LabOrderFormProps
   const { orders, setOrders } = useOrderBasket<LabOrderBasketItem>('labs', prepLabOrderPostData);
   const [inProgressLabOrder, setInProgressLabOrder] = useState(initialOrder as LabOrderBasketItem);
   const { testTypes, isLoading: isLoadingTestTypes, error: errorLoadingTestTypes } = useTestTypes();
+  const config = useConfig() as ConfigObject;
```
ibacher, line 28:
```suggestion
  const config = useConfig<ConfigObject>();
```

packages/esm-patient-labs-app/src/config-schema.ts
```diff
@@ -41,16 +41,53 @@ export const configSchema = {
...
+        _type: Type.UUID,
+        _description: "UUID for the 'Lab' Test that require order reason",
+        _default: '',
+      },
+      labTestOrderReasons: {
+        _type: Type.Array,
+        _elements: {
+          uuid: {
+            _type: Type.UUID,
+            _description: 'Lab test concept UUID',
```
ibacher, line 57: Could we get a better description here? It's really unclear how this differs from `labTestUuid` just from the text and that description should be clear.

packages/esm-patient-labs-app/src/config-schema.ts
```diff
@@ -41,16 +41,53 @@ export const configSchema = {
...
+          },
+          label: {
```
ibacher, line 59: As a general rule, we should avoid labels in configuration if we can. Labels like this are only usable in monolingual environments. In general, especially if this is concept-backed, we should prefer to use the display strings from the concepts themselves rather than inventing new text for them.

That said, what we store in the backend is just text, so I'm a little unclear on why we even have concepts involved in this at all? If we want to capture coded answers or coded or free text answers, we should actually update the backend and data model to reflect that.

ibacher: At a minimum, labels should be run through the translation system.

makombe: @ibacher orders model already provides the ability to capture order reason as coded or non coded/ free text. In this case we are using coded answers which are represented by concepts

ibacher: @makombe Thanks. I didn't realise it was using that pattern. If we're using concepts, then the label should come from the concept itself, right? Ideally, we want whatever we're using for drop-down labels to be translatable. Having strings in the configuration schema is, unfortunately, not translatable.

packages/esm-patient-labs-app/src/config-schema.ts
```diff
@@ -41,16 +41,53 @@ export const configSchema = {
       _default: '52a447d3-a64a-11e3-9aeb-50e549534c5e',
     },
   },
+  labTests: {
+    _type: Type.Array,
+    _elements: {
+      labTestUuid: {
+        _type: Type.UUID,
+        _description: "UUID for the 'Lab' Test that require order reason",
```
denniskigen, line 49:
```suggestion
        _description: "UUID for the lab test order that requires an order reason",
```

packages/esm-patient-labs-app/src/config-schema.ts
```diff
@@ -41,16 +41,53 @@ export const configSchema = {
...
+        _default: '',
+      },
+      labTestOrderReasons: {
+        _type: Type.Array,
+        _elements: {
+          uuid: {
+            _type: Type.UUID,
+            _description: 'Concept UUID representing the order reason',
```
denniskigen, line 57:
```suggestion
            _description: 'Array of concept UUIDs that represent answers for an order reason concept',
```

packages/esm-patient-labs-app/src/config-schema.ts
```diff
@@ -41,16 +41,53 @@ export const configSchema = {
...
+          },
+          label: {
+            _type: Type.String,
+            _default: null,
+            _description: 'The label for order reason',
```
denniskigen, line 62: Defer to Ian's line of reasoning above, but if we're to keep this property, a more appropriate description could be:

```suggestion
            _description: 'The label for order reason concept answer',
```

packages/esm-patient-labs-app/src/config-schema.ts
```diff
@@ -41,16 +41,53 @@ export const configSchema = {
...
+          },
+        },
+        _default: [],
+        _description: 'Coded Lab test order reason options',
+      },
+    },
+    _default: [],
+    _description: 'Whether to allow for provision of coded order reason',
```
denniskigen, line 70: Not sure this accurately communicates what this config property does. It looks like we have an array that contains concepts for lab test orders that require a reason for ordering whose structure is something like:

```ts
// labTests array 
[ 
  {
    "labTestUuid": "52a447d3-a64a-11e3-9aeb-50e549534c5e",
    "labTestOrderReasons": [
      {
        "uuid": "52a447d3-a64a-11e3-9aeb-50e549534c5e",
        "label": "Test Reason 1"
      },
      {
        "uuid": "52a447d3-a64a-11e3-9aeb-50e549534c5e",
        "label": "Test Reason 2"
      }
    ] 
  }
]
```

Perhaps the config property would be better named `labTestOrdersWithReasonsForOrdering`.

packages/esm-patient-labs-app/src/config-schema.ts
```diff
@@ -41,16 +41,53 @@ export const configSchema = {
       _default: '52a447d3-a64a-11e3-9aeb-50e549534c5e',
     },
   },
+  labTests: {
```
denniskigen, line 44: Maybe?

```suggestion
  labTestsRequiringReasonsForOrdering: {
```

packages/esm-patient-labs-app/src/config-schema.ts
```diff
@@ -41,16 +41,53 @@ export const configSchema = {
...
+    _type: Type.Array,
+    _elements: {
+      labTestUuid: {
+        _type: Type.UUID,
+        _description: "UUID for the 'Lab' Test that require order reason",
+        _default: '',
+      },
+      labTestOrderReasons: {
```
denniskigen, line 52: Could just be `reasonsForOrdering` because the array name already establishes the lab test order context.

```suggestion
      reasonsForOrdering: {
```

packages/esm-patient-labs-app/src/config-schema.ts
```diff
@@ -41,16 +41,53 @@ export const configSchema = {
...
+            _description: 'Concept UUID representing the order reason',
+          },
+          label: {
+            _type: Type.String,
+            _default: null,
+            _description: 'The label for order reason',
+          },
+        },
+        _default: [],
+        _description: 'Coded Lab test order reason options',
```
denniskigen, line 66:
```suggestion
        _description: 'Coded answers for order reason concepts',
```

ojwanganto:
pinging @brandones

packages/esm-patient-labs-app/src/config-schema.ts
brandones: Agree with @ibacher here, we should be getting labels from the concepts, right?

brandones: Then we can simplify the config schema to be just an array of string concept UUIDs. Another alternative, which could be preferable (but I don't feel strongly about it, I'll leave it up to you) would be just to have `orderReasonsConceptSet`, which would be a UUID of a concept set that has the reasons as members.

packages/esm-patient-labs-app/src/lab-orders/add-lab-order/lab-order-form.component.tsx
```diff
@@ -70,6 +72,11 @@ export function LabOrderForm({ initialOrder, closeWorkspace }: LabOrderFormProps
       ...initialOrder,
     },
   });
+  const config = useConfig<ConfigObject>();
+  const selectedLabTest =
+    config && config.labTestsWithOrderReasons
+      ? config.labTestsWithOrderReasons.find((p) => p.labTestUuid === defaultValues?.testType?.conceptUuid)
+      : null;
```
brandones, line 79:
```suggestion
  const orderReasons = config.labTestsWithOrderReasons.find((c) => c.labTestUuid === defaultValues?.testType?.conceptUuid).map(c => c.orderReasons);
```

packages/esm-patient-labs-app/src/lab-orders/add-lab-order/lab-order-form.component.tsx
```diff
@@ -187,6 +194,30 @@ export function LabOrderForm({ initialOrder, closeWorkspace }: LabOrderFormProps
               </InputWrapper>
             </Column>
           </Grid>
+          {selectedLabTest && (
```
brandones, line 197:
```suggestion
          {orderReasons.length && (
```

packages/esm-patient-labs-app/src/config-schema.ts
ojwanganto: While this would be a good approach, it forces one to have concepts whose displays read exactly as the reasons provided in the drop-down. For instance, there could be a concept for indicating a baseline test with a display of **_initial_** while on the ground folks use **_baseline_** a lot. What would be the workaround for such? Do we support some degree of overrides?

brandones: Are you not able to change names on your concepts?

makombe: @brandones we normally try to avoid editing of concept directly as to be in sync with CIEL dictionary. We consume concepts as they come from CIEL.

packages/esm-patient-labs-app/src/lab-orders/add-lab-order/lab-order-form.component.tsx
makombe: updated.

packages/esm-patient-labs-app/src/config-schema.ts
ibacher: You can always request that CIEL adds a synonym or even preferred name for a concept. Alternatively, you can use Iniz to provide custom concept names or change the preferred name in a reproducible way without needing to deviate from CIEL.

In any case, labels _must never_ appear in the configuration because it is unworkable in a multilingual environment. If we need something in the configuration to represent a user-facing string it needs to be a translation key rather than a label.

ojwanganto: Great points, thanks. I know adding synonyms in CIEL may be good but still not sustainable more so if preferences/synonyms differ across implementations. A major challenge we face with labels is trying to match the labels as they are provided in clinical forms or even reporting tools. Even though the concept is the same, we are forced to display it differently depending on the type of user, and how they commonly use the concept. 

The concept to use the iniz module is great but I don't see how it will address the label overrides for concepts that may require 2 or 3 synonyms. I may be wrong, but I was hoping that the idea behind a config is to somehow address a local context. For instance, I don't expect the same config to apply in Kenya and Tanzania where English and Kiswahili are used. I would expect the Kenyanized config to be in English and that for Tanzania in Kiswahili. Sorry if I may have missed the point.

makombe: @ibacher  @brandones I have removed the label from the config as suggested and made necessary adjustment. Thanks

ojwanganto: Thanks @makombe. @ibacher @brandones @denniskigen kindly help with review. Thanks

ojwanganto:
@denniskigen kindly help us review this. This PR will unblock the use of o3 lab order form for HIV monitoring lab tests

denniskigen (approved):

brandones:
Thank you @denniskigen for finishing the review, thanks @ojwanganto and @makombe for your work!

PR #1461, opened 2023-11-10
(feat) Preload most extensions and pages

ibacher:
## Requirements

## Summary

This is similar to other PRs in other monorepos. The goal here is to ensure that most extensions and pages are part of the main chunk for the app. While I've tried to be consistent about this, it's a little more visible in this PR that certain extensions are still imported asynchronously. I've tried to keep it so that "secondary" extensions, like confirmation dialogs or the like, are loaded asynchronously (so when they are first needed), on the assumption that there's some value in having less-used components loading deferred.

<s>I'm not sure this makes great strides in reducing the overall JS file loading. It looks like it eliminates about 2 file loads just getting to the patient chart, though it should remove quite a few more if you actually navigate through the patient chart.</s>

Forgot to also update the Webpack config, which knocks another 6 files, so this reduces the initial load by about 8 files and should reduce loads during navigation quite a bit more.

There's also a very small bugfix which will reset the dashboard title for dashboards that do not declare a title (which is most of the dashboards in the patient chart; only appointments seems to use this feature).


denniskigen (approved):
LGTM. Thanks, @ibacher!

PR #1468, opened 2023-11-16
(fix) `Add to basket` action on drug tile should mark the order as an incomplete order

vasharma05:
## Requirements

## Summary
This PR fixes the broken workflow, whenever the user adds the drug directly to the basket. The drug added to the order basket should be marked as an incomplete order and the user shouldn't be able to save the order until all the details are added.

Additional changes:
1. Sign and close will be disabled if there are any incomplete orders.
2. After filling the form, the incomplete order will be marked as New order

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-chart/assets/51502471/e0759a28-a634-43d9-8392-beb7b58b90fe


ibacher (commented):
I'm fine with having some state tracking in the `OrderBasketItem`, but I think it shouldn't be part of the `action` field. There are two reasons:

1. This maps directly to a backend field and it seems easier over the long-term to keep these in sync
2. It's not clear to me that the only "incomplete" orders are "new" orders. In essence, every order added to the order basket (whether its a new / revise / discontinue or renew) should be incomplete until the clinician reviews it.

vasharma05:
> I'm fine with having some state tracking in the `OrderBasketItem`, but I think it shouldn't be part of the `action` field. There are two reasons:
> 
> 1. This maps directly to a backend field and it seems easier over the long-term to keep these in sync
> 2. It's not clear to me that the only "incomplete" orders are "new" orders. In essence, every order added to the order basket (whether its a new / revise / discontinue or renew) should be incomplete until the clinician reviews it.
> 
> Otherwise, nice work!

Hi @ibacher !
I also agree with the points you put forward, hence I've separated the INCOMPLETE status from action into a separate key `isOrderIncomplete` which should cover all the various action flows.

Please review the changes.
Thanks!

packages/esm-patient-medications-app/src/drug-order-basket-panel/drug-order-basket-panel.extension.tsx
```diff
@@ -19,10 +19,41 @@ export default function DrugOrderBasketPanelExtension() {
...
+  } = useMemo(() => {
+    const incompleteOrderBasketItems: Array<DrugOrderBasketItem> = [];
+    const newOrderBasketItems: Array<DrugOrderBasketItem> = [];
+    const renewedOrderBasketItems: Array<DrugOrderBasketItem> = [];
+    const revisedOrderBasketItems: Array<DrugOrderBasketItem> = [];
+    const discontinuedOrderBasketItems: Array<DrugOrderBasketItem> = [];
+
+    orders.forEach((order) => {
+      if (order?.isOrderIncomplete) {
+        incompleteOrderBasketItems.push(order);
```
ibacher, line 37: Q: Should incomplete items be ordered predictably by action? This is probably fine for now.

vasharma05: My thought here was to show all incomplete orders at the top.

ibacher: Oh, that completely makes sense and I get that. I was more of wondering if the incomplete orders themselves should be sorted like the other orders while being on top.

brandones:
Thanks @vasharma05 ! Linked to ticket https://issues.openmrs.org/browse/O3-2427 .

PR #1475, opened 2023-11-19
(test) Refactor e2e tests to adopt best practices

denniskigen:
## Requirements

## Summary

In this PR, I've refactored tests in our e2e suite to leverage more robust locators (based on role attributes) and improved some wording in the test descriptions and the spec names. Additionally: 

- I've added aria-roles to some elements across various components for greater discoverability when writing e2e tests. A future iteration of this e2e work will involve swapping out test ids in favour of these accessibility attributes. 
- I've refactored some components to use semantic HTML tags that better describe what they do, e.g. using the `header` tag for the patient header instead of a plain `div`.
- I've renamed the `Clinical forms` action button label from `Clinical form` to `Clinical forms`.  

## Screenshots

CleanShot 2023-11-19 at 11  37 24@2x

## Related Issue
*None*

## Other
*None*


jayasanka-sack (commented):
Thanks, @denniskigen! This is awesome!

Quick question: Is there a specific reason for using regular expressions? I'd prefer using the exact strings since we know how it appears on the UI and don't need to consider case matching. What do you think?

The rest looks good for me.

denniskigen:
> Thanks, @denniskigen! This is awesome!
> 
> Quick question: Is there a specific reason for using regular expressions? I'd prefer using the exact strings since we know how it appears on the UI and don't need to consider case matching. What do you think?
> 
> The rest looks good for me.

Thanks, @jayasanka-sack. I think regex matches are preferable for their ability to match case-insensitive strings. Given how strings in code are prone to change due to i18n refactors, it's preferable to guard against test failures when the case of translation changes. They're also a bit more robust owing to the ability to do partial matching e.g. the text content of an element changes slightly (e.g., due to a change in wording or formatting), a regex can still match the new text if it follows the same pattern.

jayasanka-sack (approved):
That makes sense. Thanks @denniskigen !

PR #1481, opened 2023-11-20
(fix) Vitals form should not render if metadata is loading

ibacher:
## Requirements

## Summary

I haven't found a way to reproduce this issue other than the e2e tests on CI, so I've created a very minimal dummy CI.

Basically this does two things:

1. Makes min / max properties of the various `NumberInput` components we use work if there are no minimum or maximum values
2. If the form tries to render while the concept metadata is being loaded, we render a small skeleton instead.

Goal here is to fix the failing e2e tests.


denniskigen (approved):
LGTM! Thanks, @ibacher

PR #1487, opened 2023-11-21
(e2e) More e2e refactors

denniskigen:
## Requirements

## Summary

This PR makes some more refactors to our e2e test suite with the goal of making them more robust by prioritising the use of accessibility attributes to locate elements.

### DIFF
--- a/e2e/pages/allergies-page.ts
+++ b/e2e/pages/allergies-page.ts
@@ -3,8 +3,7 @@ import { Page } from '@playwright/test';
 export class PatientAllergiesPage {
   constructor(readonly page: Page) {}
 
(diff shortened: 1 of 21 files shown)


ibacher (approved):
Generally looks fine, although I think we're being a little aggressive with case-insensitive matches in cases where case-sensitivity is actually part of the design.

Also, I'm happy with using `aria-label`s, but that turns the text into user-facing text, so it would be best if it were translatable.

ibacher:
@denniskigen I'd mostly be concerned about things that are headers or links on the left nav. I think we want to make sure that those are consistent with the designs. Not a real issue, though.

PR #1508, opened 2023-11-27
(chore) Fix the missing translations in workspace registration, workspace files, and hooks 

vasharma05:
## Requirements

## Summary
The PR includes the following changes:
1. The translation extractor should scan all the `ts` and `tsx` files for translations because the translations are usually present in `index.ts` and hooks residing in `.ts` files.
2. Missing `title` translations when registering workspaces.

## Screenshots
None

## Related Issue
None


denniskigen:
Looks OK, but I would greatly prefer specifying exact file names instead of the `src/**/*.{ts,tsx}` glob pattern which includes things like ambient declaration files and config schemas in the i18next parser lookup. In practice, we'd typically expect translations strings to exist in: 

- React components (`*.component.tsx`)
- App entry points (`index.ts`)
- Workspaces with the `.workspace` suffix (`*.workspace.tsx`)
- Extensions with the `.extension` suffix (`*.extension.tsx`)

Things we don't typically expect to extract translation strings from include:

- Tests
- Resource files (with data fetching logic)
- Config files

vasharma05:
Some of the translations are also present in the hooks as well, which is usually present in either `.ts` files or `.resource.ts` files.

ibacher:
We should also include `.resource.ts` files then, though this probably represents an antipattern (`reource` files are meant to contain code for interacting with the API, not the UI). Any file with just `.ts` in an app that isn’t `index.ts` is probably not following the naming convention and should be changed.

PR #1535, opened 2023-12-12
(feat) Add billing function on start visit form

donaldkibet:
## Requirements

## Summary
This PR introduces a new billing workflow aimed at charging paying patients once they are checked in. The intention is to streamline the billing process. Most of the logic handling billing is in a custom [package](https://github.com/palladiumkenya/kenyaemr-esm-3.x/pull/96) 

To note this functionality is also an opt in feature, by default its disabled.

## Screenshots
https://github.com/openmrs/openmrs-esm-patient-chart/assets/28008754/dcf9ade5-45b5-4a13-90b2-014dfff7b5f5


ibacher:
So broad comments on here:

* I think it's fine to add a new slot, but I'd resist calling it anything to do with "billing". Slots are spaces on the screen and shouldn't be labelled for the feature they support.
* It would be preferable not to have the `billingInfo` state at all. I think that's a little too use-case specific. Instead, it might make sense to provide a state that could track callbacks to be performed during submission, and pass a `registerSubmissionCallback()` function to the extensions in the slot via the state, then have any registered submission callbacks called. That way, when we have 7 other things to do, we don't need to keep adding options.
* The above removes the need for a configuration option at all, since the slot will either be used or not used.

As another general caveat: I'm very wary of having so many different things kicked off in response to a single frontend call. It would be substantially better to build a backend service that was extensible and just handled the submission of a visit. The issue here is transactionality. For example, in the current setup, it's easily possible for the browser to, for example, kick-off creating the queue entry and then be killed before the appointment is created, leading to inconsistent state.

packages/esm-patient-chart-app/src/config-schema.ts
```diff
@@ -118,6 +118,11 @@ export const esmPatientChartSchema = {
       "Whether the visit location field in the Start Visit form should be view-only. If so, the visit location will always be set to the user's login location.",
     _default: false,
   },
+  showBillingSlot: {
+    _type: Type.Boolean,
+    _description: 'Whether on start visit form to show billing extension and enable billing functionality',
```
IamSAL, line 123: Description seems ambiguous.

suggestion: "Whether to show the billing extension and enable billing functionality, on the start visit form."

IamSAL (commented):
changes look okay. suggested a grammar change.

PR #1543, opened 2023-12-15
(fix) O3-2528: Added a prompt to open an encounter for editing if the form is still open

elimm:
## Requirements

## Summary
If an encounter is already open for editing and you try to open another encounter for editing without closing the form, the values ​​in the form are not updated.

## Screenshots
[Before](https://monosnap.com/file/1WeUZEZpXGM0AwkMkdHRvTVJdgaONc)
[After](https://monosnap.com/file/Qsjozsj7RUi9cWSIaUJwzDaO2ETffM)

## Related Issue
[O3-2665](https://openmrs.atlassian.net/browse/O3-2665)


vasharma05:
I believe you have confused encounter with workspaces, or your PR/ ticket desciption is not accurate.

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
```diff
@@ -189,14 +189,19 @@ export function launchPatientWorkspace(name: string, additionalProps?: object) {
...
+      : {
+        ...openWorkspaces[workspaceIndexInOpenWorkspaces],
+        additionalProps: newWorkspace.additionalProps
+      };
+    const restOfWorkspaces = !!openedWorkspaceWithSameType
+      ? openWorkspaces.filter((w) => w.type != newWorkspace.type)
+      : openWorkspaces.filter((w) => w.name != name);
+
+    updateStoreWithNewWorkspace(workspaceToApply, restOfWorkspaces);
+    promptBeforeLaunchingWorkspace(workspaceToApply, {
```
ibacher, line 204: So, a few things:

1. It's preferable (I think) to leave the existing logic alone and add a new else clause that handles the case where a workspace of the same type is already open.
2. The prompting needs to happen _before_ any changes to the store and if the prompt is cancelled, no state updates should be made.
3. Please verify that all existing tests pass and, if necessary, update the failing tests to reflect the new flow.

vasharma05: Hi Ian, there is already a flow handled if a workspace of the same type is already opened. There is a prompt whenever user tries to open another workspace of same type.

denniskigen: How should this work move forward, @vasharma05? Could you give @elimm direction on how to leverage the existing functionality to achieve the goals @ibacher outlined above?

ibacher: @vasharma05 This _is_ the code that implements that flow. What @elimm is trying to add here is the ability to open _the same workspace_ with new settings, basically to be able to edit an existing encounter when a form is open in the workspace, but not enough attention was paid to the ensure that the tests that rely on the existing behavior pass.

elimm: updated the tests, made the state change dependent on the result of the prompt

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
```diff
@@ -187,6 +189,7 @@ export function launchPatientWorkspace(name: string, additionalProps?: object) {
   const openWorkspaces = store.getState().openWorkspaces;
   const workspaceIndexInOpenWorkspaces = openWorkspaces.findIndex((w) => w.name === name);
   const isWorkspaceAlreadyOpen = workspaceIndexInOpenWorkspaces >= 0;
+  const isEditingEncounter = newWorkspace?.additionalProps?.hasOwnProperty('formInfo');
```
vasharma05, line 192: Can we have something better to have a comparision that the workspace is being opened with different state?

elimm: No, unfortunately, or maybe you have some suggestions?

vasharma05: Pulling in @denniskigen for suggestion here.
Thanks!

elimm: @denniskigen @vasharma05 any updates?

senthil-k8s: const isEditingEncounter = !!newWorkspace?.additionalProps?.formInfo;
(nullish coalescing approach + optional chanining method) try with this. but the time complexity seems to be similar for both.

denniskigen:
Sorry, I'm late to this, @elimm, do you mind resolving the conflicts?

elimm:
@denniskigen any updates?

brandones:
Sorry we let this fall through, @elimm . Thanks for your work on this. It will need to be done in the new [workspace system](https://github.com/openmrs/openmrs-esm-core/tree/main/packages/framework/esm-styleguide/src/workspaces). I have updated [the ticket](https://openmrs.atlassian.net/browse/O3-2665) for clarity. Note that this does not require special-casing any workspace type—the correct logic for the edit encounter workspace is the same as for all other workspace types: if the user has not made edits, just update the workspace; if the user has made edits, then show the usual prompt.

PR #1549, opened 2023-12-19
(fix) O3-2666: Order getting removed after clicking on Back to order basket

usamaidrsk:
## Requirements

## Summary

- Fix Order getting removed after clicking on `Back to order basket`

## Related Issue

[O3-2666](https://openmrs.atlassian.net/browse/O3-2666)


packages/esm-patient-medications-app/src/add-drug-order/add-drug-order.test.tsx
```diff
@@ -147,17 +147,6 @@ describe('AddDrugOrderWorkspace drug search', () => {
...
-      expect.objectContaining({
-        ...getTemplateOrderBasketItem(
-          mockDrugSearchResultApiData[0],
-          undefined,
-          mockDrugOrderTemplateApiData[mockDrugSearchResultApiData[0].uuid][0],
-        ),
-        startDate: expect.any(Date),
-      }),
-    ]);
-
```
vasharma05, line 160: Please update the test, we should never remove a test, we should modify the test file.

packages/esm-patient-medications-app/src/add-drug-order/add-drug-order.test.tsx
```diff
@@ -172,16 +161,6 @@ describe('AddDrugOrderWorkspace drug search', () => {
...
-    expect(hookResult.current.orders).toEqual([
-      expect.objectContaining({
-        ...getTemplateOrderBasketItem(
-          mockDrugSearchResultApiData[0],
-          undefined,
-          mockDrugOrderTemplateApiData[mockDrugSearchResultApiData[0].uuid][0],
-        ),
-        startDate: expect.any(Date),
-      }),
-    ]);
```
vasharma05, line 184: Same here

packages/esm-patient-medications-app/src/add-drug-order/add-drug-order.test.tsx
usamaidrsk: Thanks, the test I removed was checking if this order was saved in orders store, which is not  the case when user visits order form and gets back to the basket, given the new requirements

CC @vasharma05

packages/esm-patient-medications-app/src/add-drug-order/add-drug-order.test.tsx
usamaidrsk: https://github.com/openmrs/openmrs-esm-patient-chart/pull/1549#discussion_r1431010787

brandones:
Hi Usama! Thanks for the PR. A note on how we handle PRs.

image

The checkboxes are there both to remind the submitter to do stuff and to communicate with others. You should not check them if they are not relevant. The one about conforming to the styleguide informs the reviewer that you have done some design implementation that references some specific designs. That isn't relevant to this ticket, so don't check the box. The one about tests requires some careful thought; the question is whether tests would fail if you didn't make the change you made. Tests were passing before (with the bad behavior), and you have only removed tests; therefore it is impossible that if your change was somehow undone that tests would fail. Therefore your change is not tested for, and you shouldn't check the box. It's not strictly a requirement, but if it's not too hard to write a test that protects us against this bug, it would be great to have.

usamaidrsk:
> Hi Usama! Thanks for the PR. A note on how we handle PRs.
> 
> image
> 
> The checkboxes are there both to remind the submitter to do stuff and to communicate with others. You should not check them if they are not relevant. The one about conforming to the styleguide informs the reviewer that you have done some design implementation that references some specific designs. That isn't relevant to this ticket, so don't check the box. The one about tests requires some careful thought; the question is whether tests would fail if you didn't make the change you made. Tests were passing before (with the bad behavior), and you have only removed tests; therefore it is impossible that if your change was somehow undone that tests would fail. Therefore your change is not tested for, and you shouldn't check the box. It's not strictly a requirement, but if it's not too hard to write a test that protects us against this bug, it would be great to have.

Thanks @brandones, It's a good guide, and I will endeavor to follow along. Also with the tests I am to add the tests as to the changes I made.

brandones (approved):
LGTM, thanks @usamaidrsk & @vasharma05

PR #1562, opened 2024-01-04
(fix) O3-2654: Unable to edit visit attributes when editing visit details

usamaidrsk:
## Requirements

## Summary

- Fix error on updating visit attributes, by adding separate handler for visit attribute update.

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-chart/assets/53287480/94917381-633c-4b1c-b07f-62686cf323ab

## Related Issue

[O3-2654](https://openmrs.atlassian.net/browse/O3-2654)

### DIFF
--- a/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
@@ -245,8 +245,32 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
     return validSubmission;
   }, [setError]);
 
+  const updateVisitAttributes = useCallback(
+    async (updatedAttributes) => {
+      for (const attribute of updatedAttributes) {
+        await openmrsFetch(`/ws/rest/v1/visit/${visitToEdit.uuid}/attribute/${attribute.uuid}`, {
+          method: 'POST',
+          headers: {
+            'Content-type': 'application/json',
+          },
+          body: {
+            value: attribute.value,
+          },
+        }).catch((err) => {
+          showSnackbar({
+            title: t('errorUpdatingVisitAttribute', 'Error updating visit attribute'),
+            kind: 'error',
+            isLowContrast: false,
+            subtitle: err?.message,
+          });
+        });
+      }
+    },
+    [visitToEdit],
+  );
+
   const onSubmit = useCallback(
-    async (data: VisitFormData, event) => {
+    (data: VisitFormData, event) => {
       if (visitToEdit && !validateVisitStartStopDatetime()) {
         return;
       }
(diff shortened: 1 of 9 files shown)


packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -316,6 +317,38 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
...
 
       if (isOnline) {
+        if (visitToEdit?.uuid) {
+          const existingAttributes = visitToEdit.attributes.map((attr) => attr.attributeType.uuid) || [];
+          const updatedAttributes = [];
+          const addedAttributes = [];
+          payload.attributes.forEach((attribute) => {
+            if (existingAttributes.includes(attribute.attributeType)) {
+              const attributeUuid = visitToEdit.attributes.filter(
+                (attr) => attr.attributeType.uuid === attribute.attributeType,
```
vasharma05, line 327: Fix: `attribute.attributeType.uuid`

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -316,6 +317,38 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
...
+                value: attribute.value,
+              },
+            }).catch((err) => {
+              showSnackbar({
+                title: t('errorUpdatingVisitAttribute', 'Error updating visit attribute'),
+                kind: 'error',
+                isLowContrast: false,
+                subtitle: err?.message,
+              });
+            });
```
vasharma05, line 349: Why are we updating the visit details even before we are updating the visit details. 

The right solution will be to first update visit details and then if the request is successful, then we update the attributes.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -316,6 +317,38 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
...
+          const existingAttributes = visitToEdit.attributes.map((attr) => attr.attributeType.uuid) || [];
+          const updatedAttributes = [];
+          const addedAttributes = [];
+          payload.attributes.forEach((attribute) => {
+            if (existingAttributes.includes(attribute.attributeType)) {
+              const attributeUuid = visitToEdit.attributes.filter(
+                (attr) => attr.attributeType.uuid === attribute.attributeType,
+              )[0].uuid;
+              updatedAttributes.push({ uuid: attributeUuid, value: attribute.value });
+            } else addedAttributes.push(attribute);
```
vasharma05, line 330: There are so many things that can fail here:
1. If attributes is undefined
2. If filter returns empty
3. If the request to update the visit attribute fails, this is assuming a very happy case, which is not always the case.

vasharma05 (changes requested):

vasharma05:
Hi @dkayiwa !
Do you know why are we not able to update the visit attributes in the visit update payload itself? Is there anything we can work out on this? Updating the visit attributes separately will bring up many edge cases.
For more info, please feel free to refer the ticket assigned.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
usamaidrsk: `attribute` is already used in the scope @vasharma05

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
usamaidrsk: > There are so many things that can fail here:
> 
> 1. If attributes is undefined
> 2. If filter returns empty
> 3. If the request to update the visit attribute fails, this is assuming a very happy case, which is not always the case.

- Upon failure, I added have a snackbar error feedback.

ibacher:
@vasharma05 @usamaidrsk The way to deal with issues like this would be to open in a Jira ticket in the Webservices.REST project that says something like: "Add ability to update attributes as part of the visit payload" with a description of what you're trying to accomplish and what syntax should be supported. You can add the tag "community-priority" in Jira and it should show up on the appropriate board. In most cases like these you're simply looking for a feature no one has ever asked for before, so ask for the feature (which is what a ticket is for).

vasharma05:
Hi @ibacher!
I actually thought that this might be a part of the well thought design, rather than not being implemented since no one asked for it. Bad assumption from my side. 

I'll create a ticket and update the ticket in the O3-2654.
Thanks a lot @ibacher !

vasharma05:
Expecting a better solution with REST Ticket: [https://openmrs.atlassian.net/browse/RESTWS-926](https://openmrs.atlassian.net/jira/software/c/projects/RESTWS/boards/59?selectedIssue=RESTWS-926)

ibacher:
@vasharma05 hmmm… actually there is an issue with updates vs creates and it comes down to ambiguity in JSON. Ideally, what we want to submit as an edit is something that records the changes we want to make. There are some specified syntaxes for this (JSON Patch and XML Patch) which we don’t support. Without either an explicit patching syntax or sending the full resource with each request, the backend has no way to differentiate between an “append this attribute” request vs a “delete all attributes but this one” request or even a “delete all attributes”.

Theoretically, the principles of REST mean that we should be sending the entire representation of the resource as part of the PUT request, so we do need to be sure (on the front end side) that the update includes the full visit payload with the modifications made.

vasharma05:
This is what I also encountered in the patient registration, when dealing with identifiers. If you see, there were cases handled where a few identifiers were updated, deleted and added simultaneously, which I belive cannot be done in a single REST request, hence I asked @usamaidrsk to follow the same steps here as well.

PR #1568, opened 2024-01-08
O3-2684: remove hard coded queue entry statuses

cioan:
@mseaton , you were correct, there was another place, the visit header where the queue statuses were hard coded. Thanks!

### DIFF
--- a/packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
@@ -59,8 +59,8 @@ const PatientInfo: React.FC<PatientInfoProps> = ({ patient }) => {
   const priority = queueEntry?.priority ?? '';
 
   const getServiceString = useCallback(() => {
-    if (queueEntry?.status) {
-      return `${queueEntry.status} - ${queueEntry.service}`;
+    if (queueEntry?.status && queueEntry.service) {
+      return `${t(queueEntry.status)} - ${t(queueEntry.service)}`;
     } else {
       return '';
     }
--- a/packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
@@ -59,15 +59,10 @@ const PatientInfo: React.FC<PatientInfoProps> = ({ patient }) => {
   const priority = queueEntry?.priority ?? '';
 
   const getServiceString = useCallback(() => {
-    switch (queueEntry?.status?.toLowerCase()) {
-      case 'waiting':
-        return `Waiting for ${queueEntry.service}`;
-      case 'in service':
-        return `Attending ${queueEntry.service}`;
-      case 'finished service':
-        return `Finished ${queueEntry.service}`;
-      default:
-        return '';
+    if (queueEntry?.status) {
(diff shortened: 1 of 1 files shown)


mseaton (approved):
It is consistent with your other changes, but would be good to get someone more familiar with this to weigh in.  @ibacher  or @denniskigen ?

packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
```diff
@@ -59,15 +59,10 @@ const PatientInfo: React.FC<PatientInfoProps> = ({ patient }) => {
...
-      case 'waiting':
-        return `Waiting for ${queueEntry.service}`;
-      case 'in service':
-        return `Attending ${queueEntry.service}`;
-      case 'finished service':
-        return `Finished ${queueEntry.service}`;
-      default:
-        return '';
+    if (queueEntry?.status) {
+      return `${queueEntry.status} - ${queueEntry.service}`;
```
ibacher, line 63: Is the `-` here necessary? We kind of want to end up in a state where it's possible to display "Waiting for xxx", "Attending xxx", etc.

ibacher: My other thought here is that it would be a good idea to run this through the translation system:

```ts
return `${t(queueEntry.status)} ${t(queueEntry.service)}`
```

This allows implementations to customize the text if they don't want to customize the concept.

mseaton: It should also be null-safe I realise.  eg. we should prepare for the possibility that queueEntry.service could be null.

cioan: Thanks @ibacher and @mseaton ! I have added the translate function, and also checked that queueEntry.service is not null.
@ibacher , the reason I added the '-' (hyphen) here is to be consistent with the other change we made to the service-queue: 
https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-service-queues-app/src/helpers/functions.ts#L20

If there is a clearer pattern for displaying statuses, I am happy to update the code in both places so the queue statuses will be displayed in a consistent way. Please let me know. Thanks!

mogoodrich: It's an interesting point about the hyphen, @ibacher , but as @cioan  pointed out we are making this consistent with another change we already made, and although I can see the desire to support the display strings of status + service to represent a phrase, I don't think we we to mandate this, if that makes any sense?

mogoodrich (approved):
Looks good to me, thanks @cioan .  Will also add @CynthiaKamau  as a reviewer.

CynthiaKamau:
Thanks @cioan for this, could you share a screenshot of how this appears now? I have something like `waiting triage`

cioan:
> Thanks @cioan for this, could you share a screenshot of how this appears now? I have something like `waiting triage`

Sure @CynthiaKamau , here how it looks on the PIH test server now after we customized the queue entry statuses:
Screenshot 2024-01-11 at 3 41 37 PM
Screenshot 2024-01-11 at 3 41 47 PM

CynthiaKamau (approved):

cioan:
Thank you @CynthiaKamau !

PR #1573, opened 2024-01-10
(refactor) Set global testTimeout in Jest config

denniskigen:
## Requirements

## Summary

This PR removes test timeouts from individual tests and moves them up into the global Jest config as a single `testTimeout` property whose value is set to `20000ms`. Hopefully this tackles the issue where tests time out non-deterministically. 

20000ms seems like a reasonable value, though it'd be worthwhile trying to lower this figure down as we move along so we strike a good balance with a value that's not too low to prematurely terminate valid long-running tests, and not too high to unnecessarily prolong the feedback loop for test failures.

## Screenshots
*None*

## Related Issue
*None*

## Other
*None*

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -39,5 +39,5 @@ module.exports = {
   testEnvironmentOptions: {
     url: 'http://localhost/',
   },
-  testTimeout: 30000,
+  testTimeout: 20000,
 };
--- a/jest.config.js
+++ b/jest.config.js
@@ -39,4 +39,5 @@ module.exports = {
   testEnvironmentOptions: {
     url: 'http://localhost/',
   },
+  testTimeout: 30000,
 };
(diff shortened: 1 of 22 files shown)


ibacher (approved):
It looks like the highest value that's being replaced here is 20000, so maybe we could try that?

Since Jest is used for unit tests, anything that's taking longer than 5000ms, IMO is either a broken test (i.e., it's executing more code than it should) or indicates that the production code is way to slow (I guess it's also possible that it means that too many tests are being executed in parallel).

samuelmale:
It also seems to work perfectly with 10000ms.

denniskigen:
Thanks for the suggestions. I'm inclined to try out a `20000ms` delay and then assess things from then on.

PR #1593, opened 2024-01-16
(chore) Enable `continue-on-error` for deploy job in CI workflow

denniskigen:
## Requirements

## Summary

~~This PR adds a new step to the CI workflow file that retries the `deploy_patient_chart` job when it fails. The `deploy_patient_chart` job takes on average around 30 minutes to deploy a build on Bamboo.~~

~~If a build is ongoing for a merged PR, and a subsequent PR gets merged and triggers the job again, the request will fail with the following error:~~

CleanShot 2024-01-16 at 11  02 33@2x

~~To mitigate this issue and avoid the need for manual intervention to re-trigger the deploy job after the pending build completes, this PR adds a step that uses the [wretry.action](https://github.com/Wandalen/wretry.action) Github Action to attempt to rerun the deploy job every 15 minutes up to 3 times. This will allow hopefully the job to complete successfully even if there are issues with the initial deployment attempt.~~

UPDATED:

This PR adds a `continue-on-error` option to the `deploy-patient-chart` job and sets its value to `true`. This makes it so that the workflow will continue to execute even if that step or job encounters an error. This is a prerequisite to additional changes earmarked for the Bamboo deployment job further downstream. See the thread below for exact details from @ibacher.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,21 +114,11 @@ jobs:
     steps:
       - name: Trigger RefApp Build
         uses: fjogeleit/http-request-action@master
+        continue-on-error: true
         with:
           url: https://ci.openmrs.org/rest/api/latest/queue/REFAPP-D3X
           method: 'POST'
           customHeaders: '{ "Authorization": "Bearer ${{ secrets.BAMBOO_TOKEN }}" }'
-
-      - name: Retry deploy job
-        uses: Wandalen/wretry.action@master
-        with:
-          action: fjogeleit/http-request-action@master
-          with: |
-            url: https://ci.openmrs.org/rest/api/latest/queue/REFAPP-D3X
-            method: 'POST'
-            customHeaders: '{ "Authorization": "Bearer ${{ secrets.BAMBOO_TOKEN }}" }'
-          attempt_limit: 3
-          attempt_delay: 900000 # ms or 15 minutes
 
   release:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,6 +119,17 @@ jobs:
           method: 'POST'
           customHeaders: '{ "Authorization": "Bearer ${{ secrets.BAMBOO_TOKEN }}" }'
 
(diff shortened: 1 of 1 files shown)


ibacher:
@denniskigen So, yeah, this needs some discussion. When I setup the project I was aware this was likely to be an issue. The underlying cause is that Bamboo only allows (for good reason) a single instance of each job to be running at once. This is one way we could fix things, but its a bit of a blunt instrument. Most of the time the Bamboo job is hanging is actually down to building the backend and most of the reason that we have only one job running at a time is that we both publish tagged images to Docker Hub and then run a deployment task. I'm thinking that a better solution here is to split the Bamboo workflow up a bit into:

1. A job that builds the backend image (responds to changes to relevant files in the distro project)
2. A job that builds the gateway image (responds to changes to relevant files in the distro project)
3. A job that builds the frontend image (responds to changes to relevant files in the distro project AND the existing GH action)
4. A job that deploys dev3

Basically, we might be able to setup Bamboo to queue up the deployment job as a child of the first three and then change this to only be submitting to the job to update the frontend image. That would likely make the existing deploy task fail less and also actually speed up getting merged changes to deploy. Plus, by queueing the slow parts separately, I think we can actually not fail the job as frequently (we may need to add an intermediate job that queues tagging the frontend and then we can allow Bamboo to run as many parallel versions of the job as we want, so this would never fail).

Ultimately, though, I think we should also update the job to not mark the build as failed if the deploy step doesn't work for this reason, because, really, it's very likely nothing went wrong.

Thoughts?

denniskigen:
Yeah, this PR was really a shot in the dark more than anything concrete. The plan you laid out is very sensible indeed and I defer to your judgement fully because of your knowledge of the Bamboo setup. Any step towards making the job less susceptible to (perceived) failure is obviously very welcome!

ibacher:
Although it kinda invalidates the description, I've updated the PR with the change I think we need here. I still need to look into the Bamboo side of things...

denniskigen:
I've updated the PR title and description

ibacher (approved):

PR #1595, opened 2024-01-17
(Fix) O3-2716: Add Workspace "unsaved changes" modal prompt on navigating away from patient chart

usamaidrsk:
## Requirements

## Summary

This PR fixes the navigation from the patient chat where the prompt on unsaved changes is not triggered. So the PR adds an event listener whenever navigating away to always prompt users if there are any open workspaces

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-chart/assets/53287480/af866801-aeed-4728-a102-4679a6f00663

## Related Issue

[O3-2716)](https://openmrs.atlassian.net/browse/O3-2716)


brandones:
Sorry @usamaidrsk , this is not the right approach. The logic for this should be contained in the workspace system itself (`workspace.ts` I think it is). Right now in the workspace system we check `promptBeforeClosing`, and then conditionally prompt, before closing the workspace. We just need to make it so we do the same before navigating. So the check and prompt can be triggered by navigation, not just by closing the workspace. Hopefully this should also be simpler than what you have implemented.

PR #1598, opened 2024-01-18
(chore) Update transifex config

befantasy:
## Requirements

## Summary

This PR updates the config to sync with the transifex project

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -1,6 +1,14 @@
 [main]
 host = https://www.transifex.com
 
+[o:openmrs:p:openmrs-esm-patient-chart:r:esm-form-engine-app]
+file_filter            = packages/esm-form-engine-app/translations/<lang>.json
+source_file            = packages/esm-form-engine-app/translations/en.json
+source_lang            = en
+type                   = KEYVALUEJSON
+replace_edited_strings = false
+keep_translations      = false
+
 [o:openmrs:p:openmrs-esm-patient-chart:r:esm-form-entry-app]
 file_filter            = packages/esm-form-entry-app/translations/<lang>.json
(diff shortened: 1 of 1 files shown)


.tx/config
```diff
@@ -1,6 +1,14 @@
 [main]
 host = https://www.transifex.com
 
+[o:openmrs:p:openmrs-esm-patient-chart:r:esm-form-engine-app]
```
denniskigen, line 4: @ibacher, is this `o:` prefix necessary?

ibacher: Yes. It's [Transifex's syntax](https://developers.transifex.com/docs/migrating-from-older-versions-of-the-client#add). (`o:` for org, `p:` for project, `r:` for resource)

ibacher (approved):

denniskigen:
Thanks, @befantasy!

.tx/config
befantasy: After checking the results and command instructions, it seems that the new version of Transifex requires an additional parameter to be configured: "resource_name". Otherwise, the new added resource name displayed on Transifex will default to "en.json". 

I will submit another PR soon.

image

ibacher: That's probably because `en.json` is treated as the "master list" of strings to be translated.

befantasy: I don't quite understand. How should we go about solving this issue?

ibacher: @befantasy Yeah, I don't think it's determined by the configuration file or it may be that the resources need to be manually renamed after they were created?

befantasy: @ibacher Oh, I see. Thank you for your patient explanation. So do you know how and by whom should the manual renaming be conducted?  I'm stuck with this issue.

PR #1607, opened 2024-01-22
(feat) O3-2716: Prompt users to close workspaces when navigating from the patient chart

usamaidrsk:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.
- [x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://zeroheight.com/23a080e38/p/880723-introduction).
- [x] My work includes tests or is validated by existing tests.

## Summary
- This PR adds a prompt if navigating from the patient chart and the user has open workspaces, to ask if they wish to close the workspaces before navigating.

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-chart/assets/53287480/b4b9a7b2-efc0-4b0f-a7d1-23c6239d521f

## Related Issue

[O3-2716](https://openmrs.atlassian.net/browse/O3-2716)

PR #1613, opened 2024-01-24
(feat) O3-2760: Present workspaces should leverage the `promptBeforeClosing` function

vasharma05:
## Requirements

## Summary
This PR updates the existing workspaces implementation to use the `promptBeforeClosing` functions in the workspaces.

This PR makes the following changes in the codebase:
1. Shifts all the workspaces registered as extensions from routes.json` to registering them via `registerWorkspace`.
2. Added workspace title translations in the index.ts
3. All the workspaces now use the `promptBeforeClosing` function. 
All the workspace forms (except the Lab order form) use React Hook Form, which returns a state `isDirty`, which tells whether the user made any change to the form or not. The `promptBeforeClosing` keeps an eye on the form state `isDirty` and decides whether to show the user a prompt when closing form without saving form changes.

### CloseWorkspace
Workspace's `closeWorkspace` now takes an object of the following keys as an argument:
1. **ignoreChanges** : Whether to just close the workspace
2. **onWorkspaceClose**: Callback function to be called when the workspace is closed.

P.S. The onWorkspaceClose function can be used in the following manner:
```
❌ 
workspace.closeWorkspace();
clearGlobalStore()

✅ 
workspace.closeWorkspace({
	onWorkspaceClose: () => clearGlobalStore();
})
```

### GetWhetherWorkspaceCanBeClosed

A new function `getWhetherWorkspaceCanBeClosed` is introduced which tells whether a workspace can be directly closed or the user will be shown a prompt before closing the function.

## Screenshots
None

## Related Issue
https://issues.openmrs.org/browse/O3-2760

## Other
Also solves [O3-2714](https://openmrs.atlassian.net/browse/O3-2714).


packages/esm-patient-common-lib/src/workspaces/workspaces.ts
```diff
@@ -23,8 +23,8 @@ export interface WorkspaceStoreState {
 
 export interface OpenWorkspace extends WorkspaceRegistration {
   additionalProps: object;
-  closeWorkspace(promptBeforeClosing?: boolean): void;
-  closeWorkspace(): void;
+  closeWorkspace(closeWorkspaceOptions?: CloseWorkspaceOptions): boolean;
+  closeWorkspace(): boolean;
```
brandones, line 27: I think this is redundant with the definition above.

packages/esm-patient-common-lib/src/types/workspace.ts
```diff
@@ -1,14 +1,21 @@
 /* The possible states a workspace window can be opened in. */
 export type WorkspaceWindowState = 'maximized' | 'hidden' | 'normal';
 
+export interface CloseWorkspaceOptions {
+  workspaceTitle?: string;
+  ignoreChanges?: boolean;
+  onWorkspaceClose?: () => void;
+  confirmBeforeClosing?: boolean;
```
brandones, line 8: These will need docstrings. `confirmBeforeClosing` appears not to be used. Let's get rid of that. And `workspaceTitle` does something strange—why would the caller of `closeWorkspace` know the title, but the title not have been provided by the workspace in registration? Maybe there should be some fallback behavior if no title is defined, but this seems strange to me.

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
```diff
@@ -103,40 +103,28 @@ function getTitleFromExtension(ext: ExtensionRegistration) {
   return ext.name;
 }
 
+function getWhetherWorkspaceCanBeClosed(name: string, ignoreChanges: boolean = false) {
+  const promptBeforeFn = getPromptBeforeClosingFcn(name);
+  return ignoreChanges || !promptBeforeFn || !promptBeforeFn();
```
ibacher, line 108: Less clever, but more readable, I think (plus this skips the function call if `ignoreChanges` is `true`).

```suggestion
function getWhetherWorkspaceCanBeClosed(name: string, ignoreChanges: boolean = false) {
  if (ignoreChanges) {
    return true;
  }
  
  const promptBeforeFn = getPromptBeforeClosingFcn(name);
  return !promptBeforeFn || !promptBeforeFn();
```

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
```diff
@@ -325,3 +315,94 @@ function getUpdatedWorkspaceWindowState(workspaceAtTop: OpenWorkspace) {
...
+    }
+
+    case 'closing-all-workspaces': {
+      const workspacesNotClosed = store
+        .getState()
+        .openWorkspaces.filter(({ name }) => !getWhetherWorkspaceCanBeClosed(name))
+        .map(({ title }, indx) => `${indx + 1}. ${title}`);
+
+      const prompt: Prompt = {
+        title: translateFrom('@openmrs/esm-patient-chart-app', 'unsavedChanges', 'You have unsaved changes'),
```
ibacher, line 354: Why is the title here different?

ibacher: I'm also a bit confused about the intended workflow here. I.e., when do we trigger this `closing-all-workspaces` prompt?

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
```diff
@@ -325,3 +315,94 @@ function getUpdatedWorkspaceWindowState(workspaceAtTop: OpenWorkspace) {
...
+        body: translateFrom(
+          '@openmrs/esm-patient-chart-app',
+          'unsavedChangesInForms',
```
ibacher, line 357: Why is this key called `unsavedChangesInForms` and not `unsavedChangesInWorkspaces`?

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
```diff
@@ -325,3 +315,94 @@ function getUpdatedWorkspaceWindowState(workspaceAtTop: OpenWorkspace) {
...
+      }));
+      return;
+    }
+    case 'closing-workspace-launching-new-workspace': {
+      const prompt: Prompt = {
+        title: translateFrom('@openmrs/esm-patient-chart-app', 'unsavedChanges', 'You have unsaved changes'),
+        body: translateFrom(
+          '@openmrs/esm-patient-chart-app',
+          'unsavedChangesInForm',
+          'There are unsaved changes in {{formName}}. Please save them before opening another form.',
```
ibacher, line 385:
```suggestion
          'There are unsaved changes in {{workspaceName}}. Please save them before opening another workspace.',
```

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
```diff
@@ -325,3 +315,94 @@ function getUpdatedWorkspaceWindowState(workspaceAtTop: OpenWorkspace) {
...
+          { formName: workspaceTitle },
+        ),
+        onConfirm: () => {
+          store.setState((prevState) => ({
+            ...prevState,
+            prompt: null,
+          }));
+          onConfirmation?.();
+        },
+        confirmText: translateFrom('@openmrs/esm-patient-chart-app', 'openAnyway', 'Open anyway'),
```
ibacher, line 395: This text should mention that the previous form will be discarded, e.g., "Discard changes and open anyway"

ibacher (commented):
Wouldn't it make sense if the `closeWorkspace()` function returned a `Promise<void>` rather than `void`? Then instead of needing all this logic to embed the `onWorkspaceClose()` callback, callers that need to do something after the workspace closes can just do:

```ts
closeWorkspace().then(/* do something */);
```

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
vasharma05: Hi @ibacher !
The `closing-all-workspaces` is being implemented in #1607 .

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
vasharma05: I agree, will make the change

usamaidrsk:
@vasharma05 order is cleared upon prompting.

https://github.com/openmrs/openmrs-esm-patient-chart/assets/53287480/9d3ce3ed-58dd-4c1c-a037-5d7249750d76

vasharma05:
Thank you @usamaidrsk , I'll fix it!

vasharma05:
> Wouldn't it make sense if the closeWorkspace() function returned a Promise<void> rather than void? Then instead of needing all this logic to embed the onWorkspaceClose() callback, callers that need to do something after the workspace closes can just do:
```
closeWorkspace().then(/* do something */);
```

HI @ibacher !
I tried implementing an async `closeWorkspace` function. The implementation of resolving the Promise is similar to `onWorkspaceClose`, but the rejection of the `closeWorkspace` promise will require changes as well, which I found a bit complex. If we don't count the rejection, the promise will always be in a pending state otherwise.
Thanks!

vasharma05:
Hi @ibacher @brandones @denniskigen !
The PR is ready for review.
Thanks!

packages/esm-patient-common-lib/src/types/workspace.ts
```diff
@@ -1,14 +1,38 @@
...
-   * prompted to save changes before closing, even if the `testFcn` passed to `promptBeforeClosing`
-   * returns `true`.
+   * Call this function to close the workspace.
+   *
+   * Whenever we close the workspace with `workspace.closeWorkspace()`, the `ignoreChanges`
+   * is set to true by default.
+   *
+   * Whenever you close the opened workspace outside the workspace window, always close the workspace as
+   * `closeWorkspace({ ignoreChanges: false })`, taking care to prompting the user if there
+   * are any pending changes in the workspace.
```
brandones, line 33: If the usually desirable behavior is `ignoreChanges: false`, why isn't that the default?

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
```diff
@@ -103,40 +102,36 @@ function getTitleFromExtension(ext: ExtensionRegistration) {
   return ext.name;
 }
 
+/**
+ *
+ * @param name Name of the workspace
+ * @param ignoreChanges Whether to forcefully ignore changes in the workspace
```
brandones, line 108:
```suggestion
 * @param ignoreChanges If set to true, the "unsaved changes" modal will never be shown, even if the `promptBeforeClosing` function returns true. The user will not be prompted before closing.
```

packages/esm-patient-labs-app/src/lab-orders/lab-order-basket-panel/lab-order-basket-panel.extension.tsx
```diff
@@ -20,11 +25,17 @@ export default function LabOrderBasketPanelExtension() {
   const [isExpanded, setIsExpanded] = useState(orders.length > 0);
 
   const openNewLabForm = useCallback(() => {
-    launchPatientWorkspace('add-lab-order');
+    closeWorkspace('order-basket', {
+      ignoreChanges: true,
```
brandones, line 29: Doesn't the `order-basket` workspace always ignore changes (it has nothing to change)?

packages/esm-patient-lists-app/src/workspaces/patient-lists.workspace.tsx
```diff
@@ -34,8 +38,9 @@ function PatientListsWorkspace() {
   const { patientLists, isLoading } = usePatientLists();
 
   const launchListDetailsWorkspace = useCallback((list) => {
-    closeWorkspace('patient-lists', true);
-    launchPatientWorkspace('patient-list-details', { list, workspaceTitle: list.name });
+    closeWorkspace({
+      onWorkspaceClose: () => launchPatientWorkspace('patient-list-details', { list, workspaceTitle: list.name }),
+    });
```
brandones, line 43: I thought `patient-lists` and `patient-list-details` have the same type and therefore `patient-lists` would automatically get closed?

packages/esm-patient-medications-app/src/add-drug-order/drug-search/drug-search.component.tsx
```diff
@@ -8,15 +8,20 @@ import styles from './order-basket-search.scss';
...
+export default function DrugSearch({ openOrderForm, promptBeforeClosing }: DrugSearchProps) {
   const { t } = useTranslation();
   const isTablet = useLayoutType() === 'tablet';
   const [searchTerm, setSearchTerm] = useState('');
   const debouncedSearchTerm = useDebounce(searchTerm);
   const searchInputRef = useRef(null);
 
+  useEffect(() => {
+    promptBeforeClosing(() => !!searchTerm);
+  }, [searchTerm]);
```
brandones, line 23:
```suggestion
  useEffect(() => {
    promptBeforeClosing(() => false);
  }, []);
```

It's a search box, there's not really anything to save there.

packages/esm-patient-medications-app/src/add-drug-order/drug-search/order-basket-search-results.component.tsx
```diff
@@ -141,8 +141,10 @@ const DrugSearchResultItem: React.FC<DrugSearchResultItemProps> = ({ drug, openO
       // Directly adding the order to basket should be marked as incomplete
       searchResult.isOrderIncomplete = true;
       setOrders([...orders, searchResult]);
-      closeWorkspace('add-drug-order', true);
-      launchPatientWorkspace('order-basket');
+      closeWorkspace('add-drug-order', {
+        ignoreChanges: true,
```
brandones, line 145:
```suggestion
```

Per the above change

brandones (changes requested):
Awesome work, @vasharma05 . Just a few suggestions and questions. Let's try to get this in quick since it covers so much ground. Let me know if there's any way I can help with that.

packages/esm-patient-lists-app/src/workspaces/patient-lists.workspace.tsx
vasharma05: They have the same type, so yes, this can be handled by the order thing, I'll get on it.
Thanks for the catch!

packages/esm-patient-medications-app/src/add-drug-order/drug-search/order-basket-search-results.component.tsx
vasharma05: This requires the `ignoreChanges: true`, since the order basket looks for unsaved orders, before it can be closed.
If there is an unsaved order and you try to open any other workspace (drug form/ lab form), it will show the prompt.

packages/esm-patient-common-lib/src/types/workspace.ts
vasharma05: Yes, it's right to have the `ignoreChanges` as false by default. The following are the reasons why it should be `true` by default:

`closeWorkspace` is mainly called at 3 places: The discard button in the form, which should close the workspace directly, the submit button, which also closes the workspace directly, after the endpoint/ process is completed, and the X button at the top. For the X button, I have already passed the `ignoreChanges: false` in the `workspace-window.component.tsx` since it shouldn't directly close the workspace.

Previously, the ignoreChanges was by default set to true, and all the implementations directly call the `closeWorkspace` to close the workspace, which was hence kept the same way in the new implementation. 

Yes, the default behavior should be, not to send the `ignoreChanges`, hence, I am thinking of adding 2 new functions, `discardChangesAndCloseWorkspace` and `closeWorkspaceWithSavedChanges`.

packages/esm-patient-medications-app/src/add-drug-order/drug-search/order-basket-search-results.component.tsx
brandones: The order basket looks for unsaved orders? I thought we're closing the add-drug-order workspace?

packages/esm-patient-common-lib/src/types/workspace.ts
brandones: I think `ignoreChanges: false` is the preferable default. I think it is fine if we break existing implementations to make it the default. It is likely what the user wants in most cases anyway. It's not going to make anyone's implementation crash or anything. If implementers want to ignore changes they can add the parameter. Interested in @denniskigen and @ibacher 's takes as well.

packages/esm-patient-medications-app/src/add-drug-order/drug-search/order-basket-search-results.component.tsx
vasharma05: My bad, actually I added the prompt for search as well, which you then suggested to remove.

brandones:
@vasharma05 Please re-request review when ready. I've got my eye on this.

packages/esm-patient-chart-app/src/root.component.tsx
```diff
@@ -44,6 +44,13 @@ export default function Root() {
  * t('error', 'Error')
  * t('seeAll', 'See all')
  * t('paginationItemsCount', `{{pageItemsCount}} / {{count}} items`, { count: totalItems, pageItemsCount });
- * t('unsavedChangesInForms', 'There are unsaved changes in the following workspaces. Do you want to discard changes in the following workspaces? {{workspaceNames}}')
- * t('closeWorkspaces', 'Discard changes in {{count}} workspaces', {count})
+
+ * t('unsavedChangesInOpenedWorkspace', 'You have unsaved changes in the opened workspace. Do you want to discard these changes?')
+ * t('unsavedChangesTitleText', 'Unsaved Changes')
```
denniskigen, line 49:
```suggestion
 * t('unsavedChangesTitleText', 'Unsaved changes')
```

packages/esm-patient-chart-app/src/root.component.tsx
```diff
@@ -44,6 +44,13 @@ export default function Root() {
  * t('error', 'Error')
  * t('seeAll', 'See all')
  * t('paginationItemsCount', `{{pageItemsCount}} / {{count}} items`, { count: totalItems, pageItemsCount });
- * t('unsavedChangesInForms', 'There are unsaved changes in the following workspaces. Do you want to discard changes in the following workspaces? {{workspaceNames}}')
- * t('closeWorkspaces', 'Discard changes in {{count}} workspaces', {count})
+
+ * t('unsavedChangesInOpenedWorkspace', 'You have unsaved changes in the opened workspace. Do you want to discard these changes?')
```
denniskigen, line 48:
```suggestion
 * t('unsavedChangesInOpenedWorkspace', 'You have unsaved changes in the currently opened workspace. Do you want to discard these changes?')
```

packages/esm-patient-chart-app/src/root.component.tsx
```diff
@@ -44,6 +44,13 @@ export default function Root() {
...
+ * t('unsavedChangesTitleText', 'Unsaved Changes')
+ * t('discard','Discard')
+ * t('closingAllWorkspacesPromptTitle','You have unsaved changes')
+ * t('closingAllWorkspacesPromptBody', 'There are unsaved changes in the following workspaces. Do you want to discard changes in the following workspaces? {{workspaceNames}}')
+ * t('closeAllOpenedWorkspaces', 'Discard changes in {{count}} workspaces', {count})
+ * t('unsavedChangesInWorkspace', 'There are unsaved changes in {{workspaceName}}. Please save them before opening another workspace.')
```
denniskigen, line 54: Nit: I'm not sure that relying on workspace names for this particular translation is the most ideal approach because they're so arbitrary. That said, I don't have a better alternative.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -442,7 +444,7 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
           (offlineVisit) => {
             //setCurrentVisit(patientUuid, offlineVisit.uuid);
```
denniskigen, line 445: Should we remove this?

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -664,7 +661,7 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
           </Stack>
         </div>
         <ButtonSet className={isTablet ? styles.tablet : styles.desktop}>
-          <Button className={styles.button} kind="secondary" onClick={() => closeWorkspace(ignoreChanges)}>
+          <Button className={styles.button} kind="secondary" onClick={() => discardChangesAndCloseWorkspace()}>
```
denniskigen, line 664:
```suggestion
          <Button className={styles.button} kind="secondary" onClick={discardChangesAndCloseWorkspace}>
```

packages/esm-patient-chart-app/translations/en.json
```diff
@@ -18,8 +18,10 @@
   "causeOfDeath_title": "Cause of Death",
   "checkFilters": "Check the filters above",
   "close": "Close",
-  "closeWorkspaces_one": "Discard changes in {{count}} workspaces",
-  "closeWorkspaces_other": "Discard changes in {{count}} workspaces",
+  "closeAllOpenedWorkspaces_one": "Discard changes in {{count}} workspaces",
+  "closeAllOpenedWorkspaces_other": "Discard changes in {{count}} workspaces",
+  "closingAllWorkspacesPromptBody": "There are unsaved changes in the following workspaces. Do you want to discard changes in the following workspaces? {{workspaceNames}}",
```
denniskigen, line 23: It doesn't seem right to repeat the phrase 'following workspaces' twice. Also, wouldn't it be better to list out the open workspaces as list items?

```suggestion
  "closingAllWorkspacesPromptBody": "There are unsaved changes in several workspaces. Do you want to discard changes in the following workspaces? {{workspaceNames}}",
```

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
```diff
@@ -103,40 +103,37 @@ function getTitleFromExtension(ext: ExtensionRegistration) {
   return ext.name;
 }
 
+/**
+ *
+ * @param name Name of the workspace
+ * @param ignoreChanges If set to true, the "unsaved changes" modal will never be shown, even if the `promptBeforeClosing` function returns true. The user will not be prompted before closing.
+ * @returns true if the workspace can be closed.
+ */
+export function getWhetherWorkspaceCanBeClosed(name: string, ignoreChanges: boolean = false) {
```
denniskigen, line 112: Nit: I wish this was named differently. Don't have a better name just yet.

packages/esm-patient-conditions-app/src/conditions/conditions-form.component.tsx
```diff
@@ -64,7 +77,7 @@ const ConditionsForm: React.FC<ConditionFormProps> = ({ closeWorkspace, conditio
       <Form className={styles.form} onSubmit={methods.handleSubmit(onSubmit, onError)}>
         <ConditionsWidget
           patientUuid={patientUuid}
-          closeWorkspace={closeWorkspace}
+          discardChangesAndCloseWorkspace={discardChangesAndCloseWorkspace}
```
denniskigen, line 80: Nit: it's usually a bit cleaner to name the function prop with the `on` prefix, e.g. `onDiscardChangesAndCloseWorkspace` here.

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
brandones: Yeah good thought @denniskigen .

`shouldPromptBeforeClosingWorkspace`
`workspaceHasUnsavedChanges`

Also why is this exported? I wouldn't want this to be part of the API.

packages/esm-patient-chart-app/src/root.component.tsx
vasharma05: Since we had this in the previous versions as well, I say we move ahead with this

vasharma05:
I had a discussion with Ciaran today in the O3 UI/UX Design hours, and he said that the `Discard` button in the workspace will not directly close the workspace, but will show the prompt if there are changes in the form and the user can click the `Discard Changes` button in the prompt to close the form.

Hence I will be reverting on the `discardChangesAndCloseWorkspace` and replace it back with `closeWorkspace` and `ignoreChanges = false`.

So sorry everyone for hanging this PR so long. Thank you for bearing with me and I'll finalize the PR for final review.
Thanks a lot!

brandones:
Good call talking to Ciaran about it! Thanks for doing such a thorough job.

denniskigen:
Do let us know when you're ready so we can do a fresh pass of reviews, @vasharma05! This is very consequential work, thanks for taking it on!

vasharma05:
Hi @denniskigen!
The final changes are pushed and the PR is ready for final review.
CC: @brandones @ibacher

packages/esm-patient-allergies-app/src/allergies/allergies-form/allergy-form.component.tsx
```diff
@@ -154,7 +166,7 @@ function AllergyForm({ closeWorkspace, patientUuid }: DefaultWorkspaceProps) {
           (response: FetchResponse) => {
             if (response.status === 201) {
               mutate();
-              closeWorkspace();
+              closeWorkspace({ ignoreChanges: true });
```
brandones, line 169: Huh, I see. Now that `ignoreChanges` is false by default, we have to pass this argument on saving & exiting, which is a little awkward. Nevertheless I still think this is the right approach.

packages/esm-patient-common-lib/src/workspaces/workspaces.ts
```diff
@@ -331,39 +333,68 @@ export function closeAllWorkspaces(onClosingWorkspaces: () => void = () => {}) {
...
-export function showWorkspacePrompts(promptType: PromptType, onConfirmation: () => void = () => {}) {
+/**
+ * Which type of prompt should be shown to the user.
+ * @param promptType 'closing-workspace' | 'closing-all-workspaces' | 'closing-workspace-launching-new-workspace'
+ * @param onConfirmation Function to be called after the user confirms to close the workspace
+ * @param workspaceTitle Workspace title to be shown in the prompt
+ * @returns
+ */
+
+export function showWorkspacePrompts(
```
brandones, line 346: I don't really like this as a pattern; I think it's inappropriate unless the same prompts are being launched from many different places in the code. It is better to have the content of the prompt housed in the context that it is relevant to. But that's just sort of a note for posterity; let's get this PR in.

packages/esm-patient-allergies-app/src/allergies/allergies-form/allergy-form.component.tsx
brandones: Actually, I guess the thing that is weird is actually that `isDirty` is still true after the form is submitted. Logically, it seems like it shouldn't be.

brandones (approved):
LGTM. I agree with @denniskigen 's suggestions about copy but this is a big PR that's been around for too long already. Requesting his forgiveness and that we fix the copy after merging this.

jayasanka-sack:
@vasharma05, the program enrollment test has been failing since this PR. Can we sort it out pronto? New PRs are feeling the heat. Let's stick to the branch protection rules and avoid merging when checks are failing.

vasharma05:
Sure @jayasanka-sack !
I'll fix this first thing in the morning

jayasanka-sack:
Thanks, Vineet! ❤️

denniskigen:
> LGTM. I agree with @denniskigen 's suggestions about copy but this is a big PR that's been around for a while already. Requesting his forgiveness and that we fix the copy after merging this.
> 
> Thanks so much for all your work on this @vasharma05 !

Coming back to this like the Proverbial grammar Sith lord. Can we fix the copy?

brandones:
I'll do this after we move workspaces into core.

I have an idea: sometimes workspace titles are a little funny. But we can make workspace names whatever we want; there's no reason they need to be lowercased-and-hyphenated. Why don't we make workspace names exactly what should appear in notifications about that workspace? e.g. the allergy form, which has title `"Record a new allergy"` could have name `"Allergy Form"`. So then when you try to close it you see "You have unsaved changes in the Allergy Form." What do you think of that, @denniskigen ?

ibacher:
I think the only issue with that is it needs to be translatable, since, presumably, we don’t want to need a separate allergy workspace per language, but I like the idea.

PR #1626, opened 2024-01-29
(fix) O3-2658: Show attachment descriptions when available

Vijaykv5:
## Requirements

## Summary

- The Title for Attachments thumbnail has being removed from every attachments.
- The description of the attachment preview is changed to `title` as we are placing the `image description` in title.
- Major fix in POST API was fixed [here](https://github.com/openmrs/openmrs-esm-core/pull/908). 

## Screenshots

Screenshot 2024-01-29 at 11 51 42 PM

## Related Issue

Screenshot 2024-01-29 at 11 54 15 PM

[03-2658 
](https://openmrs.atlassian.net/browse/O3-2658)
Reference issue in API : https://github.com/openmrs/openmrs-esm-core/pull/908
## Other

The Name of the file attached is removed as it doesn't make much sense for the clinician.


packages/esm-patient-attachments-app/src/attachments/attachment-thumbnail.component.tsx
```diff
@@ -13,7 +13,7 @@ export default function AttachmentThumbnail(props: AttachmentThumbnailProps) {
 function ImageThumbnail(props: ImageProps) {
   return (
     <div className={styles.imageThumbnail} role="button" tabIndex={0} onClick={props?.onClick}>
-      <img src={props.src} alt={props.title} style={props.style} />
+      <img src={props.src} style={props.style} />
```
denniskigen, line 16: We'll still need to show some sort of `alt` text.

Vijaykv5: Oh I see!
Since we are not using imageProps anywhere else I though to clean up the code by removing the title property.
What can be done on case of `alt` ? Make the title in image props?

denniskigen: You want to have some sort of alt text to show in case the image doesn't get rendered for whatever reason. Also, could you show the image caption or description just above the date so things are aligned with the [design](https://zpl.io/bWG8yg1)?

Vijaykv5: cool!
I've made it back `alt` to image.title as we don't have description and changed the typography too
@denniskigen

denniskigen: Excellent. Thanks, @Vijaykv5!

denniskigen:
@ibacher it appears that the attachments resource doesn't have a provision for recording a description of an attachment. In lieu of this, should we weigh whether to show the image description instead of the image title (which might be not be particularly informative in many cases, unless the user explicitly amends it to make it so).

ibacher:
@denniskigen Problem is that the title is also treated as, effectively, the name of the file when it's downloaded, so I don't know that we'd want to do that. There is an `instructions` field that might be useful for holding a description? I'm not sure if it's exposed via the REST endpoint, but that would be easy to fix.

denniskigen:
Cool, can we do that then, @ibacher?

Vijaykv5:
> @denniskigen Problem is that the title is also treated as, effectively, the name of the file when it's downloaded, so I don't know that we'd want to do that. There is an `instructions` field that might be useful for holding a description? I'm not sure if it's exposed via the REST endpoint, but that would be easy to fix.

I think as of now we are not having an `instruction` field in Endpoint

ibacher:
Actually, things will work properly if we just wire the description field to the "comments" attribute and display that. What appears to be happening is I upload a file, give it an image name and that is set as both the file name (`title` on the backend) and `comments`.

Basically a file upload looks like this:

```
-----------------------------37729685656089254093382668808
Content-Disposition: form-data; name="fileCaption"

bytes.png
-----------------------------37729685656089254093382668808
Content-Disposition: form-data; name="patient"

eb0ea38e-f414-4cec-819d-6e9ae9239d4a
-----------------------------37729685656089254093382668808
Content-Disposition: form-data; name="file"; filename="bytes.png"
Content-Type: image/png

CONTENTS
```

As you can see, we're setting the name `bytes.png` twice. The first one (the `fileCaption` field) should be set to whatever is in the description.

ibacher:
And here's a ticket I created almost a year ago to get this added: https://openmrs.atlassian.net/browse/ATT-52

denniskigen:
Is it fair to say that this should be a prerequisite for merging this in?

ibacher:
> Is it fair to say that this should be a prerequisite for merging this in?

I think so. I think there's a little bit more revision of the attachments widgets required once that change gets made.

Vijaykv5:
So do we need to wait until once the prerequisite to complete? 

> Actually, things will work properly if we just wire the description field to the "comments" attribute and display that. What appears to be happening is I upload a file, give it an image name and that is set as both the file name (`title` on the backend) and `comments`.
> 
> Basically a file upload looks like this:
> 
> ```
> -----------------------------37729685656089254093382668808
> Content-Disposition: form-data; name="fileCaption"
> 
> bytes.png
> -----------------------------37729685656089254093382668808
> Content-Disposition: form-data; name="patient"
> 
> eb0ea38e-f414-4cec-819d-6e9ae9239d4a
> -----------------------------37729685656089254093382668808
> Content-Disposition: form-data; name="file"; filename="bytes.png"
> Content-Type: image/png
> 
> CONTENTS
> ```
> 
> As you can see, we're setting the name `bytes.png` twice. The first one (the `fileCaption` field) should be set to whatever is in the description.
> 
> We might need a ticket to be able to display the _filename_, but I don't know that that's critical.

denniskigen:
Work on the ticket you shared would likely obviate https://github.com/openmrs/openmrs-esm-core/pull/908 too, @ibacher

ibacher:
So I fixed that ticket and updated the distribution. Should be available on dev3 shortly.

Vijaykv5:
Took a look!
Great @ibacher 
So can I close this PR now?

Vijaykv5:
@denniskigen
May I know what I need to do with this PR?

denniskigen:
State of things following the most recent commit: 

https://github.com/openmrs/openmrs-esm-patient-chart/assets/8509731/3f6b4c87-9f47-447c-99fb-70a6a401ce0b

- The image description is now showing in the image preview UI.
- The attachments grid is now set to render 3 columns. Previously, it was set to show 4 columns. The UI would break if you had 4 attachment previews rendered in the grid as the width of the fourth item would be constrained by the max-width of the parent container element.
- The attachments app is now using the most up to date type annotations from the framework.
- The overview widget now has a loading state that's shown when data is getting fetched from the API.

denniskigen:
@ibacher just noticed that the global property `attachments.allowedFileExtensions` doesn't seem configured to support any specific extensions. Do we want to configure that?

denniskigen:
I should probably remove this as it's misleading without any corresponding checks:

CleanShot 2024-02-03 at 12  09 48@2x

denniskigen:
Thanks for working on this, @Vijaykv5. I've tacked on many other useful refactors on top of your work.

denniskigen (approved):

Vijaykv5:
Thank you @denniskigen :)

PR #1636, opened 2024-02-03
(fix) O3-2807: Quantity units default to dose unit when the quantity is not zero

mccarthyaaron:
## Requirements

## Summary
In the drug-order-form component, the dispensing quantity units will default to the dosage units if the quantity to be dispensed is greater than 0 while still giving the user the option to select their preferred quantity units. I attempt to achieve the result in two ways. Firstly changing the dosage units will trigger a corresponsing change in the quantity units but only if the quantity to be dispensed is not 0. Secondly, a change in the quantity to be dispensed to a value greater than 0 will trigger the quantity units to take on the value of the dosage units at the time. In both cases the user can still change the quantity units to a value other than the one defined in the dosing units.

## Related Issue
https://openmrs.atlassian.net/browse/O3-2807

### DIFF
--- a/packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
+++ b/packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
@@ -166,6 +166,10 @@ export function DrugOrderForm({ initialOrderBasketItem, onSave, onCancel }: Drug
     },
   });
 
+  const {
+    field: { onChange: dispensingUnitsOnChange },
+  } = useController<MedicationOrderFormData>({ name: 'quantityUnits', control });
+
   const routeValue = watch('route')?.value;
   const unitValue = watch('unit')?.value;
   const dosage = watch('dosage');
@@ -368,6 +372,7 @@ export function DrugOrderForm({ initialOrderBasketItem, onSave, onCancel }: Drug
                         control={control}
                         name="unit"
                         type="comboBox"
+                        dispensingUnitsOnChange={dispensingUnitsOnChange}
                         size={isTablet ? 'lg' : 'md'}
                         id="dosingUnits"
                         items={drugDosingUnits}
@@ -732,16 +737,32 @@ const ControlledFieldInput = ({ name, control, type, ...restProps }) => {
       );
 
     if (type === 'comboBox')
-      return (
-        <ComboBox
-          selectedItem={value}
-          onChange={({ selectedItem }) => onChange(selectedItem)}
-          onBlur={onBlur}
-          ref={ref}
-          className={fieldState?.error?.message && styles.fieldError}
-          {...restProps}
-        />
-      );
+      if (name === 'unit') {
+        return (
+          <ComboBox
+            selectedItem={value}
+            onChange={({ selectedItem }) => {
+              onChange(selectedItem);
+              restProps.dispensingUnitsOnChange?.(selectedItem);
+            }}
+            onBlur={onBlur}
+            ref={ref}
+            className={fieldState?.error?.message && styles.fieldError}
+            {...restProps}
+          />
+        );
+      }
+
+    return (
+      <ComboBox
+        selectedItem={value}
+        onChange={({ selectedItem }) => onChange(selectedItem)}
+        onBlur={onBlur}
+        ref={ref}
+        className={fieldState?.error?.message && styles.fieldError}
+        {...restProps}
+      />
+    );
 
     return null;
   }, [fieldState?.error?.message, onBlur, onChange, ref, restProps, type, value]);
--- a/packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
+++ b/packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
@@ -372,6 +372,7 @@ export function DrugOrderForm({ initialOrderBasketItem, onSave, onCancel }: Drug
                         control={control}
                         name="unit"
                         type="comboBox"
+                        watch={watch}
                         dispensingUnitsOnChange={dispensingUnitsOnChange}
                         size={isTablet ? 'lg' : 'md'}
                         id="dosingUnits"
@@ -743,7 +744,10 @@ const ControlledFieldInput = ({ name, control, type, ...restProps }) => {
             selectedItem={value}
             onChange={({ selectedItem }) => {
               onChange(selectedItem);
-              restProps.dispensingUnitsOnChange?.(selectedItem);
+              const dispensingQuantity = restProps.watch?.('pillsDispensed');
+              if (dispensingQuantity && Number(dispensingQuantity) > 0) {
+                restProps.dispensingUnitsOnChange?.(selectedItem);
+              }
             }}
             onBlur={onBlur}
             ref={ref}
--- a/packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
+++ b/packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo, useRef, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import classNames from 'classnames';
 import { useTranslation } from 'react-i18next';
 import capitalize from 'lodash-es/capitalize';
@@ -169,6 +169,16 @@ export function DrugOrderForm({ initialOrderBasketItem, onSave, onCancel }: Drug
   const {
     field: { onChange: dispensingUnitsOnChange },
   } = useController<MedicationOrderFormData>({ name: 'quantityUnits', control });
+  const {
+    field: { value: dispensingQuantity },
+  } = useController<MedicationOrderFormData>({ name: 'pillsDispensed', control });
+
+  useEffect(() => {
+    if (dispensingQuantity && Number(dispensingQuantity) > 0) {
+      const currentDosageUnit = watch('unit');
+      dispensingUnitsOnChange(currentDosageUnit);
+    }
+  }, [dispensingQuantity]);
 
   const routeValue = watch('route')?.value;
   const unitValue = watch('unit')?.value;


vasharma05:
> only if the quantity to be dispensed is not 0.
Any specific reason why we are looking into quantity here?
Thanks!

vasharma05:
Also, the ticket is not understood properly:
The title says: `Quantity Units should be required when a quantity to dispense is specified`, which is more about the form validation, which is not taken care of here.

denniskigen:
Could you add a test as well?

vasharma05:
Hi, @denniskigen!
Currently, there are no tests written for the drug order form right now.
I can write the tests in a separate file.
Thanks!

mccarthyaaron:
@ibacher could you kindly take a look at the changes I have a pushed and suggest any modifications

vasharma05:
Hi @ibacher!
 > By default neither dose unit nor quantity unit is set (unless specified in an order template)
 
No, it's also defined in the drug's object, we define the default dose unit from drug's `dosageForm`:https://github.com/openmrs/openmrs-esm-patient-chart/blob/7222e818a1c158497a27f66965c059167f9cc309/packages/esm-patient-medications-app/src/add-drug-order/drug-search/drug-search.resource.tsx#L136

vasharma05 (approved):

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
```diff
@@ -177,8 +177,8 @@ export function DrugOrderForm({ initialOrderBasketItem, onSave, onCancel }: Drug
     newValue: MedicationOrderFormData['unit'],
     prevValue: MedicationOrderFormData['unit'],
   ) => {
-    if (prevValue?.valueCoded === watch('quantityUnits')?.valueCoded) {
-      setValue('quantityUnits', newValue);
+    if (!watch('quantityUnits')) {
```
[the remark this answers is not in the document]
mccarthyaaron: > * If the dose unit is set (including by order template), but the quantity unit is unset, the quantity unit is set to the dose unit
> * In no other circumstance do we automatically change either the dose unit or quantity unit

@vasharma05 with the previous condition, if the dose unit and quantity unit were the same, then changing the dose units was also changing the quantity unit to the new value of the dose unit. Yet according to @ibacher the dose unit should only change the quantity unit if the quantity unit is unset. Also bullet 2 above emphasizes it again

denniskigen:
Maybe tangential to this work, but the forms inputs ought to have an error outline when validation fails, which is not the case here:

CleanShot 2024-02-09 at 2  44 38@2x

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
```diff
@@ -68,21 +68,34 @@ const schemaFields = {
...
+  ])
+  .refine(
+    (formValues) => {
+      if (formValues.pillsDispensed > 0) {
+        return Boolean(formValues.quantityUnits);
+      }
+      return true;
+    },
+    {
+      message: 'Please select Quantity unit',
```
denniskigen, line 95:
```suggestion
      message: 'Please select a quantity unit',
```

@vasharma05, could you guide @mccarthyaaron on how to make these error messages translatable? We should probably add an example to our [Coding Conventions](https://o3-docs.openmrs.org/docs/coding-conventions).

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
```diff
@@ -675,31 +705,55 @@ const CustomNumberInput = ({ setValue, control, name, labelText, ...inputProps }
...
-const ControlledFieldInput = ({ name, control, type, ...restProps }) => {
+interface ControlledFieldInputProps {
+  name: keyof MedicationOrderFormData;
+  type: 'toggle' | 'checkbox' | 'number' | 'textArea' | 'textInput' | 'comboBox';
+  handleAfterChange?: (
+    newValue: MedicationOrderFormData[keyof MedicationOrderFormData],
+    prevValue: MedicationOrderFormData[keyof MedicationOrderFormData],
+  ) => void;
+  control: Control<MedicationOrderFormData>;
+  [x: string]: any;
```
denniskigen, line 716: Can we get a better type here than `any`. `any` completely bails out of the type system.

denniskigen (commented):
Thanks for working on this, @mccarthyaaron!

@ibacher I've tested things out locally against the functional requirements you laid out, and they seem to work OK for me. Do you have any qualms about the approach?

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
```diff
@@ -143,7 +156,7 @@ export function DrugOrderForm({ initialOrderBasketItem, onSave, onCancel }: Drug
     return initialOrderBasketItem?.startDate as Date;
   }, [initialOrderBasketItem?.startDate]);
 
-  const { handleSubmit, control, watch, setValue } = useForm<MedicationOrderFormData>({
+  const methods = useForm<MedicationOrderFormData>({
```
ibacher, line 159: Is there a reason for the `methods` variable?

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
```diff
@@ -165,6 +178,22 @@ export function DrugOrderForm({ initialOrderBasketItem, onSave, onCancel }: Drug
...
+    watch,
+    setValue,
+    formState: { errors },
+  } = methods;
+
+  const handleUnitAfterChange = (
+    newValue: MedicationOrderFormData['unit'],
+    prevValue: MedicationOrderFormData['unit'],
+  ) => {
+    if (prevValue?.valueCoded === watch('quantityUnits')?.valueCoded) {
```
ibacher, line 193: This can probably be substituted (more correctly) with `getValues()` (also returned from `useForm()`. `watch()` is intended to subscribe to input changes, but here it's used to get a point-in-time value.

ibacher: Since this is passed as a prop to a component, it would be best to wrap this in `useCallback()`. Otherwise, each render will re-render the `ControlledFieldInput`.

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
ibacher: Properly typing this requires a discriminated union, since the values of `restProps` depends on the type. Something like:

```ts
type ControlledFieldInputProps = BaseControlledFieldInputProps & (
  ToggleFieldProps | CheckboxFieldProps | NumberFieldProps | TextAreaProps | TextInputProps | ComboBoxProps
)

interface BaseControlledFieldInputProps {
  name: keyof MedicationOrderFormData;
	handleAfterChange?: (
	   newValue: MedicationOrderFormData[keyof MedicationOrderFormData],
	   prevValue: MedicationOrderFormData[keyof MedicationOrderFormData],
	 ) => void;
   control: Control<MedicationOrderFormData>;
}

interface ToggleFieldProps extends ToggleProps /* part of Carbon */ {
  type: 'toggle',
}

// repeat for other types
```

And that should give us something fully typed.

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
```diff
@@ -675,31 +705,55 @@ const CustomNumberInput = ({ setValue, control, name, labelText, ...inputProps }
...
+  handleAfterChange,
+  ...restProps
+}: ControlledFieldInputProps) => {
   const {
     field: { onBlur, onChange, value, ref },
     fieldState,
   } = useController<MedicationOrderFormData>({ name: name, control });
 
+  const handleChange = (newValue: MedicationOrderFormData[keyof MedicationOrderFormData]) => {
+    const prevValue = watch?.(name);
```
ibacher, line 733: Again, for this purpose, `getValues()` is a better choice than `watch`. This should also be wrapped in a `useCallback()` hook.

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
ibacher: So, I don't think the logic here avoids the edge-case where the user has selected a `quantityUnits` value that happens to be the same as the `dosingUnits`, but then changes the `dosingUnits`. I'm ok with automatically changing the value as long as its been automatically determined, but user input (especially for medication instructions) needs to always override automatic determinations.

ibacher (commented):
I'm mostly fine with the logic. A few small nits and one edge-case requirement. Thanks for the great work @mccarthyaaron!

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
mccarthyaaron: Will be eager to learn how to do that. Thanks

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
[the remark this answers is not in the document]
mccarthyaaron: @vasharma05 @ibacher could you kindly offer some clarity on this; should a change in the dose unit trigger a change in the quantity unit if they the values of both are the same? Which happens to be what @vasharma05 wants. What I understood from @ibacher is that the dose unit should only change the quantity unit if the quantity unit is not set.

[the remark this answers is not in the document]
ibacher: Dosing units are only rarely dispensing units. For example, a relatively common in-patient order might be for 1000 ml of saline, which is commonly dispensed as either 1 or 2 bags, depending on availability. Here the dosing is 1000 ml, but the dispensing units are in "bags". Similarly, for paracetamol, you might be given a dose of 650 mg, but that's dispensed in the form of 2 325 mg tablets per dose. Again, the dosing is in mg, but the dispensing is in tablets. There are a few narrow cases, though, where the dose ordered is the same units as the thing dispensed. This feature is targeted at those narrow cases.

So, my opinion is that once a user has selected a dispensing unit, we should not change that, because in most cases, the dispensing unit is not going to be the same as the dosing unit. Prior to the user making that choice, however, its fine to automatically update, as for a few cases this will make people's lives easier.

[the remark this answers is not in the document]
mccarthyaaron: Understood. Thanks

mccarthyaaron:
@ibacher pushed the requested changes. anymore recommended changes?

ibacher:
@mccarthyaaron Could you take a look into the Typescript thing?

mccarthyaaron:
> @mccarthyaaron Could you take a look into the Typescript thing?

@ibacher could you kindly direct me to a resource with the carbon types if possible?

ibacher:
@mccarthyaaron The Carbon [source code](https://github.com/carbon-design-system/carbon/tree/v11.37.0/packages/react/src/components) is quite readable and has the types. Just make sure you're looking at the v11.37.0 tag.

mccarthyaaron:
> @mccarthyaaron The Carbon [source code](https://github.com/carbon-design-system/carbon/tree/v11.37.0/packages/react/src/components) is quite readable and has the types. Just make sure you're looking at the v11.37.0 tag.

For educative purposes, is there a reason it has to be that specific tag?

mccarthyaaron:
@ibacher I am running into this error: Cannot use namespace 'ToggleProps' as a type. I think I imported the correct type because the import statement bringing it in isn't throwing any errors. I don't know whats causing that

mccarthyaaron:
@ibacher I have something for the types thing. Feels makeshift-ish especially with the importing of the types. For some reason I can't use the carbon types imported from '@carbon/react'. I get the error `Cannot use namespace as type` so I imported each of them individualy in a format like this;
 `import { type ComboBoxProps } from '@carbon/react/lib/components/ComboBox/ComboBox'; 
`
There are some other details that would need some polishing and feedback from you. Can I push the saidchanges here?

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
vasharma05: Hi @mccarthyaaron!
You can refer #1652.
Thanks!

vasharma05 (approved):

vasharma05 (approved):
LGTM!

PR #1638, opened 2024-02-05
(fix) O3-2735: Display units for patient age in header

Vijaykv5:
## Requirements

## Summary
 - Patients header is rendered according to the patient's age in terms of years, months and weeks.
 - Optional chaining is added in `birthdate` and `gender`.

Screenshot 2024-02-05 at 8 40 58 PM

> After 

https://github.com/openmrs/openmrs-esm-patient-chart/assets/94985341/e1248b17-de01-4fb3-98fa-5ddebf356dea

## Related Issue

[O3-2735](https://openmrs.atlassian.net/browse/O3-2735)

## Other
parseInt `parses a value as a string and returns the first integer` is not required as we are registering patients with current age in integer or in terms of number of month( Not as a whole value )

### DIFF
--- a/packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
@@ -85,9 +85,7 @@ const PatientInfo: React.FC<PatientInfoProps> = ({ patient }) => {
           label={
             <>
               <p className={styles.tooltipPatientName}>{name}</p>
-              <p className={styles.tooltipPatientInfo}>{`${parseInt(age(patient?.birthDate))}, ${getGender(
-                patient?.gender,
-              )}`}</p>
+              <p className={styles.tooltipPatientInfo}>{`${age(patient?.birthDate)}, ${getGender(patient?.gender)}`}</p>
             </>
           }
         >
@@ -98,9 +96,7 @@ const PatientInfo: React.FC<PatientInfoProps> = ({ patient }) => {
       ) : (
         <span className={styles.patientName}>{name} </span>
       )}
-      <span className={styles.patientInfo}>
-        {parseInt(age(patient.birthDate))}, {getGender(patient.gender)}
-      </span>
+      <span className={styles.patientInfo}>{`${age(patient?.birthDate)}, ${getGender(patient?.gender)}`}</span>
       {queueEntry && (
         <>
           <div className={styles.navDivider} />


Vijaykv5:
@denniskigen May you can review this?
Thanks

vasharma05:
HI @Vijaykv5 !
As per the designs [here](https://app.zeplin.io/project/60d59321e8100b0324762e05/screen/61bb8a40a95c745f60e24390). the age is meant to be a number only. As described by Fiona in the ticket description, we need to show just the age of patients >2 yrs.

In this [age](https://github.com/openmrs/openmrs-esm-core/blob/97feb47ebb0c2f79170163f454c180b6bf856e2f/packages/framework/esm-utils/src/age-helpers.ts#L38) function, we can add a new function `shortAge` (feel free to add a better name) which can return just number for patients > 2yrs, else returns the output from the `age` function, and can use that function here.
Thanks!

Vijaykv5:
> HI @Vijaykv5 ! As per the designs [here](https://app.zeplin.io/project/60d59321e8100b0324762e05/screen/61bb8a40a95c745f60e24390). the age is meant to be a number only. As described by Fiona in the ticket description, we need to show just the age of patients >2 yrs.
> 
> In this [age](https://github.com/openmrs/openmrs-esm-core/blob/97feb47ebb0c2f79170163f454c180b6bf856e2f/packages/framework/esm-utils/src/age-helpers.ts#L38) function, we can add a new function `shortAge` (feel free to add a better name) which can return just number for patients > 2yrs, else returns the output from the `age` function, and can use that function here. Thanks!

Thanks @vasharma05 !
But is this `patients > 2yrs` is applicable only to patient header? 
what I mean is if we  directly change in [age](https://github.com/openmrs/openmrs-esm-core/blob/97feb47ebb0c2f79170163f454c180b6bf856e2f/packages/framework/esm-utils/src/age-helpers.ts#L38) , it will change the main part of patient header (shown below)
Screenshot 2024-02-06 at 1 26 48 PM

So I think it would be better to change  our `age` function [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx) if the above scenario is considered

vasharma05:
No @Vijaykv5 , age is being used at multiple places. `age` shouldn't be touched. Patient Header is the only place this needs to be implemented right now, so it's better to have a separate function for this.
Thanks!

vasharma05:
Any update @Vijaykv5?
Thanks!

ibacher (approved):

Vijaykv5:
@ibacher But this would work on the way like @vasharma05 suggested right? 

and I was on way for it or is this fine?

ibacher:
No, but I don't think the designs lacking `yrs` should be a real blocker here. I think its better if the UI is explicit about the units. If this eats up too many pixels, we can always re-assess.

denniskigen:
Thanks, @Vijaykv5!

PR #1680, opened 2024-02-19
(fix) O3-2877: Make the drug search debounce delay value configurable 

jwnasambu:
## Requirements

## Summary
I customized the debounce time by passing the debounceTime prop when using the DrugSearch component by setting `debounceTime={300 milliseconds} `

## Related Issue
https://openmrs.atlassian.net/browse/O3-2877


packages/esm-patient-medications-app/src/add-drug-order/drug-search/drug-search.component.tsx
```diff
@@ -8,13 +8,14 @@ import styles from './order-basket-search.scss';
...
+  debounceTime?: number;
 }
 
-export default function DrugSearch({ openOrderForm }: DrugSearchProps) {
+export default function DrugSearch({ openOrderForm, debounceTime = 300 }: DrugSearchProps) {
   const { t } = useTranslation();
   const isTablet = useLayoutType() === 'tablet';
   const [searchTerm, setSearchTerm] = useState('');
-  const debouncedSearchTerm = useDebounce(searchTerm);
+  const debouncedSearchTerm = useDebounce(searchTerm, debounceTime);
```
denniskigen, line 18: [useDebounce](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-react-utils/src/useDebounce.ts#L32) already defaults to a timeout of 300ms. Maybe a better approach here would be to make the debounce value a configurable property in the medications app's config schema and then using whatever value is provided, or the default otherwise.

jwnasambu: @denniskigen Thanks so much for the quick review let me try the fix.

packages/esm-patient-medications-app/src/config-schema.ts
```diff
@@ -41,3 +41,11 @@ export interface ConfigObject {
   showPrintButton: boolean;
   maxDispenseDurationInDays: number;
 }
+
+export default {
+  debounceTime: {
+    type: 'integer',
+    default: 500,
+    description: 'Debounce time for medication search input in milliseconds',
+  },
```
denniskigen, line 50: This block of config is in the wrong place. It should exist inside the `configSchema` object. Also, the object's keys don't match the schema format. It should read as follows:

```suggestion
export const configSchema = {
  // existing config schema properties...
  drugSearchDebounceTimeInMs: {
    _type: 'integer',
    _description: 'Number of milliseconds to delay the search operation in the drug search input by after the user starts typing. The useDebounce hook delays the search by 300ms by default',
    _default: 300,
  }
}
```

And then the new `debounceTimeInMs` configuration property should be added to the `ConfigObject` interface as shown below:
```suggestion
  debounceTime: {
    type: 'integer',
    default: 500,
    description: 'Debounce time for medication search input in milliseconds',
  },
```

```tsx
export interface ConfigObject {
  daysDurationUnit: {
    uuid: string;
    display: string;
  };
  drugOrderTypeUUID: string;
  showPrintButton: boolean;
  maxDispenseDurationInDays: number;
  drugSearchDebounceTimeInMs: number;
}
```

denniskigen (changes requested):

packages/esm-patient-medications-app/src/config-schema.ts
jwnasambu: @denniskigen thanks so so much for the guidance. Am so glad I have learnt something new

packages/esm-patient-medications-app/src/config-schema.ts
```diff
@@ -30,6 +30,12 @@ export const configSchema = {
     _default: 99,
     _description: 'The maximum number of days for medication dispensing.',
   },
+  debounceDelayInMs: {
+    _type: 'integer',
+    _description:
+      'Number of milliseconds to delay the search operation in the drug search input by after the user starts typing. The useDebounce hook delays the search by 500ms by default',
+    _default: 500,
```
denniskigen, line 37: Sorry, @jwnasambu, I should've said the default was 300ms. Could you change it to say that in both the `default` and the `description` keys above?

packages/esm-patient-medications-app/src/config-schema.ts
```diff
@@ -30,6 +30,12 @@ export const configSchema = {
     _default: 99,
     _description: 'The maximum number of days for medication dispensing.',
   },
+  debounceDelayInMs: {
+    _type: 'integer',
```
denniskigen, line 34: [_type](https://github.com/openmrs/openmrs-esm-core/blob/ff09ef39a8477ca2b7ddeb9c0d94ca5efc076d50/packages/framework/esm-config/src/types.ts#L1) must be one of the properties in the `Type` enum. In this case, it should be:

```suggestion
    _type: Type.Number,
```

packages/esm-patient-medications-app/src/add-drug-order/drug-search/drug-search.component.tsx
```diff
@@ -14,7 +14,8 @@ export default function DrugSearch({ openOrderForm }: DrugSearchProps) {
   const { t } = useTranslation();
   const isTablet = useLayoutType() === 'tablet';
   const [searchTerm, setSearchTerm] = useState('');
-  const debouncedSearchTerm = useDebounce(searchTerm);
+  const { debounceTime } = useConfig();
+  const debouncedSearchTerm = useDebounce(searchTerm, debounceTime);
```
denniskigen, line 18: Cool. So the next thing to do is annotate the `useConfig` import with the `ConfigObject` from the medications app schema like so:

```tsx
import { type ConfigObject } from '../../config-schema';

// ... more code omitted for brevity
  const { debounceDelayInMs } = useConfig<ConfigObject>();
  const debouncedSearchTerm = useDebounce(searchTerm, debounceDelayInMs ? debounceDelayInMs : 300);
```

denniskigen (approved):

denniskigen:
This LGTM, @jwnasambu. Thanks. I'll let @donaldkibet look through it and merge.

jwnasambu:
@denniskigen Thanks for the guidance and mentorship.I have learnt a new skill from you.

packages/esm-patient-medications-app/src/add-drug-order/drug-search/drug-search.component.tsx
```diff
@@ -14,7 +15,8 @@ export default function DrugSearch({ openOrderForm }: DrugSearchProps) {
   const { t } = useTranslation();
   const isTablet = useLayoutType() === 'tablet';
   const [searchTerm, setSearchTerm] = useState('');
-  const debouncedSearchTerm = useDebounce(searchTerm);
+  const { debounceDelayInMs } = useConfig<ConfigObject>();
+  const debouncedSearchTerm = useDebounce(searchTerm, debounceDelayInMs ? debounceDelayInMs : 300);
```
ibacher, line 19:
```suggestion
  const debouncedSearchTerm = useDebounce(searchTerm, debounceDelayInMs ?? 300);
```

jwnasambu: @ibacher thanks for the suggestion. I have fixed the issue

ibacher (approved):

PR #1686, opened 2024-02-22
(feat) O3 2888: Show non-coded allergen and allergic reaction names in summary table

jnsereko:
## Requirements

## Summary

After selecting `other` and specifying the name of the non-coded allergy and saving, the allergies table in the reactions columns displays other.  It should be better if they display the actual name of the non-coded  allergy  cc @denniskigen @vasharma05 @michaelbontyes 

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-chart/assets/58003327/9b0ead16-0fef-4431-a109-b6724bf2295b

## Related Issue

https://openmrs.atlassian.net/browse/O3-2888

### DIFF
--- a/packages/esm-patient-allergies-app/src/allergies/allergy-intolerance.resource.ts
+++ b/packages/esm-patient-allergies-app/src/allergies/allergy-intolerance.resource.ts
@@ -53,17 +53,17 @@ export function useAllergies(patientUuid: string): UseAllergies {
 }
 
 function mapAllergyProperties(allergy: FHIRAllergy): Allergy {
-  const manifestations = allergy?.reaction[0]?.manifestation?.map((coding) => coding.coding[0]?.display);
+  const manifestations = allergy?.reaction[0]?.manifestation?.map((coding) => coding?.text);
   return {
     id: allergy?.id,
     clinicalStatus: allergy?.clinicalStatus?.coding[0]?.display,
     criticality: allergy?.criticality,
-    display: allergy?.code?.coding[0]?.display,
+    display: allergy?.code?.text,
     recordedDate: allergy?.recordedDate,
     recordedBy: allergy?.recorder?.display,
     recorderType: allergy?.recorder?.type,
     note: allergy?.note?.[0]?.text,
-    reactionToSubstance: allergy?.reaction[0]?.substance?.coding[1]?.display,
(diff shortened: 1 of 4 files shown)


packages/esm-patient-allergies-app/src/allergies/allergy-intolerance.resource.ts
```diff
@@ -53,7 +53,11 @@ export function useAllergies(patientUuid: string): UseAllergies {
 }
 
 function mapAllergyProperties(allergy: FHIRAllergy): Allergy {
-  const manifestations = allergy?.reaction[0]?.manifestation?.map((coding) => coding.coding[0]?.display);
+  const manifestations = allergy?.reaction[0]?.manifestation?.map((coding) => {
+    const display = coding.coding[0]?.display;
+    const text = coding.text;
+    return display === 'Other' ? text : display;
```
ibacher, line 59: This is fragile and unmergable as-is. I _think_ what we actually need here is:

```ts
coding.coding[0]?.display ?? coding.text
```

ibacher (changes requested):
Thanks for this @jnsereko! It would definitely be good to get this in.

In addition to the issue I flagged below, there are two things that need to be corrected with this PR as a PR:

1. We _require_ a conventional commit for ESM repos. This is mentioned in the PR template; please read it and don't just check the boxed.
2. The title does not accurately reflect the change. Specifically the code here deals with _non-coded allergy reactions_, which are _not_ _non-coded allergy name_ (which is a separate concept that we also support, though not on the form).
3. It would be nice to have a unit test for this that covers the behaviour.

packages/esm-patient-allergies-app/src/allergies/allergy-intolerance.resource.ts
ibacher: Basically, the string "Other" is a concept name with is locale-specific. If the EMR was in French and the text here was "Autre", this would fail.

jnsereko: thank you @ibacher for taking a look at this.
I realised that FHIR AllergyIntolerance response contains a text property which would be better using concepts whose names are language dependent

denniskigen:
@ibacher, Reckon this is good to go following @jnsereko's changes?

PR #1718, opened 2024-03-06
(chore) Switch from lerna to yarn for publishing

denniskigen:
## Requirements

## Summary

This PR switches this repo from using lerna for publishing releases and pre-releases to using Yarn. This just brings this repo in line with others like Core and Patient Management that use Yarn for publishing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,7 +84,7 @@ jobs:
           server-token: ${{ env.TURBO_TOKEN }}
 
       - name: Version
-        run: yarn lerna version "$(node -e "console.log(require('semver').inc(require('./lerna.json').version, 'patch'))")-pre.${{ github.run_number }}" --no-git-tag-version --no-push --yes
+        run: yarn workspaces foreach --worktree --topological version "$(node -e "console.log(require('semver').inc(require('./package.json').version, 'patch'))")-pre.${{ github.run_number }}"
 
       - name: Build
         run: yarn turbo build --color --concurrency=5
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -84,7 +84,7 @@ jobs:
           server-token: ${{ env.TURBO_TOKEN }}
 
       - name: Version
-        run: yarn lerna version "$(node -e "console.log(require('semver').inc(require('./lerna.json').version, 'patch'))")-pre.${{ github.run_number }}" --no-git-tag-version --no-push --yes
+        run: yarn workspaces foreach --worktree --topological version "$(node -e "console.log(require('semver').inc(require('./package.json').version, 'patch'))")-pre.${{ github.run_number }}"
```
jayasanka-sack, line 87: We can exclude the root app here, right? 

ex: 

https://github.com/openmrs/openmrs-esm-patient-management/blob/cdf20e84511a8054d2ebdabe272b024721509740/.github/workflows/ci.yml#L89

denniskigen: No, because we want to version the `@openmrs/esm-patient-chart-app` package as well, right?

ibacher: @denniskigen The "root app" here is `@openmrs/esm-patient-chart`, i.e., the workspace package.json itself. 

That said, as long as this works, I don't think it's wrong (the difference is, at most, a couple of milliseconds of execution time). If I did it again, I probably wouldn't write the patient-management CI script the same way; it would be nice to have symmetry between the command run here and the command used in the `version` script.

ibacher: Hmmm... there was a reason for the exclusion, it appears: https://github.com/openmrs/openmrs-esm-patient-chart/actions/runs/8189908003/job/22395772588

denniskigen: I can push a PR to fix the exclusion. 

> If I did it again, I probably wouldn't write the patient-management CI script the same way; it would be nice to have symmetry between the command run here and the command used in the version script. 

Should I change the command as well?

ibacher: So, actually, I think the pertinent place for the exclusion is in the `ci:publish` and `ci:prepublish` scripts in the package.json. It makes sense to add that here, I guess, for the same symmetry I was referring to.

PR #1744, opened 2024-03-19
(fix) Various fixes for the visit notes form

denniskigen:
## Requirements

## Summary

This PR fixes the various aspects of the visit notes form, including:

 - Fixing a deprecation in how observers are handled in the `fetchConceptDiagnosisByName` function (which I've renamed `fetchDiagnosisConceptsByName`). I've elected to refactor the function to use `openmrsFetch` instead of `openmrsObservableFetch` under the hood.
 
    CleanShot 2024-03-19 at 1  34 46@2x

- Showing an error notification when there is a problem fetching diagnoses from the backend:
    
    CleanShot 2024-03-19 at 1  24 34@2x

- Rendering the diagnoses tags above the Primary diagnosis section so its clear that the diagnoses are not just linked to that section:
 
    CleanShot 2024-03-20 at 11  15 02@2x

## Screenshots

See above.


packages/esm-patient-notes-app/src/notes/visit-notes-form.component.tsx
```diff
@@ -423,7 +425,18 @@ const VisitNotesForm: React.FC<DefaultWorkspaceProps> = ({
...
+                      <InlineNotification
+                        className={styles.errorNotification}
+                        lowContrast
+                        title={t('error', 'Error')}
+                        subtitle={t('errorFetchingConcepts', 'There was a problem fetching concepts') + '.'}
+                        onClose={() => setError(null)}
+                      />
+                    );
+                  }
+                  if (loadingPrimary) {
```
vasharma05, line 439: Can we remove that (() => {})() and simply put if()... statements?

My concern here seems to be that we create a function, then we call it, which is redundant.

packages/esm-patient-notes-app/src/notes/visit-notes-form.component.tsx
```diff
@@ -495,7 +509,18 @@ const VisitNotesForm: React.FC<DefaultWorkspaceProps> = ({
...
+                      <InlineNotification
+                        className={styles.errorNotification}
+                        lowContrast
+                        title={t('error', 'Error')}
+                        subtitle={t('errorFetchingConcepts', 'There was a problem fetching concepts') + '.'}
+                        onClose={() => setError(null)}
+                      />
+                    );
+                  }
+                  if (loadingSecondary) {
```
vasharma05, line 523: Same here please

packages/esm-patient-notes-app/src/notes/visit-notes.resource.ts
```diff
@@ -104,10 +110,11 @@ export function useInfiniteVisits(patientUuid: string) {
...
-export function fetchConceptDiagnosisByName(searchTerm: string, diagnosisConceptClass: string) {
-  return openmrsObservableFetch<Array<Concept>>(
-    `${restBaseUrl}/concept?name=${searchTerm}&searchType=fuzzy&class=${diagnosisConceptClass}&name=&v=custom:(uuid,display)`,
-  ).pipe(map(({ data }) => data['results']));
+export function fetchDiagnosisConceptsByName(searchTerm: string, diagnosisConceptClass: string) {
+  const customRepresentation = 'custom:(uuid,display)';
+  const url = `${restBaseUrl}/concept?name=${searchTerm}&searchType=fuzzy&class=${diagnosisConceptClass}&v=${customRepresentation}`;
+
+  return openmrsFetch<Array<Concept>>(url).then(({ data }) => Promise.resolve(data['results']));
 }
```
vasharma05, line 118: Can't we implement `useOpenMRSSWR` in this?

packages/esm-patient-notes-app/src/notes/visit-notes-form.component.tsx
denniskigen: Fixed in the latest commit.

packages/esm-patient-notes-app/src/notes/visit-notes.resource.ts
denniskigen: So, refactoring the data fetching logic into a hook seems a lot less straightforward after looking at the ergonomics of how the primary diagnosis and secondary diagnosis search work.  It's easier when you only have one piece of state to run through the hook, but here we have both the primary and secondary diagnoses search, so at a minimum, we'd need to track more state for both. The code is fairly readable now, even without hooks, especially after switching to a promise-based implementation instead of using reactive logic via observables.

I'm happy to try and look at this in a separate PR though.

vasharma05:
I have one more suggestion, but it should be verified with @duffman
Instead of search input boxes, we should use Combo boxes, with a placeholder saying "Enter 2 or 3 characters to search", reason being that the dropdown that we are showing, it doesn't support keyboard navigation and all, which is natively supported by Combo boxes

denniskigen:
Is this good to go otherwise? @vasharma05

vasharma05:
Yes @denniskigen

vasharma05 (approved):

PR #1745, opened 2024-03-19
fix (O3-2717): Form workspace shrinks when opened while on the Vitals & Biometrics page

mccarthyaaron:
## Requirements

## Summary
This is a wacky solution which seems to work and I am hoping we can figure it out why and come up with a better one. Can walk you through my thought process as well if need be.

## Screenshots
https://github.com/openmrs/openmrs-esm-patient-chart/assets/121826239/f764b072-2e65-4883-8e15-95f6dc0e26b8

## Related Issue
https://openmrs.atlassian.net/browse/O3-2717

### DIFF
--- a/packages/esm-patient-vitals-app/src/routes.json
+++ b/packages/esm-patient-vitals-app/src/routes.json
@@ -20,7 +20,8 @@
       "slot": "patient-chart-vitals-biometrics-dashboard-slot",
       "meta": {
         "title": "Vitals",
-        "view": "vitals"
+        "view": "vitals",
+        "columnSpan": 2
       },
       "order": 1
     },
@@ -44,7 +45,8 @@
       "slot": "patient-chart-vitals-biometrics-dashboard-slot",
       "meta": {
         "view": "biometrics",
-        "title": "Biometrics"
+        "title": "Biometrics",
+        "columnSpan": 2
       }
     },
     {


brandones:
Yeah why is it wacky and why does it work? Sorry, it's been a long time since I thought about the dashboard columns.

Twiineenock:
@mccarthyaaron , did you find another way to get around this, otherwise it looks fine too me!

mccarthyaaron:
@brandones @Twiineenock . I failed to get a valid reason as to why it works. But this is why I came up with the above solution. The issue only arises when looking at "vitals and biometrics" and when you inspect the elements, the only thing that really changes in all the different scenarios is the slot used. The "patient-chart-vitals-biometrics-dashboard-slot" has `grid-template-columns: 1fr;` with the children spanning 1 column each which should work. But then I tried changing it to `grid-template-columns: 2fr;` and have each child span 2 columns which worked. It is literally the same thing from a CSS point of view but one works and the other does not. Maybe it has something to do with the particular slot.


https://github.com/openmrs/openmrs-esm-patient-chart/assets/121826239/cdc97ce5-9257-4413-b020-969498353db0

 See attached video

brandones:
Thanks for the explanation @mccarthyaaron , that makes sense! Well, your explanation makes sense, but yeah, the issue maybe doesn't :laughing: . Nice job investigating and fixing, anyway.

brandones (approved):

mccarthyaaron:
@brandones there is actually something small missing which is changing the "patient-chart-vitals-biometrics-dashboard-slot" to have 2 columns not the 1 column it currently has. Not sure exactly how to do that. It will be confusing for someone with no prior knowledge about this issue to see a grid container with 1 column whose children span 2 columns each

PR #1685, opened 2024-02-22
(feat) O3-2885: Fix allowed image types in visit notes form image uploader

jnsereko:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.
- [x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://zeroheight.com/23a080e38/p/880723-introduction).
- [x] My work includes tests or is validated by existing tests.

## Summary

Corrected and added `.jpg` image extensions cc @denniskigen @vasharma05 @michaelbontyes 

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-chart/assets/58003327/bb7a41f6-c9ec-46d7-a0a7-a64f112992f8

## Related Issue

https://openmrs.atlassian.net/browse/O3-2885

PR #1620, opened 2024-01-25
(feat) Specify upload size limit and supported file types in file uploader

denniskigen:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.
- [ ] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://zeroheight.com/23a080e38/p/880723-introduction).
- [ ] My work includes tests or is validated by existing tests.

## Summary

This PR is a follow-up to https://github.com/openmrs/openmrs-esm-patient-chart/pull/1522. It tweaks the file uploader UI in the following ways:

- Sets the max file upload size limit to 1 MB (the related global property, `attachments.maxStorageFileSize` caps the limit at 1.2 MBs, but I felt that that is an awkward value to show in the UI)

	attachment-upload-global-properties

- Shows information about the accepted file types in the File Uploader modal (JPEG, PNG, and WEBP)
- Sets the `allowedExtensions` array of the uploader's context to `['image/jpeg', 'image/png', 'image/webp']` so that the uploader knows to only show files matching those extensions in the upload UI. Files that don't match will be blurred out and incapable of being selected.
- Fixes an issue with the sizing of the ImagePreview component's `Close` button and the `Delete image` overflow menu item (which I now realize does not need to be an overflow menu that has only one option. It should probably be a button instead). 

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-chart/assets/8509731/803a95a6-696a-41cb-8ce0-c795ff879349

PR #1584, opened 2024-01-15
(feat) Restrict attachments to configured allowed extensions

icrc-loliveira:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-dev.docs.openmrs.org/#/getting_started/contributing?id=your-pr-title-should-indicate-the-type-of-change-it-is) label. See existing PR titles for inspiration.
- [x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://zeroheight.com/23a080e38/p/880723-introduction).
- [x] My work includes tests or is validated by existing tests.

## Summary
This PR created a new configuration 'allowedExtensions' that allow filter by the extensions that can be attached to openMRS.

## Screenshots
image
image

PR #1748, opened 2024-03-19
(fix) O3-2627: Improvements to the visit notes image upload functionality

jwnasambu:
## Requirements

## Summary
I modified` showImageCaptureModal` function before setting the value of the image field to checks if the file type is 'image'. If it's not an image, to displays a Snackbar error message indicating that only image files are allowed.
It rejects the promise with an error if the file type is not an image, preventing the modal from closing and the value from being set and if the file type is an image, it proceeds to set the image field value and close the modal as before.

## Screenshots

Screenshot 2024-03-19 at 10 14 57 PM

## Related Issue
https://openmrs.atlassian.net/browse/O3-2627


vasharma05:
Why are we even allowing all types of documents when we only need images?

denniskigen:
So, I've tested the existing functionality on the refapp, and it looks like the validation is already happening courtesy of work done in #1620, #1685, and #1584.

https://github.com/openmrs/openmrs-esm-patient-chart/assets/8509731/d51b4c6a-78d3-4007-84e1-a86b5c484139

jwnasambu:
@denniskigen You are correct! I believe this PR should be closed unless there is something that needs fixing. Yesterday's internet was so poor that I couldn't see the Snackbar error message. Please forgive me.

hadijahkyampeire:
@jwnasambu @vasharma05 I think in this case we want inline error notifications, not a snackbar, right? From what @denniskigen mentioned above, this is working as expected.

packages/esm-patient-notes-app/src/notes/visit-notes-form.component.tsx
```diff
@@ -213,6 +213,15 @@ const VisitNotesForm: React.FC<DefaultWorkspaceProps> = ({
   const showImageCaptureModal = useCallback(() => {
     const close = showModal('capture-photo-modal', {
       saveFile: (file: UploadedFile) => {
+        if (file.fileType !== 'image') {
+          showSnackbar({
```
hadijahkyampeire, line 217: Because this is a validation error, we should use an inline error notification with low contrast instead of a snack bar. cc @denniskigen am I correct here?

denniskigen: Technically, yes. But we shouldn't need to add anything here because validation is already handled in the [MediaUploader](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-attachments-app/src/camera-media-uploader/media-uploader.component.tsx#L14-L15) component.

mogoodrich:
> Why are we even allowing all types of documents when we only need images?

I'm not familiar entirely with this functionality, but a generic attachments functionality is useful, to be able to upload things like reports in PDF format, etc.

denniskigen:
@mogoodrich @vasharma05 it definitely makes sense to open this up to different file formats. This current implementation is guided by the design, specifically the Add Image button, which suggests that file uploads in the visit note form should be limited to images.

ojwanganto:
@denniskigen it would be nice to understand the design requirements. It would be nice to open this to some common file formats unless there are clear reasons for not opening it up. I see a lot of use cases beyond just images though

denniskigen:
Yeah, I'll seek Ciaran and Paul's feedback on this tomorrow @ojwanganto.

denniskigen (approved):
I'm merging this as it meets the requirements in the [ticket](https://openmrs.atlassian.net/browse/O3-2627). I'll file a new ticket with the new requirement of uploading different file formats via the visit note form.

mogoodrich:
Yeah, sorry, didn't mean to derail this ticket, just wanted to note that there is a significant use case for other kinds of uploads like PDFs.

icrc-loliveira:
Hello @jwnasambu,
This PR makes the error messages lock a lot nicer, which I approve :)
However since you changed the configuration to require the dots, it broke our configuration.

- Old: 'jpeg', 'jpg', 'png', 'webp'
- New: '.jpeg', '.jpg', '.png', '.webp'

This configuration comes from the backend which for security reasons also validates the allowed extensions. And now if we set the global property `attachments.allowedFileExtensions` you will not be able to upload an attachment because of that difference.
PR: https://github.com/openmrs/openmrs-module-attachments/pull/61

Do you have any reason to keep this change or can I revert it to the old one?

cc: @mogoodrich, @denniskigen, @ojwanganto, @hadijahkyampeire, @ibacher

denniskigen:
Sorry about that, @icrc-loliveira. What does the request for the `attachment.allowedFileExtensions` GP return on your end?

icrc-loliveira:
This is an example of the values that is expecting:
```
<globalProperty>
  <property>attachments.allowedFileExtensions</property>
  <value>txt,pdf,jpeg,jpg,png,gif</value>
</globalProperty>
```

denniskigen:
@ibacher does the `allowedFileExtensions` GP allow arbitrary values? Can one specify file extensions prefixed with a `.`?

ibacher:
@denniskigen All GPs are basically just string values. However, changing the value there would require updates to the attachments module code, since we're currently [lopping the extension off separately](https://github.com/openmrs/openmrs-module-attachments/pull/61/files#diff-ea166714ece8fa15e4ef6b139aa9473ef1f80d3a767e636ef1ba1ce4db07e007R122).

denniskigen:
Please feel free to revert to the old convention, @icrc-loliveira. Sorry for the inconvenience.

PR #1750, opened 2024-03-20
(fix) O3-2977: Make end date field editable for inactive conditions

denniskigen:
## Requirements

## Summary

This PR makes it possible to modify the end date property of an existing condition when specified. Currently, the end date property in the update payload maps to an end date property in the FHIR Conditions resource. This property does not exist, however. The closest matching element is [abatementDateTime](https://hl7.org/fhir/condition-definitions.html#Condition.abatement_x_), which is described as:

> 	The date or estimated date the condition resolved or went into remission. This is called "abatement" because of the many overloaded connotations associated with "remission" or "resolution" - Some conditions, such as chronic conditions, are never really resolved, but they can abate.

This PR renames the end date property in the Conditions payload to abatementDatetime, which gets things working as expected. 

I've also made the following useful fixes:

- Ensuring that the `launchConditionsForm` callback is invoked with an extra parameter, `{ formContext: 'creating' }`. This is required to get the condition name validation to work [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/94738af458ebadd96291da55d6b8b7949dffd10f/packages/esm-patient-conditions-app/src/conditions/conditions-form.component.tsx#L70). I've made this change in both the ConditionsOverview and ConditionsDetailedSummary components.
- Renaming the `search` property in the conditions form schema to `conditionName` to reflect its purpose more accurately.
- Adding a top and bottom border to search results in the Condition name search box to separate them visually.
- Renaming the `current status` field to `clinical status` to get it in conformance with the FHIR Conditions spec.

### Before

- Validation doesn't kick in when I save the form without providing a condition.
- When I mark a condition as inactive and enter an end date, I don't see that value when I subsequently go to edit the condition. What's not shown here is the response from the backend, which doesn't contain an `endDate` property.

https://github.com/openmrs/openmrs-esm-patient-chart/assets/8509731/c12055d6-a3d7-4a28-8688-2f7469e15448

### After

- Validation now kicks in when a condition is not provided.
- The end date field is now shown for conditions marked as inactive.

https://github.com/openmrs/openmrs-esm-patient-chart/assets/8509731/a512335c-d380-4a65-ae4e-923b7cfb0cf5

## Related Issue

[O3-2977](https://openmrs.atlassian.net/browse/O3-2977)

### DIFF
--- a/packages/esm-patient-conditions-app/src/conditions/conditions-detailed-summary.component.tsx
+++ b/packages/esm-patient-conditions-app/src/conditions/conditions-detailed-summary.component.tsx
@@ -19,9 +19,9 @@ import {
 import { Add } from '@carbon/react/icons';
 import { formatDate, parseDate, useLayoutType } from '@openmrs/esm-framework';
 import { CardHeader, EmptyState, ErrorState, launchPatientWorkspace } from '@openmrs/esm-patient-common-lib';
-import { useConditions } from './conditions.resource';
 import { ConditionsActionMenu } from './conditions-action-menu.component';
 import { compare } from './utils';
+import { useConditions } from './conditions.resource';
 import styles from './conditions-detailed-summary.scss';
 
 function ConditionsDetailedSummary({ patient }) {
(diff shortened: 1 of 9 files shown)


hadijahkyampeire (approved):
LGTM, thanks @denniskigen

denniskigen:
Thanks, @hadijahkyampeire!

PR #1751, opened 2024-03-21
(refactor) Trivial clean-up in test results viewer

mccarthyaaron:
## Summary

Very minor clean-up in the results viewer implementation—removing unused variables, that sort of thing. No UI changes.


chimanwadike (commented):
@mccarthyaaron Is there any reason this PR may be failing e2e tests?
Does it pass when the e2e tests when executed on your local, please?

packages/esm-patient-labs-app/src/test-results/filter/filter-set.tsx
```diff
@@ -42,8 +42,7 @@ const FilterSet: React.FC<FilterSetProps> = ({ hideFilterSetHeader = false }) =>
   const [showSearchInput, setShowSearchInput] = useState(false);
 
   const handleInputChange = useCallback(
-    (e) => {
-      setSearchTerm(searchTerm);
+    (searchTerm) => {
```
vasharma05, line 45: What is the reason for this change?
Changing the event to a string will not work, until you make change in the said `onClick` function.

mccarthyaaron: @vasharma05 should I revert it to the original? Also help explain abit more as I haven't understood what you mean by the "the said onClick function". But this was my thought process: When the such term changes, the handleInputChange function is called with the search term passed as an argument. We then use that search term to get the filtered results. Previously 'e' was not being used anywhere in the handleInputChange function and I didn't understand why we are using an event as a parameter yet handleInputChange is being passed a string. Also though that maybe setting the search term( setSearchTerm(searchTerm) ) to the same value is kinda redudant. Thats how I thought about it. I am willing to be corrected

brandones: Yeah I think @vasharma05 mistakenly thought this is an input callback. It is actually being used in a `useEffect` and passed the `searchTerm` explicitly. But so this is all sort of nonsense. The code should probably just read 

```tsx
  useEffect(() => {
      const filteredData = roots.filter((node) => filterTreeNode(searchTerm, node));
      setTreeDataFiltered(filteredData);
  }, [searchTerm, root]);
```

The `handleInputChange` function can be gotten rid of entirely.

packages/esm-patient-labs-app/src/test-results/grouped-timeline/grouped-timeline.tsx
```diff
@@ -216,7 +215,7 @@ const TimelineDataGroup = ({ parent, subRows, xScroll, setXScroll, panelName, se
     }
   }, [handleScroll]);
 
-  const onIntersect = (entries, observer) => {
+  const onIntersect = (entries) => {
```
brandones, line 218: What on earth. This block does nothing. Time for `git blame`

brandones: Oh, @vasharma05 two years ago. He was new then.

Please delete this whole block.

brandones:
@mccarthyaaron are you available to continue this work?

mccarthyaaron:
@brandones Yes. But could you take a look at this here: https://openmrs.atlassian.net/browse/O3-2973?focusedCommentId=139939 . But if its okay for me to continue working on it then we can start

brandones:
I see, thanks @mccarthyaaron ! Sure let's just unlink this PR from that ticket and keep working on this independently.

mccarthyaaron:
@brandones could you take a look at the updates. Also, what is your taking on us using debouncing for the search functionality in filter-set.tsx. Might not be a huge improvement but just inquiring

brandones (approved):
LGTM. No you don't need to debounce inputs that are entirely handled on the frontend.

denniskigen:
(refactor) would be the appropriate conventional commit label here.

PR #1757, opened 2024-03-25
(fix) O3-2937: Make clinical status mandatory in the conditions form

denniskigen:
## Requirements

## Summary

`Clinical status` should be a mandatory field in the conditions form. The FHIR [Conditions resource terminology bindings](https://www.hl7.org/fhir/condition.html#tx) mark it as a required field. It also doesn't make much sense to record a condition without recording the status of the condition.

This PR makes that possible by adding the necessary validation to the form so that an error gets shown when the form is submitted without providing a condition name and a clinical status.

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-chart/assets/8509731/42b6ceb1-2903-4078-85da-8cf66bce1090

## Related Issue

https://openmrs.atlassian.net/browse/O3-2937

## Other

From the [FHIR Condition resource elements](https://www.hl7.org/fhir/condition-definitions.html#9.2.7), `Verification status` is another required field:

CleanShot 2024-03-25 at 12  49 44@2x

Should we add this to the Conditions form?


packages/esm-patient-conditions-app/src/conditions/conditions-widget.component.tsx
```diff
@@ -363,11 +367,11 @@ const ConditionsWidget: React.FC<ConditionsWidgetProps> = ({
...
 };
 
-function RequiredLabel({ isRequired, text, t }: RequiredLabelProps) {
+function RequiredFieldLabel({ editing, label, t }) {
   return (
     <>
-      <span>{text}</span>
-      {isRequired && (
+      <span>{label}</span>
+      {!editing && (
```
ibacher, line 374: I'm not sure I quite understand the reason for changing things based on whether someone is editing or not? It seems like a required field is required regardless and should be marked that way no?

ibacher (approved):
LGTM

packages/esm-patient-conditions-app/src/conditions/conditions-widget.component.tsx
denniskigen: That's right, @ibacher. A bad rebase was the culprit here.

denniskigen:
Thanks, @ibacher!

PR #1767, opened 2024-03-31
(feat) O3-2430: User should be able to look up frequency in dropdown by abbreviation

piyushmishra1416:
## Requirements

## Summary

The medication frequency dropdown menu displays options such as "three times per day." With this update, users can now input "tid" in the dropdown menu and have "three times per day" as an option.

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-chart/assets/91911367/77a4c88c-4e5e-421b-be5a-a1bbfb96c860

## Related Issue

 https://issues.openmrs.org/browse/O3-2430


piyushmishra1416:
Hello @brandones, could you kindly review this Pull Request?

brandones:
Thank you for your work on this, @piyushmishra1416 !

Unfortunately, I think this ticket is much more non-trivial than I expected. I don't think we want to hard-code the frequencies, if we can avoid it. Ideally we would want to pull the abbreviations from the short names of the [frequency concepts](https://dev3.openmrs.org/openmrs/dictionary/concept.htm?conceptId=525). These concepts right now are obtained using the `orderentryconfig` REST endpoint, which only includes `display` and `UUID`. The backend code for this endpoint is in [OrderEntryResourceConfig](https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderEntryConfigResource1_10.java). Looking at it, it kind of looks like one could pass a [custom representation](https://openmrs.atlassian.net/wiki/spaces/docs/pages/25469830/REST+Web+Services+API+For+Clients#RESTWebServicesAPIForClients-CustomRepresentations) in the URL? But honestly I don't really know, and don't know how you would structure a custom rep correctly for this endpoint. But to start you could try adding a `v=custom:(uuid)` parameter to the request and see if the endpoint uses that parameter at all.

Requesting support from @denniskigen, @ibacher, or @mogoodrich. If it's too much of a pain I say we just hard-code the abbreviations (though we'll need to make it more robust to non-default metadata).

mogoodrich:
Thanks @brandones  I agree we want to avoid hardcoding.  What do we need beyond uuid and display?  Can we build the order around uuid rather than abbrievations (sorry if the answer is obvious, I haven't dug too deeply into the code yet)

brandones:
@mogoodrich The thing we would need besides uuid and display would be the concept short names, which would be the actual abbreviations. The idea of this ticket is that we want to make it so the user can type an abbreviation into the input box and it will select the correct frequency concept. We do indeed build the order using the uuids.

ibacher:
Seems doable without even hardcoding the UUIDs per se, just changing how we load the order configuration to something like: `/ws/rest/v1/orderentryconfig?v=custom:(uuid,display,names:(display,conceptNameType,locale))`. That will get you all the concept names (unfortunately, it will load all the names for every concept, even ones where we don't want this feature). It seems like it's necessary to drop those that aren't in the current locale (but, I guess, include English if the current locale doesn't have any available names), but that should also include, e.g., "BID / TID", etc. for frequencies.

brandones:
Thanks so much @ibacher !

> The short name isn't always the correct thing. For example, the short name for twice daily is "BD".

That's... odd. Should we forge ahead, and consider adding BID as a short name in the twice daily concept ?

ibacher:
> consider adding BID as a short name in the twice daily concept ?

I mean, it makes sense to me. We probably need to talk to Andy about this, since I think we're just using a CIEL concept.

brandones:
@piyushmishra1416 Is it clear to you how to move forward with this, and will you move it forward?

piyushmishra1416:
> @piyushmishra1416 Is it clear to you how to move forward with this, and will you move it forward?

I am on this @brandones. I am unable to fetch the short names from the backend.  I was trying to work around this URL: `/ws/rest/v1/orderentryconfig?v=custom:(uuid,display,names:(display,conceptNameType,locale))` and tried to fetch the short names using various combinations of this URL but unable to hit the correct URL. Could you please provide me some help with how to fetch the short names correctly? Thanks

denniskigen:
We're you able to fetch the order frequency shortnames from the `orderentryconfig` endpoint, @piyushmishra1416?

brandones:
Yeah sorry for the lack of response from my end; this is not my area of expertise. Perhaps @mogoodrich would know how to get the `custom:` format right.

denniskigen:
Yeah. Some extra context:`/ws/rest/v1/orderentryconfig?v=custom:(uuid,display,names:(display,conceptNameType,locale))` doesn't return an `orderFrequencies` property in its response, whereas `/ws/rest/v1/orderentryconfig?v=full` does. A `custom:` format would do the trick here.

piyushmishra1416:
Yes @denniskigen, I was getting the short names. `ws/rest/v1/orderentryconfig?v=custom:(uuid,display,concept:(names:(display,uuid))` this URL does this task.

packages/esm-patient-medications-app/src/api/order-config.ts
```diff
@@ -9,9 +9,16 @@ import {
...
+export interface Names {
+  uuid: string;
+  display: string;
+}
 export interface CommonConfigProps {
   uuid: string;
   display: string;
+  concept?: {
+    names: Names[];
+  }; //
```
brandones, line 21:
```suggestion
  };
```

packages/esm-patient-medications-app/src/api/order-config.ts
```diff
@@ -34,7 +41,7 @@ export function useOrderConfig(): {
   };
 } {
   const { data, error, isLoading, isValidating } = useSWRImmutable<{ data: OrderConfig }, Error>(
-    `${restBaseUrl}/orderentryconfig`,
+    `${restBaseUrl}/orderentryconfig?v=full`,
```
brandones, line 44: Were you not able to get it working with the custom format @piyushmishra1416 ? If not, please report the request payload size difference between the old call and the new one.

piyushmishra1416:
Fetching the data through this URL `ws/rest/v1/orderentryconfig?v=custom:(uuid,display,concept:(names:(display,uuid))` only retrieves data for orderFrequency, not for other required data such as drugRoutes and drugDosingUnits. So now I have two options from here: 
- Either use this URL `ws/rest/v1/orderentryconfig?v=full` which has a payload of `870kb`.

- Call two different URLS; use the custom URL (`ws/rest/v1/orderentryconfig?v=custom:(uuid,display,concept:(names:(display,uuid))`) for orderFrequency and the other URL (`ws/rest/v1/orderentryconfig`) for the remaining fields.

What would be the correct approach to proceed from here @brandones ?

brandones:
Hmm, that is annoying. And that's a pretty big payload for `full`, so probably best to avoid that. I guess the ideal thing is to just modify the custom format to include whatever is needed for drug routes and drug dosing units. Do you think you can do that? Just look at what properties the default object includes and try adding those. You should only need to add the top-level properties, I think it will default to including the correct sub-properties for them.

ibacher:
What about using `?v=custom:(uuid,display)`. Still requires two calls (one to get everything else, then one to get the concepts for `orderFrequencies`)...

I'm inclined to think that this should be reported as an issue to the backend. The problem is that only the `orderFrequencies` have a `concept` property, so a request that has `concept` in the custom format will fail.

piyushmishra1416:
> What about using `?v=custom:(uuid,display)`. Still requires two calls (one to get everything else, then one to get the concepts for `orderFrequencies`)...
> 
> I'm inclined to think that this should be reported as an issue in the backend. The problem is that only the `orderFrequencies` have a `concept` property, so a request that has `concept` in the custom format will fail for most other categories.

Should I create a ticket for this @ibacher?

ibacher:
Yes, that's what I meant.

brandones:
In the meantime, can we do this for this PR so we can close it?

> What about using ?v=custom:(uuid,display). Still requires two calls (one to get everything else, then one to get the concepts for orderFrequencies)...

piyushmishra1416:
Hello @brandones, could you please review this PR? I have followed the suggested approach.

packages/esm-patient-medications-app/src/api/order-config.ts
```diff
@@ -57,16 +72,24 @@ export function useOrderConfig(): {
...
+          const getAbbreviation = (names: Names[]) => {
+            return names.map((name) => name.display);
+          };
+
+          const abbreviation = getAbbreviation(concept.names);
+
+          return {
+            valueCoded: uuid,
+            value: display,
+            abbreviation: abbreviation,
```
brandones, line 85:
```suggestion
          return {
            valueCoded: uuid,
            value: display,
            names: concept.names.map(name => name.display),
```

An array should generally have a pluralized name, so `abbreviations` rather than `abbreviation`. But also this isn't just the abbreviations. This is all the names. So let's just call it names. Please adjust the rest of the code accordingly.

brandones (changes requested):
See my suggested changes. You can commit a suggestion using the buttons there. But then you'll need to make the `abbreviation` --> `names` change throughout.

Also, please add a test.

I'm really glad we've found a way to get this working! Thank you so much for your work and responsiveness on this.

piyushmishra1416:
Hello @brandones, can you please review this PR?

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
```diff
@@ -302,10 +302,13 @@ export function DrugOrderForm({ initialOrderBasketItem, onSave, onCancel, prompt
...
 
-  const orderFrequencies: Array<MedicationFrequency> = useMemo(
-    () => orderConfigObject?.orderFrequencies ?? [],
-    [orderConfigObject],
-  );
+  const orderFrequencies: Array<MedicationFrequency> = useMemo(() => {
+    return orderConfigObject?.orderFrequencies ?? [];
+  }, [orderConfigObject]);
+
+  const filterItems = (menu) => {
```
ibacher, line 309: Probably a good idea to wrap this in `useMemo()`?

packages/esm-patient-medications-app/src/api/order-config.ts
```diff
@@ -9,9 +9,16 @@ import {
   type QuantityUnit,
 } from '../types';
 
+export interface Names {
+  uuid: string;
+  display: string;
+}
```
ibacher, line 15:
```suggestion
}

```

packages/esm-patient-medications-app/src/api/order-config.ts
```diff
@@ -9,9 +9,16 @@ import {
   type QuantityUnit,
 } from '../types';
 
+export interface Names {
```
ibacher, line 12: Interface names should generally be singular.

```suggestion
export interface ConceptName {
```

piyushmishra1416:
Hello @ibacher, I fixed the changes could you please review the code?

piyushmishra1416:
Hello @brandones @ibacher @denniskigen, could you please review this PR?
Thanks a lot for your patience, guidance, and support on this PR.

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
```diff
@@ -302,9 +302,15 @@ export function DrugOrderForm({ initialOrderBasketItem, onSave, onCancel, prompt
...
-    [orderConfigObject],
+  const orderFrequencies: Array<MedicationFrequency> = useMemo(() => {
+    return orderConfigObject?.orderFrequencies ?? [];
+  }, [orderConfigObject]);
+
+  const filterItems = useMemo(
+    () => (menu) => {
+      return menu?.item?.names?.some((abbr) => abbr.toLowerCase().includes(menu?.inputValue?.toLowerCase()));
+    },
+    [],
```
ibacher, line 313: Since this is wrapping a function, its preferable to use `useCallback()` instead of `useMemo()`.

I'd probably also do something like:

```
menu?.inputValue?.length ? menu.item?.names?.some((abbr) => abbr.toLowerCase().includes(menu.inputValue.toLowerCase())) : menu.item?.names ?? [];
```

And the complexity of that I think points to the fact that this shouldn't be a one-liner.

denniskigen (approved):
Thanks!

brandones:
Yes!!! So glad to see this having made it across the finish line. I want to highlight that while often it's a bad sign if PRs get as old as this one, I think this is a clear exception— @piyushmishra1416 I really appreciate your responsiveness and collaborative effort as we figured out how to fix this one. That quick turnaround time makes a world of difference. Great work, much appreciated.

piyushmishra1416:
Thank you so much @brandones for your kind words and recognition. It means a lot.

piyushmishra1416:
I'm so happy that we have finalized this PR. It was a great team effort, and your constant guidance and support on this PR was invaluable. It was filled with lots of learning. Thanks a ton, @ibacher, @denniskigen, and @brandones for all your help and support.

PR #1771, opened 2024-04-02
(feat) O3-3507 Add revised medications table to the medications dashboard

arodidev:
## Requirements

## Summary

This PR introduces a table which tracks revised medications within the medications dashboard. The data presented in the table differs in that it is filtered based on the `action` property to only return data where `action = 'REVISE'`.

`const revisedPatientOrders = patientOrders?.filter((order) => order.action == 'REVISE');`

The table component has been slotted into the `patient-chart-medications-dashboard-slot` for configurability's sake.

## Screenshots

Screenshot 2024-04-02 at 09 14 00

Screenshot 2024-04-02 at 11 05 37

## Related Issue

https://openmrs.atlassian.net/browse/O3-3507


ojwanganto:
@arodidev thanks for this work. I am wondering how this is eventually used on the ground. My raw interpretation of this task was to highlight medications which have been revised at dispensing or something of the sort and for that you'd need to highlight what exactly was changed/revised i.e. quantity, duration, etc. As it is now, one will still be asking how the revised order is different from the original one. Do you think this is a valid case?

arodidev:
> @arodidev thanks for this work. I am wondering how this is eventually used on the ground. My raw interpretation of this task was to highlight medications which have been revised at dispensing or something of the sort and for that you'd need to highlight what exactly was changed/revised i.e. quantity, duration, etc. As it is now, one will still be asking how the revised order is different from the original one. Do you think this is a valid case?

Hello @ojwanganto, yes I do agree that this is a valid case made. During the discussions phase we did not fully discuss what this would look like in it's final form, hence I opted for an initial MVP which could be build upon further down the line. I am open to further discussions on what this could look like as a final product and how it can best be modified to serve the user's needs.
cc: @Rugute

packages/esm-patient-medications-app/src/routes.json
```diff
@@ -12,6 +12,14 @@
         "columnSpan": 1
       }
     },
+    {
+      "name": "revised-medications-id",
```
brandones, line 16:
```suggestion
      "name": "revised-medications-widget",
```

brandones:
@arodidev Since this is a WIP and it's not clear that it will be included in the RefApp, please either (A) remove the "slot" property from the extension declaration in the `routes.json` file, so the extension does not appear by default but can be added using configuration (/ the implementer tools), or (B) create a feature flag which, when disabled, causes the widget to return `null`. I think (A) is probably the better option.

brandones:
CC @ibacher @denniskigen if they have thoughts about how we should go about maintaining these kinds of extensions, which are not attached in the default RefApp. Would be great not to have this be a way that dead code proliferates.

askanter:
I really don't get the use case here. Is revised medications currently active medications that have been changed since originally prescribing? Are Past Medications, medications that are no longer active in any form/dose? Is there some definitions somewhere in the design documents I could review?

brandones:
@askanter 

Right now the official designs only contain an "Active Medications" and "Past Medications" table; this "Revised Medications" table is the invention of the Ampath team.

> Are revised medications currently active medications that have been changed since originally prescribing?

Yes. The medication will appear both under "active medications" (with its updated prescription details) and "revised medications" (with the old details).

> Are Past Medications, medications that are no longer active in any form/dose?

Since the official designs only have "Active Medications" and "Past Medications," the Past Medications table contains all non-active prescriptions, including those that have been revised.

So Ampath will, with their "Revised Medications" table, see some of the same prescriptions duplicated in that table and the Past Medications table.

Hopefully that clears things up. Thanks for your participation in this, please keep the questions coming (though the UX Advisory slack channel is more likely to produce the kind of engagement that will be useful). This PR won't affect the RefApp.

brandones:
@arodidev Is it clear to you how to move forward with this?

brandones:
Closing since @arodidev is unresponsive

PR #1059, opened 2023-03-19
(feat) O3-1942: Lift form privileges to useForms hook

Jexsie:
## Requirements

- [x] This PR has a title that briefly describes the work done, including the ticket number if there is a ticket.
- [x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://zeroheight.com/23a080e38/p/880723-introduction).
- [x] My work includes tests or is validated by existing tests.

## Summary

 - Each encounter type has associated user privileges (viewPrivilege and editPrivilege. Previously, we used these to filter 
    forms in the [forms component](https://github.com/openmrs/openmrs-esm-patient-chart/blob/40a0b56f6bf6905061a24a93225e6353c0a9d874/packages/esm-patient-forms-app/src/forms/forms.component.tsx#L40); 
   Lifted this logic to be handled directly from the **useForms** hook with the offline awareness logic
## Related Issue

https://issues.openmrs.org/browse/O3-1942

PR #1806, opened 2024-04-23
(feat) 03-3090: Enable restricting which forms to show based on user privileges

gitcliff:
## Requirements

## Summary

This PR restricts users' access to forms based on their privileges. More specifically, this logic checks that the user has the same privileges as those specified in the form's encounterType `editPrivilege` property. 

## Related Issue

https://openmrs.atlassian.net/browse/O3-3090


packages/esm-patient-forms-app/src/forms/forms-dashboard.component.tsx
```diff
@@ -40,16 +44,27 @@ const FormsDashboard: React.FC<DefaultWorkspaceProps> = () => {
     [currentVisit, htmlFormEntryForms, mutateForms, patientUuid],
   );
 
+  if (Array.isArray(forms)) {
+    forms.forEach((item) => {
+      const editPrivilege = item.form.encounterType?.editPrivilege?.name;
+      sessionPrevillages.forEach((item) => {
```
ojwanganto, line 50: a small typo here

ibacher (dismissed):
It would probably be better to check `userHasAccess()`, since that also includes the logic for super users, but this is otherwise good.

packages/esm-patient-forms-app/src/forms/forms-dashboard.component.tsx
```diff
@@ -21,6 +21,10 @@ const FormsDashboard: React.FC<DefaultWorkspaceProps> = () => {
   const { patient, patientUuid } = usePatient();
   const { data: forms, error, mutateForms } = useForms(patientUuid, undefined, undefined, !isOnline, config.orderBy);
   const { currentVisit } = useVisitOrOfflineVisit(patientUuid);
+  const sessionUser = useSession();
+  const sessionPrivileges = sessionUser?.user?.privileges;
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  let newForms = [];
```
pirupius, line 27: Minor issue with the naming here because it's a little misleading. The forms aren't new but rather a filtered version based on privileges. Maybe `filteredForms` could be an option but not sure if that is the best description. 

Am sure @denniskigen has a bot for naming 🤖

ibacher: Or possibly `editableForms`.

gitcliff: @ibacher @pirupius changed it 
Thanks

packages/esm-patient-forms-app/src/forms/forms-dashboard.component.tsx
```diff
@@ -40,16 +44,27 @@ const FormsDashboard: React.FC<DefaultWorkspaceProps> = () => {
     [currentVisit, htmlFormEntryForms, mutateForms, patientUuid],
   );
 
+  if (Array.isArray(forms)) {
```
ibacher, line 47: Use `userHasAccess()` here instead of this, please.

gitcliff: @ibacher updated it ,,thanks

packages/esm-patient-forms-app/src/forms/forms-dashboard.component.tsx
```diff
@@ -21,6 +28,10 @@ const FormsDashboard: React.FC<DefaultWorkspaceProps> = () => {
   const { patient, patientUuid } = usePatient();
   const { data: forms, error, mutateForms } = useForms(patientUuid, undefined, undefined, !isOnline, config.orderBy);
   const { currentVisit } = useVisitOrOfflineVisit(patientUuid);
+  const sessionUser = useSession();
+  const sessionPrivileges = sessionUser?.user?.privileges;
```
ibacher, line 32: I think we can forgo this now?

packages/esm-patient-forms-app/src/forms/forms-dashboard.component.tsx
```diff
@@ -21,6 +28,10 @@ const FormsDashboard: React.FC<DefaultWorkspaceProps> = () => {
...
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
ibacher, line 33: I don't think this comment does anything as it's not around a hook anyways.

packages/esm-patient-forms-app/src/forms/forms-dashboard.component.tsx
```diff
@@ -21,6 +28,10 @@ const FormsDashboard: React.FC<DefaultWorkspaceProps> = () => {
...
+  let editableForms = [];
```
ibacher, line 34: I'd consider doing the filtering inside a `useMemo()` hook, since otherwise, we may re-render the FormsList on every render that hits this.

packages/esm-patient-forms-app/src/forms/forms-list.component.tsx
```diff
@@ -63,13 +63,15 @@ const FormsList: React.FC<FormsListProps> = ({ completedForms, error, sectionNam
...
     () =>
       filteredForms?.map((formData) => {
-        return {
-          id: formData.form.uuid,
-          lastCompleted: formData.lastCompleted ? formatDatetime(formData.lastCompleted) : undefined,
-          formName: formData.form.display ?? formData.form.name,
-          formUuid: formData.form.uuid,
-          encounterUuid: formData?.associatedEncounters[0]?.uuid,
-        };
+        if (formData) {
```
ibacher, line 66: Is this necessary? If so, it seems to indicate we're getting `null` or `undefined` entries in the `filteredForms` array and we should fix that.

ibacher (dismissed):
Thanks @gitcliff! A few more minor points.

denniskigen:
@ibacher @gitcliff the more I look at this, the more I think it could benefit from being put behind a config property so its opt-in. Something like:

```ts
export const configSchema = {
	// other properties omitted for brevity
	  filterFormsByPrivilege: {
	    _type: Type.Boolean,
	    _default: false,
	    _description: 'Whether to filter forms based on user privileges.',
	  },
}
```

And then we could have a hook that fetches all the forms and then filters them based on whether the `filterFormsByPrivilege` is truthy and we have a user session:

```tsx
function useFormFilters(patientUuid, session, isOnline, config) {
  const { data: forms, error, mutateForms } = useForms(patientUuid, undefined, undefined, !isOnline, config.orderBy);

  const filteredForms = useMemo(() => {
    if (session?.user && config.filterFormsByPrivilege) {
      return forms?.filter((formInfo) =>
        userHasAccess(formInfo?.form?.encounterType?.editPrivilege?.name, session.user),
      );
    }
    return forms;
  }, [forms, session.user, config.filterFormsByPrivilege]);

  return { forms: filteredForms, error, mutateForms };
}
``` 

Then the component is unchanged except for the call to fetch forms, which changes to:

```tsx
const { forms, error, mutateForms } = useFormFilters(patientUuid, session, isOnline, config);
```

packages/esm-patient-forms-app/src/forms/forms-list.component.tsx
gitcliff: @ibacher it was just a sanitary check , but i have removed it

packages/esm-patient-forms-app/src/forms/forms-list.component.tsx
```diff
@@ -64,11 +64,11 @@ const FormsList: React.FC<FormsListProps> = ({ completedForms, error, sectionNam
     () =>
       filteredForms?.map((formData) => {
```
denniskigen, line 65: nit:

```suggestion
      filteredForms
      	?.filter(formData => Boolean(formData))
      	.map((formData) => {
```

denniskigen (approved):
LGTM. Thanks, @gitcliff!

denniskigen:
Very late in the day, but I've just realized that we may have duplicated functionality here. See https://github.com/openmrs/openmrs-esm-patient-chart/pull/1059

denniskigen:
Isn't [this](https://github.com/openmrs/openmrs-esm-patient-chart/blob/3a4f3493ad997ba525594ac7dd62e4ace2ab433e/packages/esm-patient-forms-app/src/hooks/use-forms.ts#L86) the exact same check?

denniskigen:
Closing as a duplicate of https://github.com/openmrs/openmrs-esm-patient-chart/pull/1059. Sorry for the confusion, @gitcliff :)

PR #1796, opened 2024-04-19
(feat) O3-1422 / O3-2724: Move workspace system into the framework

brandones:
DEPENDS ON: https://github.com/openmrs/openmrs-esm-core/pull/979

## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspiration.
- [x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://om.rs/o3ui).
- [x] My work includes tests or is validated by existing tests.

## Summary

Apologies for the big PR. It is generally quite trivial. I have gone through and tested every single app that has a workspace.

## Screenshots
See the main core change.

## Related Issue
https://openmrs.atlassian.net/browse/O3-1421
https://openmrs.atlassian.net/browse/O3-2724

## Other

Depends on https://github.com/openmrs/openmrs-esm-core/pull/979

PR #1812, opened 2024-04-25
(feat) O3-2953: Add the ability to edit and delete vitals and biometrics

senthil-k8s:
## Requirements

## Summary

This PR enables editing or deleting vitals and biometrics data from the Patient Chart. In edit mode, obs are retrieved using the encounter UUID and displayed in the form. When deleting data, the encounter UUID is used to void observations.

## Screenshots
https://www.loom.com/share/b47c4ddf00724b6d97c17d71513d37c8?sid=19a6b632-fea9-4c09-9c84-03e41a7f6847

## Related Issue
https://openmrs.atlassian.net/browse/O3-2953


senthil-k8s:
pulling @denniskigen @ibacher @vasharma05 , as due to register workspace deprecated in this [PR](https://github.com/openmrs/openmrs-esm-patient-chart/pull/1796/files). The workspace doesn't seem to work for me from my side. is that same on your side as well, couldn't access the workspace?
Screenshot (977)
Screenshot (978)

denniskigen:
> pulling @denniskigen @ibacher @vasharma05 , as due to register workspace deprecated in this [PR](https://github.com/openmrs/openmrs-esm-patient-chart/pull/1796/files). The workspace doesn't seem to work for me from my side. is that same on your side as well, couldn't access the workspace? Screenshot (977) Screenshot (978)

Rebasing against main and then doing a hard reload in your browser should fix this.

senthil-k8s:
thanks @denniskigen. Working as expected!

packages/esm-patient-vitals-app/src/common/data.resource.ts
```diff
@@ -353,6 +361,16 @@ export function updateVitalsAndBiometrics(
...
 
+export function deleteVitalsAndBiometrics(encounterUuid: string, abortController: AbortController) {
+  return openmrsFetch(`${restBaseUrl}/encounter/${encounterUuid}`, {
+    method: 'DELETE',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    signal: abortController.signal,
+  });
+}
```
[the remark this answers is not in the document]
ibacher: Do note that while for the example data in dev3, vitals are mostly captured as a separate encounter, this is absolutely _not_ a pattern we can guarantee, i.e., vitals captured on other forms are also reported in this widget.

packages/esm-patient-vitals-app/src/vitals-biometrics-form/delete-vitals-biometrics-modal.component.tsx
```diff
@@ -0,0 +1,64 @@
...
+      createErrorHandler();
+      showSnackbar({
+        title: t('vitalsAndBiometricsDeleteError', 'Error Deleting Vitals and Biometrics'),
+        kind: 'error',
+        isLowContrast: false,
+      });
+    } finally {
+      closeDeleteModal();
+      setIsDeleting(false);
+      abortController.abort();
```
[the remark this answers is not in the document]
ibacher: Yeah, at this point, the requests have all been settled, so it's effectively an expensive no-op.

denniskigen:
@senthil-k8s so one of the main things to take care of here is that deleting vitals and biometrics data should void the observations instead of deleting the encounter.

packages/esm-patient-vitals-app/src/vitals/vitals-overview.component.tsx
```diff
@@ -217,6 +217,7 @@ const VitalsOverview: React.FC<VitalsOverviewProps> = ({ patientUuid, pageSize,
                 <div ref={contentToPrintRef}>
                   <PrintComponent subheader={headerTitle} patientDetails={patientDetails} />
                   <PaginatedVitals
+                    patientUuid={patientUuid}
```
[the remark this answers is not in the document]
ibacher: This is one of those points where the multiple React trees causes things to not work as we might like (only the patient chart component actually parses the URL. Maybe this would be a good case for modifying the component decorator to put the props passed to an extension in a React context and then we could create a `useExtensionProps()` hook to retrieve them?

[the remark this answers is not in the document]
denniskigen: Only if it's more performant than drilling the prop down. Otherwise, it sounds like a lot of overhead.

packages/esm-patient-vitals-app/src/vitals-biometrics-form/delete-vitals-biometrics-modal.component.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const deleteVitalsBiometricsDialog = async () => {
+    const abortController = new AbortController();
+    try {
+      const response = await deleteVitalsAndBiometrics(encounterUuid, abortController);
+      invalidateCachedVitalsAndBiometrics();
+      showSnackbar({
+        isLowContrast: true,
+        kind: 'success',
+        title: t('vitalsAndBiometricsDeleted', 'Vitals and Biometrics Deleted'),
+      });
```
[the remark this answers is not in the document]
denniskigen: Right. Feel free to ignore this suggestion then, @senthil-k8s.

packages/esm-patient-vitals-app/src/vitals/vitals-overview.component.tsx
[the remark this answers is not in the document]
ibacher: Hmmm... I think it'd be slightly more _sane_ than deep prop drilling, but it's unlikely to be more performant. Technically, I think that `usePatient()` does parse the URL correctly for the chart, but I also don't think it's better to do that multiple times.

I'm going to go with: this is one of the things that the AppContext feature is intended to fix, so that you can just call `usePatient()` and it will contextually "know" which patient is active (at least in the chart!).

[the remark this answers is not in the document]
denniskigen: Sounds good!

denniskigen:
@ibacher so what's the ideal approach for @senthil-k8s to pursue w.r.t to voiding the observations using the Delete action?

packages/esm-patient-vitals-app/src/vitals-biometrics-form/delete-vitals-biometrics-modal.component.tsx
```diff
@@ -0,0 +1,64 @@
...
+    } catch (error) {
+      createErrorHandler();
```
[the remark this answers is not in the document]
brandones: Huh we should really document [that function](https://github.com/openmrs/openmrs-esm-core/blame/c10b70bf30e3ecbb240ac6abc81b1a61caa99d55/packages/framework/esm-error-handling/src/index.ts#L30). It definitely does nothing as used here. As the name would suggest, the thing that it does is to create an error handler, which it returns. `reportError` would be appropriate, and then you could get rid of the `showSnackbar` here, because I think `reportError` handles the notification.

senthil-k8s:
@denniskigen editing the vitals and biometrics i.e., editing the existing obs. problem in my approach is creating new obs every time when user editing vitals. Editing vitals could lead to [POST request for each obs](https://github.com/openmrs/openmrs-esm-patient-chart/blob/c5cb4cc5023704d8b0b06425336736bd8d7df4cd/packages/esm-patient-orders-app/src/lab-results/lab-results.resource.ts#L126). Is there any better approach to edit the vitals here.

packages/esm-patient-vitals-app/src/common/data.resource.ts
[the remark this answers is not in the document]
senthil-k8s: @ibacher @denniskigen while deleting the encounter, it deleted the obs only and set voided as true. We could retrieve the same encounter by it's uuid. Does this approach lead to data loss?



https://github.com/openmrs/openmrs-esm-patient-chart/assets/93466587/b0fb60ce-de30-43c8-981f-e217b04637c8

denniskigen:
@samuelmale do you mind clarifying what the approach should look like?

ibacher:
I think this is where EMR API could actually be helpful, but I'm not sure what the specific API to use there is. Otherwise, yes, I think the only real option is a request-per-obs.

samuelmale:
I think we just have to grep and group obs by their category ie. "vitals" vs "biometrics" and then simply mark them as `voided`:

```ts

function getVitalsObs(encounter: Encounter): Array<OpenmrsResource> {
   // return vitals specific observations
}

function deleteVitals(encounter: Encounter) {
   const vitalsToBeVoided = getVitalsObs(encounter).map(obs => ({ uuid: obs.uuid, voided: true }));
   // You can either save them all at once as part of the encounter OR save them individually using the Obs endpoint.
}

```

I prefer saving all the voided obs apart of the encounter object in a single shot.

senthil-k8s:
@samuelmale, @denniskigen, could you please let me know if my approach makes sense?

https://www.loom.com/share/c85f115c1e8b417a9f920b2f3b93e8b5

denniskigen:
@senthil-k8s I've squashed your commits into a single one to ensure a cleaner commit history. This should simplify future rebasing and merges.

Thank you for your understanding and contributions!

packages/esm-patient-vitals-app/src/vitals-biometrics-form/delete-vitals-biometrics-modal.component.tsx
```diff
@@ -0,0 +1,99 @@
...
+        showSnackbar({
+          isLowContrast: false,
+          kind: 'error',
+          title: t('encounterNotFound', 'Encounter not found'),
+        });
+      }
+
+      const updatedEncounter = {
+        ...response.data,
+        obs: response.data.obs?.map((obs) => ({ ...obs, voided: true })) || [],
```
ibacher, line 34: This is still incorrect, in that it assumes all `obs` in the encounter are vitals, which cannot always be guaranteed to be the case.

packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.workspace.tsx
```diff
@@ -164,7 +194,100 @@ const VitalsAndBiometricsForm: React.FC<DefaultPatientWorkspaceProps> = ({
...
+          config.concepts,
+          patientUuid,
+          formData,
+          date,
+          abortController,
+          vitalsBiometricsFormData?.uuid,
+          session?.sessionLocation?.uuid,
+        );
+
+        if (response.status === 200) {
```
ibacher, line 210: Since this calls `openmrsFetch()` you should not have to do this. The promise will reject if the response is not ok.

packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.workspace.tsx
```diff
@@ -164,7 +194,100 @@ const VitalsAndBiometricsForm: React.FC<DefaultPatientWorkspaceProps> = ({
...
+        createErrorHandler();
+        showSnackbar({
+          title: t('vitalsAndBiometricsEditError', 'Error editing vitals and biometrics'),
+          kind: 'error',
+          isLowContrast: false,
+          subtitle: t('checkForValidity', 'Some of the values entered are invalid'),
+        });
+      } finally {
+        abortController.abort();
+      }
```
ibacher, line 231:
```suggestion
      }
```

ibacher: And other places. This is unnecessary.

packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.workspace.tsx
```diff
@@ -164,7 +194,100 @@ const VitalsAndBiometricsForm: React.FC<DefaultPatientWorkspaceProps> = ({
...
+          patientUuid,
+          formData,
+          date,
+          abortController,
+          vitalsBiometricsFormData?.uuid,
+          session?.sessionLocation?.uuid,
+        );
+
+        if (response.status === 200) {
+          invalidateCachedVitalsAndBiometrics();
```
ibacher, line 211: Shouldn't we call `setIsSubmitting(false);` here too?

packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.workspace.tsx
```diff
@@ -93,6 +105,26 @@ const VitalsAndBiometricsForm: React.FC<DefaultPatientWorkspaceProps> = ({
   const [showErrorNotification, setShowErrorNotification] = useState(false);
   const [showErrorMessage, setShowErrorMessage] = useState(false);
 
+  const vitalsBiometricsFormData = vitalsBiometrics?.filter(
+    (vitalsBiometrics) => vitalsBiometrics.uuid === encounterUuid,
+  )[0];
+
+  const defaultValues =
+    formContext === 'editing' && vitalsBiometricsFormData
+      ? {
```
ibacher, line 114: I don't really understand why we're creating this object here? Would be harmful to do:

```ts
const defaultValues = formContext === 'editing' ? vitalsBiometricsFormData ?? {} : {};
```

__mocks__/vitals-and-biometrics.mock.ts
```diff
@@ -1,3 +1,5 @@
+import { PatientVitalsAndBiometrics } from '../packages/esm-patient-vitals-app/src/common';
```
ibacher, line 1:
```suggestion
import { type PatientVitalsAndBiometrics } from '../packages/esm-patient-vitals-app/src/common';
```

packages/esm-patient-vitals-app/src/vitals-biometrics-form/delete-vitals-biometrics-modal.component.tsx
denniskigen: What should happen here instead, @ibacher?

ibacher: I think we have to look at the obs concept and make sure it's one in the set of vitals.

packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.workspace.tsx
denniskigen: The more concise approach you suggested doesn't work as well if `vitalsBiometricsFormData` has nested undefined values. The detailed object construction approach appears to work better when `vitalsBiometricsFormData` is partially populated as it excludes undefined values.

packages/esm-patient-vitals-app/src/vitals-biometrics-form/delete-vitals-biometrics-modal.component.tsx
```diff
@@ -0,0 +1,106 @@
...
+      }
+
+      const conceptUuids = new Set(Object.values(concepts));
+
+      const updatedEncounter = {
+        ...response.data,
+        obs: (response.data.obs || [])
+          .filter((obs) => conceptUuids.has(obs.concept.uuid))
+          .map((obs) => ({ ...obs, voided: true })),
+      };
```
denniskigen, line 41: Is this the right approach, @ibacher?

packages/esm-patient-vitals-app/src/vitals-biometrics-form/delete-vitals-biometrics-modal.component.tsx
denniskigen: See [this](https://github.com/openmrs/openmrs-esm-patient-chart/pull/1812#discussion_r1639704029) below.

packages/esm-patient-vitals-app/src/vitals-biometrics-form/delete-vitals-biometrics-modal.component.tsx
ibacher: Seems right to me!

ibacher (approved):

denniskigen:
Ok, thanks, guys! Looks like this is good to go!

denniskigen (approved):

PR #1813, opened 2024-04-25
(fix) Use order reasons should correctly chunk concept references and  load all of them

ibacher:
## Requirements

## Summary

This just fixes the loading of concept references here so that:

1. They are chunked properly
2. We actually load all the chunks via SWR instead of just the first


denniskigen:
The approach looks good to me @ibacher, thanks! Do you know which test configured in O3 has reasons for ordering associated with it?

denniskigen (approved):

PR #1815, opened 2024-04-29
(fix) O3-3133: Clicking the cancel button should not close the workspace

denniskigen:
## Requirements

## Summary

Presently, clicking the `Cancel` button on an active form closes the workspace panel when using the React Form Engine. This feels wrong. 

Users launch forms by launching the Clinical forms workspace first and then clicking a form they’d like to fill. Clicking the Cancel button could be assumed to mean “Hey, I won’t fill this form, and want to go back to the forms list to view a different form”. 

To that end, the `Cancel` button logic should be amended to:

- Close the active form
- Return to the forms list

In that order.

### Current behaviour

https://github.com/openmrs/openmrs-esm-patient-chart/assets/8509731/c1548c45-de35-42a1-849b-3eb4a744aab1

### Desired behaviour
 
https://github.com/openmrs/openmrs-esm-patient-chart/assets/8509731/b819f7a1-2123-48aa-8a01-941404cd820c

## Related Issue

https://openmrs.atlassian.net/browse/O3-3133?atlOrigin=eyJpIjoiMTc0MjA2OTdkNWM0NGM4M2I1MmEyYWU1ZmNlMGZkNWEiLCJwIjoiaiJ9

## Other

Looks like this is the existing behaviour for the AFE as well. This change should be propagated there too.

### DIFF
--- a/packages/esm-form-engine-app/src/form-renderer/form-renderer.component.tsx
+++ b/packages/esm-form-engine-app/src/form-renderer/form-renderer.component.tsx
@@ -1,36 +1,46 @@
-import React from 'react';
+import React, { useCallback } from 'react';
 import { useTranslation } from 'react-i18next';
 import { InlineLoading } from '@carbon/react';
 import { FormEngine } from '@openmrs/openmrs-form-engine-lib';
 import { type Visit } from '@openmrs/esm-framework';
-import useFormSchema from '../hooks/useFormSchema';
+import { launchPatientWorkspace, type DefaultWorkspaceProps } from '@openmrs/esm-patient-common-lib';
 import FormError from './form-error.component';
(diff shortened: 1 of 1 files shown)


pirupius (approved):
Not a blocker to this PR; would it be right to assume the same behaviour should be with save form? Should saving form take the user back to the forms workspace or close it?

ibacher:
> Should saving form take the user back to the forms workspace or close it?

I think the answer here varies. If the user opened the form workspace from, e.g., editing an existing encounter or clicking a link in the center, the workspace should close if cancelled or saved. However, if the user opens the form list and selects a form, we should probably send them back to the form list.

ibacher (approved):

PR #1823, opened 2024-05-07
(feat) O3-3158: Conditionally render Patient Header in the Start Visit form

mogoodrich:
## Requirements

## Summary

See: https://openmrs.atlassian.net/browse/O3-3158


ibacher:
Code looks fine. Could we make the "optional" part of this something controllable via the config system?

mogoodrich:
@ibacher  can you comment more on what you mean by making optional part of the config?  FWIW, in our use case where we display the header is context-dependent, for instance in our specific use case I would simply add a "showPatientHeader: true" here (which is opening the check in form within the appointments app, which is outside the patient chart):

https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-appointments-app/src/appointments/common-components/checkin-button.component.tsx#L34

ibacher:
@mogoodrich Well, I mostly meant that since the default value is `showPatientHeader = true`, so the existing uses of this will automatically add the header. Basically, I think it would be cleaner if the default is `false`.

mogoodrich:
Ah, thanks @ibacher , that was straight out bug on my part, it was meant to have a default of false, nice catch, I have updated.

denniskigen:
@mogoodrich I think Ian means we should add this as a configurable property to the config schema for the chart app.

mogoodrich:
@denniskigen  see my response (and fix) based on @ibacher follow-up comment... I don't think this should actually be a configuration property because it's more context-specific (you want this component to render the patient header when opened outside the patient chart, but not when opened within the chart) that configuration-specific.  Hope that makes sense.

denniskigen (approved):

denniskigen:
Yeah, makes sense. Thanks!

mogoodrich:
Thanks @denniskigen !

PR #1835, opened 2024-05-15
(fix) O3-3195: Fix rendering logic for the weight tile

jona42-ui:
## Requirements

## Summary

We show a weight tile above the form UI in the tablet viewport. The logic that determines what content to show in it is currently incorrect. Instead of checking whether the entire biometrics array is empty, it should check whether weight values exist in the array instead. The rest of the logic should continue to work as expected.

TL;dr: Changed the check in the WeightTile component that renders the weight data to check the length of weightData instead of biometrics.

## Screenshots

https://openmrs.atlassian.net/browse/O3-3195


denniskigen (approved):
LGTM. Thanks, @jona42-ui!

PR #1843, opened 2024-05-21
(fix) O3-3158: Restore abortControllers

ibacher:
## Requirements

## Summary

Fixes some boneheaded issues I accidentally introduced in #1831 by keeping the abort controller local to the function. For most places, I've still removed the abort controller since we never used it.

While working on this, I also noticed an issue with how the contents are justified on very "long" screens. See the attached screen shots.

## Screenshots

**Before**

Screenshot showing contents far apart

**After**

Screenshot showing contents closer together

### DIFF
--- a/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.scss
+++ b/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.scss
@@ -12,12 +12,12 @@
 }
 
 .heading {
-  @include type.type-style("heading-03");
+  @include type.type-style('heading-03');
   margin: spacing.$spacing-05;
 }
 
 .sectionTitle {
-  @include type.type-style("heading-compact-02");
+  @include type.type-style('heading-compact-02');
   color: $text-02;
   margin: 0 0 spacing.$spacing-03 0;
 }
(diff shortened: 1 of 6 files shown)


mogoodrich (commented):
Are the yarn.lock changes intentional?

ibacher:
@mogoodrich I updated the framework and openmrs to include [this fix](https://github.com/openmrs/openmrs-esm-core/commit/ae063dfeaccda2efcc3d9268f149c735ddef7ec4) while testing, so yes.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.scss
```diff
@@ -44,10 +44,15 @@
 .form {
   display: flex;
   flex-direction: column;
-  justify-content: space-between;
+  justify-content: start;
   height: var(--desktop-workspace-window-height);
```
denniskigen, line 48: This should probably be propagated to more places. We use this logic in most workspace forms.

denniskigen (approved):
LGTM. Thanks, @ibacher.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.scss
ibacher: Definitely. I think it's more visible here because we have a slightly larger number of direct children. The other trick here was the `justify-self: end` and `margin-top: auto` applied to the buttons... Otherwise the rendered somewhere in the middle of the form.

PR #1844, opened 2024-05-22
(feat) O3-3253: Move the details panel button below the Patient Header in the workspace

usamaidrsk:
## Requirements

## Summary

 This PR moves the show details button below the patient details.

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-chart/assets/53287480/e4376600-fff4-439c-b98b-78868c93beef

## Related Issue

[O3-3253](https://openmrs.atlassian.net/browse/O3-3253)

## Other

[O3-3253]: https://openmrs.atlassian.net/browse/O3-3253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/esm-patient-banner-app/src/banner/patient-banner.component.tsx
+++ b/packages/esm-patient-banner-app/src/banner/patient-banner.component.tsx
@@ -43,6 +43,9 @@ const PatientBanner: React.FC<PatientBannerProps> = ({ patient, patientUuid, hid
   }, []);
 
   const isDeceased = Boolean(patient?.deceasedDateTime);
+  // render details button below patient details for workspaces
+  // 520px is the maximum width a workspace occupies
+  const renderDetailsButtonBelow = patientBannerRef.current?.scrollWidth <= 520;
 
   return (
     <header
@@ -65,7 +68,7 @@ const PatientBanner: React.FC<PatientBannerProps> = ({ patient, patientUuid, hid
               isDeceased={patient.deceasedBoolean}
             />
           ) : null}
-          {!isTabletViewport ? (
+          {!renderDetailsButtonBelow ? (
             <PatientBannerToggleContactDetailsButton
               className={styles.toggleContactDetailsButton}
               toggleContactDetails={toggleContactDetails}
@@ -74,7 +77,7 @@ const PatientBanner: React.FC<PatientBannerProps> = ({ patient, patientUuid, hid
           ) : null}
         </div>
       </div>
-      {isTabletViewport ? (
+      {renderDetailsButtonBelow ? (
         <PatientBannerToggleContactDetailsButton
           className={styles.toggleContactDetailsButton}
           toggleContactDetails={toggleContactDetails}
--- a/packages/esm-patient-banner-app/src/banner/patient-banner.component.tsx
+++ b/packages/esm-patient-banner-app/src/banner/patient-banner.component.tsx
@@ -43,10 +43,8 @@ const PatientBanner: React.FC<PatientBannerProps> = ({ patient, patientUuid, hid
   }, []);
 
   const isDeceased = Boolean(patient?.deceasedDateTime);
-  // render details button below patient details for workspaces
-  // 520px is the maximum width a workspace occupies
-  const workspaceWidth = 520;
-  const showDetailsButtonBelowHeader = patientBannerRef.current?.scrollWidth <= workspaceWidth;
+  const maxDesktopWorkspaceWidthInPx = 520;
+  const showDetailsButtonBelowHeader = patientBannerRef.current?.scrollWidth <= maxDesktopWorkspaceWidthInPx;
 
   return (
     <header
@@ -64,9 +62,9 @@ const PatientBanner: React.FC<PatientBannerProps> = ({ patient, patientUuid, hid
         <div className={styles.buttonCol}>
           {!hideActionsOverflow ? (
             <PatientBannerActionsMenu
-              patientUuid={patientUuid}
-              actionsSlotName={'patient-actions-slot'}
+              actionsSlotName="patient-actions-slot"
               isDeceased={patient.deceasedBoolean}
+              patientUuid={patientUuid}
             />
           ) : null}
           {!showDetailsButtonBelowHeader ? (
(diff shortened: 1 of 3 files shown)


mogoodrich:
@usamaidrsk  can we have a screenshot of how this would look when used as the patient banner on the patient chart, and also with the actions button present.

denniskigen:
@ibacher is this the implementation you envisaged?

ibacher:
I would've anticipated only moving the button in the condensed version of the patient banner. I.e., in the chart, this actually makes the banner take up more visual space, which seems undesirable.

denniskigen:
I agree. Is there a contextual prop @usamaidrsk should be checking for to know whether the banner is a condensed version or not?

ibacher:
`patientBannerRef.current.scrollWidth` maybe? If it's <= 500px it should be being rendered in the workspace (we need to talk to Paul or Ciarán about the workspace size... It looks like we're mostly using the wide version everywhere?).

denniskigen:
It looks like we could also conditionally render based on [isTabletViewport](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-banner-app/src/banner/patient-banner.component.tsx#L20) which is computed from `ResizeObserver`. Gives you this:


https://github.com/openmrs/openmrs-esm-patient-chart/assets/8509731/a729f31c-93dc-4e05-bca1-fdd9a1df1924

usamaidrsk:
> It looks like we could also conditionally render based on [isTabletViewport](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-banner-app/src/banner/patient-banner.component.tsx#L20) which is computed from ResizeObserver. Gives you this:

Yes, @denniskigen that does do the trick.

packages/esm-patient-banner-app/src/banner/patient-banner.component.tsx
```diff
@@ -64,13 +64,22 @@ const PatientBanner: React.FC<PatientBannerProps> = ({ patient, patientUuid, hid
               isDeceased={patient.deceasedBoolean}
             />
           ) : null}
-          <PatientBannerToggleContactDetailsButton
-            className={styles.toggleContactDetailsButton}
-            toggleContactDetails={toggleContactDetails}
-            showContactDetails={showContactDetails}
-          />
+          {!isTabletViewport ? (
```
ibacher, line 67: So, I was expecting the condition here to be something like:

```ts
const renderDetailsButtonBelow = !isTabletViewport && (!Boolean(patientBannerRef.current) || patientBannerRef.current.scrollWidth <= 500);
```

Basically:

1. We do not want to move the details button in tablet mode, because tablet mode is actually wider than the workspace, so this should be a non-issue. (We may want to move it in tablet mode when there are less than maybe 800px of screen width, but let's cross that if it comes up).
2. We only want to move the icon if the `<header />` element is 500px or less, because that indicates that we're being rendered in the workspace (tablet mode is a minimum of 600px of screen real-estate).
3. If the ref has not yet been set, we don't know which of these two modes to use, so we just assume the default (render to the right).

mogoodrich: Defer to @ibacher  on this....

ibacher: This is the technical version of "we leave the patient banner as-is, but where there’s 500px or less of width to render the banner in, we move the “Show Details“ to a new line" 😁

usamaidrsk: I have pushed an update that uses
```typescript
const renderDetailsButtonBelow = patientBannerRef.current?.scrollWidth <= 520;
``` 

As per [zeroheight workspace documentation](https://zeroheight.com/23a080e38/p/483a22-workspace).
> On desktop screens, the workspace sits on the right side of the page. Its width is 420px in most cases, and 520px if it's a clinical form.

packages/esm-patient-banner-app/src/banner/patient-banner.component.tsx
```diff
@@ -42,6 +42,9 @@ const PatientBanner: React.FC<PatientBannerProps> = ({ patient, patientUuid, hid
   }, []);
 
   const isDeceased = Boolean(patient?.deceasedDateTime);
+  // render details button below patient details for workspaces
+  // 520px is the maximum width a workspace occupies
+  const renderDetailsButtonBelow = patientBannerRef.current?.scrollWidth <= 520;
```
denniskigen, line 47: A naming suggestion:

```suggestion
	const maxWorkspaceWidth = 520;
	
  const showDetailsButtonBelowHeader = patientBannerRef.current?.scrollWidth <= maxWorkspaceWidth;
```

denniskigen: The current logic doesn't account for two of the conditions Ian suggested.

denniskigen, line 47:
```suggestion
	const maxWorkspaceWidth = 520;
	
  const showDetailsButtonBelowHeader = !isTabletViewport && (!Boolean(patientBannerRef.current) || patientBannerRef.current.scrollWidth <= maxWorkspaceWidth);
```

The current logic doesn't account for two of the situations Ian mentioned above.

usamaidrsk: Yes @denniskigen. Thanks for pointing this out. I tested it with the condition provided, but the appearance on the desktop view was not as expected. 

image

Therefore, I decided to only check the width of the banner instead. Whenever the width is less than 520px, the button will be displayed below the header. This way, the button will never appear under the header on tablets, as they are greater than 520px in width.

cc @ibacher

denniskigen:
So I've just tested this and it works well with workspaces:

https://github.com/openmrs/openmrs-esm-patient-chart/assets/8509731/ce6387c9-eacc-42a8-a4d6-303f65055ab2

It doesn't work so well with overlays (such as what we use for the Appointments form). Is it because we don't constrain the max-width @brandones? Perhaps setting a max-width and using that value for the condition is the ideal approach.

CleanShot 2024-05-27 at 8  56 49@2x

usamaidrsk:
Thanks @denniskigen!,
> It doesn't work as well with overlays (such as what we use for the Appointments form). Is it because we don't constrain the max-width 
image

I think the screenshot looks okay, given that the patient details on the banner are not compressed/squeezed as it was why @mogoodrich  suggested to improve the layout.

denniskigen (approved):
Ok. Approving and deferring to @mogoodrich to merge if it meets the acceptance criteria. Thanks, @usamaidrsk!

PR #1847, opened 2024-05-23
(feat) O3-3276: Add support for printing patient identification stickers

jnsereko:
## Requirements

## Summary

Print a sticker at the registration desk with basic patient information like Patient ID, First name, last name, age, gender.

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-chart/assets/58003327/fd73ae0e-871a-460d-8d34-941a481a0464

## Related Issue

https://openmrs.atlassian.net/browse/O3-3276

## Other

cc @michaelbontyes @pirupius @denniskigen @ibacher @hadijahkyampeire


jnsereko:
> Thanks @jnsereko for this. I think it would be great to also configure the identifiers to show on the print view, some are very sensitive

@CynthiaKamau, do you mean the implementer choosing which identifier types to include?  something like 
`idToInclude[ 'uuid_for_openmrs_identifier', 'uuid_for_someOther' ]`

packages/esm-patient-banner-app/src/config-schema.ts
```diff
@@ -24,11 +24,33 @@ export const configSchema = {
     _description: "Whether to use the relationship name as a link to the person's patient chart",
     _default: false,
   },
+  showPrintIdentifierStickerButton: {
+    _type: Type.Boolean,
+    _description: "Whether to display the 'Print Patient Identifier' in the patient Banner",
```
denniskigen, line 29:
```suggestion
    _description: "Whether to display the 'Print identifier sticker' button in the patient banner",
```

packages/esm-patient-banner-app/src/config-schema.ts
```diff
@@ -24,11 +24,33 @@ export const configSchema = {
...
+    _default: false,
+  },
+  printIdentifierStickerFields: {
+    _type: Type.Array,
+    _description: 'Patient fields to include in the identifier sticker to print',
```
denniskigen, line 34:
```suggestion
    _description: 'Patient demographics to include in the identifier sticker printout',
```

packages/esm-patient-banner-app/src/config-schema.ts
```diff
@@ -24,11 +24,33 @@ export const configSchema = {
...
+    _default: ['name', 'dateOfBirth', 'age', 'gender', 'identifier'],
+    _elements: {
+      _type: Type.String,
+    },
+  },
+  printIdentifierStickerSize: {
+    _type: Type.String,
+    _description:
+      'Paper size of the sticker to print.  Specify units ie mm or in or use named sizes like `148mm 210mm`, A1, A2, A4 ... A5',
```
denniskigen, line 43:
```suggestion
      'Specifies the paper size for printing the sticker. You can define the size using units (e.g., mm, in) or named sizes (e.g., "148mm 210mm", "A1", "A2", "A4", "A5").'
```

packages/esm-patient-banner-app/src/banner/patient-banner.component.tsx
```diff
@@ -57,13 +73,25 @@ const PatientBanner: React.FC<PatientBannerProps> = ({ patient, patientUuid, hid
...
-              patientUuid={patientUuid}
-              actionsSlotName={'patient-actions-slot'}
-              isDeceased={patient.deceasedBoolean}
-            />
-          ) : null}
+          <div className={styles.buttonRow}>
+            {showPrintIdentifierStickerButton && (
+              <Button
+                kind="ghost"
+                hasIconOnly={true}
```
denniskigen, line 80:
```suggestion
                hasIconOnly
```

Truthy boolean prop values can be omitted.

packages/esm-patient-banner-app/src/banner/patient-banner.component.tsx
```diff
@@ -57,13 +73,25 @@ const PatientBanner: React.FC<PatientBannerProps> = ({ patient, patientUuid, hid
...
+                renderIcon={(props) => <Printer size={16} {...props} />}
+                iconDescription={t('printIdentifierSticker', 'Print Identification Sticker')}
+                className={styles.printButton}
+                onClick={openModal}
+              ></Button>
```
denniskigen, line 85:
```suggestion
             />
```

packages/esm-patient-banner-app/src/banner/patient-banner.component.tsx
```diff
@@ -43,6 +50,15 @@ const PatientBanner: React.FC<PatientBannerProps> = ({ patient, patientUuid, hid
 
   const isDeceased = Boolean(patient?.deceasedDateTime);
 
+  const { showPrintIdentifierStickerButton } = useConfig() as ConfigObject;
```
denniskigen, line 53:
```suggestion
  const { showPrintIdentifierStickerButton } = useConfig<ConfigObject>();
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    expect(mockedCloseModal).toHaveBeenCalled();
+  });
+
+  test('calls the print function when print button is clicked', async () => {
+    const handlePrint = jest.fn();
+    (useReactToPrint as jest.Mock).mockReturnValue(handlePrint);
+
+    render(<PrintIdentifierSticker patient={mockPatient} closeModal={mockedCloseModal} />);
+    fireEvent.click(screen.getByText('Print'));
```
denniskigen, line 52: Prefer [userEvent](https://testing-library.com/docs/user-event/intro/). Also we should probably be querying the button by role instead.

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    await waitFor(() => {
+      expect(handlePrint).toHaveBeenCalled();
+    });
+  });
+});
```
denniskigen, line 58:
```ts
function renderPrintIdentifierSticker() {
	render(<PrintIdentifierSticker patient={mockPatient} closeModal={mockedCloseModal} />);
}
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    render(<PrintIdentifierSticker patient={mockPatient} closeModal={mockedCloseModal} />);
+
+    expect(screen.getByText('Print Identifier Sticker')).toBeInTheDocument();
+    expect(screen.getByText('Wilson John')).toBeInTheDocument();
+    expect(screen.getByText('100GEJ')).toBeInTheDocument();
+    expect(screen.getByText('1972-04-04')).toBeInTheDocument();
+  });
+
+  test('calls closeModal when cancel button is clicked', () => {
+    render(<PrintIdentifierSticker patient={mockPatient} closeModal={mockedCloseModal} />);
```
denniskigen, line 41:
```suggestion
    renderPrintIdentifierSticker();
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    fireEvent.click(screen.getByText('Cancel'));
+
+    expect(mockedCloseModal).toHaveBeenCalled();
+  });
+
+  test('calls the print function when print button is clicked', async () => {
+    const handlePrint = jest.fn();
+    (useReactToPrint as jest.Mock).mockReturnValue(handlePrint);
+
+    render(<PrintIdentifierSticker patient={mockPatient} closeModal={mockedCloseModal} />);
```
denniskigen, line 51:
```suggestion
    renderPrintIdentifierSticker();
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    openmrsFetch: jest.fn(),
+    useConfig: jest.fn().mockImplementation(() => ({
+      printIdentifierStickerFields: ['name', 'identifier', 'age', 'dateOfBirth', 'gender'],
+    })),
+  };
+});
+
+describe('PrintIdentifierSticker', () => {
+  test('renders the component', () => {
+    render(<PrintIdentifierSticker patient={mockPatient} closeModal={mockedCloseModal} />);
```
denniskigen, line 32:
```suggestion
    renderPrintIdentifierSticker();
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.scss
```diff
@@ -0,0 +1,34 @@
+@use '@carbon/styles/scss/spacing';
```
denniskigen, line 1: The indentation in this file looks off.

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,171 @@
...
+interface PrintIdentifierStickerProps {
+  patient: fhir.Patient;
+  closeModal: () => void;
+}
+
+const PrintIdentifierSticker: React.FC<PrintIdentifierStickerProps> = ({ patient, closeModal }) => {
+  const { t } = useTranslation();
+  const [isPrinting, setIsPrinting] = useState(false);
+  const contentToPrintRef = useRef(null);
+  const { printIdentifierStickerFields, printIdentifierStickerSize } = useConfig() as ConfigObject;
```
denniskigen, line 20:
```suggestion
  const { printIdentifierStickerFields, printIdentifierStickerSize } = useConfig<ConfigObject>();
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,171 @@
...
+  };
+
+  const patientDetails = {
+    id: patient?.id,
+    photo: patient?.photo,
+    name: patient?.name,
+    dateOfBirth: patient?.birthDate,
+    age: calculateAge(patient?.birthDate),
+    gender: patient?.gender,
+    adderesses: patient?.address,
```
denniskigen, line 45:
```suggestion
    address: patient?.address,
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,171 @@
...
+  const onBeforeGetContentResolve = useRef(null);
+
+  useEffect(() => {
+    if (isPrinting && onBeforeGetContentResolve.current) {
+      onBeforeGetContentResolve.current();
+    }
+  }, [isPrinting]);
+
+  const calculateAge = (birthDate) => {
+    const ageDifMs = Date.now() - new Date(birthDate).getTime();
```
jayasanka-sack, line 33: Just a small suggestion: how about using a more descriptive name for ageDifMs? Something like `ageDifferenceMs` or `msSinceBirth` could make it a bit clearer.

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,171 @@
...
+    const ageDate = new Date(ageDifMs);
+    return Math.abs(ageDate.getUTCFullYear() - 1970);
+  };
+
+  const patientDetails = {
+    id: patient?.id,
```
jayasanka-sack, line 39: the patient object shouldn't be null, right?

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.test.tsx
jayasanka-sack: Totally agree! `userEvent` is better when it comes to mimicking real user interactions.

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
jnsereko: @jayasanka-sack, what might happen in terms of low internet connection?  You know users will never wait for the whole page to load.  As long as the button is visible, they will always click it.

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.scss
jnsereko: Looks better now?

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+    renderPrintIdentifierSticker();
+
+    const printButton = screen.getByRole('button', { name: /Print/ });
+    expect(printButton).toBeInTheDocument();
+
+    await user.click(printButton);
+    await waitFor(() => {
+      expect(handlePrint).toHaveBeenCalled();
+    });
```
denniskigen, line 67:
```suggestion
    expect(handlePrint).toHaveBeenCalled();
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.test.tsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { mockPatient } from 'tools';
+import { useReactToPrint } from 'react-to-print';
+import PrintIdentifierSticker from './print-identifier-sticker.component';
+
+const mockedCloseModal = jest.fn();
```
denniskigen, line 8:
```suggestion
const mockedCloseModal = jest.fn();
const mockedUseReactToPrint = jest.mocked(useReactToPrint);
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+    useReactToPrint: jest.fn(),
+  };
+});
+
+jest.mock('@openmrs/esm-framework', () => {
+  const originalModule = jest.requireActual('@openmrs/esm-framework');
+
+  return {
+    ...originalModule,
+    openmrsFetch: jest.fn(),
```
denniskigen, line 24:
```suggestion
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.test.tsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
```
denniskigen, line 2:
```suggestion
import { render, screen } from '@testing-library/react';
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.scss
```diff
@@ -0,0 +1,32 @@
...
+@use '@carbon/styles/scss/type';
+@import '~@openmrs/esm-styleguide/src/vars';
```
denniskigen, line 3:
```suggestion
```

packages/esm-patient-banner-app/translations/en.json
```diff
@@ -1,12 +1,23 @@
...
   "country": "Country",
   "countyDistrict": "District",
   "district": "District",
+  "female": "Female",
+  "male": "Male",
+  "other": "Other",
+  "patientIdentifierSticker": "Patient Identifier Sticker",
   "postalCode": "Postal code",
+  "print": "Print",
+  "printIdentifierSticker": "Print Identifier Sticker",
```
denniskigen, line 18:
```suggestion
  "printIdentifierSticker": "Print identifier sticker",
```

packages/esm-patient-banner-app/src/config-schema.ts
```diff
@@ -24,11 +24,33 @@ export const configSchema = {
...
 };
 
 export interface ConfigObject {
   contactAttributeTypes: Array<string>;
   excludePatientIdentifierCodeTypes: Array<String>;
   customAddressLabels: Object;
   useRelationshipNameLink: boolean;
+  showPrintIdentifierStickerButton: boolean;
+  printIdentifierStickerFields: Array<String>;
+  printIdentifierStickerSize: String;
```
denniskigen, line 55:
```suggestion
  excludePatientIdentifierCodeTypes: Array<string>;
  customAddressLabels: Object;
  useRelationshipNameLink: boolean;
  showPrintIdentifierStickerButton: boolean;
  printIdentifierStickerFields: Array<string>;
  printIdentifierStickerSize: string;
```

https://www.typescriptlang.org/docs/handbook/basic-types.html#about-number-string-boolean-symbol-and-object

packages/esm-patient-banner-app/translations/en.json
```diff
@@ -1,12 +1,23 @@
...
+  "cancel": "Cancel",
   "city": "City",
   "cityVillage": "city",
   "country": "Country",
   "countyDistrict": "District",
   "district": "District",
+  "female": "Female",
+  "male": "Male",
+  "other": "Other",
+  "patientIdentifierSticker": "Patient Identifier Sticker",
```
denniskigen, line 15:
```suggestion
  "patientIdentifierSticker": "Patient identifier sticker",
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,178 @@
...
+    if (currentPair.length === 1) {
+      pairs.push(currentPair);
+    }
+
+    return pairs;
+  };
+
+  return (
+    <div>
+      <ModalHeader closeModal={closeModal} title={t('printIdentifierSticker', 'Print Identifier Sticker')} />
```
denniskigen, line 127:
```suggestion
      <ModalHeader closeModal={closeModal} title={t('printIdentifierSticker', 'Print identifier sticker')} />
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,178 @@
...
+      <ModalBody>
+        <div className={styles.stickerContainer} ref={contentToPrintRef}>
+          {printIdentifierStickerFields.includes('name') && (
+            <div key="name" className={styles.patientName}>
```
denniskigen, line 131:
```suggestion
            <div key="name" className={styles.patientName}>
```

Isn't the patient UUID more likely to be unique?

denniskigen (commented):
This is very close, @jnsereko. I've left a few more suggestions.

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.scss
```diff
@@ -0,0 +1,32 @@
+@use '@carbon/styles/scss/spacing';
+@use '@carbon/styles/scss/type';
```
denniskigen, line 2:
```suggestion
@use '@carbon/layout';
@use '@carbon/type';
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.scss
```diff
@@ -0,0 +1,32 @@
...
+@import '~@openmrs/esm-styleguide/src/vars';
+
+.stickerContainer {
+  padding: 1rem 1rem 0 1rem;
+}
+
+.row {
+  margin: spacing.$spacing-03 0rem 0rem;
```
denniskigen, line 10:
```suggestion
  margin: layout.$spacing-03 0rem 0rem;
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.scss
```diff
@@ -0,0 +1,32 @@
...
+  display: flex;
+  flex-flow: row wrap;
+  gap: spacing.$spacing-05;
```
denniskigen, line 13:
```suggestion
  gap: layout.$spacing-05;
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+      printIdentifierStickerFields: ['name', 'identifier', 'age', 'dateOfBirth', 'gender'],
+    })),
+  };
+});
+
+describe('PrintIdentifierSticker', () => {
+  test('renders the component', () => {
+    renderPrintIdentifierSticker();
+
+    expect(screen.getByText('Print Identifier Sticker')).toBeInTheDocument();
```
denniskigen, line 35:
```suggestion
    expect(screen.getByText(/Print Identifier Sticker/i)).toBeInTheDocument();
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+    expect(screen.getByText('100GEJ')).toBeInTheDocument();
+    expect(screen.getByText('1972-04-04')).toBeInTheDocument();
+  });
+
+  test('calls closeModal when cancel button is clicked', async () => {
+    const user = userEvent.setup();
+
+    renderPrintIdentifierSticker();
+
+    const cancelButton = screen.getByRole('button', { name: /Cancel/ });
```
denniskigen, line 46:
```suggestion
    const cancelButton = screen.getByRole('button', { name: /Cancel/i });
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  test('calls the print function when print button is clicked', async () => {
+    const handlePrint = jest.fn();
+    (useReactToPrint as jest.Mock).mockReturnValue(handlePrint);
+
+    const user = userEvent.setup();
+
+    renderPrintIdentifierSticker();
+
+    const printButton = screen.getByRole('button', { name: /Print/ });
```
denniskigen, line 61:
```suggestion
    const printButton = screen.getByRole('button', { name: /Print/i });
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+    const cancelButton = screen.getByRole('button', { name: /Cancel/ });
+    expect(cancelButton).toBeInTheDocument();
+
+    await user.click(cancelButton);
+    expect(mockedCloseModal).toHaveBeenCalled();
+  });
+
+  test('calls the print function when print button is clicked', async () => {
+    const handlePrint = jest.fn();
+    (useReactToPrint as jest.Mock).mockReturnValue(handlePrint);
```
denniskigen, line 55:
```suggestion
    mockedUseReactToPrint.mockReturnValue(handlePrint);
```

packages/esm-patient-banner-app/src/index.ts
```diff
@@ -28,6 +28,8 @@ export const deceasedPatientTag = getSyncLifecycle(deceasedPatientTagComponent,
 
 export const patientBanner = getSyncLifecycle(patientBannerComponent, options);
 
+export const printIdentifierSticker = getSyncLifecycle(printIdentifierStickerComponent, options);
```
denniskigen, line 31:
```suggestion
export const printIdentifierSticker = getSyncLifecycle(printIdentifierStickerComponent, options);
```

`getSyncLifecycle` or `getAsyncLifecycle` here, @ibacher? I forget.

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const handleBeforeGetContent = useCallback(() => {
+    return new Promise((resolve) => {
+      if (patient && headerTitle) {
+        closeModal();
+        onBeforeGetContentResolve.current = resolve;
+        setIsPrinting(true);
+        const printStyles = `@media print { @page { size: ${printIdentifierStickerSize}; } }`;
+
+        const style = document.createElement('style');
+        style.type = 'text/css';
```
denniskigen, line 73:
```suggestion
```

`type` is deprecated.

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
jnsereko: I didn't use the patient's id because that div is only for the patient name and not for the whole patient.

denniskigen:
Please fix the conflicts with `main`. Also, have you figured out why printing doesn't work the first time round?

packages/esm-patient-banner-app/src/index.ts
jnsereko: @denniskigen, I always prefer alining components like the already existing to avoid hours of debugging.  Since all elements in `esm-banner` are using `getSyncLifecycle` i decided to use that too

jnsereko:
> Please fix the conflicts with `main`. Also, have you figured out why printing doesn't work the first time round?

@denniskigen I have been banging my head on the wall since yesterday but not yet successful

packages/esm-patient-banner-app/src/index.ts
denniskigen: Thanks, @ibacher.

denniskigen:
A few other random notes:

- We're using the `modal` suffix for modal components - so `PrintIdentifierSticker` should be renamed `print-identifier-sticker.modal.tsx`.

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,198 @@
...
+        case 'unknown':
+          return t('unknown', 'Unknown');
+        default:
+          return gender;
+      }
+    };
+
+    const identifiers =
+      patient?.identifier?.filter(
+        (identifier) => !excludePatientIdentifierCodeTypes?.uuids.includes(identifier.type.coding[0].code),
```
ibacher, line 50: I think an allow-list of identifiers to include is better than a list of identifiers to exclude. It would also be nice to not necessarily tie this directly to UUIDs, e.g., it would be nice to be able to specify these by identifier name.

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,198 @@
...
+  }, [patient, t, excludePatientIdentifierCodeTypes?.uuids]);
+  const handleBeforeGetContent = useCallback(() => {
+    return new Promise<void>((resolve) => {
+      if (patient && headerTitle) {
+        onBeforeGetContentResolve.current = resolve;
+        setIsPrinting(true);
+        const printStyles = `@media print { @page { size: ${printIdentifierStickerSize}; } }`;
+
+        const style = document.createElement('style');
+        style.appendChild(document.createTextNode(printStyles));
```
ibacher, line 72: This seems messy. Let's move this into actual CSS and use a CSS var to set the size, that way we have something that's useful in more context than this. It would also avoid the issue where every time this callback is called, it will create yet another style node.

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,198 @@
...
+  };
+
+  return (
+    <div>
+      <ModalHeader closeModal={closeModal} title={t('printIdentifierSticker', 'Print identifier sticker')} />
+      <ModalBody>
+        <div className={styles.stickerContainer} ref={contentToPrintRef}>
+          {printIdentifierStickerFields.includes('name') && (
+            <div key="name" className={styles.patientName}>
+              {patientDetails?.name}
```
ibacher, line 152: It's weird to me that this renders at all if `patientDetails` is null or undefined. We should move that check up.

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,198 @@
...
+            });
+            setprintRetryCount(1);
+          }
+          resolve();
+        }
+      });
+    },
+  });
+
+  const renderElementsInPairs = (elements) => {
```
ibacher, line 113: We should convert this from a function to a component. That will behave better with React/

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,198 @@
...
+      <ModalBody>
+        <div className={styles.stickerContainer} ref={contentToPrintRef}>
+          {printIdentifierStickerFields.includes('name') && (
+            <div key="name" className={styles.patientName}>
+              {patientDetails?.name}
+            </div>
+          )}
+          {renderElementsInPairs(
+            [
+              patientDetails?.identifiers?.map((identifier, index) => (
```
ibacher, line 157: It seems like we should pre-check whether there are any identifiers before calling `renderElementsInPairs()` (or its component replacement).

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,198 @@
...
+                <div key={`identifier-${index}`}>
+                  <p key={`identifier-text-${index}`}>
+                    {identifier?.type?.text}: <strong>{identifier?.value}</strong>
```
ibacher, line 160: There shouldn't be a way for the identifier to be undefined or null here. Add a filter if necessary.

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,198 @@
...
+        <div className={styles.stickerContainer} ref={contentToPrintRef}>
+          {printIdentifierStickerFields.includes('name') && (
+            <div key="name" className={styles.patientName}>
+              {patientDetails?.name}
+            </div>
+          )}
+          {renderElementsInPairs(
+            [
+              patientDetails?.identifiers?.map((identifier, index) => (
+                <div key={`identifier-${index}`}>
```
ibacher, line 158: Prefer using something unique from the identifier, i.e., the `type.text` instead of `index` here.

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,198 @@
...
+                  <p key={`identifier-text-${index}`}>
+                    {identifier?.type?.text}: <strong>{identifier?.value}</strong>
+                  </p>
+                </div>
+              )),
+              <p key="gender">
+                {t('sex', 'Sex')}: <strong>{patientDetails?.gender}</strong>
+              </p>,
+              <p key="dateOfBirth">
+                {t('bod', 'DOB')}: <strong>{patientDetails?.dateOfBirth}</strong>
```
ibacher, line 168: Why `bod` here?

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,198 @@
...
+                {t('age', 'Age')}: <strong>{patientDetails?.age}</strong>
+              </p>,
+            ].flat(),
+          ).map((pair, index) => (
+            <Grid className={styles.gridRow} key={`grid-${index}`}>
+              <Column lg={8} md={4} sm={4}>
+                <div key={`pair-0-${index}`}>{pair[0]}</div>
+              </Column>
+              <Column lg={8} md={4} sm={4}>
+                <div key={`pair-1-${index}`}>{pair[1] || <div />}</div>
```
ibacher, line 180: Why do we render two empty divs if `pair[1]` is `null`, `undefined` or `false`?

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
ibacher: Also, sequence-wise, I think DOM modifications like this need to be made in a `useEffect()` hook. As [the docs say](https://react.dev/reference/react/useEffect): "useEffect is a React Hook that lets you [synchronize a component with an external system.](https://react.dev/learn/synchronizing-with-effects)". In React's model, the DOM is always an "external system".

ibacher: Basically  `handleBeforeGetContent()` should probably just call `setIsPrintingTrue()` and a `useEffect()` hook should load the CSS, I guess (really it should just set the CSS variable).

ibacher (commented):
Bunch of miscellaneous thoughts. I think, though, that the main thing is to look at [this example](https://github.com/MatthewHerbst/react-to-print?tab=readme-ov-file#setting-state-in-onbeforegetcontent) vs the code here. Note in particular, we are doing _nothing_ that ensures `isPrinting` is `true` _before_ the promise resolves.

denniskigen:
> Bunch of miscellaneous thoughts. I think, though, that the main thing is to look at [this example](https://github.com/MatthewHerbst/react-to-print?tab=readme-ov-file#setting-state-in-onbeforegetcontent) vs the code here. Note in particular, we are doing _nothing_ that ensures `isPrinting` is `true` _before_ the promise resolves.

Could you clarify this a bit more, @ibacher?

packages/esm-patient-banner-app/src/banner/patient-banner.component.tsx
```diff
@@ -60,13 +76,26 @@ const PatientBanner: React.FC<PatientBannerProps> = ({ patient, patientUuid, hid
...
         <PatientBannerPatientInfo patient={patient} />
         <div className={styles.buttonCol}>
-          {!hideActionsOverflow ? (
-            <PatientBannerActionsMenu
-              actionsSlotName="patient-actions-slot"
-              isDeceased={patient.deceasedBoolean}
-              patientUuid={patientUuid}
-            />
-          ) : null}
+          <div className={styles.buttonRow}>
```
ibacher, line 79: The print button should be moved into the actions menu rather than being a standalone thing here.

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,198 @@
...
+  patient: fhir.Patient;
+  closeModal: () => void;
+}
+
+const PrintIdentifierSticker: React.FC<PrintIdentifierStickerProps> = ({ patient, closeModal }) => {
+  const { t } = useTranslation();
+  const [isPrinting, setIsPrinting] = useState(false);
+  const contentToPrintRef = useRef(null);
+  const [printRetryCount, setprintRetryCount] = useState(0);
+  const { printIdentifierStickerFields, printIdentifierStickerSize, excludePatientIdentifierCodeTypes } = useConfig();
```
ibacher, line 20:
```suggestion
  const { printIdentifierStickerFields, printIdentifierStickerSize, excludePatientIdentifierCodeTypes } = useConfig<ConfigObject>();
```

ibacher: Note that the type for `excludePatientIdnetifierCodeTypes` is incorrect.

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+  const headerTitle = t('patientIdentifierSticker', 'Patient Identifier Sticker');
+
+  const onBeforeGetContentResolve = useRef(null);
+
+  useEffect(() => {
+    if (isPrinting && onBeforeGetContentResolve.current) {
+      onBeforeGetContentResolve.current();
+    }
+  }, [isPrinting]);
```
ibacher, line 30: It would be best to move the DOM manipulation into here. Something like:

```ts
  const styleTag = useRef<HTMLStyleElement | null>(null);

  useEffect(() => {
    if (isPrinting) {
      const printStyles = `@media print { @page { size: ${printIdentifierStickerSize}; } }`;

      const style = document.createElement('style');
      style.appendChild(document.createTextNode(printStyles));

      styleTag.current = document.head.appendChild(style);

      onBeforeGetContentResolve.current?.();
    } else {
      if (styleTag.current !== null) {
        document.head.removeChild(styleTag.current);
        styleTag.current = null;
      }
    }
  }, [isPrinting, printIdentifierStickerSize]);
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,198 @@
...
+const PrintIdentifierSticker: React.FC<PrintIdentifierStickerProps> = ({ patient, closeModal }) => {
+  const { t } = useTranslation();
+  const [isPrinting, setIsPrinting] = useState(false);
+  const contentToPrintRef = useRef(null);
+  const [printRetryCount, setprintRetryCount] = useState(0);
+  const { printIdentifierStickerFields, printIdentifierStickerSize, excludePatientIdentifierCodeTypes } = useConfig();
+
+  const headerTitle = t('patientIdentifierSticker', 'Patient Identifier Sticker');
+
+  const onBeforeGetContentResolve = useRef(null);
```
ibacher, line 24:
```suggestion
  const onBeforeGetContentResolve = useRef<() => void | null>(null);
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+  useEffect(() => {
+    if (isPrinting && onBeforeGetContentResolve.current) {
+      onBeforeGetContentResolve.current();
+    }
+  }, [isPrinting]);
+
+  const patientDetails = useMemo(() => {
```
ibacher, line 32: Patient details should start with:

```ts
    if (!patient) {
      return {};
    }
```

Which allows you to get rid of all of the `patient?` bits.

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+    return {
+      id: patient?.id,
+      photo: patient?.photo,
+      name: patient ? displayName(patient) : '',
+      dateOfBirth: patient.birthDate,
+      age: age(patient?.birthDate),
+      gender: getGender(patient?.gender),
+      address: patient?.address,
+      identifiers: identifiers?.length ? identifiers.map(({ value, type }) => ({ value, type })) : [],
```
ibacher, line 61: `identifiers` here _always_ has a value. I don't really understand the point of the `map()` call?

```suggestion
      identifiers: identifiers.length > 0 ? identifiers.map(({ value, type }) => ({ value, type })) : [],
```

ibacher: Maybe you wanted:

```ts
identifiers: [...identifiers],
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,198 @@
...
+interface PrintIdentifierStickerProps {
+  patient: fhir.Patient;
+  closeModal: () => void;
+}
+
+const PrintIdentifierSticker: React.FC<PrintIdentifierStickerProps> = ({ patient, closeModal }) => {
+  const { t } = useTranslation();
+  const [isPrinting, setIsPrinting] = useState(false);
+  const contentToPrintRef = useRef(null);
+  const [printRetryCount, setprintRetryCount] = useState(0);
```
ibacher, line 19:
```suggestion
  const [printRetryCount, setPrintRetryCount] = useState(0);
```

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker.component.tsx
```diff
@@ -0,0 +1,198 @@
...
+    onPrintError: (errorLocation, error) => {
+      new Promise<void>((resolve) => {
+        if (error) {
+          if (printRetryCount < 2) {
+            setprintRetryCount(printRetryCount + 1);
+            handleBeforeGetContent().then(() => {
+              handlePrint();
+            });
+          } else {
+            showSnackbar({
```
ibacher, line 99: Seems like the promise should reject on failure?

ibacher:
@denniskigen I think I was wrong about that...

denniskigen:
About?

ibacher:
Sorry responding to this:

> Could you clarify this a bit more, @ibacher?

ibacher:
Something like that. The ReactToPrint library works by copying the DOM of the passed-in `ref` into an IFrame and printing that. I'm sort of wondering whether between that and Carbon's tendency to sometimes clone nodes, especially when using some composed bits is resulting in ReactToPrint looking at a `ref` that has no content.

denniskigen:
@jnsereko I've squashed your commits into a single commit to ensure a cleaner commit history. This should simplify future rebasing and merges.

Thank you for your understanding and contributions!

denniskigen:
@ibacher FWIW, this is the actual error (from logging inside the `onPrintError` callback):

CleanShot 2024-06-14 at 4  12 05@2x

jnsereko:
> It's happening in the `onBeforeGetContent` callback (per the `errorLocation` argument).

exactly @denniskigen

ibacher:
So here's my best guess of what's going on:

* The patient chart runs React 18.3.1
* ReactToPrint relies on the deprecated [findDOMNode](https://react.dev/reference/react-dom/findDOMNode) function
* Somewhere along the way `findDOMNode()` tries to warn about this deprecation or something similar, but only in development mode
* Because the tree being used by ReactToPrint when it calls `findDOMNode()` is incomplete, an error occurs while generating the deprecation message / warning / etc.
* Built in production mode, which disables the React dev warnings, the print function works as expected

denniskigen:
Thanks for the deep dive, @ibacher. Is the takeaway then that merging this in will mean it will work as expected in production? If so, then let's focus on getting it in.

ibacher:
Yeah, I think it's very likely to work as expected in production. We just may need to have a medium-term strategy to either update or replace ReactToPrint too.

packages/esm-patient-banner-app/src/banner/patient-banner.component.tsx
denniskigen: Do you agree with this suggestion, @jnsereko?

ibacher: Just channeling [Paul's comment here](https://openmrs.slack.com/archives/CKS32D55G/p1716539108400769?thread_ts=1716473583.247999&cid=CKS32D55G).

ibacher:
Anyway we can set the print rotation to default to landscape? I think that will likely work better. For future work, we probably want to allow implementations to specify custom CSS that could be used to style the identifier badge to meet local requirements, but LGTM.

ibacher (approved):

packages/esm-patient-banner-app/src/banner-tags/print-identifier-sticker-modal.component.tsx
```diff
@@ -0,0 +1,168 @@
...
+const PrintComponent = ({ patientDetails, printIdentifierStickerFields, printIdentifierStickerSize, printRef, t }) => {
+  return (
+    <div className={styles.stickerContainer} ref={printRef}>
+      <style type="text/css" media="print">
+        {`
+          @page {
+            size: ${printIdentifierStickerSize};
+          }
+        `}
+      </style>
```
denniskigen, line 144: @ibacher it's trivial to add a config property that governs which orientation to default to and then it would leverage [this approach](https://github.com/MatthewHerbst/react-to-print#set-the-page-orientation).

denniskigen: Scratch that, it appears that it's not so straightforward to change the orientation for custom sizes (like the default `4in 6in` value). For standard [page sizes](https://developer.mozilla.org/en-US/docs/Web/CSS/%40page/size#page-size), something like this works well:

```tsx
const  { printPageOrientation } = useConfig();

<style type="text/css" media="print">
  {`
    @page {
      size: A3 ${printPageOrientation};
    }
  `}
</style>
```

denniskigen: It looks like this is something else worth looking into when evaluating whether to move away from this library. I'm proceeding to merge to unblock Madiro for now.

denniskigen:
FYI https://github.com/MatthewHerbst/react-to-print/issues/706#issuecomment-2169838777

MatthewHerbst:
`react-to-print` maintainer here. Cool to see the tool being used by y'all! Will be getting React 19 support, as well as general modernization (ESM module support for example), figured out in the next few days

PR #1727, opened 2024-03-10
(feat) O3-2117: Save orders on a new encounter if there is no encounter

usamaidrsk:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspiration.
- [ ] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://om.rs/o3ui).
- [ ] My work includes tests or is validated by existing tests.

## Summary

This PR adds the functionality to save orders along with creating a new encounter if one didn't exist. Before two transactions existed, i.e one to create encounter first and then create orders.

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-chart/assets/53287480/a68fbd48-a32b-4e0d-977c-94b7e53161a1

## Related Issue

[O3-2117](https://openmrs.atlassian.net/browse/O3-2117)

## Other

[O3-2117]: https://openmrs.atlassian.net/browse/O3-2117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1855, opened 2024-05-30
Use local esm-patient-common-lib in E2E tests

jayasanka-sack:
## Requirements

### DIFF
--- a/e2e/support/github/run-e2e-docker-env.sh
+++ b/e2e/support/github/run-e2e-docker-env.sh
@@ -5,11 +5,13 @@ script_dir=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 # create a temporary working directory
 working_dir=$(mktemp -d "${TMPDIR:-/tmp/}openmrs-e2e-frontends.XXXXXXXXXX")
 # get a list of all the apps in this workspace
-apps=$(yarn workspaces list --json | jq -r 'if ((.location == ".") or (.location | test("form-engine-app")) or (.location | test("-app") | not)) then halt else .name end')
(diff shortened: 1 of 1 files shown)


jayasanka-sack:
@brandones @ibacher @denniskigen @vasharma05 Any idea on how to include `esm-patient-common-lib`? 

I'm trying to find a way to avoid situations like this: https://github.com/openmrs/openmrs-esm-patient-chart/pull/1727#issuecomment-2138824459

brandones:
I worry this problem might actually be caused by non-determinism in webpack module federation resolution. This is an old problem. The gist is that each app has its own copy of esm-patient-common-lib, and the first app to load will have its copy loaded, and all other apps will just use that copy. So if the app you're working on loads first, you're good, but if some other app loads first, then your app will try to use the old version of esm-patient-common-lib and it will break. I'm not sure whether this is causing the problem or not.

Here's [the issue](https://openmrs.atlassian.net/issues/O3-1274) I made for it a long time ago. Actually it looks like I had a good idea about how to solve it, but we never implemented it. Messing with that stuff is challenging work.

ibacher:
This might actually be easier to solve with module federation 2.0, since there's now a sane way to add plugins to the runtime, e.g., [the resolveShare](https://module-federation.io/plugin/dev/index.html#resolveshare) hook looks like it'd do the job.

denniskigen:
Does https://github.com/openmrs/openmrs-esm-patient-chart/pull/2356 solve this, albeit in a roundabout way?

ibacher:
@denniskigen I don't think so. We had that setup before I added the auto-update script. Unfortunately, the long-term solution here is somehow ensuring the patient-chart-app (if it exists) is loaded before anything else requesting the patient-common-lib. at least that way, we'd have a consistent version loaded and a straight-forward fix ("upgrade your patient-chart-app version")

denniskigen:
Ok. Closing this as stale.

PR #1872, opened 2024-06-18
(refactor) Refactor Mark patient deceased form and Mark patient alive modal

denniskigen:
## Requirements

## Summary

This PR makes the following refactors to the `Mark patient deceased form` and `Mark patient alive` modal:

### Mark patient deceased form
- Should leverage React hook form with a Zod schema for validation. Submitting the form without specifying a death date or a cause of death should bring up a validation error below the field.
- The form component should be a workspace instead of an extension in the routes.json file.
- Should trigger a full patient reload (or another revalidation mechanism for the patient object) upon successful form submission.
- Should have better loading and empty states for the cause-of-death concepts section of the form.

### Mark patient alive modal
- The modal should be defined using the modal system instead of the extension system.
- Should trigger a full page reload (or another revalidation mechanism for the patient object) upon successfully marking the patient as alive.

## Screenshots
https://github.com/openmrs/openmrs-esm-patient-chart/assets/8509731/d88ddbb5-1823-4956-8690-c622f0c97d03

## Related Issue

https://openmrs.atlassian.net/browse/O3-3452


packages/esm-patient-chart-app/src/mark-patient-deceased/deceased-form.workspace.tsx
```diff
@@ -0,0 +1,240 @@
...
+          });
+        }
+        setIsSubmitting(false);
+        return;
+      }
+
+      markPatientDeceased(deathDate, patientUuid, causeOfDeath)
+        .then(() => {
+          closeWorkspace();
+          window.location.reload();
```
denniskigen, line 102: @ibacher, @brandones is there a better revalidation mechanism for the patient object? [fetchCurrentPatient](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-api/src/shared-api-objects/current-patient.ts#L22) doesn't use SWR so I assume calling mutate with that key won't work. Triggering a full page reload here ensures that the patient's deceased status gets reflected in the Patient Header.

packages/esm-patient-chart-app/src/mark-patient-deceased/deceased-form.workspace.tsx
```diff
@@ -0,0 +1,240 @@
...
+    <Form className={styles.form} onSubmit={handleSubmit(onSubmit, onError)}>
+      <div>
+        {isTablet && (
+          <Row className={styles.headerGridRow}>
+            <ExtensionSlot className={styles.dataGridRow} name="visit-form-header-slot" state={memoizedPatientUuid} />
+          </Row>
+        )}
+        <div className={styles.container}>
+          <span className={styles.warningContainer}>
+            <WarningFilled aria-label={t('warning', 'Warning')} className={styles.warningIcon} size={20} />
```
denniskigen, line 131: @ibacher should we switch to the new icons in the framework across the board?

packages/esm-patient-chart-app/src/mark-patient-deceased/deceased-form.workspace.tsx
```diff
@@ -0,0 +1,238 @@
...
+              {isLoadingCausesOfDeath ? <StructuredListSkeleton /> : null}
+              {causesOfDeath?.length > 10 ? (
+                <ResponsiveWrapper>
+                  <Search
+                    onChange={(event) => setSearchTerm(event.target.value)}
+                    placeholder={t('searchForCauseOfDeath', 'Search for a cause of death')}
+                    labelText=""
+                  />
+                </ResponsiveWrapper>
+              ) : null}
```
denniskigen, line 188: This change is very arbitrary. On dev3, we currently have 6 cause-of-death concept answers configured. In practice, would we expect implementations to configure a lot more through this form? Or would most implementers elect to use a custom form for Marking patients as deceased? At AMPATH, we used a custom clinical form for this purpose, so I imagine most will be doing something similar. Should I default to showing the Search input anyway, regardless of the number of options in the list?

packages/esm-patient-chart-app/src/deceased/confirmation-dialog.component.tsx
denniskigen, line 1: This component rendered a confirmation modal after clicking the `Save and close` button in the Mark patient deceased form:

CleanShot 2024-06-19 at 10  17 47@2x

I deleted it because I felt that the intent of submitting the form meant that the user must have been sure they wanted to mark a patient deceased. Do you agree? Should we explicitly require the user to confirm that they want to mark a patient deceased after submitting the form?

denniskigen:
CleanShot 2024-06-19 at 10  18 22@2x

Would it be useful to show the cause-of-death here?

packages/esm-patient-chart-app/src/mark-patient-deceased/deceased-form.workspace.tsx
denniskigen: Also, I've elected to not show a success notification here as it wouldn't stay on the screen long enough before the page reloads. Should I show it anyway and maybe set an artificial delay to communicate to the user explicitly that marking the patient as deceased succeeded?

ibacher: Currently not, but that's horrible. We should probably update `usePatient()` to return something similar to the `mutate()` hook (i.e., to force it to reload).

packages/esm-patient-chart-app/src/mark-patient-deceased/deceased-form.workspace.tsx
ibacher: As much as possible! I haven't created a component for every SVG yet and there are some gaps (warning--filled is one of the icons not included in CarbonMRS and I'm not entirely certain we can just re-use Carbon icons, but it's something I'll look into).

packages/esm-patient-chart-app/src/mark-patient-deceased/deceased-form.workspace.tsx
ibacher: We probably should just show the search input. Yeah, I don't necessarily expect implementations to use this feature as there is very likely to be other perimortem information they need to capture.

packages/esm-patient-chart-app/src/deceased/confirmation-dialog.component.tsx
ibacher: IIRC, this modal was originally added when there was no form. It was just a "Mark patient as deceased" link. As long as we have the form, I don't see any issue.

ibacher:
> Would it be useful to show the cause-of-death here?

In the tooltip? I like the idea.

denniskigen:
CleanShot 2024-06-21 at 10  21 42@2x

ibacher (approved):

denniskigen:
Thanks, @ibacher!

PR #1885, opened 2024-06-28
(test) Fix flaky drug and lab E2E tests

jayasanka-sack:
## Requirements

## Summary

This PR addresses the flakiness in the following two E2E tests:

### Lab Results:

Issue: The test compared values from table rows instead of individual cells. This caused random test failures based on rendering differences and could produce false positives as rows include reference values.
Fix: Updated the test to compare values within specific cells, ensuring accurate and consistent validation.

### Drug Orders:

Issue: The test asserted values to be present anywhere on the page, leading to failures due to word duplication. For instance, the drug name appears twice on the modify drug order form if the user scrolls, causing the test to fail when the view is slightly scrolled.
Fix: Modified the test to assert values within specific sub-elements, preventing false failures due to duplicate display of drug names.

This update ensures more reliable and accurate E2E tests by refining the assertions to target precise elements.

### DIFF
--- a/e2e/specs/drug-orders.spec.ts
+++ b/e2e/specs/drug-orders.spec.ts
@@ -14,6 +14,8 @@ test.beforeEach(async ({ api }) => {
 
 test('Record, edit and discontinue a drug order', async ({ page }) => {
   const medicationsPage = new MedicationsPage(page);
+  const form = page.locator('#drugOrderForm');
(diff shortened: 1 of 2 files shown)


denniskigen (approved):
Excellent. Thanks, @jayasanka-sack!

PR #1895, opened 2024-07-04
(refactor) O3-3522: Replace module-local translations with core translations

jwnasambu:
## Requirements

## Summary

I replaced the t() calls with getCoreTranslation() calls knowing  that getCoreTranslation behaves similarly to t() and returns the translated text.
## Related Issue
https://openmrs.atlassian.net/browse/O3-3522


jwnasambu:
@jnsereko, @denniskigen, @ibacher, could you please help me review my PR please? Besides, I am encountering this error: Argument of type '"patientIdentifierSticker"' is not assignable to parameter of type '"address" | "male" | "female" and more additional strings that are not present in the [esm-translations file](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-translations/src/translations.ts) that I believe I should add Kindly feel free to correct me if necessary. Thank you!

ibacher:
Short version is not _everything_ goes into Core Translations, only strings that we use in several places. `patientIdentifierSticker` should be translated using the `t()` function as usual.

jwnasambu:
@ibacher thanks for the review and guidance

jwnasambu:
> Short version is not _everything_ goes into Core Translations, only strings that we use in several places. `patientIdentifierSticker` should be translated using the `t()` function as usual.

@ibacher I have fixed the proposed changes.

jwnasambu:
> Short version is not _everything_ goes into Core Translations, only strings that we use in several places. `patientIdentifierSticker` should be translated using the `t()` function as usual.

@ibacher, would it be okay for me to close this PR? I have already fixed the frequently used strings in the core translations. The specific or less frequently used strings are to be translated directly using the t() function in the code where they are used, based on your comment above which makes me believe these changes are not needed? kindly correct me please!

ibacher:
@jwnasambu We merged in the PR with changes. I think having standardized translations for printing, etc. makes sense. However, in this PR you will need to update the versions of the CLI and framework as outlined [here](https://o3-docs.openmrs.org/docs/frontend-modules/development.en-US#i-see-errors-about-missing-apis-when-i-run-the-app) for this code to be able to "see" those changes.

jwnasambu:
@ibacher Kindly I have tried out something here and I will be glad for more guidance and review.

PR #1911, opened 2024-07-12
(feat) O3-3608: Adjust step precision for the dose field in the DrugOrder form

its-kios09:
## Requirements

## Screenshots

medication

## Related Issue

https://openmrs.atlassian.net/browse/O3-3608


makombe:
@its-kios09 can you please ticket this in the OpenMRs jira. Thanks

mogoodrich (approved):

brandones:
@its-kios09 echoing @makombe , please file a Jira ticket for record-keeping. In the ticket, please name the organization that is requesting this.

My only concern about the implementation is that this will make the up and down arrows increment the dose by `0.1`, which is strange. Unfortunately this is the only way to support this with the current carbon react. There is a [bug from 2020](https://github.com/carbon-design-system/carbon/issues/6930) which would allow fixing this by changing the step to `step="any"`. Looks like an easy ticket for anyone with a dev environment set up for Carbon.

its-kios09:
@brandones @makombe sharing in a few

brandones:
Thank you @its-kios09 . Please fix CI.

its-kios09:
> Thank you @its-kios09 . Please fix CI.

@brandones it is working now

denniskigen:
Strictly speaking, the [NumberInput](https://react.carbondesignsystem.com/?path=/docs/components-numberinput--overview) component supports decimal values out of the box. Without a defined `step` attribute, the default increment/decrement buttons (not enabled here because of the `hideSteppers` attribute) typically adjust the value by 1. However, this doesn't prevent users from manually entering decimal values. Setting the `step` to 0.1 makes it so that: 

- When the input is focused, using the up and down arrow keys will increase or decrease the value by 0.1.
- Users can still type in any value, including those that don't align with the `step` (e.g. 0.05).
- The input will be considered valid for multiples of 0.1 from the min value.

brandones:
@denniskigen based on the video in the ticket, it would appear that the `NumberInput` does not accept decimal values when the `step` prop is not set.

denniskigen:
Could you try in this demo example here https://carbondesignsystem.com/components/number-input/usage/

ibacher:
It's not a validation on the level of Carbon, but one implemented by the browser for `<input type="number" />` with a `step` attribute (see [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#step)). Unfortunately, Carbon [both decides `step` must be a number](https://github.com/carbon-design-system/carbon/blob/cc284f3cb6f7c289c95cf56dd602e67e6044a796/packages/react/src/components/NumberInput/NumberInput.tsx#L177) and [defaults it to `1` if unset](https://github.com/carbon-design-system/carbon/blob/cc284f3cb6f7c289c95cf56dd602e67e6044a796/packages/react/src/components/NumberInput/NumberInput.tsx#L225), which is useful for the custom steppers, but does mean that by default Carbon's `NumberInput`s don't actually allow you to _submit_ a form with decimal values.

denniskigen:
Related: https://github.com/openmrs/openmrs-esm-patient-chart/commit/7f84c0364dddbc52a1e78d19718b49eeb5e42e9e

PR #1920, opened 2024-07-16
(fix) Abatement date for a condition should be later than the onset date

denniskigen:
## Requirements

## Summary

This PR fixes a regression in the Conditions form where a user could be able to select an abatement date for a condition that is earlier than the onset date. This happened because the `minDate` validation logic was referencing the wrong target input. Setting this back to the onset date gets the validation working as expected.

## Screenshots

https://github.com/user-attachments/assets/3fd642c6-2eec-4743-93ef-236c4df524e9


brandones (approved):
LGTM

PR #1923, opened 2024-07-17
Revert "(feat) O3-2953: Add the ability to edit and delete vitals and biometrics"

denniskigen:
## Requirements

## Summary

Reverts #1812

The current approach to editing vitals and biometrics creates new observations under the same encounter instead of updating existing ones. This results in many duplicate, non-voided observations.

When updating observations, we need a way to semantically indicate that the new observation is an updated version of an existing one. Each new observation in the payload should have the same UUID as the existing observation. This will ensure that we update observations in place instead of creating new ones.

This feature will be reintroduced after the 3.1.0 release.


ibacher (approved):
As discussed, unfortunately, this is necessary.

PR #1932, opened 2024-07-22
O3-3633 - Reusing the  visitable renderer for encounter and its obs without defining the components again

its-kios09:
## Requirements

## Summary

https://openmrs.atlassian.net/browse/O3-3633

## Screenshots

job

### DIFF
--- a/packages/esm-patient-chart-app/src/index.ts
+++ b/packages/esm-patient-chart-app/src/index.ts
@@ -30,6 +30,7 @@ import startVisitActionButtonOnPatientSearch from './visit/start-visit-button.co
 import startVisitFormComponent from './visit/visit-form/visit-form.component';
 import stopVisitActionButtonComponent from './actions-buttons/stop-visit.component';
 import visitAttributeTagsComponent from './patient-banner-tags/visit-attribute-tags.component';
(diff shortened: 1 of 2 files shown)


ojwanganto:
Thanks, @its-kios09. 
Tagging @ibacher @denniskigen @brandones for their reviews

ibacher:
Exporting something from an app and importing it into another app is always a programming error, because it couples two apps. If you want to reuse the component, extract it into the common library and use it from there.

brandones:
It's really hard to tell what you're actually trying to accomplish, since all that's been written seems to be suggestions about how it should be implemented. Which anyway, as @ibacher indicated, isn't the right solution.

I would suggest adding this table to [esm-generic-patient-widgets-app](https://github.com/openmrs/openmrs-esm-patient-chart/tree/main/packages/esm-generic-patient-widgets-app). Then it can be added to dashboards as an extension and configured as needed. For reference, check out the existing extension, [`obs-by-encounter-widget`](https://github.com/openmrs/openmrs-esm-patient-chart/blob/8f117b157710f4eee51c6684fdb28119d42a49d9/packages/esm-generic-patient-widgets-app/src/routes.json), and the details of its implementation.

ojwanganto:
The guidance should be clear. Thanks, @ibacher @brandones.
@its-kios09 please take note.

PR #1933, opened 2024-07-22
(test) Remove unnecessary partial mocks; improve typing in tests

denniskigen:
## Requirements

## Summary

This PR refactors various tests in the Patient Chart to fix the issue of unnecessary [partial mocks](https://jestjs.io/docs/mock-functions#mocking-partials). The [mock.tsx](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-framework/mock.tsx) manual mock in Core aggregates stubs for most functions in Core. We should only extend the manual mock if we want to include stubs for things that exist in Core, but are not included in the manual mock. Otherwise, we should extend the stubs already provided by the manual mock. This PR enables just that.

## Other

Documentation updates will follow shortly.

### DIFF
--- a/packages/esm-patient-chart-app/src/config-schema.ts
+++ b/packages/esm-patient-chart-app/src/config-schema.ts
@@ -1,42 +1,91 @@
 import { Type } from '@openmrs/esm-framework';
 
 export const esmPatientChartSchema = {
-  freeTextFieldConceptUuid: {
-    _default: '5622AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
-    _type: Type.ConceptUuid,
-  },
-  visitDiagnosisConceptUuid: {
-    _default: '159947AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
-    _type: Type.ConceptUuid,
-  },
-  notesConceptUuids: {
-    _type: Type.Array,
-    _default: ['162169AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'],
-  },
-  offlineVisitTypeUuid: {
-    _type: Type.UUID,
-    _description: 'The UUID of the visit type to be used for the automatically created offline visits.',
-    _default: 'a22733fa-3501-4020-a520-da024eeff088',
-  },
-  showRecommendedVisitTypeTab: {
-    _type: Type.Boolean,
-    _description: 'Whether start visit form should display recommended visit type tab. Requires `visitTypeResourceUrl`',
-    _default: false,
-  },
-  visitTypeResourceUrl: {
+  defaultFacilityUrl: {
     _type: Type.String,
-    _default: '/etl-latest/etl/patient/',
-    _description: 'Custom URL to load resources required for showing recommended visit types',
+    _default: '',
+    _description: 'Custom URL to load default facility if it is not in the session',
+  },
+  disableChangingVisitLocation: {
+    _type: Type.Boolean,
+    _description:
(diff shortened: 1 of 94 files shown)


brandones (approved):
LGTM, thanks @denniskigen .

I wonder if we can make a nice pattern that ensures that typescript or the test breaks if `mockUseConfig.mockReturnValue` is set to something incompatible with the config type. Maybe like

```
    mockUseConfig.mockReturnValue({
      concepts: {
        pulseUuid: '5087AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
      },
    } as Partial<ConfigObject>);
```

?

denniskigen:
@brandones if we want strict conformance with the source function's type definitions, we can do something like:

```tsx
import type { ConfigObject } from './config-schema';
 
const mockUseConfig = jest.mocked<() => ConfigObject>(useConfig);
``

If we want to satisfy the type contract with a partial config, we could then do something like:

```tsx
beforeEach(() => {
  mockUseConfig.mockReturnValue({
    concepts: {
      pulseUuid: false,
    },
  } as ConfigObject);
});
```

This will fail type checking because the `ConfigObject` type expects `pulseUuid` to be a string. Does this seem reasonable?

brandones:
Yeah I think the latter is reasonable. I wonder if it also works well with

```tsx
beforeEach(() => {
  mockUseConfig.mockReturnValue({
    ...getDefaultsFromConfigSchema(schema),
    concepts: {
      pulseUuid: "asdfasdfasdf",
    },
  } as ConfigObject);
});
```

which IMO may be the best way to do config overrides

packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.test.tsx
```diff
@@ -38,9 +44,13 @@ jest.mock('../common', () => ({
   })),
 }));
 
+mockUseConfig.mockReturnValue({
+  ...getDefaultsFromConfigSchema(configSchema),
+  mockVitalsConfig,
```
brandones, line 49: This seems like it would create a key called `mockVitalsConfig` on the object, which I would really hope would produce a type error

denniskigen: Fixed in the latest push, I think.

denniskigen:
I'm sorry for blowing the blast radius of this PR a fair bit after review. The main motivators are:

- Replacing most occurrences of`const mockedSourceFn = sourceFn as jest.Mock` with `jest.mocked(sourceFn)`. [jest.mocked](https://jestjs.io/docs/mock-function-api#jestmockedsource-options) infers type definitions from the source function. This refactor has unearthed some issues with misaligned mock data, where the shape of the data doesn't match what the actual implementation expects. TypeScript helps enforce these contracts, so it makes sense to leverage it wherever possible.
- Specifying that mock calls, instances, contexts and results should be automatically cleared between tests by setting `clearMocks: true` in the root Jest config. More here jestjs.io/docs/configuration#clearmocks-boolean.
- Annotating `useConfig` mocks with types from the associated config schemas for improved type inference.
- Fixing some tests that were not working as intended.

packages/esm-patient-chart-app/src/config-schema.ts
```diff
@@ -1,42 +1,91 @@
 import { Type } from '@openmrs/esm-framework';
 
 export const esmPatientChartSchema = {
+  defaultFacilityUrl: {
```
brandones, line 4: Yesss thank you

brandones (approved):

PR #1946, opened 2024-07-30
(feat) O3-3683: Add aria-labels to Start and End visit buttons in the visit header

Vijaykv5:
## Requirements

## Summary

This PR adds aria-label property for the easy **accessibility** of Start a Visit and End visit button in the Navbar.

## Related Issue

[O3-3683](https://openmrs.atlassian.net/browse/O3-3683)

## Other

cc : @jayasanka-sack 

[O3-3683]: https://openmrs.atlassian.net/browse/O3-3683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx
@@ -182,12 +182,21 @@ const VisitHeader: React.FC = () => {
           <>
             <ExtensionSlot name="visit-header-right-slot" />
             {!isLoading && !currentVisit && !isDeceased && (
-              <Button className={styles.startVisitButton} onClick={launchStartVisitForm} size="lg">
+              <Button
+                className={styles.startVisitButton}
+                onClick={launchStartVisitForm}
+                size="lg"
+                aria-label={t('startAVisit', 'Start a visit')}
+              >
                 {t('startAVisit', 'Start a visit')}
               </Button>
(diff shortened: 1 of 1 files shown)


jayasanka-sack:
@denniskigen @ibacher we need to provide CSS selectors for the onboarding tutorials. It's straightforward for most elements since we can use extension IDs and other selectors. However Vijay is having a hard time querying some elements like the start visit button. What’s the best way to define selectors for them? Using aria-label isn’t ideal because it changes with the locale. Maybe we should add an ID?

ibacher:
Bah! It's as easy as `[data-extension-slot-name="visit-header-right-slot"] + button`\</s>. 

In all seriousness, it makes sense to include something that makes it easier to select the appropriate value (though the above will sort of work, though it may also select the "End Visit" button). However, I'm hesitant about defining `id`s. While these are easy to work with from an element selection standpoint, they are extremely hard to work with from a microfrontend stand-point (because we need to ensure that no other MF renders an element with the same id, since ids must be unique within a DOM tree).

Maybe we should add a `data-openmrs-role` attribute to button those buttons so the selector could be `button[data-openmrs-role="Start Visit"]`.

jayasanka-sack (approved):

PR #1954, opened 2024-08-07
(chore) Conditionally set server-token for CI workflow jobs

denniskigen:
## Requirements

## Summary

Github Action jobs that depend on remote caching require a `server-token` input. Because forks don't have access to environment variables, workflow runs from PRs made from forks are far slower. Conditionally setting a default value in the case where the `server-token` cannot be obtained is a possible solution to mitigate the slow builds. I'm not too sure though about how this affects security and whether this can be leveraged by bad actors.

### DIFF
--- a/.github/workflows/bundle-size.yml
+++ b/.github/workflows/bundle-size.yml
@@ -23,7 +23,7 @@ jobs:
         uses: felixmosh/turborepo-gh-artifacts@v3
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


denniskigen:
Per the [action author](https://github.com/felixmosh/turborepo-gh-artifacts/issues/1), `server-token` isn't particularly special, and setting it to an empty string would suffice. I'll push a commit that does exactly that.

ibacher (approved):

NethmiRodrigo (approved):
LGTM, thanks @denniskigen!

PR #1957, opened 2024-08-08
(feat) O3-3779: Should not be able to order labs with no configured price

its-kios09:
## Requirements

## Summary

see: https://openmrs.atlassian.net/browse/O3-3779

## Screenshots

fix-lab

### DIFF
--- a/packages/esm-patient-labs-app/src/lab-orders/add-lab-order/lab-order-form.component.tsx
+++ b/packages/esm-patient-labs-app/src/lab-orders/add-lab-order/lab-order-form.component.tsx
@@ -154,6 +154,7 @@ export function LabOrderForm({
   }, [isDirty, promptBeforeClosing]);
 
   const responsiveSize = isTablet ? 'lg' : 'sm';
+  const [hasPrice, setHasPrice] = useState(true);
 
   return (
     <>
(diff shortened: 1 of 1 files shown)


donaldkibet:
@its-kios09 running this PR am getting this error 
Screenshot 2024-08-27 at 13 25 58


Kindly share a video for these changes

brandones:
I do not like that this introduces billing logic explicitly in patient-chart.

Moreover, why would we want users to ever be confronted with the experience of trying to order something, being able to click it and add it to the basket, but being totally unable to actually fill out or complete the order? It is a bad user experience. The provider can do nothing about the problem. Should the order type simply not appear in the order search? And if it is a configuration error, do we log things to the console? It seems like it would be best if this were handled on the backend. CC @ibacher 

Obviously the code is no good but I think we need a broader discussion of the approach here.

ibacher:
@brandones The requirement here is less around "billing logic" and more around displaying prices to clinicians.

However, I absolutely agree that logic for loading prices for items **cannot** be in the lab, orders, or medications app. This is because we **cannot** require that O3 systems include prices for things. The logic for the linkage to billing and stock management **must** be in its own, separate app and the UI, at least, needs to be added as appropriate via extensions. Anything else is absolutely unacceptable in the community-maintained libraries and will be removed.

This specific PR appears to implement a business logic rule that cannot be part of the default setup (remember, we as a community need to support implementations with no billing logic). This implies that we have to make some design decisions about how to do this, which we need to actually design.

For the above reasons, I think we should close this PR until we have a community agreement on the approach here. The first step to building that out is a post on Talk describing the goals / requirements here.

mogoodrich:
Yeah, I haven't been following this at all, but just to second/third that we certainly can't require that the O3 system include prices for things!

PR #1961, opened 2024-08-13
(fix) O3-3715: Add phone number person attribute to patient banner config

denniskigen:
## Requirements

## Summary

Telephone numbers recorded during patient registration don't show up in the patient header. This is because the default contact person attribute types array in the Patient Banner config is set to an empty array. That's problematic because the logic that determines which attributes to show under the contact details section checks the patient's attributes against that array. This PR adds a person attribute UUID for the Telephone number attribute as a default value in the `contactAttributeTypes` array, which fixes the root issue.

### Bug

https://github.com/user-attachments/assets/4f847ddc-87e4-44f9-919a-28ceeefca732

### Fix

Telephone number in patient banner contact details section

## Related Issue
https://openmrs.atlassian.net/browse/O3-3715


ibacher (approved):
We had an issue before where I think the UUID for this wasn't stable, but it seems fine. Note, though, that if we're using the FHIR Patient object, it should already have the phone number in the `telecom` field, assuming that the FHIR2 module is configured (i.e., we should try to avoid needing to make unnecessary requests).

denniskigen:
You're absolutely right! This uses the FHIR Patient object, and the telephone number is indeed included in the `telecom` object. I'll make the necessary amendments.

denniskigen:
We should probably get this in, then @ibacher.

ibacher (approved):

PR #1963, opened 2024-08-15
(feat) O3-3799: Open HTML Form Entry Forms within Form Workspace

mogoodrich:
## Requirements


packages/esm-patient-common-lib/src/form-entry-interop.ts
```diff
@@ -24,29 +22,18 @@ export function launchFormEntryOrHtmlForms(
...
-    } else {
-      if (encounterUuid) {
-        navigate({
-          to: `\${openmrsBase}/htmlformentryui/htmlform/${htmlForm.formEditUiPage}.page?patientId=${patientUuid}&visitId=${visitUuid}&encounterId=${encounterUuid}&definitionUiResource=${htmlForm.formUiResource}&returnUrl=${window.location.href}`,
-        });
-      } else {
-        navigate({
-          to: `\${openmrsBase}/htmlformentryui/htmlform/${htmlForm.formUiPage}.page?patientId=${patientUuid}&visitId=${visitUuid}&definitionUiResource=${htmlForm.formUiResource}&returnUrl=${window.location.href}`,
-        });
-      }
```
mogoodrich, line 48: Previously, if this form had been configured as an HTML Form, we simply redirected to the O2 app,

packages/esm-patient-forms-app/src/forms/form-entry.workspace.tsx
```diff
@@ -82,7 +91,21 @@ const FormEntry: React.FC<FormEntryComponentProps> = ({
...
+      )}
+      {showForm && formInfo && patientUuid && patient && htmlForm && !encounterUuid && (
+        <HtmlFormEntryWrapper
+          closeWorkspace={closeWorkspace}
+          src={`${window.openmrsBase}/htmlformentryui/htmlform/${htmlForm.formUiPage}.page?patientId=${patientUuid}&visitId=${visitUuid}&definitionUiResource=${htmlForm.formUiResource}&returnUrl=post-message:close-workspace`}
+        />
+      )}
+      {showForm && formInfo && patientUuid && patient && !htmlForm && (
+        <ExtensionSlot name="form-widget-slot" state={state} />
+      )}
```
mogoodrich, line 108: Now, instead of navigating to the O3 app, we render a new HTML Form Entry Wrapper component, which basically just loads the O2 app in an iframe (see below).  

The way I've set up here, by just referencing the Html Form Entry Wrapper within form-entry.workspace, is likely wrong, but I wasn't sure what the best approach was.  Should I simply render a different extension slot ("html-form-entry-widget-slot") and define a new extension thta uses HtmlFormEntryWrapper?  Should there still just be one extension slot ("form-widget-slot") and what gets wired into it varies?  Or should we have an entirely different "html-form-entry.workspace".    Apologies for my naivety with best practices/uses of workspaces. :)

One issue I'm seeing right now, that may factor into this decision, is that if you shift between the form workspace and another workspace (ie the order workspace) the data you can currently entered into the form, but not saved, is *sometimes* lost.  Once we get this set up the "right" way, I can troubleshoot this some more and hopefully we can come up with a workable solution. 

Also note that we pass in a "returnUrl" of "post-message:close-workspace", more on this below.

packages/esm-patient-forms-app/src/htmlformentry/html-form-entry-wrapper.component.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { useEffect } from 'react';
+import styles from './html-form-entry-wrapper.scss';
+
+interface HtmlFormEntryWrapperProps {
+  closeWorkspace: () => void;
+  src: string;
+}
+
+const HtmlFormEntryWrapper: React.FC<HtmlFormEntryWrapperProps> = ({ closeWorkspace, src }) => {
```
mogoodrich, line 9: So basically all this does is load the O2 app into an iframe and hides the headers and the breadcrumbs

packages/esm-patient-forms-app/src/htmlformentry/html-form-entry-wrapper.component.tsx
```diff
@@ -0,0 +1,36 @@
...
+      dashboard.querySelector('.patient-header')?.remove();
+      dashboard.querySelector('#breadcrumbs')?.remove();
+      iframe.style.display = 'block';
+    });
+
+    // register a listener to close the workspace when HFE sends a "close-workspace" message
+    window.addEventListener('message', (event) => {
+      if (event.data === 'close-workspace') {
+        closeWorkspace();
+      }
```
mogoodrich, line 25: I've made a small change to HTML Form Entry UI so that the if the "returnUrl" message is in the format: 'post-message:some-message' it posts a message "some-message" instead of attempting to redirect to "returnUrl'

Also: I see there is closeWorkspace and a closeWorkspaceWithSavedChanges.  Is there an explanation of the different between these two (and which one should I be calling here)

packages/esm-patient-forms-app/src/htmlformentry/html-form-entry-wrapper.scss
```diff
@@ -0,0 +1,8 @@
+.wrapper {
+  width: 95%;
+  border: none;
+  margin: 0;
+  padding: 0;
+  height: 3000px; // TODO: how to make this dynamic?
```
mogoodrich, line 6: My limited CSS skills were failing here... :) I basically want the iframe height to be equal to the height of the content it is displaying.

mogoodrich (commented):
Definitely not ready for review, but I'm looking for initial feedback on the best way to go about this.

Screencast of how this looks:

[Screencast from 08-14-2024 03:35:54 PM.webm](https://github.com/user-attachments/assets/3e9bcca4-637e-4e95-b26d-4b6c60748587)

Also link to ticket: https://openmrs.atlassian.net/browse/O3-3799

mogoodrich:
Related HFE-UI PR to support closing the workspace when saving/cancelling:  
https://github.com/openmrs/openmrs-module-htmlformentryui/pull/63

packages/esm-patient-forms-app/src/forms/form-entry.workspace.tsx
ibacher: I think the solution to the "losing form data" _may_ be to have a unique workspace for this? Not sure, but I think that's the case.

mseaton (commented):
Generally looks good to me @mogoodrich - nice!  Defer to others on the semantic questions you are asking to get this done.

packages/esm-patient-forms-app/src/htmlformentry/html-form-entry-wrapper.component.tsx
```diff
@@ -0,0 +1,36 @@
...
+import styles from './html-form-entry-wrapper.scss';
+
+interface HtmlFormEntryWrapperProps {
+  closeWorkspace: () => void;
+  src: string;
+}
+
+const HtmlFormEntryWrapper: React.FC<HtmlFormEntryWrapperProps> = ({ closeWorkspace, src }) => {
+  useEffect(() => {
+    const iframe = document.querySelector('iframe');
```
ibacher, line 11: So the React way to do this is to create a ref. Basically something like:

```ts
const iframeRef = useRef<HTMLIFrameElement>();

// then in use effect
useEffect(() => {
  const iframe = iframeRef.current;
 // rest of the stuff goes here.
```

packages/esm-patient-forms-app/src/htmlformentry/html-form-entry-wrapper.component.tsx
```diff
@@ -0,0 +1,36 @@
...
+    iframe.addEventListener('load', () => {
```
ibacher, line 12: Rather than using `addEventListener`, etc. it's probably better to add the function to the `iframe`'s `onLoad` propery. That way the event handler is correctly cleaned up when the iframe is removed from the DOM.

```ts
onLoad={() => {
  const dashboard = iframe.current.contentDocument;
 // etc.
}}
```

packages/esm-patient-forms-app/src/htmlformentry/html-form-entry-wrapper.component.tsx
```diff
@@ -0,0 +1,36 @@
...
+      // hide the header and breadcrumbs
+      const dashboard = iframe.contentDocument;
+      dashboard.querySelector('header')?.remove();
+      dashboard.querySelector('.patient-header')?.remove();
+      dashboard.querySelector('#breadcrumbs')?.remove();
+      iframe.style.display = 'block';
+    });
+
+    // register a listener to close the workspace when HFE sends a "close-workspace" message
+    window.addEventListener('message', (event) => {
```
ibacher, line 22: Calling `addEventListener()` without `removeEventListener()` leaves the event listener around. This is usually not to be a deal in a multi-page application, since each page load will remove the listeners, but in a SPA it's essential to clean things up. I'd probably use a separate `useEffect()` hook for this like this:

```ts
useEffect(() => {
	const callback = (event) => event?.data === 'close-workspace' && closeWorkspace();
  window.addEventListener('message', callback);
  // functions returned from `useEffect()` get called when the component containing the hook is unmounted
  return () => window.removeEventListener('message', callback);
}, [closeWorkspace])
```

packages/esm-patient-forms-app/src/htmlformentry/html-form-entry-wrapper.component.tsx
ibacher: The difference is basically that `closeWorkspace()` will notify the user if they close the workspace with unsaved changes whereas `closeWorkspaceWithSavedChanges()` doesn't check if there are unsaved changes.

packages/esm-patient-forms-app/src/htmlformentry/html-form-entry-wrapper.scss
ibacher: I actually think this is quite hard! _Most_ DOM elements have some way of determining the height of their contents so you could just use `fit-content`, but iframes are special, because they can load cross-origin content. The best I can think of is adding to the `onLoad` function so it becomes something like:

```ts
const onLoad = useCallback(() => {
  const iframe = iframeRef.current;
  const dashboard = iframe.contentDocument;
  dashboard.querySelector('header')?.remove();
  dashboard.querySelector('.patient-header')?.remove();
  dashboard.querySelector('#breadcrumbs')?.remove();

  iframe.style.display = 'block';
  iframe.height = iframe.contentWindow.document.body.scrollHeight;
}, []);
```

ibacher (commented):
Wow! This is _way_ easier than I thought it would be!

mogoodrich:
> Wow! This is _way_ easier than I thought it would be!

Agreed @ibacher ... we will see, I'm still worried about jinxing ourselves and missing a dealbreaker issue :)

Thanks for the helpful comments... I'll let others weigh in and then test them out.

packages/esm-patient-forms-app/src/forms/form-entry.workspace.tsx
```diff
@@ -82,7 +91,21 @@ const FormEntry: React.FC<FormEntryComponentProps> = ({
 
   return (
     <div>
-      {showForm && formInfo && patientUuid && patient && <ExtensionSlot name="form-widget-slot" state={state} />}
+      {showForm && formInfo && patientUuid && patient && htmlForm && encounterUuid && (
```
chibongho, line 94: Maybe simplify to something like this?

```
    const showFormAndLoadedData = showForm && formInfo && patientUuid && patient;
    return (
      <div>
        {showFormAndLoadedData && (
           htmlForm? 
             <ExtensionSlot name="form-widget-slot" state={state} /> :
             <HtmlFormEntryWrapper src={encounterUuid? url1 : url2} />
        )}
       </div>
    )
```

mogoodrich:
No hurry, @denniskigen  but would be great to have your thoughts on this as well before I start  making the suggested changes.

packages/esm-patient-forms-app/src/htmlformentry/html-form-entry-wrapper.scss
mogoodrich: And this worked too, once I futzed around a bit and added a little padding... I will fiddle with this a little more....

packages/esm-patient-forms-app/src/forms/form-entry.workspace.tsx
mogoodrich: Yeah, let me experiment with doing this in the it's own workspace.

packages/esm-patient-forms-app/src/htmlformentry/html-form-entry-wrapper.component.tsx
```diff
@@ -0,0 +1,35 @@
...
+  }, []);
+
+  return (
+    <div>
+      <iframe ref={iframeRef} src={src} className={styles.wrapper} onLoad={onLoad}></iframe>
+    </div>
+  );
+};
+
+export default HtmlFormEntryWrapper;
```
mogoodrich, line 35: Thanks for all the tips @ibacher ... the scrollHeight trick generally worked, though I needed to add some padding.

mogoodrich:
Ok, I think this is getting close!  Updated screencast here:

[htnmlformO3.webm](https://github.com/user-attachments/assets/b23f7db8-40b1-48f3-a6aa-a4f7437a1552)

Some notes, thoughts:

* After recording the video I changed the workspace size from "wider" to "extra-wide"
* I can/should add some basic tests if we like this approach
* I tried to fiddle around with holding HTML Form state when switching between workspaces, but ultimately decided it was best/safest to make the HTML Form workspace one that you can't hide.  You'll note, however, that it gives the potentially incorrect warning dialog ("you have unsaved changes...") because unfortunately there's no way for O3 to track the changes in the O2 form.  I'm thinking we can come with a custom dialog that says something like "Please click the Cancel or Save button to close the "xxxx Form" workspace before clicking on another workspace. Thoughts?
* When editing a form (from the "Visits" view) you will see in the video that the Visit state isn't mutated.  Do we know if this works correctly for non-HTML Forms?

Anyway ready for re-review @ibacher  ... and I'd love your thoughts @denniskigen  @brandones

ibacher:
>  I'm thinking we can come with a custom dialog that says something like "Please click the Cancel or Save button to close the "xxxx Form" workspace before clicking on another workspace. Thoughts?

It's a little bit unfortunate that there doesn't seem to be a way to customize the messages here.

ibacher (approved):
LGTM!

mogoodrich:
> I don't think it does. I could be wrong about that, but the Visits view is one of the most expensive to render.

Great, I was hoping that fell under the category of "pre-existing feature" and not something I had to worry about for this PR... :)

denniskigen:
This is looking pretty clutch. Thanks, @mogoodrich. Might it be useful to have an error state to show if the form can't load for whatever reason? Or does HFE have an error state that can render correctly within the iframe?

denniskigen (approved):

mogoodrich:
> This is looking pretty clutch. Thanks, @mogoodrich. Might it be useful to have an error state to show if the form can't load for whatever reason? Or does HFE have an error state that can render correctly within the iframe?

It stack traces within the iframe... :)  Not the greatest behavior, but not different from what happens in O2 now.

mogoodrich:
Merging this in as the E2E failure is *not* related to this change

PR #1969, opened 2024-08-22
(feat) O3-3773 Support for Patient Program Summary in Patient Chart

kajambiya:
## Requirements

## Summary
This PR introduces summary program summary  to patient chart.

## Screenshots
Screenshot 2024-09-06 at 10 03 33


packages/esm-patient-chart-app/src/encounter-tile/utils/helpers.ts
```diff
@@ -0,0 +1,146 @@
+import dayjs from 'dayjs';
+import { formatDate, parseDate } from '@openmrs/esm-framework';
+
+export function getEncounterValues(encounter, param: string, isDate?: Boolean) {
+  if (isDate) return dayjs(encounter[param]).format('DD-MMM-YYYY');
+  else return encounter[param] ? encounter[param] : '--';
+}
+
+export function formatDateTime(dateString: string): any {
```
hadijahkyampeire, line 9: Let's use the already existing Date formatters from the framework

packages/esm-patient-chart-app/src/encounter-tile/utils/helpers.ts
```diff
@@ -0,0 +1,146 @@
...
+  const format = 'YYYY-MM-DDTHH:mm:ss';
+  if (dateString.includes('.')) {
+    dateString = dateString.split('.')[0];
+  }
+  return dayjs(dateString, format, true).toDate();
+}
+
+export function obsArrayDateComparator(left, right) {
+  return formatDateTime(right.obsDatetime) - formatDateTime(left.obsDatetime);
```
hadijahkyampeire, line 18: Same here, we can leverage the framework functions, check on my PR, I got a way around this

hadijahkyampeire (commented):
Mostly looks good, thanks @kajambiya , just a few comments

packages/esm-patient-chart-app/src/index.ts
```diff
@@ -112,6 +113,11 @@ export const stopVisitPatientSearchActionButton = getSyncLifecycle(stopVisitActi
   moduleName,
 });
 
+export const patientProgramSummary = getSyncLifecycle(PatientSummaryOverviewList, {
```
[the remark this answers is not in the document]
CynthiaKamau: The data is related to a program but some people may classify it differently.

packages/esm-patient-chart-app/src/encounter-tile/utils/helpers.ts
```diff
@@ -0,0 +1,146 @@
...
+    if (typeof obs.value === 'object' && obs.value?.names) {
+      return formatDate(parseDate(obs.obsDatetime), { mode: 'wide' });
+    } else {
+      return formatDate(parseDate(obs.value), { mode: 'wide' });
+    }
+  }
+
+  if (typeof obs.value === 'object' && obs.value?.names) {
+    return (
+      obs.value?.names?.find((conceptName) => conceptName.conceptNameType === 'SHORT')?.name || obs.value.name.name
```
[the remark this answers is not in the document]
CynthiaKamau: do you mean `concept.display` or `concept.preferredName` ?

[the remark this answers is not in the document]
ibacher: `display` is always the preferred name in the user's locale.

packages/esm-patient-chart-app/src/index.ts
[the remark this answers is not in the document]
CynthiaKamau: It is not program specific, I think the name might be misleading , it’s more of a clinical view summary. I have updated the name

packages/esm-patient-chart-app/src/index.ts
```diff
@@ -112,6 +113,11 @@ export const stopVisitPatientSearchActionButton = getSyncLifecycle(stopVisitActi
   moduleName,
 });
 
+export const clinicalViewsSummary = getSyncLifecycle(ClinicalViewsSummary, {
```
ibacher, line 116: Please use `getAsyncLifecycle()` here.

packages/esm-patient-chart-app/src/index.ts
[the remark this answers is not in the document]
ibacher: Part of my point is that this is a public interface intended to be consumed by multiple configurations written by implementations that may or may not understand what the original purpose of a widget is; thus naming things so that they are consistent, discoverable, and not confusing is of vital importance here.

packages/esm-patient-chart-app/src/clinical-views/components/clinical-views-summary.component.tsx
```diff
@@ -0,0 +1,46 @@
...
+  columns: Array<EncounterTileColumn>;
+}
+
+const ClinicalViewsSummary: React.FC<OverviewListProps> = ({ patientUuid }) => {
+  const config = useConfig();
+  const tilesDefinitions = config.tilesDefinitions;
+
+  const tilesData = useMemo(
+    () =>
+      tilesDefinitions?.map((tile: any) => ({
```
ibacher, line 24: This should be correctly typed.

```suggestion
      tilesDefinitions?.map((tile=> ({
```

packages/esm-patient-chart-app/src/clinical-views/components/clinical-views-summary.component.tsx
```diff
@@ -0,0 +1,46 @@
...
+        title: tile.tileHeader,
```
ibacher, line 25: If this is coming straight from config, it should be translatable.

packages/esm-patient-chart-app/src/clinical-views/components/clinical-views-summary.component.tsx
```diff
@@ -0,0 +1,46 @@
...
+}
+
+interface TileDefinition {
+  title: string;
+  columns: Array<EncounterTileColumn>;
+}
+
+const ClinicalViewsSummary: React.FC<OverviewListProps> = ({ patientUuid }) => {
+  const config = useConfig();
+  const tilesDefinitions = config.tilesDefinitions;
```
ibacher, line 20:
```suggestion
  const { tileDefinitions } = useConfig</* ConfigType goes here */>();
```

packages/esm-patient-chart-app/src/clinical-views/components/clinical-views-summary.component.tsx
```diff
@@ -0,0 +1,46 @@
+import React, { useMemo } from 'react';
+import { useConfig } from '@openmrs/esm-framework';
+import { getEncounterTileColumns } from '../../encounter-tile/utils/encounter-tile-config-builder';
```
ibacher, line 3: Organization-wise here, the encounter-tile is part of the clinical views, so surely it should be under the components for that?

packages/esm-patient-chart-app/src/encounter-tile/components/encounter-tile.component.tsx
```diff
@@ -0,0 +1,79 @@
...
+          {columns.map((column, ind) => (
+            <EncounterValuesTile key={ind} patientUuid={patientUuid} column={column} />
+          ))}
+        </Column>
+      </Tile>
+    </div>
+  );
+};
+
+export const MemoizedEncounterTile = React.memo(EncounterTile);
```
ibacher, line 44:
```suggestion
export const EncounterTile = React.memo(EncounterTileInternal);
```

ibacher: It's better to reserve weird names for the non-exported bits of an API.

packages/esm-patient-chart-app/src/encounter-tile/components/encounter-tile.component.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  if (isLoading || isValidating) {
+    return <CodeSnippetSkeleton type="multi" data-testid="skeleton-text" />;
+  }
+
+  if (error || lastEncounter === undefined) {
+    return (
+      <div className={styles.tileBox}>
+        <div className={styles.tileBoxColumn}>
+          <span className={styles.tileTitle}> {column.header} </span>
```
ibacher, line 57: Spaces here seem wrong?

packages/esm-patient-chart-app/src/encounter-tile/hooks/useLastEncounter.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import useSWR from 'swr';
+
+export const encounterRepresentation =
+  'custom:(uuid,encounterDatetime,encounterType,location:(uuid,name),' +
+  'patient:(uuid,display,age,identifiers,person),encounterProviders:(uuid,provider:(uuid,name)),' +
+  'obs:(uuid,obsDatetime,voided,groupMembers,concept:(uuid,name:(uuid,name)),value:(uuid,name:(uuid,name),' +
+  'names:(uuid,conceptNameType,name))),form:(uuid,name))';
+
+export function useLastEncounter(patientUuid: string, encounterType: string) {
```
ibacher, line 12:
```suggestion
export function useMostRecentEncounter(patientUuid: string, encounterType: string) {
```

packages/esm-patient-chart-app/src/encounter-tile/components/tile.scss
```diff
@@ -0,0 +1,96 @@
...
+  width: 100%;
+  display: 'flex';
+  flex-direction: 'row';
+}
+
+.tileSubTitle {
+  height: layout.$spacing-05;
+  font-size: 16px;
+  line-height: 1.33;
+  letter-spacing: 0.32px;
```
ibacher, line 61: Shouldn't we be using Carbon's typography defaults here and everywhere else in this file?

ibacher (changes requested):

packages/esm-patient-chart-app/src/clinical-views/components/clinical-views-summary.component.tsx
CynthiaKamau: I added the `Tile` type instead

packages/esm-patient-chart-app/src/clinical-views/components/encounter-tile/encounter-tile.component.tsx
```diff
@@ -0,0 +1,100 @@
+import { CodeSnippetSkeleton, Tile, Column, Layer } from '@carbon/react';
+import React, { useMemo } from 'react';
+import styles from './tile.scss';
+import { groupColumnsByEncounterType } from '../../utils/helpers';
+import { useLastEncounter } from '../../hooks/useLastEncounter';
+import { LazyCell } from '../../../lazy-cell/lazy-cell.component';
+import { useTranslation } from 'react-i18next';
+import { type FormattedCardColumn } from '../../utils/encounter-tile-config-builder';
+import { useLayoutType } from '@openmrs/esm-framework';
```
ibacher, line 9: Please review our Coding Conventions on code organization and how imports should be ordered.

packages/esm-patient-chart-app/src/clinical-views/components/encounter-tile/encounter-tile.component.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+const EncounterTileInternal: React.FC<EncounterTileProps> = ({ patientUuid, columns, headerTitle }) => {
+  const columnsByEncounterType = useMemo(() => groupColumnsByEncounterType(columns), [columns]);
+  const isTablet = useLayoutType() === 'tablet';
+
+  return (
+    <Layer className={styles.layer}>
+      <Tile className={styles.tile}>
+        <div className={isTablet ? styles.tabletHeading : styles.desktopHeading}>
+          <h4 className={styles.title}>{headerTitle}</h4>
```
ibacher, line 39: Any string displayed to the user needs to be translatable.

packages/esm-patient-chart-app/src/clinical-views/components/encounter-tile/encounter-tile.component.tsx
```diff
@@ -0,0 +1,100 @@
...
+              columns={columns as FormattedCardColumn[]}
+            />
+          ))}
+        </Column>
+      </Tile>
+    </Layer>
+  );
+};
+
+export const EncounterTile = React.memo(EncounterTileInternal);
```
ibacher, line 56: Why create two separate `const`s for this?

```ts
export const EncounterTile = React.memo(({ patientUuid, columns, headerTitle }: EncounterTileProps) => {
   // code goes here
});
```

Also works.

packages/esm-patient-chart-app/src/clinical-views/components/encounter-tile/encounter-tile.component.tsx
```diff
@@ -0,0 +1,100 @@
...
+        <div className={isTablet ? styles.tabletHeading : styles.desktopHeading}>
+          <h4 className={styles.title}>{headerTitle}</h4>
+        </div>
+        <Column className={styles.columnContainer}>
+          {Object.entries(columnsByEncounterType).map(([encounterType, columns]) => (
+            <EncounterData
+              key={encounterType}
+              patientUuid={patientUuid}
+              encounterType={encounterType}
+              columns={columns as FormattedCardColumn[]}
```
ibacher, line 47: Why do we need to cast the column types here? This seems like an unnecessary escape hatch when we have full control over:

1. The types in question
2. The thing that's being cast to a type
3. The function returning the thing that needs to be cast

packages/esm-patient-chart-app/src/clinical-views/components/encounter-tile/encounter-tile.component.tsx
```diff
@@ -0,0 +1,100 @@
...
+  if (error || lastEncounter === undefined) {
+    return (
+      <div className={styles.tileBox}>
+        {columns.map((column, ind) => (
+          <div key={ind} className={styles.tileBoxColumn}>
+            <span className={styles.tileTitle}>{t(column.title)}</span>
+            <span className={styles.tileValue}>--</span>
+          </div>
+        ))}
+      </div>
```
ibacher, line 79: So, here, an error happens and instead of reporting that to the user we just render a blank set of cells? That doesn't seem right... I'd expect us to render a message to the user explaining that the encounter data could not be loaded.

packages/esm-patient-chart-app/src/config-schema.ts
```diff
@@ -133,6 +133,11 @@ export const esmPatientChartSchema = {
     _default: '/etl-latest/etl/patient/',
     _description: 'Custom URL to load resources required for showing recommended visit types',
   },
+  trueConceptUuid: {
+    _type: Type.String,
+    _description: 'Default concept uuid for true in forms',
```
ibacher, line 138:
```suggestion
    _description: 'The UUID of the concept for true',
```

packages/esm-patient-chart-app/src/clinical-views/components/encounter-tile/encounter-tile.component.tsx
```diff
@@ -0,0 +1,100 @@
...
+    );
+  }
+
+  return (
+    <div className={styles.tileBox}>
+      {columns.map((column, ind) => (
+        <div key={ind} className={styles.tileBoxColumn}>
+          <span className={styles.tileTitle}>{column.title}</span>
+          <span className={styles.tileValue}>
+            <LazyCell lazyValue={column.getObsValue(lastEncounter)} />
```
ibacher, line 89: What value does the `LazyCell` add here? We've already loaded the data from the backend, so what does asynchronicity get us here?

packages/esm-patient-chart-app/src/clinical-views/components/encounter-tile/encounter-tile.component.tsx
```diff
@@ -0,0 +1,100 @@
...
+          </span>
+          {column.hasSummary && (
+            <span className={styles.tileTitle}>
+              <LazyCell lazyValue={column.getSummaryObsValue(lastEncounter)} />
```
ibacher, line 93: Ditto

packages/esm-patient-chart-app/src/clinical-views/components/encounter-tile/tile.scss
```diff
@@ -0,0 +1,96 @@
...
+    padding-top: 0.188rem;
+    border-bottom: 0.375rem solid var(--brand-03);
+  }
+}
+
+.heading:after {
+  content: '';
+  display: block;
+  width: layout.$spacing-07;
+  padding-top: 0.188rem;
```
ibacher, line 84: Seems like this could just be `layout.$spacing-01`, not exactly the same, but close enough, unless we have a really strong reason for using non-standard values?

packages/esm-patient-chart-app/src/clinical-views/components/encounter-tile/tile.scss
```diff
@@ -0,0 +1,96 @@
...
+  @include type.type-style('body-compact-01');
+}
+
+.tileValue {
+  @include type.type-style('body-long-01');
+  margin: layout.$spacing-05 layout.$spacing-10 layout.$spacing-06 0;
+}
+
+.tileBox {
+  width: 25%;
```
ibacher, line 45: This seems like a bad assumption here. Surely we can use flex-box to figure out the right width for this element?

ibacher: Otherwise, this widget winds up being much less flexible than it seems to be...

packages/esm-patient-chart-app/src/clinical-views/hooks/useLastEncounter.ts
```diff
@@ -0,0 +1,27 @@
...
+  const query = `encounterType=${encounterType}&patient=${patientUuid}&limit=1&order=desc&startIndex=0`;
+  const endpointUrl = `/ws/rest/v1/encounter?${query}&v=${encounterRepresentation}`;
+
+  const { data, error, isValidating } = useSWR<{ data: { results: Array<OpenmrsEncounter> } }, Error>(
+    endpointUrl,
+    openmrsFetch,
+    { dedupingInterval: 5000, refreshInterval: 0 },
+  );
+  return {
+    lastEncounter: data ? data?.data?.results.shift() : null,
```
ibacher, line 22: Even with this change, this can still yield `undefined` as well as `null`...

```suggestion
    lastEncounter: data ? data.data?.results.shift() : null,
```

packages/esm-patient-chart-app/src/clinical-views/utils/encounter-tile-config-builder.ts
ibacher, line 1: Can we come up with a better name for this file? Like, why isn't this just `utils/index.ts` or something? (The types and methods in here are specifically tied to configuration or anything).

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -0,0 +1,189 @@
...
+  obs: Array<Observation>;
+  form?: {
+    uuid: string;
+  };
+  visit?: string;
+}
+
+export function getEncounterValues(encounter: Encounter, param: string, isDate?: Boolean) {
+  if (isDate) return formatDate(encounter[param]);
+  else return encounter[param] ? encounter[param] : '--';
```
ibacher, line 46:
```suggestion
  else return encounter[param] ?? '--';
```

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -0,0 +1,189 @@
...
+}
+
+export function obsArrayDateComparator(left: Observation, right: Observation) {
+  return new Date(right.obsDatetime).getTime() - new Date(left.obsDatetime).getTime();
+}
+
+export function findObs(encounter: Encounter, obsConcept: string): Observation {
+  const allObs = encounter?.obs?.filter((observation) => observation.concept.uuid === obsConcept) || [];
+  return allObs?.length == 1 ? allObs[0] : allObs?.sort(obsArrayDateComparator)[0];
```
ibacher, line 55:
```suggestion
  return allOb?.length == 1 ? allObs[0] : allObs.sort(obsArrayDateComparator)[0];
```

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -0,0 +1,189 @@
...
+    const concept = fallbackConcepts.find((c) => findObs(encounter, c) != null);
+    obs = findObs(encounter, concept);
+  }
+
+  if (!obs) {
+    return '--';
+  }
+
+  if (isDate) {
+    if (typeof obs.value === 'object' && obs.value?.names) {
```
ibacher, line 168: This could use a comment for what it's checking for...

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -0,0 +1,189 @@
...
+) {
+  let obs = findObs(encounter, obsConcept);
+  if (!encounter || !obsConcept) {
+    return '--';
+  }
+
+  if (isTrueFalseConcept) {
+    if (typeof obs?.value === 'object') {
+      if (
+        (obs?.value?.uuid != esmPatientChartSchema.trueConceptUuid._default && obs?.value?.name?.name !== 'Unknown') ||
```
ibacher, line 122: So we only support the default value in the config schema? That's definitely an abuse of the config schema, since it wouldn't change anything if we hard-coded the value inline. This needs to be updated to taken the `trueConceptUuid` as an argument.

ibacher: Also, the check for `name !== 'Unknown` is incorrect as we cannot rely on the name returned for the "Unknown" concept being "Unknown" unless we are only working in English. Ideally, we'd be capturing the UUIDs of _both_ the `true` and `unknown` concepts (and maybe false) in the config.

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -0,0 +1,189 @@
...
+        obs?.value?.name?.name === t('FALSE')
```
ibacher, line 123: This is a bad use of the translation system, since it requires that the translation for `FALSE` be whatever the name of the concept is in the locale. Just collect the UUID for the false concept via config.

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -0,0 +1,189 @@
...
+      ) {
+        return t('No');
+      } else if (obs?.value?.uuid == esmPatientChartSchema.trueConceptUuid._default) {
+        return t('Yes');
```
ibacher, line 127: We should probably renamed these to "Yes concept" and "No concept" if that's the string that we're using for them.

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -0,0 +1,189 @@
...
+    return encounter.location.display;
+  }
+
+  if (type === 'provider') {
+    return encounter.encounterProviders.map((p) => p.provider.name).join(' | ');
+  }
+
+  if (type === 'visitType') {
+    return encounter.encounterType.name;
+  }
```
ibacher, line 144: All of these seem like a mixture of concerns that should be moved somewhere else, since these are not about obs from an encounter, but properties of the encounter itself.

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  if (type === 'ageAtHivTest') {
```
ibacher, line 146: Can we rename this to `ageAtEncounter`?

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -0,0 +1,189 @@
...
+    const obs = getObsFromEncounter(encounter, concept);
+    if (obs !== '--') {
+      observations.push(obs);
+    }
+  });
+
+  return observations.length ? observations.join(', ') : '--';
+}
+
+export function getObsFromEncounter(
```
ibacher, line 104: I feel like this function needs documentation explaining what it's supposed to do, etc.

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -0,0 +1,189 @@
...
+  encounter: Encounter,
+  obsConcept: string,
+  isDate?: Boolean,
+  isTrueFalseConcept?: Boolean,
+  type?: string,
```
ibacher, line 109: We only support a fixed set of types, so the type here should probably reflect that.

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -0,0 +1,189 @@
...
+  fallbackConcepts?: Array<string>,
+  secondaryConcept?: string,
```
ibacher, line 111: What is a "secondaryConcept"? How does this differ from "fallbackConcepts"? These are the things that we need to explain via comments.

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -0,0 +1,189 @@
...
+    return encounter.encounterType.name;
+  }
+
+  if (type === 'ageAtHivTest') {
+    return age(encounter.patient.birthDate, encounter.encounterDatetime);
+  }
+
+  if (secondaryConcept && typeof obs.value === 'object' && obs.value.names) {
+    const primaryValue = obs.value.name.display;
+    if (primaryValue === t('Other non-coded')) {
```
ibacher, line 152: This is, again, not what the translation system should be used for. Capture the UUID of the concept via the configuration system.

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -0,0 +1,189 @@
...
+}
+
+export function findObs(encounter: Encounter, obsConcept: string): Observation {
+  const allObs = encounter?.obs?.filter((observation) => observation.concept.uuid === obsConcept) || [];
+  return allObs?.length == 1 ? allObs[0] : allObs?.sort(obsArrayDateComparator)[0];
+}
+
+export function getObsFromEncounters(encounters: Encounter, obsConcept: string) {
+  const filteredEnc = encounters?.find((enc) => enc.obs.find((obs) => obs.concept.uuid === obsConcept));
+  return getObsFromEncounter(filteredEnc, obsConcept);
```
ibacher, line 60: This is one of the many reasons that I think `getObsFromEncounter` needs refactoring. It's doing too many things with too many concerns.

packages/esm-patient-common-lib/src/types/index.ts
```diff
@@ -1,3 +1,5 @@
+import { OpenmrsResource } from '@openmrs/esm-framework';
```
ibacher, line 1:
```suggestion
import { type OpenmrsResource } from '@openmrs/esm-framework';
```

ibacher (changes requested):
Not sure I've covered everything, but here are some things that need to be cleaned-up before this is mergeable. Goal here is to make sure that the code is in a format that the community can support over the long term. Thanks!

packages/esm-patient-chart-app/src/clinical-views/components/encounter-tile/encounter-tile.component.tsx
CynthiaKamau: The strings from the config are translated already

packages/esm-patient-chart-app/src/clinical-views/components/encounter-tile/encounter-tile.component.tsx
CynthiaKamau: The data loaded here is the whole encounter, so we need to fetch a specific obs from it and format the obs value

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
CynthiaKamau: I have added another function to address this

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
CynthiaKamau: A fallback concept is used in instances where a tile has an alternative value, for example, enrollment date and re-enrollment date. If there is no enrollment date, then the value of the re-enrollment date can be displayed instead 
cc : @eudson

ibacher: The main point here was that these use-cases need to be clear in the code, so please add some comments explaining that distinction.

packages/esm-patient-chart-app/src/clinical-views/components/encounter-tile/encounter-tile.component.tsx
ibacher: I'm not sure I follow? At this point in the code, `lastEncounter` is loaded and has a value. `getObsValue()` is not asynchronous in any way, so why do we need asynchronicity here?

gracepotma:
I'm confused, is this ready for rull re-review?

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
CynthiaKamau: I have added comments to explain params and some specific functions

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
CynthiaKamau: i have seperated some few things, like data that is not fetched from the encounter eg a visit type

eudson:
> I'm confused, is this ready for rull re-review?

@gracepotma what does ready for full re-review means?

gracepotma:
Discussed on today's Content Package check-in call w/ Eudson, Ian, Cynthia, & Jonathan O; apparently this is ready for re-review :) We'll make this a priority to review along with the Tabbed Encounter Widget PR: https://github.com/openmrs/openmrs-esm-patient-chart/pull/1987

ibacher (commented):
At a high-level, this looks like it isn't massively distinct from #1987, at least in the backing helpers, etc. Since I'd prefer we didn't duplicate code, is it possible to rework this PR to use the same definitions from there at least for `getObsFromEncounter()` and similar shared functionality?

packages/esm-patient-chart-app/src/clinical-views/components/encounter-tile/encounter-tile.component.tsx
```diff
@@ -0,0 +1,100 @@
...
+  columns: EncounterTileColumn[];
+}> = ({ patientUuid, encounterType, columns }) => {
+  const { lastEncounter, isLoading, error, isValidating } = useLastEncounter(patientUuid, encounterType);
+  const { t } = useTranslation();
+
+  if (isLoading || isValidating) {
+    return <CodeSnippetSkeleton type="multi" role="progressbar" />;
+  }
+
+  if (error || lastEncounter == undefined) {
```
ibacher, line 63: Shouldn't this just be `!lastEncounter`?

packages/esm-patient-chart-app/src/clinical-views/hooks/useLastEncounter.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  const cacheKey = endpointUrl;
+
+  const { data, error, isValidating, isLoading } = useSWR<{ results: Array<Encounter> }, Error>(
+    cacheKey,
+    async (url) => {
+      const cachedData = cache.get(url);
+      if (cachedData) {
+        return cachedData;
+      }
```
ibacher, line 26: The manual caching here seems unnecessary. It would be simpler to use `useSWRImmutable`, I think.

ibacher (commented):
Couple more small points.

CynthiaKamau:
I have restructured a few files that cut across this pr and the encounter list PR

eudson:
> I have restructured a few files that cut across this pr and the encounter list PR

@ibacher @gracepotma kindly re-review the PR

packages/esm-patient-chart-app/src/clinical-views/encounter-list/encounter-list.component.tsx
```diff
@@ -6,12 +6,11 @@ import { EncounterListDataTable } from './table.component';
...
 import { Add } from '@carbon/react/icons';
 import { launchEncounterForm } from '../utils/helpers';
-import { deleteEncounter } from '../encounter-list.resource';
+import { deleteEncounter } from '../utils/encounter-list.resource';
 import { useEncounterRows, useFormsJson } from '../hooks';
 
 import styles from './encounter-list.scss';
-import { type TableRow, type Encounter, type Mode, type ColumnValue } from '../types';
-import { type FormattedColumn } from '../utils/encounter-list-config-builder';
+import { type TableRow, type Encounter, type Mode, type ColumnValue, type FormattedColumn } from '../types';
```
ibacher, line 13: Can we reorder the imports here and elsewhere to conform to [our code organization standards](https://o3-docs.openmrs.org/docs/coding-conventions/code-organization), please?

packages/esm-patient-chart-app/src/clinical-views/encounter-tile/clinical-views-summary.component.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { getEncounterTileColumns } from '../utils';
+import { EncounterTile } from './encounter-tile.component';
+import { type MenuCardProps } from '../types';
+
+interface OverviewListProps {
+  patientUuid: string;
+}
+
+const ClinicalViewsSummary: React.FC<OverviewListProps> = memo(({ patientUuid }) => {
+  const config = useConfig<ConfigObject>();
```
ibacher, line 14: There's not really any point to using the `ConfigObject` type here. It's more useful to use an app-specific type, but if there is not, declare the type here appropriately.

```suggestion
  const config = useConfig();
```

ibacher: It also seems like it could be cleaner to write:

```ts
const {
  tileDefinitions,
  trueConceptUuid,
  falseConceptUuid,
  otherConceptUuid
} = useConfig();
```

packages/esm-patient-chart-app/src/clinical-views/encounter-tile/clinical-views-summary.component.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+    return tileDefinitions?.map((tile: MenuCardProps) => ({
+      title: t(tile.tileHeader),
+      columns: getEncounterTileColumns(tile, configConcepts),
+    }));
+  }, [tileDefinitions, t, config.trueConceptUuid, config.falseConceptUuid, config.otherConceptUuid]);
+
+  return (
+    <>
+      {tilesData?.length > 0 &&
```
ibacher, line 33: This feels like it should be part of the original return, e.g.,

```
return tilesData?.length > 0 ? /* current stuff */ : null;
```

packages/esm-patient-chart-app/src/clinical-views/encounter-tile/clinical-views-summary.component.tsx
```diff
@@ -0,0 +1,41 @@
...
+        tilesData?.map((tile, index) => (
```
ibacher, line 34: At this point, we know `tilesData` is not null or undefined.

```suggestion
        tilesData.map((tile, index) => (
```

packages/esm-patient-chart-app/src/clinical-views/encounter-tile/encounter-tile.component.tsx
```diff
@@ -0,0 +1,84 @@
+import React, { useMemo } from 'react';
+import { useLayoutType } from '@openmrs/esm-framework';
+import { CodeSnippetSkeleton, Tile, Column, Layer } from '@carbon/react';
+import styles from './tile.scss';
+import { groupColumnsByEncounterType } from '../utils/helpers';
+import { useTranslation } from 'react-i18next';
+import { type EncounterTileColumn, type EncounterTileProps } from '../types';
+import { useLastEncounter } from '../hooks';
+
+export const EncounterTile = React.memo(({ patientUuid, columns, headerTitle }: EncounterTileProps) => {
```
ibacher, line 10:
```suggestion
export const EncounterTile = memo(({ patientUuid, columns, headerTitle }: EncounterTileProps) => {
```

packages/esm-patient-chart-app/src/clinical-views/encounter-tile/encounter-tile.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const isTablet = useLayoutType() === 'tablet';
+
+  return (
+    <Layer className={styles.layer}>
+      <Tile className={styles.tile}>
+        <div className={isTablet ? styles.tabletHeading : styles.desktopHeading}>
+          <h4 className={styles.title}>{headerTitle}</h4>
+        </div>
+        <Column className={styles.columnContainer}>
+          {Object.entries(columnsByEncounterType).map(([encounterUuid, columns]) => (
```
ibacher, line 21:
```suggestion
          {Object.entries(columnsByEncounterType).map(([encounterTypeUuid, columns]) => (
```

packages/esm-patient-chart-app/src/clinical-views/encounter-tile/encounter-tile.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+  if (error || lastEncounter == undefined) {
+    return (
+      <div className={styles.tileBox}>
+        {columns.map((column, ind) => (
+          <div key={ind}>
+            <span className={styles.tileTitle}>{t(column.title)}</span>
+            <span className={styles.tileValue}>{error?.message}</span>
+          </div>
+        ))}
+      </div>
```
ibacher, line 56: If there's only a single error, why do we render the error once per column? I'd just display the error message in the box, maybe with something that says "Error loading most recent encounter" or something

packages/esm-patient-chart-app/src/clinical-views/encounter-tile/encounter-tile.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+              <p>{column.getObsValue(lastEncounter)}</p>
+            </span>
+            {column.hasSummary && (
+              <>
+                <span className={styles.tileValue}>
+                  <p>{column.getSummaryObsValue(lastEncounter)}</p>
+                </span>
+                <span className={styles.tileValue}>
+                  <p>{column.getObsValue(lastEncounter)}</p>
+                </span>
```
ibacher, line 76: Why do we render the current value a second time here?

packages/esm-patient-chart-app/src/clinical-views/utils/index.ts
```diff
@@ -0,0 +1,99 @@
+import { getConceptFromMappings, getObsFromEncounter } from './helpers';
+import { type ColumnDefinition, type ConfigConcepts, type EncounterTileColumn, type MenuCardProps } from '../types';
+
+const calculateDateDifferenceInDate = (givenDate: string): string => {
+  const dateDifference = new Date().getTime() - new Date(givenDate).getTime();
+  const totalDays = Math.floor(dateDifference / (1000 * 3600 * 24));
+  return `${totalDays} days`;
```
ibacher, line 7: Seems like this could be:

```suggestion
   return `${dayjs().diff(dayjs(givenDate), 'days')} days`
```

ibacher: Also, the `days` here needs to be translatable. I'd probably have this function calculate the difference in days and then use i18next's format capabilities.

packages/esm-patient-chart-app/src/clinical-views/utils/index.ts
```diff
@@ -0,0 +1,99 @@
...
+};
+
+export const getEncounterTileColumns = (tileDefinition: MenuCardProps, config: ConfigConcepts) => {
+  const columns: Array<EncounterTileColumn> = tileDefinition.columns?.map((column: ColumnDefinition) => ({
+    key: column.title,
+    header: column.title,
+    concept: column.concept,
+    encounterUuid: column.encounterType,
+    hasSummary: column.hasSummary || false,
+    getObsValue: (encounter) => {
```
ibacher, line 17: Can we have a type for `encounter` here, please?

packages/esm-patient-chart-app/src/clinical-views/utils/index.ts
```diff
@@ -0,0 +1,99 @@
...
+  const totalDays = Math.floor(dateDifference / (1000 * 3600 * 24));
+  return `${totalDays} days`;
+};
+
+export const getEncounterTileColumns = (tileDefinition: MenuCardProps, config: ConfigConcepts) => {
+  const columns: Array<EncounterTileColumn> = tileDefinition.columns?.map((column: ColumnDefinition) => ({
+    key: column.title,
+    header: column.title,
+    concept: column.concept,
+    encounterUuid: column.encounterType,
```
ibacher, line 15:
```suggestion
    encounterTypeUuid: column.encounterType,
```

packages/esm-patient-chart-app/src/clinical-views/utils/index.ts
```diff
@@ -0,0 +1,99 @@
...
+          column.isTrueFalseConcept,
+          column.type,
+          column.fallbackConcepts,
+          column.summaryConcept?.secondaryConcept,
+          config,
+        );
+      } else {
+        obsValue = getObsFromEncounter(encounter, column.concept, column.isDate, null, null, null, null, config);
+      }
+      return typeof obsValue === 'string' ? obsValue : obsValue?.name?.name || '--';
```
ibacher, line 34:
```suggestion
      return typeof obsValue === 'string' ? obsValue : obsValue?.name?.name ?? '--';
```

packages/esm-patient-chart-app/src/clinical-views/utils/index.ts
```diff
@@ -0,0 +1,99 @@
...
+    },
+    getSummaryObsValue: column.hasSummary
+      ? (encounter) => {
```
ibacher, line 37: Ditto on type

packages/esm-patient-chart-app/src/clinical-views/utils/index.ts
```diff
@@ -0,0 +1,99 @@
...
+              encounter,
+              column.summaryConcept.primaryConcept,
+              null,
+              null,
+              null,
+              null,
+              null,
+              config,
+            );
+            if (primaryConceptType !== '--') {
```
ibacher, line 51:
```suggestion

            if (primaryConceptType !== '--') {
```

packages/esm-patient-chart-app/src/clinical-views/hooks/useLastEncounter.ts
```diff
@@ -0,0 +1,27 @@
...
+  const endpointUrl =
+    patientUuid && encounterType ? `/ws/rest/v1/encounter?${query}&v=${encounterRepresentation}` : null;
+
+  const { data, error, isLoading, isValidating } = useSWRImmutable<{ data: { results: Array<Encounter> } }, Error>(
+    endpointUrl,
+    openmrsFetch,
+  );
+
+  return {
+    lastEncounter: data ? data?.data.results?.length > 0 && data?.data.results[0] : null,
```
ibacher, line 22: This expression seems incorrect to me. What this does is:

* If the data is falsy, return `null`
* If the data is truthy, but `results` is undefined, null, or length 0, return `false`
* Otherwise, return the first result

Specifically, the middle item looks unintentional.

packages/esm-patient-chart-app/src/clinical-views/hooks/useLastEncounter.ts
```diff
@@ -0,0 +1,27 @@
...
+import useSWRImmutable from 'swr/immutable';
+
+const encounterRepresentation =
+  'custom:(uuid,encounterDatetime,encounterType,location:(uuid,name),' +
+  'patient:(uuid,display,age,identifiers,person),encounterProviders:(uuid,provider:(uuid,name)),' +
+  'obs:(uuid,obsDatetime,voided,groupMembers,concept:(uuid,display,name:(uuid,name)),value:(uuid,name:(uuid,name,display),' +
+  'names:(uuid,conceptNameType,name,display))),form:(uuid,name))';
+
+export function useLastEncounter(patientUuid: string, encounterType: string) {
+  const query = `encounterType=${encounterType}&patient=${patientUuid}&limit=1&order=desc&startIndex=0`;
```
ibacher, line 12: I'm confused as to why this query includes fixed parts, but then we append a semidynamic value (the representation) to the end. This feels like it could be better organized.

packages/esm-patient-chart-app/src/clinical-views/hooks/useLastEncounter.ts
```diff
@@ -0,0 +1,27 @@
...
+  const endpointUrl =
+    patientUuid && encounterType ? `/ws/rest/v1/encounter?${query}&v=${encounterRepresentation}` : null;
+
+  const { data, error, isLoading, isValidating } = useSWRImmutable<{ data: { results: Array<Encounter> } }, Error>(
```
ibacher, line 16: This seems like an incorrect use of `useSWRImmutable()`. Please just `useSWR()` here. We can adjust the SWR config if it needs to not reload under some circumstances or something.

ibacher (changes requested):
Thanks @CynthiaKamau! A few more nitpicky things that need to addressed before this is mergeable. Thanks!

packages/esm-patient-chart-app/src/clinical-views/encounter-tile/encounter-tile.component.tsx
CynthiaKamau: The fist block displays the actual concept answer, but in some instances, if a user wants to diaply additional information captured in a different concept, they can do it ussing a summary concept like :
```
  {
      "id": "currentRegimen",
      "hasSummary": true,
      "title": "currentRegimenTitle",
      "encounterType": "74bf4fe6-8fdb-4228-be39-680a93a9cf6d",
      "concept": "dfbe256e-30ba-4033-837a-2e8477f2e7cd",
      "summaryConcept": {
        "primaryConcept": "164515AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
      }
 },
```

packages/esm-patient-chart-app/src/clinical-views/utils/index.ts
CynthiaKamau: whats the difference here ?

packages/esm-patient-chart-app/src/clinical-views/encounter-tile/encounter-tile.component.tsx
CynthiaKamau: Each column can have a different encounter type, so an endpoint request will be made for each. One request can have an error and another successful, the purpose of displaying each per box is to help the person creating the columns to know why that specific value is not being displayed.

packages/esm-patient-chart-app/src/clinical-views/encounter-list/encounter-list.component.tsx
CynthiaKamau: Done

packages/esm-patient-chart-app/src/clinical-views/hooks/useLastEncounter.ts
CynthiaKamau: Would this suffice ?
```
const query = `encounterType=${encounterType}&patient=${patientUuid}`;
  const endpointUrl =
    patientUuid && encounterType
      ? `/ws/rest/v1/encounter?${query}&v=${encounterRepresentation}&limit=1&order=desc&startIndex=0`
      : null;
```

packages/esm-patient-chart-app/src/clinical-views/utils/index.ts
ibacher: Just adding a blank line. I just had it here so I can commit it.

packages/esm-patient-chart-app/src/clinical-views/hooks/useLastEncounter.ts
ibacher: Yeah, that's better.

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -98,56 +102,47 @@ export function getMultipleObsFromEncounter(encounter: Encounter, obsConcepts: A
...
+ * @param obsConcept - The main concept to search for in the encounter's observations.
+ * @param isDate - Optional flag to indicate if the observation value should be treated as a date.
+ * @param isTrueFalseConcept - Optional flag to check if the observation concept represents a true/false value.
+ * @param type - Optional property type for filtering or fetching additional encounter properties.
+ * @param fallbackConcepts - Optional list of alternative concepts to use if the primary concept is not found.
+ * @param secondaryConcept - Optional secondary concept to check if the primary value matches a specific condition.
+ * @param t - Optional translation function.
+ * @returns The value of the observation, formatted appropriately, or '--' if not found or applicable.
+ */
 export function getObsFromEncounter(
```
NethmiRodrigo, line 118: Could we use named parameters and do something like:
```
// Define an interface for the parameters
interface GetObsFromEncounterParams {
  encounter: Encounter;
  obsConcept: string;
  isDate?: boolean;
  isTrueFalseConcept?: boolean;
  type?: EncounterPropertyType;
  fallbackConcepts?: Array<string>;
  secondaryConcept?: string;
  config?: ConfigConcepts;
}

export function getObsFromEncounter({
  encounter,
  obsConcept,
  isDate = false,
  isTrueFalseConcept = false,
  type,
  fallbackConcepts = [],
  secondaryConcept,
  config,
}: GetObsFromEncounterParams) {}
```
Since I've seen in multiple places that the function is called with many null params `getObsFromEncounter(encounter, concept, null, null, null, null, null, config);` and it can be confusing so having function calls with named params:
 `getObsFromEncounter({
  encounter: myEncounter,
  obsConcept: "someConceptString",
  isDate: true,
  fallbackConcepts: ["concept1", "concept2"]
});`
would be easier to read.

NethmiRodrigo (commented):
Thank you @CynthiaKamau! Haven't gone through it in its entirety but left some minor nit picks and suggestions:

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
CynthiaKamau: Done

ibacher (approved):
Thanks for all the work on this!

ibacher:
Hi @CynthiaKamau! I intentionally left this unmerged because we're in a release cycle right now. I'm going to need to back this out and we're going to need to revive this PR.

PR #1987, opened 2024-09-02
(feat) O3-3797: Add the Encounter List Table and Tabs to Patient Chart

hadijahkyampeire:
## Requirements

## Summary
- This PR adds the Encounter list table, tabs and schema builder to patient chart in regards to the HIV Care and Treatment content package.

## Screenshots

https://github.com/user-attachments/assets/e566356e-699c-4f7d-84ea-298feede52af

## Related Issue
- https://openmrs.atlassian.net/browse/O3-3788
- https://openmrs.atlassian.net/browse/O3-3797


packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -0,0 +1,188 @@
...
+      encounterUuid,
+      formUuid: form.name,
+      patientUuid: patientUuid,
+      visitTypeUuid: '',
+      visitUuid: '',
+      visitStartDatetime: '',
+      visitStopDatetime: '',
+      additionalProps: {
+        mode: action === 'add' ? 'enter' : action,
+        formSessionIntent: intent,
```
kajambiya, line 31:
```suggestion
        formSessionIntent: intent,
        openClinicalFormsWorkspaceOnFormClose: false,
```

kajambiya: Add this flag to prevent auto opening of the clinical forms workspace when a form is cancelled.

packages/esm-patient-chart-app/src/encounter-list/components/encounter-list.component.tsx
```diff
@@ -0,0 +1,300 @@
...
+          name: formsJson[0]?.name,
+        },
+        mode: 'delete',
+        intent: '*',
+      },
+    ],
+    [formsJson, t],
+  );
+
+  const createLaunchFormAction = (encounter, mode) => () =>
```
ibacher, line 113: This should be wrapped in `useCallback()`. Also, `encounter`, and `mode` should be typed.

packages/esm-patient-chart-app/src/encounter-list/components/encounter-list.component.tsx
```diff
@@ -0,0 +1,300 @@
...
+            .catch(() => {
+              showSnackbar({
+                isLowContrast: false,
+                title: t('error', 'Error'),
+                subtitle: `Encounter ${t('failedDeleting', "couldn't be deleted")}`,
+                kind: 'error',
+              });
+            })
+            .finally(() => {
+              mutate();
```
ibacher, line 142: `mutate()` only needs to be called if the operation succeeded.

packages/esm-patient-chart-app/src/encounter-list/components/encounter-list.component.tsx
```diff
@@ -0,0 +1,300 @@
...
+              close();
+            });
+        },
+      });
+    },
+    [onFormSave, t, mutate],
+  );
+
+  const tableRows = encounters.map((encounter) => {
```
ibacher, line 151: Ideally, this would be a separate component (in this same file) or, at the very least, wrapped in a `useMemo()`.

packages/esm-patient-chart-app/src/encounter-list/components/encounter-list.component.tsx
```diff
@@ -0,0 +1,300 @@
...
+      }
+      tableRow[column.key] = val;
+    });
+
+    const actions = tableRow.actions?.length ? tableRow.actions : defaultActions;
+
+    tableRow['actions'] = (
+      <OverflowMenu flipped className={styles.flippedOverflowMenu} data-testid="actions-id">
+        {actions.map((actionItem, index) => {
+          const form = formsJson.length ? formsJson?.find((form) => form.name === actionItem?.form?.name) : null;
```
ibacher, line 185:
```suggestion
          const form = formsJson.length && actionItem?.form?.name ? formsJson.find((form) => form.name === actionItem.form.name) : null;
```

packages/esm-patient-chart-app/src/encounter-list/components/encounter-list.component.tsx
```diff
@@ -0,0 +1,300 @@
...
+              close();
+            });
+        },
+      });
+    },
+    [onFormSave, t, mutate],
+  );
+
+  const tableRows = encounters.map((encounter) => {
+    const tableRow: { id: string; actions: any } = { id: encounter.uuid, actions: null };
```
ibacher, line 152: Again, we shouldn't use an explicit `any` when the type is known:

```
{ form: { name: string }, label: string, mode: 'view' | 'edit' | 'delete' }

**Review by ibacher — CHANGES_REQUESTED** [review]
I haven't made it through this whole PR, but here are some of the more obvious either issues, things that weren't clear, or places where we should have improvements.

File: packages/esm-patient-chart-app/src/encounter-list/hooks/usePatientDeathStatus.ts [hunk 7]
```diff
@@ -0,0 +1,14 @@
+import { openmrsFetch, restBaseUrl } from '@openmrs/esm-framework';
+import useSWRImmutable from 'swr';
+
+export function usePatientDeathStatus(patientUuid: string) {
```
ibacher, line 4: Why do we need a hook just to load the patient's death status? This is overkill. The extension will already have the patient fed into it as a prop; use that.

packages/esm-patient-chart-app/src/encounter-list/hooks/usePatientDeathStatus.ts
```diff
@@ -0,0 +1,14 @@
+import { openmrsFetch, restBaseUrl } from '@openmrs/esm-framework';
+import useSWRImmutable from 'swr';
```
ibacher, line 2: nit, but this applies to all imports from `swr` in this PR.

```suggestion
import useSWR from 'swr';
```

packages/esm-patient-chart-app/src/encounter-list/hooks/usePatientDeathStatus.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export function usePatientDeathStatus(patientUuid: string) {
+  const {
+    data: response,
+    isLoading,
+    error,
+  } = useSWRImmutable<any, Error>(`${restBaseUrl}/person/${patientUuid}?v=custom:(dead)`, openmrsFetch);
```
ibacher, line 9: Don't explicitly use `any` if possible.

packages/esm-patient-chart-app/src/index.ts
```diff
@@ -238,3 +239,8 @@ export const activeVisitActionsComponent = getAsyncLifecycle(
   () => import('./visit/visits-widget/active-visit-buttons/active-visit-buttons'),
   { featureName: 'active-visit-actions', moduleName },
 );
+
+export const encounterListTableTabs = getSyncLifecycle(encounterListTabsComponent, {
```
ibacher, line 243: Use `getAsyncLifecycle()` here.

packages/esm-patient-chart-app/src/routes.json
```diff
@@ -13,6 +13,12 @@
       "online": true,
       "offline": true
     },
+    {
+      "name": "encounter-list-table-tabs",
```
ibacher, line 17: Since this is the name people will need to add to their actual configs, it would be better if we were using a more understandable name like `encounter-list` or `encounter-table` (my preference)

packages/esm-patient-chart-app/src/encounter-list/components/tag.component.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { Tag } from '@carbon/react';
+import { getObsFromEncounter, findObs } from '../../clinical-views/utils/helpers';
+
+export const renderTag = (encounter, concept, statusColorMappings) => {
```
ibacher, line 5: Please add types here.

packages/esm-patient-chart-app/src/encounter-list/components/encounter-list.component.tsx
```diff
@@ -0,0 +1,323 @@
...
+import { deleteEncounter } from '../encounter-list.resource';
+import { useEncounterRows, useFormsJson, usePatientDeathStatus } from '../hooks';
+
+import styles from './encounter-list.scss';
+import { type TableRow, type Encounter, type Mode } from '../types';
+
+export interface EncounterListColumn {
+  key: string;
+  header: string;
+  getValue: (encounter: any) => string;
```
ibacher, line 27: Encounter should be typed.

packages/esm-patient-chart-app/src/encounter-list/components/encounter-list.component.tsx
```diff
@@ -0,0 +1,323 @@
...
+    excludedIntents?: Array<string>;
+    fixedIntent?: string;
+    isDefault?: boolean;
+  }>;
+  launchOptions: {
+    hideFormLauncher?: boolean;
+    displayText?: string;
+    workspaceWindowSize?: 'minimized' | 'maximized';
+  };
+  filter?: (encounter: any) => boolean;
```
ibacher, line 49: Ditto here.

packages/esm-patient-chart-app/src/encounter-list/components/encounter-list.component.tsx
```diff
@@ -0,0 +1,323 @@
...
+  key: string;
+  header: string;
+  getValue: (encounter: any) => string;
+  link?: any;
+}
+
+export interface EncounterListProps {
+  patientUuid: string;
+  encounterType: string;
+  columns: Array<any>;
```
ibacher, line 34: This should be explicitly typed.

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -0,0 +1,192 @@
...
+type LaunchAction = 'add' | 'view' | 'edit' | 'embedded-view';
+
+export function launchEncounterForm(
+  form: FormSchema,
+  action: LaunchAction = 'add',
+  onFormSave: () => void,
+  title?: string,
+  encounterUuid?: string,
+  intent: string = '*',
+  workspaceWindowSize?: 'minimized' | 'maximized',
```
ibacher, line 17: This property isn't used at all, which seems silly because it is often given a value.

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -0,0 +1,192 @@
...
+import { esmPatientChartSchema } from '../../config-schema';
+import { type TFunction } from 'i18next';
+
+type LaunchAction = 'add' | 'view' | 'edit' | 'embedded-view';
+
+export function launchEncounterForm(
+  form: FormSchema,
+  action: LaunchAction = 'add',
+  onFormSave: () => void,
+  title?: string,
```
ibacher, line 14: This property also isn't used here at all. Consider removing it altogether.

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -0,0 +1,192 @@
...
+  // There might be a better way to use a generic value
+  if (isTrueFalseConcept) {
+    if (typeof obs?.value === 'object') {
+      if (
+        (obs?.value?.uuid != esmPatientChartSchema.trueConceptUuid._default && obs?.value?.name?.name !== 'Unknown') ||
+        obs?.value?.name?.name === 'FALSE'
+      ) {
+        return 'No';
+      } else if (obs?.value?.uuid == esmPatientChartSchema.trueConceptUuid._default) {
+        return 'Yes';
```
ibacher, line 129: Surely these strings need to be translatable? It would almost be better to use the actual names from the concept.

ibacher (changes requested):
High-level stuff:

* There's still too much reliance on `any` types
* This should all be consolidated into a single `encounter-list` folder. We don't clearly need both `clinical-views` and `encounter-list`
* The component in `index.ts` must use `getAsyncLifecycle()`.
* I can't get this working by adding the component to a slot; we really need a screenshot / video of _this_ component in use, not just the version in OHRI

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -0,0 +1,192 @@
...
+    workspaceTitle: form.name,
+    mutateForm: onFormSave,
+    formInfo: {
+      encounterUuid,
+      formUuid: form.name,
+      patientUuid: patientUuid,
+      visitTypeUuid: '',
+      visitUuid: '',
+      visitStartDatetime: '',
+      visitStopDatetime: '',
```
ibacher, line 30: For the purposes of the RefApp, we should require a visit before a form gets launched, as with the typical form-entry flow.

brandones:
Ping @hadijahkyampeire , is it clear to you how to move forward with this?

hadijahkyampeire:
@brandones I think @CynthiaKamau or @ODORA0 will be taking this on as I am currently unavailable to work. Thanks

brandones:
@CynthiaKamau or @ODORA0 will you be taking this on?

CynthiaKamau:
> @CynthiaKamau or @ODORA0 will you be taking this on?

i will take it up

packages/esm-patient-chart-app/src/encounter-list/hooks/useFormsJson.ts
```diff
@@ -0,0 +1,35 @@
+import { useEffect, useState } from 'react';
+import useSWRImmutable from 'swr';
+import { fetchOpenMRSForms } from '../encounter-list.resource';
+import { type FormSchema } from '@openmrs/esm-form-engine-lib';
+
+export function useFormsJson(formUuids: string[]) {
```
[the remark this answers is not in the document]
CynthiaKamau: This was mean to handle situations where an encounter has many forms affiliated to it, but we can start with the default use case and figure out how to do specific cases after

packages/esm-patient-chart-app/src/encounter-list/components/tag.component.tsx
CynthiaKamau: Converting it to a component becomes a bit hard to return it in a nested function

packages/esm-patient-chart-app/src/encounter-list/components/encounter-list-tabs.component.tsx
```diff
@@ -0,0 +1,53 @@
...
+        <TabList contained>
+          {tabsConfig.map((tab) => (
+            <Tab key={tab.name}>{t(tab.name)}</Tab>
+          ))}
+        </TabList>
+        <TabPanels>
+          {tabsConfig.map((tab) => (
+            <TabPanel key={tab.name}>
+              <EncounterList
+                filter={tab.hasFilter ? (encounter) => filter(encounter, tab.formList[0].uuid) : null}
```
NethmiRodrigo, line 34: Probably better to wrap the filter function part in a useCallback

packages/esm-patient-chart-app/src/encounter-list/components/encounter-list.component.tsx
NethmiRodrigo: In the `EncounterListTabsComponent`, this gets the type `FormattedColumn`, couldn't we use the same here, instead of any.

packages/esm-patient-chart-app/src/encounter-list/components/encounter-list-tabs.component.tsx
```diff
@@ -0,0 +1,53 @@
...
+                patientUuid={patientUuid}
+                formList={tab.formList}
+                columns={tab.columns}
+                encounterType={tab.encounterType}
+                launchOptions={tab.launchOptions}
+                headerTitle={tab.headerTitle}
+                description={tab.description}
+                currentVisit={currentVisit}
+                deathStatus={patient?.patient?.deceasedBoolean}
```
NethmiRodrigo, line 43: Would be better to default this to false, since this condition could result in null or undefined, and EncounterList component expects a boolean.

packages/esm-patient-chart-app/src/encounter-list/components/encounter-list.component.tsx
```diff
@@ -0,0 +1,296 @@
+import React, { useCallback, useMemo, useState } from 'react';
+import { navigate, showModal, showSnackbar, Visit } from '@openmrs/esm-framework';
```
NethmiRodrigo, line 2:
```suggestion
import { navigate, showModal, showSnackbar, type Visit } from '@openmrs/esm-framework';
```

packages/esm-patient-chart-app/src/encounter-list/components/encounter-list.component.tsx
```diff
@@ -0,0 +1,296 @@
...
+  afterFormSaveAction,
+  currentVisit,
+  deathStatus,
+}) => {
+  const { t } = useTranslation();
+
+  const [currentPage, setCurrentPage] = useState(1);
+  const [pageSize, setPageSize] = useState(10);
+
+  const { formsJson, isLoading: isLoadingFormsJson, error: errorFormJson } = useFormsJson(formList[0].uuid);
```
NethmiRodrigo, line 64: Wouldn't doing `formList[0].uuid` break in the (maybe unlikely) scenario that formList is an empty array?

NethmiRodrigo: Also `errorFormJson` is unused so we could remove it
```suggestion
  const { formsJson, isLoading: isLoadingFormsJson } = useFormsJson(formList[0].uuid);
```

packages/esm-patient-chart-app/src/encounter-list/components/encounter-list.component.tsx
```diff
@@ -0,0 +1,296 @@
...
+  const { encounters, total, isLoading, onFormSave, mutate } = useEncounterRows(
+    patientUuid,
+    encounterType,
+    filter,
+    afterFormSaveAction,
+    pageSize,
+    currentPage,
+  );
+
+  const { workspaceWindowSize, displayText, hideFormLauncher } = launchOptions;
```
NethmiRodrigo, line 74: Also unused
```suggestion
  const { displayText, hideFormLauncher } = launchOptions;
```

packages/esm-patient-chart-app/src/encounter-list/components/encounter-list.component.tsx
```diff
@@ -0,0 +1,296 @@
...
+
+  const createLaunchFormAction = useCallback(
+    (encounter: Encounter, mode: Mode) => () => {
+      launchEncounterForm(formsJson, currentVisit, mode, onFormSave, encounter.uuid, null, patientUuid);
+    },
+    [formsJson, onFormSave, patientUuid, currentVisit],
+  );
+
+  const handleDeleteEncounter = useCallback(
+    (encounterUuid, encounterTypeName) => {
```
NethmiRodrigo, line 114: Can we specify the types of these two as `string` (or the correct type, if I'm mistaken)?

packages/esm-patient-chart-app/src/encounter-list/components/encounter-list.component.tsx
```diff
@@ -0,0 +1,296 @@
...
+  }
+
+  return (
+    <>
+      {tableRows?.length > 0 || encounters.length > 0 ? (
+        <>
+          <div className={styles.widgetContainer}>
+            <div className={styles.widgetHeaderContainer}>
+              <h4 className={`${styles.productiveHeading03} ${styles.text02}`}>{t(headerTitle)}</h4>
+              {/* @ts-ignore */}
```
NethmiRodrigo, line 267: Purely for understanding, why do we have a `ts-ignore` here?

NethmiRodrigo (commented):
Thanks @CynthiaKamau! Didn't quite make it through the entire PR yet, but a left a few comments about suggestions or questions I had

packages/esm-patient-chart-app/src/encounter-list/utils/helpers.ts
```diff
@@ -0,0 +1,188 @@
...
+        obs?.value?.name?.name === t('FALSE')
+      ) {
+        return t('No');
+      } else if (obs?.value?.uuid == esmPatientChartSchema.trueConceptUuid._default) {
+        return t('Yes');
+      } else {
+        return obs?.value?.name?.name;
+      }
+    }
+  }
```
ibacher, line 130: I've made these comments before but:

1. This mechanism of loading the the concept UUID from the schema is effectively the same as hard-coding the UUID. This _must_ be made configurable, and that means calling either `useConfig()` and passing it in or `getConfig()`.
2. `Unknown` and `FALSE` values _must_ be determinable by _concept_ and not by matching translations.

packages/esm-patient-chart-app/src/encounter-list/utils/helpers.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  // TODO: Need to get a better place for this
+  if (type === 'ageAtHivTest') {
+    return age(encounter.patient.birthDate, encounter.encounterDatetime);
+  }
+
+  if (secondaryConcept && typeof obs.value === 'object' && obs.value.names) {
+    const primaryValue =
+      obs.value.names.find((conceptName) => conceptName.conceptNameType === t('SHORT'))?.name || obs.value.name.name;
```
ibacher, line 156: `conceptNameType` does not need to be translated. It's metadata, not data.

packages/esm-patient-chart-app/src/index.ts
```diff
@@ -27,6 +27,7 @@ import startVisitActionButtonOnPatientSearch from './visit/start-visit-button.co
 import startVisitFormComponent from './visit/visit-form/visit-form.component';
 import stopVisitActionButtonComponent from './actions-buttons/stop-visit.component';
 import visitAttributeTagsComponent from './patient-banner-tags/visit-attribute-tags.component';
+import encounterListTabsComponent from './encounter-list/components/encounter-list-tabs.component';
```
ibacher, line 30:
```suggestion
```

packages/esm-patient-chart-app/src/encounter-list/utils/encounter-list-config-builder.ts
```diff
@@ -0,0 +1,137 @@
...
+  );
+};
+
+export const getTabColumns = (columnsDefinition: Array<ColumnDefinition>) => {
+  const columns: Array<FormattedColumn> = columnsDefinition.map((column: ColumnDefinition) => ({
+    key: column.id,
+    header: column.title,
+    concept: column.concept,
+    getValue: (encounter) =>
+      getColumnValue(encounter, column) as string | number | JSX.Element | NamedColumn[] | FormColumn[],
```
ibacher, line 110: Shouldn't this be the return type for `getColumnValue()` instead of a cast?

packages/esm-patient-chart-app/src/encounter-list/components/encounter-list-tabs.component.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+interface EncounterListTabsComponentProps {
+  patientUuid: string;
+}
+
+const EncounterListTabsComponent: React.FC<EncounterListTabsComponentProps> = ({ patientUuid }) => {
+  const config = useConfig();
+  const { tabDefinitions = [] } = config;
+  const { t } = useTranslation();
+  const tabsConfig = getMenuItemTabsConfiguration(tabDefinitions);
```
ibacher, line 19: So, in `getObsFromEncounter()`, we still have multiple references like this:

```ts
esmPatientChartSchema.falseConceptUuid._default
```

As I've pointed out a few times, this is hard-coding the values that are defined in the schema. To address this, we need to ensure that the `trueConceptUuid`, `falseConceptUuid`, and `otherConceptUuid` are actually resolved from the config. To do that, we need to pass in the config from somewhere, like here where we get it from `useConfig()`.

Basically, if we can pass the config as a property here, then in `getMenuItemTabsConfiguration()` we can associate the config with each tab definition, and then we should be able to use it where necessary.

ibacher (changes requested):

packages/esm-patient-chart-app/src/encounter-list/components/encounter-list-tabs.component.tsx
CynthiaKamau: Resulted to passing it as a prop in the function since they are not components.

ibacher (approved):

PR #2014, opened 2024-09-17
(fix) O3-3992: Add safeguards for service queue entry creation to visit form

makombe:
## Requirements

## Summary

This PR adds checks for `queueLocation`, `service`, and `priority` as a safeguard to ensure that all necessary information is available before attempting to add a patient to the service queue when starting or editing a visit using the Visit form. Even if the `showServiceQueueFields` config property is truthy, there might be cases where the user hasn't selected all required queue fields. This check ensures that we only proceed with queue entry creation when all necessary data has been provided.

## Related Issue

https://openmrs.atlassian.net/browse/O3-3992

### DIFF
--- a/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
@@ -823,10 +823,9 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
           </Stack>
         </div>
         <ButtonSet
-          className={classNames({
+          className={classNames(styles.buttonSet, {
             [styles.tablet]: isTablet,
             [styles.desktop]: !isTablet,
-            [styles.buttonSet]: true,
           })}
         >
           <Button className={styles.button} kind="secondary" onClick={closeWorkspace}>
--- a/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
@@ -464,7 +464,7 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
           .subscribe({
             next: (response) => {
               if (response.status === 201) {
-                if (config.showServiceQueueFields) {
+                if (config.showServiceQueueFields && queueLocation && service && priority) {
                   // retrieve values from the queue extension
                   setVisitUuid(response.data.uuid);
 
--- a/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.test.tsx
+++ b/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.test.tsx
@@ -568,7 +568,7 @@ describe('Visit form', () => {
     expect(mockSaveVisit).not.toHaveBeenCalled();
   });
 
-  it('should disable the submit button show an inline error notification if required visit attribute fields fail to load', async () => {
+  it('should disable the submit button and display an inline error notification if required visit attribute fields fail to load', async () => {
     mockUseVisitAttributeType.mockReturnValue({
       isLoading: false,
       error: new Error('failed to load'),


ojwanganto:
LGTM. Thanks, @makombe

ojwanganto (approved):

denniskigen:
Tangential, but we should make the logic transactional so that if, say, saving a queue entry fails, all the previous operations, like creating or updating the visit would be rolled back.

vasharma05:
Hey @makombe, can you please add screenshots displaying the different scenarios?
Thanks!

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -464,7 +464,7 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
           .subscribe({
             next: (response) => {
               if (response.status === 201) {
-                if (config.showServiceQueueFields) {
+                if (config.showServiceQueueFields && queueLocation && service && priority) {
```
denniskigen, line 467: Shouldn't we extend the validation schema as well so that if the user fails to select values for the given fields in the form, they get shown a validation error message and the form fails to submit? Is it mandatory to fill those fields when `showServiceQueueFields` is set to `true`?

denniskigen (approved):

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
denniskigen: That would entail extending the schema like so, for example: 

```tsx
  const visitFormSchema = useMemo(() => {
    const visitAttributes = (config.visitAttributeTypes ?? [])?.reduce(
      (acc, { uuid, required }) => ({
        ...acc,
        [uuid]: required
          ? z
              .string({
                required_error: t('fieldRequired', 'This field is required'),
              })
              .refine((value) => !!value, t('fieldRequired', 'This field is required'))
          : z.string().optional(),
      }),
      {},
    );

    const baseSchema = {
      visitStartDate: z.date().refine(
        (value) => {
          const today = dayjs();
          const startDate = dayjs(value);
          return displayVisitStopDateTimeFields ? true : startDate.isSameOrBefore(today, 'day');
        },
        t('invalidVisitStartDate', 'Start date needs to be on or before {{firstEncounterDatetime}}', {
          firstEncounterDatetime: formatDatetime(new Date()),
          interpolation: {
            escapeValue: false,
          },
        }),
      ),
      visitStartTime: z
        .string()
        .refine((value) => value.match(time12HourFormatRegex), t('invalidTimeFormat', 'Invalid time format')),
      visitStartTimeFormat: z.enum(['PM', 'AM']),
      visitStopDate: displayVisitStopDateTimeFields ? z.date() : z.date().optional(),
      visitStopTime: displayVisitStopDateTimeFields
        ? z
            .string()
            .refine((value) => value.match(time12HourFormatRegex), t('invalidTimeFormat', 'Invalid time format'))
        : z.string().optional(),
      visitStopTimeFormat: displayVisitStopDateTimeFields ? z.enum(['PM', 'AM']) : z.enum(['PM', 'AM']).optional(),
      programType: z.string().optional(),
      visitType: z.string().refine((value) => !!value, t('visitTypeRequired', 'Visit type is required')),
      visitLocation: z.object({
        display: z.string(),
        uuid: z.string(),
      }),
      visitAttributes: z.object(visitAttributes),
    };

    return z.object(config.showServiceQueueFields ? { ...baseSchema, ...serviceQueueFieldsSchema } : baseSchema);
  }, [
    config.visitAttributeTypes,
    config.showServiceQueueFields,
    t,
    displayVisitStopDateTimeFields,
    serviceQueueFieldsSchema,
  ]);
```

And then we could pass the schema down into the slot as a piece of state:

```tsx
{config.showServiceQueueFields && (
              <section>
                <div className={styles.sectionTitle}></div>
                  <ExtensionSlot
                    name="visit-form-queue-slot"
                    state={{ setFormFields: setVisitFormFields, schema: serviceQueueFieldsSchema }}
                  />
                </div>
              </section>
            )}
```

And then we'd need to modify the [VisitFormQueueFields component](https://github.com/openmrs/openmrs-esm-patient-management/blob/3a0f42beaf8972d1d060473d8bf888f829ca789f/packages/esm-service-queues-app/src/patient-search/visit-form-queue-fields/visit-form-queue-fields.component.tsx) to leverage the RHF and the zod schema prop.

denniskigen: All of this is not strictly necessary for now, though, which is why I proposed addressing it in a subsequent ticket.

makombe: Thanks @denniskigen I agree we can do this in a follow up PR. @ojwanganto  Whenever queue fields are configured to show on visit form, would it be wise to make it mandatory for them to fill the queue fields or there are exceptional cases?

ojwanganto: I may not be objective in this. Let's engage the BAs further for a correct and practical position

donaldkibet: We should rethink form extensibility to avoid hacky solutions like those we've used for billing & queues. Adding features like skip logic and the ability to hide/show fields would help make the forms more flexible and extandable for various workflows. Thoughts?

cc: @denniskigen @ibacher @vasharma05

donaldkibet (approved):
Thanks @makombe, we should also think of validating the billing section fields

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
ibacher: I think, ultimately, we should redesign these forms to actually use the form engine. In fact, I thought there was a way of doing that? Fundamentally, hard-coded forms in an EMR are always going to be a pain-point.

PR #2015, opened 2024-09-17
(fix) Pass patient as a prop where-ever possible in the chart

ibacher:
## Requirements

## Summary

The bulk of the changes here focus on ensuring that we are using the `fhir.Patient` object from the Patient Chart state wherever possible, meaning that most components do not require a separate call to `usePatient()`. I've re-written a few hooks to adjust for this (i.e., `usePatientDeceasedStatus()` is basically just a hook that gets a boolean result based on the patient).

There are a few miscellaneous changes:

1. The chart header and side panel are now part of the "Patient Chart" component rather than the "root" component
2. I've added a white background to the panel view in the test results viewer

## Other

Unchanged for now are most workspaces and panels.

This removes about 7 requests when loading the chart summary and 1-per-row when loading the result viewer, so the chart should feel snappier.

I'm unsure about some of the action buttons functioning when the banner is rendered in the search results, since this uses dummied FHIR objects. However, I'm of the view that actions in the search results are bad UX and should be removed.


denniskigen (approved):
LGTM

packages/esm-patient-attachments-app/src/camera-media-uploader/capture-photo.component.tsx
```diff
@@ -58,7 +57,7 @@ const CapturePhoto: React.FC<CapturePhotoProps> = ({ initialState, onCapturePhot
           className={styles.editButton}
           kind="ghost"
           onClick={showCam}
-          renderIcon={(props) => <Edit {...props} />}
+          renderIcon={(props: Partial<Parameters<typeof EditIcon>[0]>) => <EditIcon {...props} />}
```
denniskigen, line 60: Comparing apples to apples, but we could maybe this signature could be simplified to:

```suggestion
          renderIcon={(props: React.ComponentProps<typeof EditIcon>) => <EditIcon {...props} />}
```

ibacher: Actually, no. That's way more explicit, even if not necessarily shorter. Let me fix that.

ibacher:
In retrospect, I probably should've done the icons in a separate PR, but here we hare.

denniskigen:
Let's see how this pans out in the refapp!

PR #2017, opened 2024-09-18
(feat) O3-3818: Extend clinical forms workspace 

usamaidrsk:
## Requirements

## Summary

This PR extends clinical forms workspace i.e(to the ward app). This it to eliminate where a user has to go to and fro from the patient chart to the ward app so as to fill forms. 

## Screenshots

html-forms
forms-list

## Related Issue

https://openmrs.atlassian.net/browse/O3-3818


usamaidrsk:
This PR depends on this PR https://github.com/openmrs/openmrs-esm-patient-chart/pull/2018

mogoodrich (commented):
Thanks @usamaidrsk !  Defer to others on this one.

chibongho (approved):
Looks good to me, but I think others with more knowledge of the workspace should review.

brandones:
Okay so it looks like the deeper problem this is trying to mitigate is that really we don't want workspaces themselves to have to say whether they are part of a sidebar family or not. That really should be defined by the workspace container. Then we will just be able to reuse the same workspaces from the patient chart in the ward app, without having to do this awkward duplicate definition thing.

I have filed [O3-4077](https://openmrs.atlassian.net/browse/O3-4077) to suggest a fix for the workspace system. CC @ibacher @denniskigen @chibongho @vasharma05 , or if someone wants to find me a contract to do it :)

packages/esm-patient-forms-app/src/index.ts
```diff
@@ -39,6 +39,10 @@ export const patientHtmlFormEntryWorkspace = getAsyncLifecycle(
 
 // t('clinicalForms', 'Clinical forms')
 export const clinicalFormsWorkspace = getAsyncLifecycle(() => import('./forms/forms-dashboard.workspace'), options);
+export const clinicalFormsWorkspaceExtension = getAsyncLifecycle(
+  () => import('./forms/forms-dashboard.workspace'),
+  options,
+);
```
brandones, line 45: This is a duplicate declaration; it doesn't add anything. The code would be easier to understand if the extension registration simply referred to the `clincialFormsWorkspace` object. Then it would be immediately obvious we are doing something weird, rather than taking time to figure out we are doing something weird :)

PR #2019, opened 2024-09-18
(fix) O3-3979: Visit start date field not populating correctly when editing a visit

Vijaykv5:
## Requirements

## Summary

This PR fixes the issue where the visit start date field was not populating correctly when editing a visit. The root issue seems to be that while the visit date is correctly captured in the field control, the date value is not being set correctly in the field because the date format is not being set correctly. This PR explicitly sets the date format to `DD/MM/YYYY` in the `DatePicker` component. It also amends the existing E2E test to check that the date is correctly captured to prevent future regressions.

## Screenshots

Before 
Screenshot 2024-09-18 at 7 23 09 PM

After 
Screenshot 2024-09-18 at 7 24 26 PM

## Related Issue

[O3-3979](https://openmrs.atlassian.net/browse/O3-3979)

## Other

[O3-3979]: https://openmrs.atlassian.net/browse/O3-3979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/e2e/specs/edit-existing-visit.spec.ts
+++ b/e2e/specs/edit-existing-visit.spec.ts
@@ -27,7 +27,12 @@ test('Edit an existing visit', async ({ page }) => {
 
   await test.step('Then I should see the `Edit Visit` form launch in the workspace', async () => {
     await expect(chartPage.page.getByText(/visit start date and time/i)).toBeVisible();
-    await expect(chartPage.page.getByPlaceholder(/dd\/mm\/yyyy/i)).toBeVisible();
+    const datePickerInput = chartPage.page.getByPlaceholder(/dd\/mm\/yyyy/i);
+    await expect(datePickerInput).toBeVisible();
+    const dateValue = await datePickerInput.inputValue();
+    expect(dateValue).not.toBe('');
+    expect(dateValue).toMatch(/^\d{2}\/\d{2}\/\d{4}$/);
+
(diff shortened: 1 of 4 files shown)


Vijaykv5 (commented):
@denniskigen Removing the minDate and maxDate here not a perfect fix.

In the [VisitComponent](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx#L228), `maxVisitStartDatetime` and `minVisitStopDatetime` is getting null as there is not object present inside _visitToEdit.encounters_

denniskigen:
Thanks for giving this a shot, @Vijaykv5. I don’t think this is the right solution—the root issue seems to be that while the `visitStartDate` DatePicker has a value, it doesn’t match the expected format, causing the input to appear blank. One way to address this is by explicitly formatting the date, like so:

```tsx
value={value ? dayjs(value).format('DD/MM/YYYY') : null}
```

However, I’m unsure if we should be setting date formats explicitly this way.

denniskigen (approved):

denniskigen:
Thanks, @Vijaykv5!

PR #2021, opened 2024-09-19
(feat) Tweak Test results dashboard sticky header UI

CynthiaKamau:
## Requirements

## Summary

- Add background color on scrolling sticky results menu

## Screenshots

https://github.com/user-attachments/assets/5ad074d2-4bb8-4164-90c9-0c6bf12703d4

### DIFF
--- a/packages/esm-patient-labs-app/src/test-results/results-viewer/results-viewer.extension.tsx
+++ b/packages/esm-patient-labs-app/src/test-results/results-viewer/results-viewer.extension.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useContext, useState } from 'react';
+import React, { useCallback, useContext, useEffect, useRef, useState } from 'react';
 import classNames from 'classnames';
 import { type TFunction, useTranslation } from 'react-i18next';
 import { useParams } from 'react-router-dom';
@@ -15,6 +15,7 @@ import Trendline from '../trendline/trendline.component';
 import type { ConfigObject } from '../../config-schema';
 import styles from './results-viewer.scss';
 import { type viewOpts } from '../../types';
+import debounce from 'lodash-es/debounce';
 
 type panelOpts = 'tree' | 'panel';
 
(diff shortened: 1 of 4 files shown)


vasharma05:
Hi @CynthiaKamau!
The header is not touching the bottom of the primary navigation bar, as is visible from the video, as when you scroll, a part of the text is visible above the test results header, can you please fix that?
Thanks!
image

CynthiaKamau:
> Hi @CynthiaKamau! The header is not touching the bottom of the primary navigation bar, as is visible from the video, as when you scroll, a part of the text is visible above the test results header, can you please fix that? Thanks! image

on it

packages/esm-patient-labs-app/src/test-results/results-viewer/results-viewer.extension.tsx
```diff
@@ -65,6 +66,34 @@ const ResultsViewer: React.FC<ResultsViewerProps> = ({ patientUuid, basePath, lo
...
+    };
+
+    window.addEventListener('scroll', handleScroll);
+    window.addEventListener('resize', setInitialPosition);
+
+    return () => {
+      window.removeEventListener('scroll', handleScroll);
+      window.removeEventListener('resize', setInitialPosition);
+    };
+  }, []);
```
denniskigen, line 97: I think it'd be preferable to achieve this using the [IntersectionObserver API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API). It's more efficient as it's optimized by the browser and doesn't fire on every scroll. It also handles the visibility calculations internally, reducing the chance of errors. We could create an IntersectionObserver that triggers when the header is no longer visible in the viewport. We'd then update some state (something like `isHeaderVisible`) based on the intersection status. Here's how it could look:

```tsx
const [isHeaderVisible, setIsHeaderVisible] = useState(true);
  const headerRef = useRef<HTMLDivElement>(null);

  useEffect(() => {
    const observerCallback: IntersectionObserverCallback = (entries) => {
      const [entry] = entries;
      setIsHeaderVisible(entry.isIntersecting);
    };

    const observerOptions: IntersectionObserverInit = {
      threshold: 1,
      rootMargin: '-1px 0px 0px 0px',
    };

    const observer = new IntersectionObserver(observerCallback, observerOptions);

    const currentHeader = headerRef.current;
    if (currentHeader) {
      observer.observe(currentHeader);
    }

    return () => {
      if (currentHeader) {
        observer.unobserve(currentHeader);
      }
      observer.disconnect();
    };
  }, []);
```

And then we could replace `div.resultsHeader` with:

```tsx
<div ref={headerRef} className={styles.headerSentinel} />
<div className={classNames(styles.resultsHeader, { [styles.resultsHeaderScrolled]: !isHeaderVisible })}>
```

And have the following styling:

```tsx
.resultsHeader {
  padding: layout.$spacing-03 0;
  z-index: 10;
  background-color: $openmrs-background-grey;
  position: sticky;
  transition: border-bottom 0.2s ease-in-out;
}

.resultsHeaderScrolled {
  border-bottom: 1px solid colors.$gray-20;
  width: 100%;
  left: 0;
  right: 0;
}

.headerSentinel {
  position: absolute;
  top: 0;
  left: 0;
  right: 0;
  height: 1px;
  visibility: hidden;
}
```

denniskigen (commented):
Thanks, @CynthiaKamau. The approach looks OK to me, though I think it'd be preferable to use an [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) instead. I've explained why in my review suggestion.

denniskigen:
One other thing we might want to do is remove the top margin applied to the `Results viewer` dashboard container [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-chart-app/src/patient-chart/chart-review/dashboard-view.scss#L18). We could add an override based on the extension slot name like so:

```scss
// See https://zpl.io/lrlmdq0 for the Visits dashboard design
[data-extension-slot-name='patient-chart-encounters-dashboard-slot'],
// See https://zpl.io/RmMXrDE for the Test Results dashboard design
[data-extension-slot-name='patient-chart-test-results-dashboard-slot'] {
  margin: 0 layout.$spacing-05;
}
````

packages/esm-patient-labs-app/src/test-results/results-viewer/results-viewer.extension.tsx
CynthiaKamau: looking into this

denniskigen (approved):
LGTM. Thanks, @CynthiaKamau!

PR #2057, opened 2024-10-07
(fix) O3-4039: Enable backdating a prescription's start date

denniskigen:
## Requirements

## Summary

This PR fixes an issue where the Start date field of the drug order form was not being set correctly. The default value for the field was being set to the current date. Instead, the field should be set to the value of the medication order's `dateActivated` property (when editing an existing order) or the current date (when creating a new order).

This PR also refactors the MedicationsDetailsTable component to leverage the new MedicationsDisplaySection sub-component. The idea with this refactor is to clean up the logic for rendering the medications section and to make the component more modular and easier to maintain.

## Related Issue

https://openmrs.atlassian.net/browse/O3-4039


ojwanganto:
Thanks, sir @denniskigen. We got this request from the field and am glad you have provided a fix. I would request that we also fix it for lab orders. Most orders done retrospectively pick the current date and it becomes an issue for sensitive tests i.e. viral load and CD4.

denniskigen:
Set to draft because this doesn't work as expected as currently implemented. The backend won't let you set a `dateActivated` value older than the current encounter's start date. So just passing a start date from the past won't work:
 
CleanShot 2024-10-07 at 11  30 37@2x

denniskigen:
Any pointers, @ibacher?

ibacher:
I don't think this issue is fixable with the order basket working the way that it does, i.e., creating an "Orders" encounter and using that. Basically, this ticket _requires_ some kind of RDE functionality to pre-exist.

The "minimal" thing to do here would be to move the start time of the orders encounter, but that cannot be moved before the start of the current visit and could potentially raise other issues if there are orders already attached to the encounter.

Instead of trying to fix things piecemeal like this, my vote is we make "support for RDE" into a community priority and start on that.

denniskigen:
Closing in favour of continuing the discussion in the ticket.

PR #2059, opened 2024-10-09
(fix) Fix `Last completed date` in forms dashboard

denniskigen:
## Requirements

## Summary

This PR fixes an issue where the `Last completed date` in the forms dashboard was not being displayed correctly. The logic that determines the last completed date was not correctly sorting the encounters by date, and was not using the most recent encounter date. This PR fixes these issues and ensures that the last completed date is the most recent encounter date for each form.

## Screenshots

https://github.com/user-attachments/assets/b0170024-f1fc-4f0f-86a3-a81a3cf1c97c

## Related Issue

https://openmrs.atlassian.net/browse/O3-4010

### DIFF
--- a/packages/esm-patient-forms-app/src/hooks/use-forms.ts
+++ b/packages/esm-patient-forms-app/src/hooks/use-forms.ts
@@ -8,9 +8,9 @@ import {
   userHasAccess,
   useSession,
 } from '@openmrs/esm-framework';
-import { type ListResponse, type Form, type EncounterWithFormRef, type CompletedFormInfo } from '../types';
+import type { ConfigObject } from '../config-schema';
+import type { ListResponse, Form, EncounterWithFormRef, CompletedFormInfo } from '../types';
 import { customEncounterRepresentation, formEncounterUrl, formEncounterUrlPoc } from '../constants';
-import { type ConfigObject } from '../config-schema';
 import { isValidOfflineFormEncounter } from '../offline-forms/offline-form-helpers';
 
 export function useFormEncounters(cachedOfflineFormsOnly = false, patientUuid: string = '') {
@@ -94,7 +94,8 @@ export function useForms(
   } else {
     formsToDisplay?.sort(
       (formInfo1, formInfo2) =>
-        (formInfo1.lastCompleted ?? MINIMUM_DATE).getDate() - (formInfo2.lastCompleted ?? MINIMUM_DATE).getDate(),
+        (formInfo1.lastCompletedDate ?? MINIMUM_DATE).getDate() -
+        (formInfo2.lastCompletedDate ?? MINIMUM_DATE).getDate(),
     );
   }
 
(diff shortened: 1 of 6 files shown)


denniskigen:
FYI @VeronicaMuthee @fanderson

packages/esm-patient-forms-app/src/types/index.ts
```diff
@@ -73,7 +73,7 @@ export interface ListResponse<T> {
 export interface CompletedFormInfo {
   form: Form;
   associatedEncounters: Array<EncounterWithFormRef>;
-  lastCompleted?: Date;
+  lastCompletedDate?: Date;
```
denniskigen, line 76: I've renamed this value to `lastCompletedDate` for clarity.

packages/esm-patient-forms-app/src/hooks/use-forms.ts
```diff
@@ -113,13 +114,15 @@ function mapToFormCompletedInfo(
 ): Array<CompletedFormInfo> {
   return allForms.map((form) => {
     const associatedEncounters = encounters.filter((encounter) => encounter.form?.uuid === form?.uuid);
-    const lastCompleted =
-      associatedEncounters.length > 0 ? new Date(associatedEncounters?.[0].encounterDatetime) : undefined;
```
denniskigen, line 117: This function determines the data of the most recent matching encounter by picking the `encounterDatetime` of the first encounter and creates a new Date object from it. If there are no associated encounters, it sets `lastCompleted` to undefined. 

This implementation is not accurate. The code assumes that `associatedEncounters` is already sorted with the most recent encounter first. However, there’s no guarantee of this sorting in the code. Second, it only checks the first encounter in the array `associatedEncounters?.[0]`, which may not be the most recent encounter if the array is not sorted. 

I've modified the code to make sure we’re getting the most recent date from `associatedEncounters`. We’d then create a new Date object from the most recent encounter's date, guaranteeing that that would be the last completed date.

VeronicaMuthee:
Thanks for the fix @denniskigen and will retest once the PR is merged.

ibacher:
This suggests to me two tickets for the backend:

1. When the list of encounters is being restricted to a date range, the backend should support sorting those encounters by date, probably defaulting to ascending order.
2. We could create a specialized endpoint to return the list of forms with the last completed date. More generally, I think an endpoint for forms that could take a "patient" and maybe "visit" would likely be generally useful for more targeted form lists.

mogoodrich:
> This suggests to me two tickets for the backend:
> 
> 1. When the list of encounters is being restricted to a date range, the backend should support sorting those encounters by date, probably defaulting to ascending order.

Agree re: this... though we should first confirm such an endpoint doesn't already exist, it's possible it does and whoever first build this didn't realize it.

denniskigen:
Will you take care of filing those, @ibacher? I'll proceed to merge this in. Thanks for the review.

PR #2062, opened 2024-10-15
(feat) O3-4087: Enable input for set-based lab tests on lab results form and handle partial set results

donaldkibet:
## Requirements

## Summary
See https://openmrs.atlassian.net/browse/O3-4087

In this PR, I have updated the logic for generating lab observation (obs) values:

- **Non-set lab order concepts**: If the lab order concept is not a set, we generate the payload. If no values are present for the concept, we return an empty array.

- **Set member lab order concepts**: For concepts that are part of a set, we only return observation payloads for set members that have values. This ensures that incomplete results (missing values for some set members) do not affect the generation process.

## Screenshots
Kapture 2024-10-15 at 11 46 28

### DIFF
--- a/packages/esm-patient-orders-app/src/lab-results/lab-results.resource.ts
+++ b/packages/esm-patient-orders-app/src/lab-results/lab-results.resource.ts
@@ -2,6 +2,7 @@ import useSWR from 'swr';
 import { type OpenmrsResource, openmrsFetch, restBaseUrl, type FetchResponse } from '@openmrs/esm-framework';
 import { type Observation, type Encounter } from '../types/encounter';
 import { type OrderDiscontinuationPayload } from '../types/order';
+import { type Order } from '@openmrs/esm-patient-common-lib';
 
 const labEncounterRepresentation =
   'custom:(uuid,encounterDatetime,encounterType,location:(uuid,name),' +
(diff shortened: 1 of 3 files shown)


packages/esm-patient-orders-app/src/lab-results/lab-results.resource.ts
```diff
@@ -157,3 +158,69 @@ export async function updateOrderResult(
   }
   throw new Error('Failed to update order');
 }
+
+export function createObservationPayload(concept: LabOrderConcept, order: Order, values: any) {
```
ibacher, line 162: I'm always suspicious of explicitly using `any`, especially when the code here implies a rather specific shape, namely that `values` is that values is something like `Record<string, unknown>`, which would be much better to use.

We should likewise type the sub functions here as this helps ensure the correctness of the code.

packages/esm-patient-orders-app/src/lab-results/lab-results.resource.ts
```diff
@@ -157,3 +158,69 @@ export async function updateOrderResult(
...
+      .filter((member) => member?.value !== null && member?.value !== undefined);
+
+    if (groupMembers.length === 0) {
+      return { obs: [] };
+    }
+
+    return { obs: [createObservation(order, groupMembers)] };
+  }
+
+  if (!concept.set && concept.setMembers.length === 0) {
```
ibacher, line 175: Why not just use `else` or `else if`?

packages/esm-patient-orders-app/src/lab-results/lab-results.resource.ts
```diff
@@ -157,3 +158,69 @@ export async function updateOrderResult(
   }
   throw new Error('Failed to update order');
 }
+
+export function createObservationPayload(concept: LabOrderConcept, order: Order, values: any) {
+  if (concept.set && concept.setMembers.length > 0) {
+    const groupMembers = concept.setMembers
+      .map((member) => createGroupMember(member, order, values))
+      .filter((member) => member?.value !== null && member?.value !== undefined);
```
ibacher, line 166:
```suggestion
      .filter((member) => member !== null && member.value !== null && member.value !== undefined);
```

packages/esm-patient-orders-app/src/lab-results/lab-results.resource.ts
```diff
@@ -157,3 +158,69 @@ export async function updateOrderResult(
...
+    value: value,
+    status: 'FINAL',
+    order: { uuid: order.uuid },
+  };
+}
+
+function createObservation(order, groupMembers = null, value = null) {
+  return {
+    concept: { uuid: order.concept.uuid },
+    status: 'FINAL',
```
ibacher, line 202: Ideally, eventually, this would be something the user can define.

packages/esm-patient-orders-app/src/lab-results/lab-results.resource.ts
```diff
@@ -157,3 +158,69 @@ export async function updateOrderResult(
...
+
+function getValue(concept, values) {
+  const { datatype, uuid } = concept;
+  const value = values[uuid];
+
+  if (value === null || value === undefined) {
+    return null;
+  }
+
+  if (['Numeric', 'Text'].includes(datatype.display)) {
```
ibacher, line 217: This is actually a case where using `display` is incorrect. `display` is a user-facing property because its localizable, but because its localizable, we cannot rely on it's value being English. In this case, the "correct" value is (for historical reasons) `hl7Abbreviation`, but `name` is less likely to be wrong.

ibacher (changes requested):
Thanks! Nice work and once again, only some nitpicking on my part.

packages/esm-patient-orders-app/src/lab-results/lab-results.resource.ts
ibacher: `hl7Abbreviation` is something like:

|Datatype|Code|
|---|---|
|Coded|CWE|
|Text|ST|
|Numeric|NM|

packages/esm-patient-orders-app/src/lab-results/lab-results.resource.ts
```diff
@@ -158,3 +159,71 @@ export async function updateOrderResult(
...
+
+  if (value === null || value === undefined) {
+    return null;
+  }
+
+  // hl7Abbreviation is NM for Numeric and ST for Text
+  if (['NM', 'ST'].includes(datatype.hl7Abbreviation)) {
+    return value;
+  }
+
```
denniskigen, line 223: nit:

```suggestion
  // hl7Abbreviation is CWE for Coded with exceptions datatype
```

denniskigen (approved):
Great work, @donaldkibet!

PR #2073, opened 2024-10-24
(refactor) Switch visit requests to use promises instead of observables

denniskigen:
## Requirements

## Summary

Related to and dependent on https://github.com/openmrs/openmrs-esm-core/pull/1184.

Ports over requests to the Visit API to use promise-based fetch requests instead of observables. Promises are generally simpler to work with than observables, and this should make the code easier to understand. This change also makes the code more consistent with other requests the frontend makes to the REST API.

I've also amended the `UpdateVisitPayload` type annotation to a partial of `NewVisitPayload`, as not all fields are required when updating a visit. We should only really need to provide the stop datetime when ending a visit, for example.

### DIFF
--- a/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
@@ -474,122 +474,120 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
           ? updateVisit(visitToEdit?.uuid, payload, abortController)
           : saveVisit(payload, abortController)
         )
-          .pipe(first())
-          .subscribe({
-            next: (response) => {
-              if (config.showServiceQueueFields && queueLocation && service && priority) {
-                // retrieve values from the queue extension
-                setVisitUuid(response.data.uuid);
+          .then((response) => {
+            if (config.showServiceQueueFields && queueLocation && service && priority) {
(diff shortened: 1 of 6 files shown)


packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -474,122 +474,120 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
...
+                  });
+                },
+              );
+
+              if (config.showUpcomingAppointments && upcomingAppointment) {
+                updateAppointmentStatus('CheckedIn', upcomingAppointment.uuid, abortController).then(
+                  () => {
                     mutateCurrentVisit();
                     mutateVisits();
                     mutateInfiniteVisits();
```
vasharma05, line 519: Can we make these 3 mutates outside of the `if` conditions since they are related to the core visit form itself?

vasharma05 (approved):

ibacher (approved):

NethmiRodrigo (approved):

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
ibacher: @vasharma05 Let's move that to it's own ticket.

PR #2080, opened 2024-10-29
(refactor) Remove the dependency of using `patientUuid` from the URL

vasharma05:
## Requirements

## Summary
This PR aims to refactor the code locations that retrieve the patient UUID from the URL of the patient chart. The primary motivation for this refactor is that the extensions, widgets, and workspaces utilized in the patient chart are no longer guaranteed to be present in the patient chart. For instance, the Patient chart summary workspaces in the IPD. To reuse these widgets/workspaces in the IPD’s patient summary workspace and potentially in other future locations, I have implemented a global Patient Chart Store, which must be accessible by all components in the patient chart. 

I also modified the `getPatientUuidFromUrl` function to get the patient UUID from the URL, but if not present it will look into the store.

I also removed the instances where `patientUuid` was being fetched from the `usePatient`, which should now be fetched from the `usePatientChartStore` hook. 

## Screenshots
None

## Related Issue
None

### DIFF
--- a/packages/esm-patient-common-lib/src/get-patient-uuid-from-url.ts
+++ b/packages/esm-patient-common-lib/src/get-patient-uuid-from-url.ts
@@ -1,4 +1,7 @@
+import { getPatientChartStore } from './store/patient-chart-store';
+
 export function getPatientUuidFromUrl(): string {
   const match = /\/patient\/([a-zA-Z0-9\-]+)\/?/.exec(location.pathname);
-  return match && match[1];
+  const patientUuidFromUrl = match && match[1];
+  return patientUuidFromUrl || getPatientChartStore().getState().patientUuid;
 }
--- a/packages/esm-patient-common-lib/src/get-patient-uuid-from-url.ts
+++ b/packages/esm-patient-common-lib/src/get-patient-uuid-from-url.ts
@@ -1,7 +1,11 @@
-import { getPatientUuidFromUrlOrStore } from './store/patient-chart-store';
+import { getPatientUuidFromStore } from './store/patient-chart-store';
 
 /**
- * @deprecated This function is now replaced with `getPatientUuidFromUrlOrStore`. This function will be removed in upcoming releases.
+ * @deprecated This function is now replaced with `getPatientUuidFromStore`. This function will be removed in upcoming releases.
  * @returns {string} patientUuid
  */
-export const getPatientUuidFromUrl = getPatientUuidFromUrlOrStore;
+export function getPatientUuidFromUrl(): string {
+  const match = /\/patient\/([a-zA-Z0-9\-]+)\/?/.exec(location.pathname);
+  const patientUuidFromUrl = match && match[1];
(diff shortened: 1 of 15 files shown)


packages/esm-patient-common-lib/src/get-patient-uuid-from-url.ts
```diff
@@ -1,4 +1,7 @@
+import { getPatientChartStore } from './store/patient-chart-store';
+
 export function getPatientUuidFromUrl(): string {
   const match = /\/patient\/([a-zA-Z0-9\-]+)\/?/.exec(location.pathname);
-  return match && match[1];
+  const patientUuidFromUrl = match && match[1];
+  return patientUuidFromUrl || getPatientChartStore().getState().patientUuid;
```
usamaidrsk, line 6: The function name `getPatientUuidFromUrl` suggests it only retrieves the UUID from the URL, but it actually has a fallback behavior that gets the UUID from the store state. I suggest either:

1. Rename the function to better reflect its complete behavior, e.g., `getActivePatientUuid` or `resolvePatientUuid`, since it resolves the patient UUID from either the URL or store
2. OR split this into two separate concerns if these use cases need to be handled differently

This would make the function's behavior more predictable from its name

packages/esm-patient-common-lib/src/store/patient-chart-store.ts
```diff
@@ -0,0 +1,19 @@
+import { createGlobalStore, useStore } from '@openmrs/esm-framework';
+
+export interface PatientChartStore {
+  patientUuid: string;
+}
+
+const patientChartStoreName = 'patient-chart-global-store';
+
+const patientChartStore = createGlobalStore<PatientChartStore>(patientChartStoreName, {
+  patientUuid: '',
```
usamaidrsk, line 10: Shouldn't this be `patientUuid: null`

usamaidrsk (commented):
Thanks @vasharma05 , generally looks great.
I just have a few comments.

packages/esm-patient-common-lib/src/get-patient-uuid-from-url.ts
vasharma05: Yes @usamaidrsk!
I agree with you on this, and am seeking @denniskigen's advice too. But do note that this is an API in the Patient Common Lib, and changing the function will require updating all the places this function is being used, and hence will become a breaking change, since this will also affect other implementations to update their code when released.

One thing is, we can deprecate this function and in the message we can ask them to update the API in their respective code. This way, it'll not be a breaking change.
Thanks!

packages/esm-patient-common-lib/src/store/patient-chart-store.ts
vasharma05: Right!

usamaidrsk (approved):
Thanks @vasharma05 LGTM.
Could you check why the E2E tests are failing

PR #2091, opened 2024-11-04
(feat) O3-4136: Move lab-results components From esm-patient-orders-app To esm-patient-tests-app

Twiineenock:
## Requirements

## Summary

The `esm-patient-orders-app` should be limited to handling generic order-management functionalities. All laboratory-specific components and features should be contained within the `esm-patient-tests-app` ([see](https://openmrs.atlassian.net/browse/O3-4136))

## Related Issue

[https://openmrs.atlassian.net/browse/O3-4136](https://openmrs.atlassian.net/browse/O3-4136)

### After the move!

[lab-results.webm](https://github.com/user-attachments/assets/88d1b9b4-3a87-4452-8089-01514052df59)


denniskigen (commented):
Good start, @Twiineenock. I’ve left some feedback.

packages/esm-patient-tests-app/src/lab-results/lab-results-form.test.tsx
```diff
@@ -8,17 +8,17 @@ import {
   type LabOrderConcept,
   updateOrderResult,
   type Datatype,
-} from './lab-results.resource';
+} from '@openmrs/esm-patient-common-lib';
 import LabResultsForm from './lab-results-form.component';
 import { type Order } from '@openmrs/esm-patient-common-lib';
-import { type Encounter } from '../types/encounter';
+import { type Encounter } from '@openmrs/esm-patient-common-lib';
```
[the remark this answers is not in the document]
Twiineenock: Thanks @denniskigen for the review!
consolidated the imports and updated the translations too.

packages/esm-patient-common-lib/src/lab-resource/types/encounter.ts
```diff
@@ -0,0 +1,65 @@
+import { type OpenmrsResource, type Visit } from '@openmrs/esm-framework';
+import { type Order } from '@openmrs/esm-patient-common-lib';
+
+export interface Encounter {
```
brandones, line 4: I'm inclined to say we should add these instead to [src/types/index.ts](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-common-lib/src/types/index.ts)

brandones:
Code looks good to me with the one caveat I mentioned. @ibacher could you just confirm that this work covers the extent of what you'd imagined with O3-4136? It really just moves the lab results entry workspace.

packages/esm-patient-common-lib/src/lab-resource/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './lab.resource';
+export * from './types/encounter';
+export * from './types/order';
```
ibacher, line 3: nit:

```suggestion
export * from './types/order';

```

packages/esm-patient-common-lib/src/lab-resource/types/order.ts
ibacher, line 1: See this is, to me, the thing that should not exist, a shared reference to "lab orders"

packages/esm-patient-common-lib/src/lab-resource/types/order.ts
```diff
@@ -0,0 +1,16 @@
+import { type Order } from '@openmrs/esm-patient-common-lib';
+
+export interface PatientMedicationFetchResponse {
```
ibacher, line 3: Why is _this_ type in `/lab-resource/types`. It seems like it definitely belongs somewhere else.

packages/esm-patient-common-lib/src/lab-resource/types/order.ts
```diff
@@ -0,0 +1,16 @@
...
+  results: Array<Order>;
+}
+
+export interface OrderDiscontinuationPayload {
+  previousOrder: string;
+  type: string;
+  action: string;
```
ibacher, line 10: I feel like order action is a restricted (enum) type we already have defined. Please re-use that.

packages/esm-patient-common-lib/src/lab-resource/types/order.ts
```diff
@@ -0,0 +1,16 @@
...
+  careSetting: string;
+  encounter: string;
+  patient: string;
+  concept: string;
+  orderer: { display: string; person: { display: string }; uuid: string };
```
ibacher, line 15: Is this the result we get back or the type for the message we send? Display strings should _never_ be sent to the backend under any circumstances.

packages/esm-patient-common-lib/src/lab-resource/types/order.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export interface PatientMedicationFetchResponse {
+  results: Array<Order>;
+}
+
+export interface OrderDiscontinuationPayload {
+  previousOrder: string;
+  type: string;
+  action: string;
+  careSetting: string;
```
ibacher, line 11: This, also, could be more tightly-bound.

packages/esm-patient-tests-app/src/lab-results/useLabResultsFormSchema.tsx
```diff
@@ -1,5 +1,5 @@
 import { z } from 'zod';
-import { type LabOrderConcept, useOrderConceptByUuid } from './lab-results.resource';
+import { type LabOrderConcept, useOrderConceptByUuid } from '@openmrs/esm-patient-common-lib';
```
ibacher, line 2: This is probably minor, but do we really _need_ a type for the concept for a lab order? That seems... decadent.

packages/esm-patient-orders-app/translations/en.json
```diff
@@ -18,15 +17,10 @@
...
   "endDate": "End date",
-  "enterTestResults": "Enter test results",
   "error": "Error",
   "errorCancellingOrder": "Error cancelling order",
-  "errorSavingLabResults": "Error saving lab results",
   "indication": "Indication",
   "launchOrderBasket": "Launch order basket",
-  "loading": "Loading",
-  "loadingInitialValues": "Loading initial values",
-  "loadingTestDetails": "Loading test details",
```
ibacher, line 29: In this case, where we're moving strings between packages, it is important that we retain the existing translations in non-English languages. Unfortunately, the only way I know to do that is some tedious copy-and-paste work, but this would be the rare instance were manually updating the translation files is for the best.

ibacher (changes requested):

packages/esm-patient-common-lib/src/lab-resource/types/order.ts
ibacher: It should hold the `uuid` of the current practitioner for the user placing the order.

brandones:
@Twiineenock is it clear to you how to move forward with this?

Twiineenock:
> @Twiineenock is it clear to you how to move forward with this?

@brandones , a lot was merged in since my first commit, I am making changes locally and pushing soon.

Twiineenock (commented):
Hi @ibacher and @brandones,  

What are your thoughts on setting up translations in the `esm-patient-common-lib`? I recently added a file with translations.

packages/esm-patient-common-lib/src/lab-resource/types/order.ts
brandones: @Twiineenock please address this feedback

brandones:
Hi @Twiineenock , I don't think it would make sense to put translations in `esm-patient-common-lib`, and we don't really have a way of doing that anyway.

Thank you for moving the translations over!

jwnasambu:
@Twiineenock Thanks for the great work! Would you happen to have some time to resolve the merge conflicts? If you're currently constrained, feel free to let me know happy to support where I can.

denniskigen:
Thanks for tackling this, @Twiineenock — moving lab-results concerns out of `esm-patient-orders-app` and into `esm-patient-tests-app` is the right direction and aligns with the ticket.

Two unresolved issues from the review make this hard to land as-is:

1. **Wrong destination for shared resources.** Lab-specific code (`lab.resource.ts`, `test-order.component.tsx`, `LabOrderConcept`) ended up in `esm-patient-common-lib` rather than `esm-patient-tests-app`. Common-lib is for cross-app shared utilities — lab domain logic belongs in the tests app directly.

2. **Non-English translations will be lost.** Moving keys between packages by only updating `en.json` drops all Transifex-translated strings for other locales. The translated strings need to be manually carried over when relocating keys across packages.

This PR is also now 6 months old and the codebase has moved on, so a rebase would be substantial. If you'd like to pick this up again, a fresh PR scoped to moving files directly into `esm-patient-tests-app` (not common-lib) and including a translation migration step would be the cleanest path forward.

Closing to avoid reviving outdated code — the original intent was solid.

PR #2094, opened 2024-11-06
(test) O3-4176: E2E test for invalid biometrics

kdaud:
## Requirements

## Related Issue

https://issues.openmrs.org/browse/O3-4176


e2e/specs/biometrics.spec.ts
```diff
@@ -66,6 +66,52 @@ test('Record biometrics', async ({ page }) => {
   });
 });
 
+test('Record abnormal biometrics', async ({ page }) => {
```
ibacher, line 69: These are not "abnormal" these are "invalid". Abnormal, in the context of vitals has a specific, clinically relevant meaning and it's best not to mess that up.

Also, weights over 250kg should not be invalid to enter. The heaviest person still living used to weight 610kg.  Vitals like `height`, `weight`, `BMI` and `MUAC`—where there is no reason that a measurement has a "natural" cap—should not have absolute maximums. This differs from values like O2 Sat, which has a natural maximum (it's a percentage).

jayasanka-sack: Yes, it makes sense to add values for Temperature, BP, heart rate, respiration rate, and SpO2 if we need to check for invalid values, since they have specific limits.

PR #2103, opened 2024-11-12
(feat) O3-4071: Improve Start Visit Form to support "Visit Locations"

mogoodrich:
(feat) O3-4071: Improve Start Visit Form to support "Visit Locations"
(feat) O3-4072: Improve Start Visit Form to Support Default Visit Type

## Requirements

## Screenshots
image

Example with restrictByVisitLocationTag set to true and emr.atFacilitiyVisitType specified... note that visit selector is hidden and only locations that are visit locations (KGH and Wellbody) appear in the location dropdown.

## Related Issue

https://openmrs.atlassian.net/browse/O3-4071
https://openmrs.atlassian.net/browse/O3-4072


packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -203,9 +206,8 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
       visitStartDate,
       visitStartTime: dayjs(visitStartDate).format('hh:mm'),
       visitStartTimeFormat: visitStartDate.getHours() >= 12 ? 'PM' : 'AM',
-
-      visitType: visitToEdit?.visitType?.uuid,
-      visitLocation: visitToEdit?.location ?? sessionLocation ?? {},
+      visitType: visitToEdit?.visitType?.uuid ?? emrConfiguration?.atFacilityVisitType?.uuid,
```
mogoodrich, line 209: Set the visit type to the "atFacilityVisitType" as defined by the EMR API module, if present.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -664,14 +672,6 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
   const minVisitStopDatetimeFallback = Date.parse(visitStartDate.toLocaleString());
   minVisitStopDatetime = minVisitStopDatetime || minVisitStopDatetimeFallback;
 
-  useEffect(() => {
```
mogoodrich, line 667: I'm no sure this was doing much of anything.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -769,38 +769,40 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
...
-                    {contentSwitcherIndex === 1 && <BaseVisitType visitTypes={allVisitTypes} />}
-                  </>
-                ) : (
-                  // Defaults to showing all possible visit types if recommended visits are not enabled
-                  <BaseVisitType visitTypes={allVisitTypes} />
-                )}
-              </div>
-            </section>
+            {/* Lists available visit types if no atFacilityVisitType enabled. The content switcher only gets shown when recommended visit types are enabled */}
+            {!emrConfiguration?.atFacilityVisitType && (
```
mogoodrich, line 773: Hide the visit type selector if atFacilityVisitType has been specified.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.test.tsx
```diff
@@ -126,8 +130,12 @@ jest.mock('../hooks/useVisitAttributeType', () => ({
   })),
 }));
 
-jest.mock('../hooks/useDefaultLocation', () => {
-  const requireActual = jest.requireActual('../hooks/useDefaultLocation');
+jest.mock('../hooks/useEmrConfiguration', () => ({
+  useEmrConfiguration: jest.fn(() => ({})),
+}));
```
mogoodrich, line 135: According to best practices, we shouldn't do this anymore because the "jest.mocked(useEmrConfiguration)" should handle this... but it doesn't, and in mocked objects in this file it appears to both use the "jest.mocked" pattern and then use the jest.mock('../hooks...')... if someone other than me can understand what is going on here, happy to fix.

mogoodrich:
And it looks like the build is failing because it's attempting to modify the yarn.lock file to set some of the OpenMRS versions back to next... I had to update to the latest version of ESM Core, did I screw something up?  @denniskigen  @ibacher ?  Thanks... (I can look into this more tomorrow as well)

ibacher:
Run `yarn` locally and commit the updated `yarn.lock` file.

mogoodrich:
> Run `yarn` locally and commit the updated `yarn.lock` file.

Ah right, thanks @ibacher , done!

packages/esm-patient-chart-app/src/visit/hooks/useDefaultVisitLocation.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Fetches the default visit location based on the location and the restrictByVisitLocationTag
+ * If restrictByVisitLocationTag is true, it fetches the nearest ancestor location that supports visits, otherwise it returns the passed-in location
+ *
+ * @param location
+ * @param restrictByVisitLocationTag
+ */
+export function useDefaultVisitLocation(location: Location, restrictByVisitLocationTag: boolean) {
+  let url = `${restBaseUrl}/emrapi/locationThatSupportsVisits?location=${location?.uuid}`;
```
chibongho, line 13: I think we need to check the feature flag for whether emrapi is installed

packages/esm-patient-chart-app/src/config-schema.ts
```diff
@@ -61,6 +60,12 @@ export const esmPatientChartSchema = {
     _description: 'The UUID of the visit type to be used for the automatically created offline visits.',
     _default: 'a22733fa-3501-4020-a520-da024eeff088',
   },
+  restrictByVisitLocationTag: {
+    _type: Type.Boolean,
+    _description:
+      'On the start visit form, whether to restrict the visit location to locations with the Visit Location tag',
+    _default: false,
```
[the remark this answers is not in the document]
chibongho: Maybe we should mentioned that this requires the EMR API to be installed on the backend? (And maybe even provide a warning if it's set to true without it installed)

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.test.tsx
chibongho: We don't need [partial mocks](https://o3-docs.openmrs.org/docs/frontend-modules/unit-and-integration-testing#you-should-not-need-to-use-partial-mocks) for hooks from core anymore, but we still need them for hooks defined within an app. 

For core hooks, I think the magic comes from these places: [jest.config.js](https://github.com/openmrs/openmrs-esm-core/blob/f1d8002b89cfe042580e6825b4b4d7def3ba1628/packages/apps/esm-devtools-app/jest.config.js), [esm-core's mock.tsx](https://github.com/openmrs/openmrs-esm-core/blob/8d4612d384f000990303365e0d8575ebf382bb4f/packages/framework/esm-framework/mock.tsx) and [esm-react-utils's mock.tsx](https://github.com/openmrs/openmrs-esm-core/blob/0120ab0a384d7a2adcd90aec88dc224802af6ab2/packages/framework/esm-react-utils/mock.tsx). It'd be nice if we have something similar for individual apps, or at least a standardized file to add partial mocks to so we don't have to do that in every *test.tsx file

packages/esm-patient-chart-app/src/visit/hooks/useDefaultVisitLocation.tsx
mogoodrich: Good point... I was going this would be "buyer beware" and the implementer would need to know that the EMR API module needed to be installed, but no reason *not* to add this check as well.

packages/esm-patient-chart-app/src/config-schema.ts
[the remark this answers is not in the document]
mogoodrich: And yes, good point, we should definitely note that, will add.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.test.tsx
mogoodrich: Thanks!  So just to confirm, the takeaway here is what I did is fine, correct?

chibongho: Yeah, I think so.

chibongho (approved):
LGTM!

mogoodrich:
Merging this in, thanks @chibongho , fyi @ibacher  @denniskigen ... (this should theorectically, I believe, have no affect on dev3 until I enable the "restrictByVisitLocationTag", but I plan to do that in a separate PR once I confirm things are still working without it)

PR #2104, opened 2024-11-13
(feat) O3-4190 Add ability to filter lab orders by type and date

CynthiaKamau:
## Requirements

## Summary

Add ability to filter lab orders by type and date

## Screenshots

https://github.com/user-attachments/assets/35504a61-9360-4784-8d08-f0079c39fad6

### DIFF
--- a/packages/esm-patient-orders-app/src/components/order-details-table.scss
+++ b/packages/esm-patient-orders-app/src/components/order-details-table.scss
@@ -192,21 +192,3 @@
   display: flex;
   align-items: center;
 }
-
-.fullWidthDatePickerContainer {
-  :global .cds--date-picker {
-    width: 40%;
-
-    :global .cds--date-picker-container {
-      width: 100%;
-
-      :global .cds--date-picker-input__wrapper {
-        width: 100%;
-
-        :global .cds--date-picker__input {
-          width: 100%;
-        }
-      }
-    }
-  }
-}
--- a/packages/esm-patient-orders-app/src/components/order-details-table.scss
+++ b/packages/esm-patient-orders-app/src/components/order-details-table.scss
@@ -192,3 +192,8 @@
   display: flex;
   align-items: center;
 }
+
+.rangeLabel {
+  @extend .helperText;
+  margin-right: layout.$spacing-02;
+}
--- a/packages/esm-patient-orders-app/src/components/order-details-table.scss
+++ b/packages/esm-patient-orders-app/src/components/order-details-table.scss
@@ -187,3 +187,26 @@
     page-break-inside: avoid;
   }
 }
+
+.filterContainer {
+  display: flex;
+  align-items: center;
+}
+
+.fullWidthDatePickerContainer {
+  :global .cds--date-picker {
+    width: 40%;
+
+    :global .cds--date-picker-container {
+      width: 100%;
+
+      :global .cds--date-picker-input__wrapper {
+        width: 100%;
+
+        :global .cds--date-picker__input {
+          width: 100%;
+        }
+      }
+    }
+  }
+}
--- a/packages/esm-patient-orders-app/src/components/orders-details-table.component.tsx
+++ b/packages/esm-patient-orders-app/src/components/orders-details-table.component.tsx
@@ -309,15 +309,15 @@ const OrderDetailsTable: React.FC<OrderDetailsProps> = ({ patientUuid, showAddBu
 
   const handleDateFilterChange = ([startDate, endDate]) => {
     if (startDate) {
-      setSelectedFromDate(startDate);
+      setSelectedFromDate(startDate.toISOString());
       if (selectedToDate && startDate && selectedToDate < startDate) {
-        setSelectedToDate(startDate);
+        setSelectedToDate(startDate.toISOString());
       }
     }
     if (endDate) {
-      setSelectedToDate(endDate);
+      setSelectedToDate(endDate.toISOString());
       if (selectedFromDate && endDate && selectedFromDate > endDate) {
-        setSelectedFromDate(endDate);
+        setSelectedFromDate(endDate.toISOString());
       }
     }
   };
@@ -352,8 +352,8 @@ const OrderDetailsTable: React.FC<OrderDetailsProps> = ({ patientUuid, showAddBu
             handleDateFilterChange([startDate, endDate]);
           }}
         >
-          <DatePickerInput id="startDatePickerInput" placeholder="dd/mm/yyyy" value={formatDate(selectedFromDate)} />
-          <DatePickerInput id="endDatePickerInput" placeholder="dd/mm/yyyy" value={formatDate(selectedToDate)} />
+          <DatePickerInput id="startDatePickerInput" placeholder="dd/mm/yyyy" />
+          <DatePickerInput id="endDatePickerInput" placeholder="dd/mm/yyyy" />
         </DatePicker>
       </div>
 
--- a/packages/esm-patient-orders-app/src/components/orders-details-table.component.tsx
+++ b/packages/esm-patient-orders-app/src/components/orders-details-table.component.tsx
@@ -24,7 +24,9 @@ import {
   TableRow,
   TableToolbarContent,
   Tile,
- DatePicker , DatePickerInput } from '@carbon/react';
+  DatePicker,
+  DatePickerInput,
+} from '@carbon/react';
 import {
   CardHeader,
   EmptyState,
@@ -307,15 +309,17 @@ const OrderDetailsTable: React.FC<OrderDetailsProps> = ({ patientUuid, showAddBu
 
   const handleDateFilterChange = ([startDate, endDate]) => {
     if (startDate) {
-      setSelectedFromDate(startDate.toISOString());
-      if (selectedToDate && startDate && selectedToDate < startDate) {
-        setSelectedToDate(startDate.toISOString());
+      const isoStartDate = startDate.toISOString();
+      setSelectedFromDate(isoStartDate);
+      if (selectedToDate && selectedToDate < startDate) {
+        setSelectedToDate(isoStartDate);
       }
     }
     if (endDate) {
-      setSelectedToDate(endDate.toISOString());
-      if (selectedFromDate && endDate && selectedFromDate > endDate) {
-        setSelectedFromDate(endDate.toISOString());
+      const isoEndDate = endDate.toISOString();
+      setSelectedToDate(isoEndDate);
+      if (selectedFromDate && selectedFromDate > endDate) {
+        setSelectedFromDate(isoEndDate);
       }
     }
   };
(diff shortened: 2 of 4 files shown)


donaldkibet:
Thanks @CynthiaKamau IMO having the backend `activatedOnOrBeforeDate` and `activatedOnOrAfterDate` would be better here.

ojwanganto:
Thanks, @CynthiaKamau. Could this be related to the print feature?

ojwanganto:
> Thanks @CynthiaKamau IMO having the backend `activatedOnOrBeforeDate` and `activatedOnOrAfterDate` would be better here.

This can save the filtering on the frontend

packages/esm-patient-orders-app/src/components/order-details-table.scss
```diff
@@ -187,3 +187,26 @@
...
+    width: 40%;
+
+    :global .cds--date-picker-container {
+      width: 100%;
+
+      :global .cds--date-picker-input__wrapper {
+        width: 100%;
+
+        :global .cds--date-picker__input {
+          width: 100%;
```
donaldkibet, line 207: Having this nesting for look not correct.

donaldkibet (approved):
Thanks @CynthiaKamau LGTM

packages/esm-patient-orders-app/src/components/orders-details-table.component.tsx
```diff
@@ -303,25 +305,53 @@ const OrderDetailsTable: React.FC<OrderDetailsProps> = ({ patientUuid, showAddBu
     [orderTypes, t],
   );
 
+  const handleDateFilterChange = ([startDate, endDate]) => {
+    if (startDate) {
+      setSelectedFromDate(startDate.toISOString());
+      if (selectedToDate && startDate && selectedToDate < startDate) {
```
denniskigen, line 311:
```suggestion
      if (selectedToDate && selectedToDate < startDate) {
```

Isn't the outer check sufficient?

packages/esm-patient-orders-app/src/components/orders-details-table.component.tsx
```diff
@@ -303,25 +305,53 @@ const OrderDetailsTable: React.FC<OrderDetailsProps> = ({ patientUuid, showAddBu
...
+        setSelectedToDate(startDate.toISOString());
+      }
+    }
+    if (endDate) {
+      setSelectedToDate(endDate.toISOString());
+      if (selectedFromDate && endDate && selectedFromDate > endDate) {
```
denniskigen, line 317: Ditto

packages/esm-patient-orders-app/src/components/orders-details-table.component.tsx
```diff
@@ -303,25 +305,53 @@ const OrderDetailsTable: React.FC<OrderDetailsProps> = ({ patientUuid, showAddBu
...
+                return;
+              }
+              setSelectedOrderTypeUuid(e.selectedItem.uuid);
+            }}
+            selectedItem={orderTypes?.find((x) => x.uuid === selectedOrderTypeUuid)}
+            titleText={t('selectOrderType', 'Select order type') + ':'}
+            type="inline"
+          />
+        </div>
+        <p className={styles.helperText}>{t('dateRange', 'Date range')} :</p>
```
denniskigen, line 343:
```suggestion
        <p className={styles.helperText}>{t('dateRange', 'Date range')}:{" "}</p>
```

packages/esm-patient-orders-app/src/components/orders-details-table.component.tsx
```diff
@@ -303,25 +305,53 @@ const OrderDetailsTable: React.FC<OrderDetailsProps> = ({ patientUuid, showAddBu
     [orderTypes, t],
   );
 
+  const handleDateFilterChange = ([startDate, endDate]) => {
+    if (startDate) {
+      setSelectedFromDate(startDate.toISOString());
+      if (selectedToDate && startDate && selectedToDate < startDate) {
+        setSelectedToDate(startDate.toISOString());
+      }
+    }
```
denniskigen, line 314:
```suggestion
    if (startDate) {
    	const isoStartDate = startDate.toISOString();
      setSelectedFromDate(isoStartDate);
      if (selectedToDate && selectedToDate < startDate) {
        setSelectedToDate(isoStartDate);
      }
    }
```

The outer `startDate` check in the if condition should be sufficient.

packages/esm-patient-orders-app/src/components/orders-details-table.component.tsx
```diff
@@ -303,25 +305,53 @@ const OrderDetailsTable: React.FC<OrderDetailsProps> = ({ patientUuid, showAddBu
...
+    if (endDate) {
+      setSelectedToDate(endDate.toISOString());
+      if (selectedFromDate && endDate && selectedFromDate > endDate) {
+        setSelectedFromDate(endDate.toISOString());
+      }
+    }
+  };
```
denniskigen, line 321:
```suggestion
    if (endDate) {
      const isoEndDate = endDate.toISOString();
      setSelectedToDate(isoEndDate);
      if (selectedFromDate && selectedFromDate > endDate) {
        setSelectedFromDate(isoEndDate);
      }
    }
  };
```

packages/esm-patient-orders-app/src/components/orders-details-table.component.tsx
denniskigen, line 343:
```suggestion
        <span className={styles.rangeLabel}>{t('dateRange', 'Date range')} :</span>
```

`<span>` or some other inline element is more ideal here as the element doesn't span the full width (because of the range datepickers). Also, this element will need some horizontal margin to the datepickers, so we could do something like:

		.rangeLabel {
		  @extend .helperText;
		  margin-right: layout.$spacing-02;
		}

denniskigen:
Thanks for working on this, @CynthiaKamau. One challenge we'll need to address is that the date comparison doesn't factor in the time. The `Datepicker` component defaults to 12:00 AM, whereas the order activation date from the backend includes the time. This discrepancy can lead to unexpected results when filtering orders by activation date. You can see this in the screenshot below where the selected date range includes orders that were activated earlier than the selected date:

CleanShot 2024-11-18 at 2  03 33@2x

CynthiaKamau:
> Thanks for working on this, @CynthiaKamau. One challenge we'll need to address is that the date comparison doesn't factor in the time. The `Datepicker` component defaults to 12:00 AM, whereas the order activation date from the backend includes the time. This discrepancy can lead to unexpected results when filtering orders by activation date. You can see this in the screenshot below where the selected date range includes orders that were activated earlier than the selected date:
> 
> CleanShot 2024-11-18 at 2  03 33@2x

Would including the time help ?

denniskigen:
It would - I'm just concerned about how clunky the UX would be. FWIW, we're trying to tackle the same problem [here](https://openmrs.atlassian.net/browse/O3-3983), so maybe we shouldn't try to solve it in this particular PR. Something worth thinking about though.

packages/esm-patient-orders-app/src/components/orders-details-table.component.tsx
```diff
@@ -303,25 +307,55 @@ const OrderDetailsTable: React.FC<OrderDetailsProps> = ({ patientUuid, showAddBu
...
+                return;
+              }
+              setSelectedOrderTypeUuid(e.selectedItem.uuid);
+            }}
+            selectedItem={orderTypes?.find((x) => x.uuid === selectedOrderTypeUuid)}
+            titleText={t('selectOrderType', 'Select order type') + ':'}
+            type="inline"
+          />
+        </div>
+        <span className={styles.rangeLabel}>{t('dateRange', 'Date range')} :</span>
```
denniskigen, line 347:
```suggestion
        <span className={styles.rangeLabel}>{t('dateRange', 'Date range')}:</span>
```

CynthiaKamau:
> It would - I'm just concerned about how clunky the UX would be. FWIW, we're trying to tackle the same problem [here](https://openmrs.atlassian.net/browse/O3-3983), so maybe we shouldn't try to solve it in this particular PR. Something worth thinking about though.

Would be great to also hear what @ojwanganto and @donaldkibet think since they have a specific use case

CynthiaKamau:
> Thanks for working on this, @CynthiaKamau. One challenge we'll need to address is that the date comparison doesn't factor in the time. The `Datepicker` component defaults to 12:00 AM, whereas the order activation date from the backend includes the time. This discrepancy can lead to unexpected results when filtering orders by activation date. You can see this in the screenshot below where the selected date range includes orders that were activated earlier than the selected date:
> 
> CleanShot 2024-11-18 at 2  03 33@2x

What if the start date is set to 12AM and end day set to 11:59 , end of day?

denniskigen:
Any suggestions, @ibacher, @mogoodrich?

ojwanganto:
> > It would - I'm just concerned about how clunky the UX would be. FWIW, we're trying to tackle the same problem [here](https://openmrs.atlassian.net/browse/O3-3983), so maybe we shouldn't try to solve it in this particular PR. Something worth thinking about though.
> 
> Would be great to also hear what @ojwanganto and @donaldkibet think since they have a specific use case

Our use case is geared towards printing of the orders after applying the filters

denniskigen:
Per @dkayiwa, this is a known backend issue. So I think we can merge this and seek a workaround in subsequent PR.

denniskigen:
Reminder to self: Update the `dateActivated` property in `Order` type from `string` to `Date`.

denniskigen:
@CynthiaKamau could we get some test coverage for the date filtering?

mogoodrich:
> Any suggestions, @dkayiwa, @ibacher, @mogoodrich?

Sorry @denniskigen  I haven't been following this PR... what's the specific question, or are you asking me for an overall review? Thanks!

denniskigen:
> Thanks for working on this, @CynthiaKamau. One challenge we'll need to address is that the date comparison doesn't factor in the time. The `Datepicker` component defaults to 12:00 AM, whereas the order activation date from the backend includes the time. This discrepancy can lead to unexpected results when filtering orders by activation date. You can see this in the screenshot below where the selected date range includes orders that were activated earlier than the selected date:

> CleanShot 2024-11-18 at 2  03 33@2x

@mogoodrich I wanted to know whether this was a known issue with the backend (specifically the Order resource) and whether there's any known workarounds for this. It's basically an issue with how timezones are (or are not) getting dealt with by the backend AFAICT.

CynthiaKamau:
> @CynthiaKamau could we add some test coverage for the date filtering?

on it

ibacher:
> I wanted to know whether this was a known issue with the backend (specifically the Order resource) and whether there's any known workarounds for this. It's basically an issue with how timezones are (or are not) getting dealt with by the backend AFAICT.

The ideal rule here is that date / times are stored in UTC in backend and localized based on the client timezone on the frontend, in which case, the "right" thing to do for consistent results display if for the "start date" to start at midnight and the "end date" to end at 11:59:59. 

While not everything is stored in UTC, I still thing using that time range is appropriate.

mogoodrich:
> The ideal rule here is that date / times are stored in UTC in backend and localized based on the client timezone on the frontend, in which case, the "right" thing to do for consistent results display if for the "start date" to start at midnight and the "end date" to end at 11:59:59.
> 
> While not everything is stored in UTC, I still thing using that time range is appropriate.

Yeah, there is no good, ideal solution for this, unfortunately.  To confirm that it's a time zone issue (sounds like it is) you could confirm that everything works as expected when the client and server are in the same time zone.  Not that this is solution.

PR #2121, opened 2024-11-28
(feat) O3-2801 - Add icons to patient chart left nav

Munyua123:
## Requirements

## Screenshots

Screenshot from 2024-11-28 14-05-54

## Related Issue

https://openmrs.atlassian.net/browse/O3-2801


Munyua123:
@denniskigen @gracepotma @CynthiaKamau I have made the changes. I had a problem with the last pull request with the build but this one should work just fine. KIndly review

gracepotma:
Asante SANA @Munyua123!! LGTM <3

packages/esm-patient-common-lib/src/dashboards/DashboardExtension.tsx
```diff
@@ -22,13 +37,40 @@ export const DashboardExtension = ({
...
+    Immunizations: SyringeIcon,
+    Attachments: DocumentAttachmentIcon,
+    Programs: ProgramsIcon,
+    Appointments: EventScheduleIcon,
+  } as const;
+
+  const menuIcon = (title: MenuTitle) => {
+    const Icon = MenuIcons[title];
+    return Icon ? <Icon className={styles.icon} /> : null;
+  };
```
ibacher, line 60: There's no good reason for these to be located inside this component. They would be better located outside of it.

packages/esm-patient-common-lib/src/dashboards/DashboardExtension.tsx
```diff
@@ -22,13 +37,40 @@ export const DashboardExtension = ({
...
+
+  const renderIcon = menuIcon(title as MenuTitle);
```
ibacher, line 62: No weird edge-cases and we don't have to lie to TypeScript:

```suggestion
  const renderIcon = Object.keys(MenuIcons).includes(title) ? menuIcon(title) : null;
```

packages/esm-patient-common-lib/src/dashboards/dashboard-extension.scss
```diff
@@ -0,0 +1,8 @@
+.menu {
+  display: flex;
+  align-items: center;
```
ibacher, line 3: I don't think centering is the right option here. The items should flex from the start regardless of the width of the container.

```suggestion
  align-items: start;
```

packages/esm-patient-common-lib/src/dashboards/dashboard-extension.scss
```diff
@@ -0,0 +1,8 @@
...
+}
+
+.icon {
+  margin-right: 0.5rem;
```
ibacher, line 7: Please use the [Carbon spacing tokens](https://carbondesignsystem.com/elements/spacing/overview/) for things like this (`use @carbon/layout; margin: layout.$spacing-03;`

packages/esm-patient-common-lib/src/dashboards/DashboardExtension.tsx
ibacher: Honestly, though, I think this might not be the right way to design things. This works, but only for providing icons for known menu titles, which is bad for extensibility. Better would be to update the `DashboardExtensionProps` to take a string of the name icon and just render it like:

```ts
document.getElementById(`omrs-icon-${icon}`) ? <Icon className={styles.icon} icon={`omrs-icon-${icon}`} /> : null;
```

ibacher (changes requested):
Nice work! Some minor nits and a redesign suggestion.

Munyua123:
@ibacher Happy New Year. I'm sorry for the delay in solving the conflicts you raised, I have been able to solve some of the conflicts you raised above. Could you review and revert with any changes?

packages/esm-patient-common-lib/src/dashboards/MenuIcons.tsx
```diff
@@ -0,0 +1,42 @@
...
+  Attachments: DocumentAttachmentIcon,
+  Programs: ProgramsIcon,
+  Appointments: EventScheduleIcon,
+} as const;
+
+interface IconProps {
+  title: MenuTitle;
+}
+
+export const IconRenderer = ({ title }: IconProps) => {
```
ibacher, line 39: So... title is an arbitrary string and you've used `as any` to work around Typescript's complaining about the mismatch. Just type things more weakly here, i.e., `title` should just be a string.

ibacher: Design-wise, I still don't love the hard-coding of menu icons here, since it makes this completely unusable except for a few dashboards we have foreknowledge of. Since this PR started, I added a more generic `MaybeIcon` component that takes a string like `omrs-report-icon` and renders the icon if that exists. Instead of relying on the `title` to chose which `icon`, I suggest we make the icon part of the props that get passed to the `DashboardExtension` via the extension meta. This allows us to support more than just a small list of known panels.

Munyua123:
@ibacher I have added the Icon as a prop to the dashboard link configuration and dashboard extension, where I have put up the Maybeicon as you had instructed and passed the icons to all the menu titles from the dashboard meta. The icons for the appointments and billing history since they are in their modules / esm once this is merged I can be able to add their icons from the modules. Could you review the above? Thank you

ibacher (approved):

PR #2124, opened 2024-12-01
(chore) Bump @carbon/react to v1.71.0

Muta-Jonathan:
## Requirements

## Summary
This PR upgrades carbon to the latest version 1.71.0. Release notes at https://carbondesignsystem.com/migrating/guide/develop/

## Related Issue
[Bump carbon version in Patient Chart](https://openmrs.atlassian.net/browse/O3-4226)

### DIFF
--- a/packages/esm-form-engine-app/package.json
+++ b/packages/esm-form-engine-app/package.json
@@ -37,7 +37,7 @@
     "url": "https://github.com/openmrs/openmrs-esm-patient-chart/issues"
   },
   "dependencies": {
-    "@carbon/react": "^1.47.0",
+    "@carbon/react": "1.71.0",
     "@openmrs/esm-form-engine-lib": "next",
     "lodash-es": "^4.17.21",
     "react-error-boundary": "^4.0.13"
--- a/packages/esm-generic-patient-widgets-app/package.json
+++ b/packages/esm-generic-patient-widgets-app/package.json
@@ -38,7 +38,7 @@
   },
   "dependencies": {
     "@carbon/charts-react": "^1.5.2",
-    "@carbon/react": "^1.12.0",
+    "@carbon/react": "1.71.0",
     "d3": "^7.6.1",
     "lodash-es": "^4.17.21"
   },
--- a/packages/esm-patient-allergies-app/package.json
+++ b/packages/esm-patient-allergies-app/package.json
@@ -37,7 +37,7 @@
     "url": "https://github.com/openmrs/openmrs-esm-patient-chart/issues"
   },
   "dependencies": {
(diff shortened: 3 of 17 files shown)


NethmiRodrigo:
@Muta-Jonathan I fixed one of the tests by fixing the Regex we had used and by adding a stub for `scrollIntoView` function - https://github.com/jsdom/jsdom/issues/1695. Please use that as reference to fix the remaining failing unit tests, you can use `screen.logTestingPlaygroundURL();` in the middle of a test case to see what roles are accessible and for seeing how a regex needs to be updated. For example, in the `visits-table.test.stx`, for the failing test case `Delete encounter > Clicking the `Delete` button deletes an encounter` that gives the error 
> Unable to find an accessible element with the role "row" and name `/18-Jan-2022, 04:25 PM Facility Visit Admission POC Consent Form -- Options/i`

adding the `screen.logTestingPlaygroundURL();` right before the line where you get the error 
(which is line number 89) - https://github.com/openmrs/openmrs-esm-patient-chart/blob/1d717fa48173f9f147771dcc195a4c7df9db9267/packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/visits-table/visits-table.test.tsx#L89-L91
would give you a url that when you open in the browser would give you a playground. The regex in the query box is what I used to update the test case. 
Screenshot 2024-12-02 at 2 24 41 PM

Muta-Jonathan:
thanks @NethmiRodrigo @denniskigen this worked ... ready for review

NethmiRodrigo:
Can you update your branch @Muta-Jonathan?

yarn_BACKUP_25899.lock
NethmiRodrigo, line 1: @Muta-Jonathan you've accidentally committed a bunch of yarn backup files

Muta-Jonathan: 🥹 sorry have removed them 😆

NethmiRodrigo:
@Muta-Jonathan could you please update your branch and fix the failing e2e tests? Please do let me know if you need any guidance on fixing the e2e test

Muta-Jonathan:
indeed @NethmiRodrigo talked to @denniskigen yesterday on the coffe-break it seems it is breaking due to changes he did in the meta-Data probably

NethmiRodrigo:
On a side note, @ibacher should this go in after the carbon upgrade in core?

ibacher:
> On a side note, @ibacher should this go in after the carbon upgrade in core?

It shouldn't matter too much. However, we may need to update the patient-management before merging this.

NethmiRodrigo (approved):

PR #2125, opened 2024-12-02
(refactor) O3-4125:  Vitals and Biometrics should not rely on concept search

D-matz:
For O3-4125, not searching for "VITALS SIGNS" concept, instead getting it from packages\esm-patient-vitals-app\src\config-schema.ts

## Requirements

## Related Issue

https://openmrs.atlassian.net/jira/software/c/projects/O3/issues/O3-4125

## Other

Not sure this is actually better, as now the concept UUID needs to be in the config file (and mockVitalsConfig). I'm new to openMRS, so if you had a different way of getting the concept please try that instead.


jwnasambu:
@D-matz kindly the commit message should read something like this `chore/O3-4125:  Vitals and Biometrics should not rely on concept search`

D-matz:
@jwnasambu thank you, this is the vitals page (should be same as before) and the updated concept set result
vitals_page
vitals_resp

Will change to that commit message

packages/esm-patient-vitals-app/src/config-schema.ts
```diff
@@ -42,6 +42,10 @@ export const configSchema = {
       _type: Type.ConceptUuid,
       _default: '1343AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
     },
+    vitalsSignsUuid: {
```
denniskigen, line 45:
```suggestion
    vitalSignsConceptSetUuid: {
```

Because this property represents a concept set:

CleanShot 2024-12-04 at 11  19 32@2x

packages/esm-patient-vitals-app/src/config-schema.ts
```diff
@@ -116,6 +120,7 @@ export interface ConfigObject {
     weightUuid: string;
     respiratoryRateUuid: string;
     midUpperArmCircumferenceUuid: string;
+    vitalsSignsUuid: string;
```
denniskigen, line 123: Ditto

packages/esm-patient-vitals-app/src/common/data.resource.ts
```diff
@@ -47,28 +47,26 @@ export interface ConceptMetadata {
...
 }
 
 function getInterpretationKey(header: string) {
   // Reason for `Render` string is to match the column header in the table
   return `${header}RenderInterpretation`;
 }
 
-export function useVitalsConceptMetadata() {
+export function useVitalsConceptMetadata(vitalsSignsConceptUUID) {
   const customRepresentation =
```
denniskigen, line 59:
```suggestion
  const { concepts } = useConfig<ConfigObject>();
  const vitalSignsConceptSetUuid = concepts.vitalSignsConceptSetUuid;
  
  const customRepresentation =
```

We could get the config property directly here from the `useConfig` hook and avoid passing it down as an argument in all invocations of `useVitalsConceptMetadata`. This approach:

- Directly destructures the specific UUID you need from the config
- Reduces unnecessary variable declarations
- Makes the code more concise and easier to follow
- Eliminates the need to pass the concept UUID as an argument through multiple function calls

packages/esm-patient-vitals-app/src/common/data.resource.ts
```diff
@@ -47,28 +47,26 @@ export interface ConceptMetadata {
...
     'custom:(setMembers:(uuid,display,hiNormal,hiAbsolute,hiCritical,lowNormal,lowAbsolute,lowCritical,units))';
 
-  const apiUrl = `${restBaseUrl}/concept/?q=VITALS SIGNS&v=${customRepresentation}`;
+  const apiUrl = `${restBaseUrl}/concept/${vitalsSignsConceptUUID}?v=${customRepresentation}`;
```
denniskigen, line 62:
```suggestion
  const apiUrl = `${restBaseUrl}/concept/${vitalSignsConceptSetUuid}?v=${customRepresentation}`;
```

denniskigen (commented):
Thanks, @D-matz! I've left some suggestions on how we could tackle this slightly differently.

packages/esm-patient-vitals-app/src/common/data.resource.ts
D-matz: @denniskigen getting the UUID directly in one place does seem better than getting it and passing it in everywhere, thanks. I'll change the pull request title to refactor as well.

PR #2130, opened 2024-12-04
(feat) Update workspace calls to handle workspace groups

vasharma05:
## Requirements

## Summary
This PR adds the necessary changes to add the order basket to the ward app:
* Add `ward-patient` group name to the order basket related workspaces registration
* Add the order basket action menu button to the ward patient action menu items

## Screenshots

https://github.com/user-attachments/assets/07cc0af2-2c34-4481-afa5-bf26dbf077f4

## Related Issue
None

## Other
Depends on https://github.com/openmrs/openmrs-esm-core/pull/1185

### DIFF
--- a/packages/esm-patient-medications-app/src/drug-order-basket-panel/drug-order-basket-panel.extension.tsx
+++ b/packages/esm-patient-medications-app/src/drug-order-basket-panel/drug-order-basket-panel.extension.tsx
@@ -58,6 +58,7 @@ export default function DrugOrderBasketPanelExtension() {
     closeWorkspace('order-basket', {
       ignoreChanges: true,
       onWorkspaceClose: () => launchPatientWorkspace('add-drug-order'),
+      closeWorkspaceGroup: false,
     });
   };
 
@@ -65,6 +66,7 @@ export default function DrugOrderBasketPanelExtension() {
     closeWorkspace('order-basket', {
       ignoreChanges: true,
       onWorkspaceClose: () => launchPatientWorkspace('add-drug-order', { order }),
+      closeWorkspaceGroup: false,
     });
   };
 
--- a/packages/esm-patient-orders-app/src/routes.json
+++ b/packages/esm-patient-orders-app/src/routes.json
@@ -36,6 +36,11 @@
       }
     },
     {
+      "name": "order-basket-ward-action-menu",
+      "component": "orderBasketActionMenu",
+      "slot": "action-menu-ward-patient-items-slot"
+    },
+    {
       "name": "patient-orders-summary-dashboard",
       "component": "ordersDashboardLink",
       "slot": "patient-chart-dashboard-slot",
@@ -61,7 +66,8 @@
       "title": "orderBasketWorkspaceTitle",
       "component": "orderBasketWorkspace",
       "type": "order",
-      "canHide": true
+      "canHide": true,
+      "groups": ["ward-patient"]
     },
     {
       "name": "patient-orders-form-workspace",
--- a/packages/esm-patient-orders-app/src/routes.json
+++ b/packages/esm-patient-orders-app/src/routes.json
@@ -24,8 +24,10 @@
     {
       "name": "order-basket-action-menu",
       "component": "orderBasketActionMenu",
-      "slot": "action-menu-patient-chart-items-slot",
-      "order": 0
+      "slots": [
+        "action-menu-patient-chart-items-slot", 
+        "action-menu-ward-patient-items-slot"
+      ]
     },
     {
       "name": "lab-result",
@@ -34,11 +36,6 @@
       "meta": {
         "fullWidth": false
       }
-    },
-    {
-      "name": "order-basket-ward-action-menu",
-      "component": "orderBasketActionMenu",
-      "slot": "action-menu-ward-patient-items-slot"
     },
     {
       "name": "patient-orders-summary-dashboard",
@@ -66,8 +63,7 @@
       "title": "orderBasketWorkspaceTitle",
       "component": "orderBasketWorkspace",
       "type": "order",
-      "canHide": true,
-      "groups": ["ward-patient"]
+      "canHide": true
     },
     {
       "name": "patient-orders-form-workspace",
--- a/packages/esm-patient-orders-app/src/routes.json
+++ b/packages/esm-patient-orders-app/src/routes.json
@@ -66,7 +66,8 @@
       "title": "orderBasketWorkspaceTitle",
       "component": "orderBasketWorkspace",
       "type": "order",
-      "canHide": true
+      "canHide": true,
+      "groups": ["ward-patient"]
     },
     {
       "name": "patient-orders-form-workspace",
--- a/packages/esm-patient-orders-app/src/routes.json
+++ b/packages/esm-patient-orders-app/src/routes.json
@@ -24,10 +24,7 @@
     {
       "name": "order-basket-action-menu",
       "component": "orderBasketActionMenu",
-      "slots": [
-        "action-menu-patient-chart-items-slot", 
-        "action-menu-ward-patient-items-slot"
-      ]
+      "slot": "action-menu-patient-chart-items-slot"
     },
     {
       "name": "lab-result",
--- a/packages/esm-patient-orders-app/src/routes.json
+++ b/packages/esm-patient-orders-app/src/routes.json
@@ -66,8 +66,7 @@
       "title": "orderBasketWorkspaceTitle",
       "component": "orderBasketWorkspace",
       "type": "order",
-      "canHide": true,
-      "groups": ["ward-patient"]
+      "canHide": true
     },
     {
       "name": "patient-orders-form-workspace",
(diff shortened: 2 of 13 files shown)


vasharma05:
Hi @brandones @ibacher @chibongho, 
The changes in https://github.com/openmrs/openmrs-esm-core/pull/1185 and https://github.com/openmrs/openmrs-esm-patient-management/pull/1398 allows the developer consuming the workspace system to make only a few lines of changes to add an existing workspace to add the respective workspace to a workspace group.

vasharma05:
Any reason by the build job is failing during the `Post cache` job?
Thanks!

ibacher:
> Any reason why the build job is failing during the Post cache job?

No

So, solution-wise, I don't think what we're doing in this PR is sustainable. We need a way of the ward app to be able to declare that these are the workspaces that it's using. This suggests to me that we need to reconsider being able to set the group during the call to launch workspace or some other mechanism that doesn't require modifying the workspace registration.

vasharma05:
> We need a way of the ward app to be able to declare that these are the workspaces that it's using.

If we pass the group name in the `launchWorkspace`, it negates the purpose of not including conditional checks for launching workspaces. To enable the ward app to indicate that the `ward-patient-workspace` is using the order basket or another workspace, I attempted to implement this. However, it became intricate because we would need to define all group workspaces in the `launchWorkspaceGroup`. Consequently, we would have two apps to manage for adding workspace from one app to another app's workspace group.

vasharma05:
> How do I go about telling the framework that when launched in the context of my app, its in its own workspace group that doesn't overlap with the workspaces in the ward app and the workspaces in the chart?

That we don't have to, because as per the design and confirmed with Paul, only one workspace group can be opened at a time. Alongside, the `order-basket` doesn't need to know in what context is it opened, it just knows that it is opened, and the workspace system has information about the workspace group from the group information stored in the workspace store.

> We allow them to register workspace groups and then list the workspaces that (if available) are included in that group?

Yes, this was also suggested by @usamaidrsk. However, the drawback is that whenever I want to add a workspace to a workspace group, I will have to go to the workspace group's registration and update it. But why does the workspace group need to handle this scenario? In case someone builds on top of the ward app, they will have to come to the ref app and define their workspace in the workspace group's registration. My question is that why does the workspace group needs to know about all the workspaces defined for that group, and if it wants the information, the opened workspace stores the [currentWorkspaceGroup](https://github.com/openmrs/openmrs-esm-core/blob/d46e219bec0d6b211cd5d9f8fb5129fc3e7fbac8/packages/framework/esm-styleguide/src/workspaces/workspaces.ts#L110) info, so that is also convinient.

chibongho:
Right now, I have a scenario where I want to include the patient search workspace from patient-search-app in a new workspace group `"ward-patient-admission-requests"` within the ward app. It seems like I have to add that group into the `groups` definition of the `patient-search-workspace` in patient-search-app's `routes.json` for it to properly display the action icons. Unless I'm missing something, I think that means any other app that want the same pattern will need to do the same, right? So the workspace declaration in `routes.json` could look like this:

```
    {
      "name": "patient-search-workspace",
      "component": "patientSearchWorkspace",
      "title": "searchPatient",
      "groups": ["ward-patient-admission-requests", "some-other-app-group", "some-other-app-group-not-in-refapp", ...]
    }
```
I think this is contrary to the separation of concern design pattern that extensions / workspaces system otherwise gives us.

image

brandones:
Right @vasharma05 at the end of the day I think this PR, as @chibongho points out, demonstrates a violation of our design principles. So we have to tweak the workspace groups system so that we can have a proper separation of concerns. The patient chart should not need to know about the existence of the ward app for any reason.

vasharma05:
Yes, @chibongho, you need to add the `routes` property for the patient search workspace.

@brandones @ibacher, let me work on defining workspace groups more explicitly as you have discussed.
Thanks!

vasharma05:
Hi @ibacher @brandones,
May we go ahead with this PR, to add the order basket to the Ward Patient workspace for now, and I can work on the improving the workspace group implementation after the new year, as this work is just building on the recent workspace group system improvement.
Thanks!

usamaidrsk:
@vasharma05 the build and e2e tests are failing

brandones:
Sorry I didn't see this til the new year, happy new year @vasharma05  :)

I would much rather we fix the workspace group system and get this right, than keep building on something we know we're going to have to breaking-change later.

rbuisson:
This is a change we'll need for a project with quite a tight timeline, in other words, we wanted this to be part of `3.3.0`

If we decide to merge this as is (which of course is far from ideal), can this be considered to be added as a patch of `3.3.0`? (we'll proceed with the proper implementation right after merge).
If not, will the "proper" implementation (using workspace groups) be accepted as a patch of `3.3.0`?
If not, when is the `3.4.0` planned?

@brandones  @ibacher.

ibacher:
> we wanted this to be part of 3.3.0

We're currently doing an RC of 3.2.0, just for clarity.

> can this be considered to be added as a patch of 3.3.0? 

Yes, but the problem is that this PR kind of demonstrates that the [prior PR ](https://github.com/openmrs/openmrs-esm-core/pull/1185) (which is included in the release candidate) isn't really fit for purpose.

Timeline-wise, we have—I think—at least a week before the next planned RC or release. Is it possible to get a "proper" implementation done in that time-frame? If it's not, we can probably discuss a way forward.

vasharma05:
Hi @ibacher @brandones!

Wishing you both a happy new year 🎉 

As per suggestions in the texts above, I have made changes https://github.com/openmrs/openmrs-esm-core/pull/1256 and https://github.com/openmrs/openmrs-esm-patient-management/pull/1428, which will now allow the workspace group's registration to define the workspaces which are part of the workspace group, and the individual workspace will not need to be aware of workspace groups that might be present in some other app.
Thanks!

packages/esm-patient-orders-app/src/routes.json
```diff
@@ -35,6 +35,11 @@
         "fullWidth": false
       }
     },
+    {
```
ibacher, line 38: This is another thing I'm going to object to, since we already have an extension for this component. For now, I guess we can just change `slot` to `slots` in the existing definition, but we need to come up with something better long-term.

packages/esm-patient-medications-app/src/drug-order-basket-panel/drug-order-basket-panel.extension.tsx
```diff
@@ -58,13 +58,15 @@ export default function DrugOrderBasketPanelExtension() {
...
       onWorkspaceClose: () => launchPatientWorkspace('add-drug-order'),
+      closeWorkspaceGroup: false,
     });
   };
 
   const openDrugForm = (order: DrugOrderBasketItem) => {
     closeWorkspace('order-basket', {
       ignoreChanges: true,
       onWorkspaceClose: () => launchPatientWorkspace('add-drug-order', { order }),
+      closeWorkspaceGroup: false,
```
ibacher, line 69: Shouldn't this be the default, assuming that there's at least one workspace in this group that hasn't been closed?

packages/esm-patient-orders-app/src/routes.json
vasharma05: Yes, ideally we should, but the original definition has an `order=0` which isn't the case with the `ward-patient` action menu.

packages/esm-patient-medications-app/src/drug-order-basket-panel/drug-order-basket-panel.extension.tsx
vasharma05: > that there's at least one workspace in this group that hasn't been closed?

Actually no, since the new workspace is launched only after the previous workspace closes. For now, it's not possible to know whether a new workspace will be opened after closing the current one or not, hence we have the `closeWorkspaceGroup` property set to `true` by default to clear the workspace group and run cleanup. In case we don't want to clear the workspace group, similar to the above use case, we can pass the `closeWorkspaceGroup` to false.
Refer: https://github.com/openmrs/openmrs-esm-core/blob/78b7975f08b975e6080c5224bb745bd8be97beba/packages/framework/esm-styleguide/src/workspaces/workspaces.ts#L249

packages/esm-patient-orders-app/src/routes.json
vasharma05: image

The patient chart action menu button should remain at the top of the slot. I tried giving the patient chart action menu button an order of -1, but it didn't work, as the minimum value required is 0.

image

Thanks!

vasharma05:
@ibacher, requesting re-review here
Thanks!

packages/esm-patient-orders-app/src/routes.json
```diff
@@ -24,8 +24,10 @@
     {
       "name": "order-basket-action-menu",
       "component": "orderBasketActionMenu",
-      "slot": "action-menu-patient-chart-items-slot",
-      "order": 0
+      "slots": [
+        "action-menu-patient-chart-items-slot", 
+        "action-menu-ward-patient-items-slot"
```
ibacher, line 29: Can we also add this extension to the slot via config please?

vasharma05: Updated the config here: https://github.com/openmrs/openmrs-distro-referenceapplication/pull/876/files

ibacher (approved):

PR #2134, opened 2024-12-05
(feat) O3-4200 Make visit form usable in other apps

chibongho:
## Requirements

## Summary

This PR refactors the patient chart visit form for it to be re-usable in different apps (like appointments and queues), without needing logic specific to those apps in the visit form. See corresponding changes in esm-patient-management [here](https://github.com/openmrs/openmrs-esm-patient-management/pull/1402). While some changes are opinionated, there is no breaking change.

Despite having extension slots for queues and appointments UIs (which can be turned on via config), the visit form still contains code related to submitting queue entry and checking in an appointment. This change provides a way for those extensions to supply their own callbacks to run after a visit has been created / updated. This allows us to decouple the visit form from any unrelated logic. 

Notes:
- Minor behavioral change. The visit form's onSubmit has a complex promise chain to start visit, update visit attributes, create queue entry, check in appointment. Previously, on success of all of them, it tries to group the notifications into a single snackbar notification. This PR separates them into individual snarkbars. (Although we still have one exception. When there are multiple visit attributes to be created/updated/deleted, we only show one success snackbar when all of them succeed. We show individual snackbar for each failure.) 
  - As we are making several separate requests on form submit, they are not atomic. Longer term, we might want to have a discussion whether we even want these actions grouped together in a single form. (We might, instead, provide a way for the visit form to open another form on success.) That said, this PR doesn't change this behavior.
- The patient-chart keeps the config values `showServiceQueueFields` and `showUpcomingAppointments` to specify whether those corresponding extensions should show, but now, the code to actually check for those values live within the extensions themselves. Arguably, we should make a breaking change to have those configs be within their respective apps.
- The `extra-visit-attribute-slot`, currently used by the Billing app, could benefit from a similar refactoring as other extension slots to supply its own onVisitCreatedOrUpdated callback, but I'm leaving it untouched for now as I'm not that familiar with that code base.
- Ideally, there should be some form validation in the queues / appointments extensions, but right now there isn't, and I'm keeping the existing behavior. (When we open the start visit form by clicking on the "Add patient to queue" button in the queues app, we ought to make the form fields for queues required instead of optional.) I think this should be possible if we pass in the React Hook Form controller to the extensions.

## Screenshots

The visit form, when opened from the queues app:
image

On submit:
image

## Related Issue

https://issues.openmrs.org/browse/O3-4200


packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -833,15 +768,19 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
...
+            {/* Queue location and queue fields. These get shown when config.showServiceQueueFields is true,
+                or when the form is opened from the queues app */}
+            <section>
+              <div className={styles.sectionTitle}></div>
+              <div className={styles.sectionField}>
+                <VisitFormExtensionSlot
+                  name="visit-form-queue-slot"
+                  patientUuid={patientUuid}
+                  visitFormOpenedFrom={openedFrom}
+                  setOnVisitCreatedOrUpdatedCallbacks={setOnVisitCreatedOrUpdatedCallbacks}
```
mogoodrich, line 780: More a question than a request for a change... why do the queues and appointments have specific extension slots, rather than a single "visit-form-extension-slot" that both queues and appointments insert into?  As we have it now, if another ESM wanted to insert it's own functionality into the visit form, would it have to add another extension slot to the form?

mogoodrich (approved):
I'll defer to other on the specifics of the code, but your description sounds great, thanks so much for taking the time to clean this up.

I did have one question, and noticed that tests were failing.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
chibongho: Yeah, with this change, those 2 extensions look more similar than before. I'm keeping it as is because where the extensions appear are different (the appointments extension is near the top, close to where the visit start time fields are, vs the queues extension is near the bottom). It might be possible to slot them into one single slot (and keeping the existing ordering) if we slice up the visit form further so that even the visit form fields are extensions themselves , but I think that's a pretty drastic change.

ibacher: I don't _love_ that the slots are named after specific apps ... It kind of violates the idea that slots are just a location on a screen that extensions can appear. I'm kind of ok with having multiple slots if it makes sense, e.g., there are two places on a screen we want extensions to appear.

ibacher (approved):

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
chibongho: Ok, I changed the slot named to `visit-form-top-slot` and `visit-form-bottom-slot`

ojwanganto:
Thanks so much @chibongho for standardizing the UX. It makes a lot of sense to me. I know this may sound off at this point but do you think we can provide better wordings for the visit attributes? I know most users will not even know what visit attributes mean.

ojwanganto (approved):

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
mogoodrich: Thanks @chibongho  and @ibacher !

ibacher:
> do you think we can provide better wordings for the visit attributes? 

"Additional visit information" maybe?

chibongho:
@ojwanganto Sure. Easy change.

PR #2139, opened 2024-12-06
(fix) Restore generic patient widget

ibacher:
## Requirements

## Summary

This fixes the generic-patient-widgets app. The functionality was broken in openmrs/openmrs-esm-core#898, which was solving an issue related to configurations being provided before their schema was loaded. Inadvertently, this _requires_ apps to call `defineConfigSchema()`is they use configuration anywhere. 

The generic-patient-widgets-app was using `defineExtensionConfigSchema()`, and using this call wrongly (since it passed the module name instead of the extension name). The goal of this function is to allow individual extensions to have config schemas separate from those that are in the app they define, not to provide the configuration schema for the module.

### DIFF
--- a/packages/esm-generic-patient-widgets-app/src/index.ts
+++ b/packages/esm-generic-patient-widgets-app/src/index.ts
@@ -1,4 +1,4 @@
-import { defineExtensionConfigSchema, getSyncLifecycle } from '@openmrs/esm-framework';
+import { defineConfigSchema, getSyncLifecycle } from '@openmrs/esm-framework';
 import { configSchema } from './config-schema';
 import obsSwitchableComponent from './obs-switchable/obs-switchable.component';
 
@@ -14,5 +14,5 @@ const options = {
 export const switchableObs = getSyncLifecycle(obsSwitchableComponent, options);
 
 export function startupApp() {
-  defineExtensionConfigSchema(moduleName, configSchema);
+  defineConfigSchema(moduleName, configSchema);
 }


brandones:
Good catch, Ian. But I think we actually do want to provide this as an extension config? Looking at the [config schema](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-generic-patient-widgets-app/src/config-schema.ts), all of that appears to be specific to a particular widget. I guess it's mostly academic right now since, with only one extension, the behavior will be the same. But that's my thought.

ibacher:
My thinking here is that although it's a config schema for a specific widget, that widget is the only public extension, so all copies of the extension have a shared schema, even though they should have different schema instances. I guess alternatively we can switch to `defineExtensionConfig('obs-by-encounter-widget', schema)`?

brandones:
> alternatively we can switch to `defineExtensionConfig('obs-by-encounter-widget', schema)`

yeah that's what I was thinking; and then when one day we have another generic widget it will be clearer how to add its config

denniskigen (dismissed):

ibacher:
Huh... so it looks like `defineExtensionConfig()` only works when the `extension` argument is the moduleName which is bad... Should hold off on merging this in until its fixed.

denniskigen:
Should we close this out? @ibacher, @vasharma05?

ibacher:
> Should we close this out?

No, but I'll turn it into a draft. We need to fix the config system so that this works.

PR #2147, opened 2024-12-11
(fix) O3-4260 fix error when starting a visit immediately after ending one

chibongho:
…g one

## Requirements

## Summary

In O3, when we start a visit, we truncate the start time’s seconds field to 0. However, when we end a visit, we do not do the same truncation. Hence, when we try to start another visit within the same minute, we get a This visit overlaps with another visit of the same patient error. 

This PR fix this by making the end time truncate the seconds field to 0 as well when ending a visit, 

Note that, however, there is still an issue with starting a visit, then ending the visit, and then starting a new visit, all within the same minute, as the backend will still think that the new visit conflicts with the old one. We’ll have to change the [backend logic](https://github.com/openmrs/openmrs-core/blob/723fb0c65a6aca9a9a7fc522f7db1c3a9f0231fb/api/src/main/java/org/openmrs/validator/VisitValidator.java#L137) (and probably also remove the seconds truncation logic when we start and stop visits) if we want to fix that as well.

## Related Issue

https://openmrs.atlassian.net/browse/O3-4260

### DIFF
--- a/packages/esm-patient-chart-app/src/visit/visit-prompt/end-visit-dialog.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit/visit-prompt/end-visit-dialog.component.tsx
@@ -9,9 +9,12 @@ import styles from './end-visit-dialog.scss';
 interface EndVisitDialogProps {
   patientUuid: string;
   closeModal: () => void;
+
+  // for test mock only
+  stopDatetime?: Date;
 }
 
-const EndVisitDialog: React.FC<EndVisitDialogProps> = ({ patientUuid, closeModal }) => {
+const EndVisitDialog: React.FC<EndVisitDialogProps> = ({ patientUuid, closeModal, stopDatetime = new Date() }) => {
   const { t } = useTranslation();
   const { currentVisit, currentVisitIsRetrospective, mutate } = useVisit(patientUuid);
   const { queueEntry } = useVisitQueueEntry(patientUuid, currentVisit?.uuid);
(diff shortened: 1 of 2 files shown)


ibacher (approved):
Might it be simpler to just rollback the date a full second? E.g.,

```ts
stopDatetime = new Date(new Date() - 1000);
 ```

mogoodrich:
> The reason we truncate the seconds in the start visit form in the first place is that we don't have a form element to ask for the seconds. I think if people want to modify the visit start time (for RDE), defaults the seconds to 0 is reasonable. However, when they want "current time" to be the visit start time, I think we shouldn't truncate.

I agree with this... I don't see the need to truncate.  That being said, I don't think we store milliseconds in the DB, so this might not fix things?

ibacher:
> Thoughts?

I agree with the optional time thing. 

We've run into issues before with time truncation in non-RDE contexts, e.g., needing to wait 1 minute between actions, which is why I'm not a huge fan. I agree that in a RDE context, truncating to 0 seconds makes complete sense, though if we do that, I would also advocate that we truncate the start date/time as well in an RDE context. For the non-RDE use case, I still like the idea of defaulting to the current date - 1 second. That is, as @mogoodrich points out, the smallest difference that can be stored in the database and should ensure that (at least for the same user) the "End Visit" is always before the "Start Visit".

chibongho:
Thanks for the input. Closing this PR for now. I'll update the corresponding ticket with our discussion.

PR #2149, opened 2024-12-11
(fix) Attempt to fix e2e test errors

ibacher:
## Requirements

## Summary

Semi-speculative attempt to fix failing e2e tests that are hard to reproduce:

1. Since `endVisit()` is called without the full existing visit, it used to always change the start date; this leads to hard-to-reproduce failures, where encounters are recorded "inside" the visit after, but close to the start date and so the attempt to stop the visit ends up rejected by the backend. Since we were already storing a reference to the visit, there's no reason we can't use that.
2. For some unknown reason—and apparently only in e2e tests—at least one entry in the `postDataPrepFunctions` winds up being not a function, which raises an error that isn't handled by the code. This PR introduces a simple check for this condition and skips the callback if it is not a function.

Hopefully, this fixes the issues with the e2e tests in GitHub Actions

### DIFF
--- a/e2e/commands/visit-operations.ts
+++ b/e2e/commands/visit-operations.ts
@@ -13,16 +13,16 @@ export const startVisit = async (api: APIRequestContext, patientId: string): Pro
     },
   });
 
-  await expect(visitRes.ok()).toBeTruthy();
+  expect(visitRes.ok()).toBeTruthy();
   return await visitRes.json();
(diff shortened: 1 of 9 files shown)


denniskigen:
Let's see how this goes! Thanks, @ibacher.

denniskigen (approved):

PR #2150, opened 2024-12-12
(fix) O3-4270: Group obs by encounter instead of datetime

vasharma05:
## Requirements

## Summary
This PR makes a change in the `obs-by-encounter-widget`. Initially, the obs were grouped with the obs' issued date time. The issue with the above approach is that a new row in the table is created when the obs value is updated.

The `obs` should be grouped by encounter rather than date and time, is because the obs are recorded as part of an encounter. Even if we group `obs` by time, the obs recorded together (as part of an encounter) will be grouped together, hence it makes sense to group the obs by related encounter, rather than recorded time.

### Configuration

```
{
  "@openmrs/esm-patient-chart-app": {
    "extensionSlots": {
      "patient-chart-summary-dashboard-slot": {
        "add": [
          "obs-by-encounter-widget#covid"
        ],
        "configure": {
          "obs-by-encounter-widget#covid": {
            "title": "COVID Results",
            "resultsName": "COVID Results",
            "data": [
              {
                "concept": "611190b4-2b83-42c5-bbb1-8d99bfaec9ed",
                "label": "Test done",
                "color": "green"
              },
              {
                "concept": "89c5bc03-8ce2-40d8-a77d-20b5a62a1ca1",
                "label": "Tested for COVID 19",
                "color": "green"
              },
              {
                "concept": "192f67ac-247c-4e1d-966a-4c223e33d38b",
                "label": "Test results",
                "color": "green"
              }
            ],
            "table": {
              "pageSize": 5
            }
          }
        }
      }
    }
  }
}
```

### Before

https://github.com/user-attachments/assets/22ad64a7-7f83-48a8-8b76-deae708b4838

### After
image

## Related Issue
https://openmrs.atlassian.net/browse/O3-4270

### DIFF
--- a/packages/esm-generic-patient-widgets-app/src/obs-table/obs-table.component.tsx
+++ b/packages/esm-generic-patient-widgets-app/src/obs-table/obs-table.component.tsx
@@ -23,8 +23,8 @@ const ObsTable: React.FC<ObsTableProps> = ({ patientUuid }) => {
   const { t } = useTranslation();
   const config = useConfig();
   const { data: obss, error, isLoading, isValidating } = useObs(patientUuid, config.showEncounterType);
-  const uniqueDates = [...new Set(obss.map((o) => o.issued))].sort();
-  const obssByDate = uniqueDates.map((date) => obss.filter((o) => o.issued === date));
+  const uniqueDates = [...new Set(obss.map((o) => o.encounter.reference.split('#')[1]))].sort();
+  const obssByDate = uniqueDates.map((date) => obss.filter((o) => o.encounter.reference.split('#')[1] === date));
 
   const tableHeaders = [
     { key: 'date', header: t('dateAndTime', 'Date and time'), isSortable: true },
--- a/packages/esm-generic-patient-widgets-app/src/obs-table/obs-table.component.tsx
+++ b/packages/esm-generic-patient-widgets-app/src/obs-table/obs-table.component.tsx
@@ -22,9 +22,11 @@ interface ObsTableProps {
 const ObsTable: React.FC<ObsTableProps> = ({ patientUuid }) => {
   const { t } = useTranslation();
   const config = useConfig();
-  const { data: obss, error, isLoading, isValidating } = useObs(patientUuid, config.showEncounterType);
-  const uniqueDates = [...new Set(obss.map((o) => o.encounter.reference.split('#')[1]))].sort();
-  const obssByDate = uniqueDates.map((date) => obss.filter((o) => o.encounter.reference.split('#')[1] === date));
+  const { data: obss } = useObs(patientUuid, config.showEncounterType);
+  const uniqueEncounterUuids = [...new Set(obss.map((o) => o.encounter.reference))].sort();
+  const obssGroupedByEncounters = uniqueEncounterUuids.map((date) =>
+    obss.filter((o) => o.encounter.reference === date),
+  );
 
   const tableHeaders = [
     { key: 'date', header: t('dateAndTime', 'Date and time'), isSortable: true },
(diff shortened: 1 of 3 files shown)


vasharma05:
One thing I also realized, we should be using the obs' `effectiveDatetime` instead of `issued` when grouping the obs.
@ibacher @dkayiwa, should we still group by encounter or by `effectiveDatetime`? Since in the end, it's also the encounter's datetime.
Thanks!

packages/esm-generic-patient-widgets-app/src/obs-table/obs-table.component.tsx
```diff
@@ -23,8 +23,8 @@ const ObsTable: React.FC<ObsTableProps> = ({ patientUuid }) => {
   const { t } = useTranslation();
   const config = useConfig();
   const { data: obss, error, isLoading, isValidating } = useObs(patientUuid, config.showEncounterType);
-  const uniqueDates = [...new Set(obss.map((o) => o.issued))].sort();
-  const obssByDate = uniqueDates.map((date) => obss.filter((o) => o.issued === date));
+  const uniqueDates = [...new Set(obss.map((o) => o.encounter.reference.split('#')[1]))].sort();
```
brandones, line 26: I'd like @ibacher's input on whether this is a legal way to get a date

brandones: The [HL7 docs](https://hl7.org/fhir/references-definitions.html#Reference.reference) don't specify much about the actual format of the reference, AFAICT

ibacher: I think this doesn't do what it's supposed to do. References are, essentially, either URLs or URL parts so something like:

https://example.org/fhir/Encounter/8081f1a1-8c64-495f-b21c-f7395484a705
Encounter/8081f1a1-8c64-495f-b21c-f7395484a705

It can also contain ids using the `#`, but it's not a standard part.

What this definitely doesn't do is extract a date or time at all, unless there's some kind of pre-processing I'm unaware of.

ibacher:
@vasharma05 We may as well group by the encounter datetime, even though this won't necessarily align with the effective date of the obs, but since each obs can have it's own effective date, I don't see how we could get the right "logical" grouping. Maybe we put the obs effective date in a tooltip or something?

packages/esm-generic-patient-widgets-app/src/obs-table/obs-table.component.tsx
vasharma05: Sorry, that's my bad, it's actually returning the encounter UUIDs, instead of dates, I didn't update the variable name. Let me just fix that.

vasharma05: From the Observation response, we get the reference of the Encounter in the form of `Encounter#<Associated encounter UUID>`.

vasharma05: Also, I don't know why I used `#`, but it's actually `/`, thanks @ibacher for the catch.
I am not splitting it anymore and using the whole encounter reference for grouping.
Thanks!

vasharma05: Pushed the latest change!

vasharma05:
> Maybe we put the obs effective date in a tooltip or something?
Actually, the obs effective date is what is being shown under the `Date and time` column.
Thanks!

brandones (approved):
LGTM, thanks @vasharma05 !

brandones:
Please fix the conflict and CI, and then you should be good to merge

PR #2151, opened 2024-12-12
(fix) O3-4271 fix callback registration in visit form

chibongho:
## Requirements

## Summary

See also [corresponding PR](https://github.com/openmrs/openmrs-esm-patient-management/pull/1411) in patient-management.

This is a regression from [O3-4200](https://openmrs.atlassian.net/browse/O3-4200). In the service queues app, the form for creating a new queue entry for patient with existing visit is broken. The actual fix is in [corresponding PR](https://github.com/openmrs/openmrs-esm-patient-management/pull/1411) in patient management; this PR makes necessary change on how the visit form registers `onVisitCreatedOrUpdated` callbacks.

Root cause:

`[state, setState = useState<Function>()` is generally a bad idea, because `setState` is overloaded to take in either 1) the new state or 2) a lambda that returns the new state. When doing `setState(callback)`, JavaScript has no way to distinguish between the two, and erroneously treats it as case 2. The visit form, which stores `onVisitCreatedOrUpdated` callbacks in a `Map<string, Callbacks>`, doesn't actually have this problem, but the form for creating a new queue entry for patient with existing visit does.

This PR addresses this by wrapping the callback states in an object.

Unrelated, the visit form has a non-functional / performance issue with infinite state updates when the extensions in `VisitFormExtensionSlot` register callbacks. (Generally, the pattern of a child component updating the parent's state is dangerous if the state is an object.)  This PR address that with a combination of `useCallbacks`, `useMemo` and `memo` in various places. (It feels like it should be possible to reduce the use of these memoing functions, but I haven't been able to simplify it.)

## Screenshots

On submission on start visit form to: start a visit, update visit attributes, check-in appointment, create queue entry.
image

On submission of form to create queue entry for patient with existing visit:
image

## Related Issue

https://openmrs.atlassian.net/browse/O3-4271

## Other

[O3-4200]: https://openmrs.atlassian.net/browse/O3-4200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
@@ -14,6 +14,7 @@ import {
 } from '@carbon/react';
 import { zodResolver } from '@hookform/resolvers/zod';
 import {
+  type AssignedExtension,
   Extension,
   ExtensionSlot,
   formatDatetime,
@@ -60,10 +61,10 @@ import VisitDateTimeField from './visit-date-time.component';
 import {
   createVisitAttribute,
   deleteVisitAttribute,
-  type OnVisitCreatedOrUpdatedCallback,
   updateVisitAttribute,
   useConditionalVisitTypes,
-  useOnVisitCreatedOrUpdatedCallbacks,
+  useVisitFormCallbacks,
+  type VisitFormCallbacks,
   type VisitFormData,
 } from './visit-form.resource';
 import styles from './visit-form.scss';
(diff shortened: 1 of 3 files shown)


ibacher:
> It feels like it should be possible to reduce the use of these memoing functions, but I haven't been able to simplify it.

Have you tried `ref`s? That's usually my first thought when replacing state with something that doesn't produce re-renders.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -817,52 +818,47 @@ interface VisitFormExtensionSlotProps {
...
+    const config = useConfig<ChartConfig>();
+
+    return (
+      <ExtensionSlot name={name}>
+        {(extension: AssignedExtension) => {
+          const state: VisitFormExtensionState = {
+            patientUuid,
+            setVisitFormCallbacks: (callbacks) => {
+              setVisitFormCallbacks((old) => {
+                return new Map(old).set(extension.id, callbacks);
```
ibacher, line 851: Do we need a new `Map` on every call? It might simplify things if it's possible to do an update-in-place.

ibacher (approved):

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
chibongho: I think we do, if we keep using `useState`, as the Map should be immutable. (This is akin of editing / adding an element to an array state). That said, I think we should move to `useRef`.

chibongho:
> > It feels like it should be possible to reduce the use of these memoing functions, but I haven't been able to simplify it.
> 
> Have you tried `ref`s? That's usually my first thought when replacing state with something that doesn't produce re-renders.

That's a good idea. Filed [O3-4274](https://openmrs.atlassian.net/browse/O3-4274) to track this. I'll commit this in the meantime as it's a blocker.

PR #2157, opened 2024-12-14
(fix) O3-4274: Use useRef instead of useState for useVisitFormCallbacks

Bharath-K-Shetty:
## Requirements

## Summary
This PR refactors the useVisitFormCallbacks hook by replacing the usage of useState with useRef. The primary goal of this change is to improve performance by avoiding unnecessary re-renders triggered by state updates. useRef is used to persist values across renders without causing re-renders, which is particularly beneficial in the case of form callbacks where state persistence is needed but without affecting component lifecycle.

This change ensures that the form callbacks are maintained correctly across renders, while reducing the overhead associated with state management in React.

Jira Issue: O3-4274

## Screenshots

https://github.com/user-attachments/assets/74b44d3c-70cf-4c01-b40e-2c2e8cfe472e

## Related Issue

https://openmrs.atlassian.net/browse/O3-4274


packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -580,6 +584,7 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
         return;
       }
     },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
brandones, line 587: I think it is preferable simply to include the ref in the dependency array, even though it will not change.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -120,7 +120,10 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
   const { visitAttributeTypes } = useVisitAttributeTypes();
   const [extraVisitInfo, setExtraVisitInfo] = useState(null);
 
-  const [visitFormCallbacks, setVisitFormCallbacks] = useVisitFormCallbacks();
+
+  const visitFormCallbacksRef = useVisitFormCallbacks();
+
+
```
brandones, line 126:
```suggestion
  const visitFormCallbacksRef = useVisitFormCallbacks();
```

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -531,7 +533,9 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
               },
             );
 
-            const onVisitCreatedOrUpdatedRequests = [...visitFormCallbacks.values()].map((callbacks) =>
+
+            const onVisitCreatedOrUpdatedRequests = [...visitFormCallbacksRef.current.values()].map((callbacks) =>
+
```
brandones, line 538:
```suggestion
            const onVisitCreatedOrUpdatedRequests = [...visitFormCallbacksRef.current.values()].map((callbacks) =>
```

Please don't add lots of unnecessary whitespace.

brandones: Please get rid of the extra whitespace elsewhere in the PR.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -660,10 +665,12 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
               <div className={styles.sectionTitle}></div>
               <div className={styles.sectionField}>
                 <VisitFormExtensionSlot
-                  name="visit-form-top-slot"
+                  name="visit-form-bottom-slot"
```
brandones, line 668: Why did you do this?

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -660,10 +665,12 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
...
                   patientUuid={patientUuid}
                   visitFormOpenedFrom={openedFrom}
-                  setVisitFormCallbacks={setVisitFormCallbacks}
+
+                  visitFormCallbacksRef={visitFormCallbacksRef} // Pass the ref instead
```
brandones, line 672: Don't include comments that won't make sense to someone reading this code later. The whole point of comments is to make code _less_ confusing.

brandones (changes requested):
Thanks for this change, @Bharath-K-Shetty . Please clean up your code before submitting it. Remove the extra whitespace and strange comments.

brandones:
Please add a video demonstrating that everything still behaves as expected.

Bharath-K-Shetty:
Hey @brandones,I’ve made the changes as requested and updated the PR. Let me know if any further changes are needed.

Thank you

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
brandones: Please don't resolve conversations without addressing the question being asked.

Why did you change the extension slot name?

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -243,7 +243,6 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
     reset,
   } = methods;
 
-  // default values are cached so form needs to be reset when they change (e.g. when default visit location finishes loading)
```
brandones, line 246: Please restore this comment; it provides useful information for understanding what the code is doing, which is not obvious.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
brandones: @Bharath-K-Shetty you actually need to respond to reviewer feedback, you can't just hit the "resolve conversation" button to make things go away

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -732,7 +731,6 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
                       {contentSwitcherIndex === 1 && <BaseVisitType visitTypes={allVisitTypes} />}
                     </>
                   ) : (
-                    // Defaults to showing all possible visit types if recommended visits are not enabled
```
brandones, line 735: As above, please restore this comment

brandones (changes requested):

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
Bharath-K-Shetty: Ok @brandones that was my fault.Thank you

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
Bharath-K-Shetty: I changed the slot name from visit-form-top-slot to visit-form-bottom-slot because the original name (visit-form-top-slot) was causing a 404 error, indicating that the server could not find the requested resource. After troubleshooting, I updated the name to visit-form-bottom-slot, which resolved the issue and allowed the form to function as intended. This suggests that the visit-form-bottom-slot aligns with the correct configuration or extension setup expected by the system.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
brandones: Ok if adding the ref in the dependency array causes a different eslint error that we would need to suppress anyway, then yes this is fine. Very strange though, that there is no dependency array that will not produce an ESLint error when a ref is used in the useEffect.

Do not click the "resolve conversation" button unless you yourself *opened* the conversation, otherwise the person you are talking with might not see your response. I opened this conversation with a request, so it's resolved either when I click the "resolved" button or when I respond with something that suggests that I've acknowledged what you've written and think my question is resolved. Hope this is clear and helpful for your future GitHub use.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
brandones: That's very surprising, and a separate bug from this ticket. I'm fine with you addressing it in this PR, but you need to document the bug. Please provide screenshots showing the error.

Whatever the problem is, this is pretty certainly not the solution, since there is already a `visit-form-bottom-slot` in existence, on line 774 of this file. If you're not sure how the extension system works, read the docs.

brandones (changes requested):

brandones:
Also, you can use the "Request re-review" button instead of pinging people on Slack.

image

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -771,10 +771,10 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
               <div className={styles.sectionTitle}></div>
               <div className={styles.sectionField}>
                 <VisitFormExtensionSlot
-                  name="visit-form-bottom-slot"
+                  name="visit-form-top-slot"
```
brandones, line 774: Basically same feedback as before... there is already a top slot, at the top of the component. This is a different slot, at the bottom of the component.

brandones (changes requested):
This just keeps getting worse... I'm going to close this PR. If you'd like to make a fresh start of it, feel free. Tag @chibongho as reviewer in that case, since he opened the ticket, I think he has the best understanding of what this should look like.

Bharath-K-Shetty:
@brandones Thank you for the feedback and your patience. I'll take a fresh look and ensure a cleaner implementation if I revisit this. I'll tag @chibongho for guidance next time.

PR #2165, opened 2024-12-19
(fix) O3-4294: Fix incorrect visit date validation in edit mode in the visit form

denniskigen:
## Requirements

## Summary

This PR fixes a bug in the visit form that allows users to set future start dates when editing visits. This happens because the start date validation was getting bypassed when the `displayVisitStopDateTimeFields` was true (which is the case when editing an existing visit).  

Start dates should always be validated to be on or before the current date, regardless of whether stop date fields are displayed. This fix ensures that that is the case.

### Visit start date validation not taking effect when editing an existing visit

https://github.com/user-attachments/assets/1bd15306-e37a-4ef2-94c2-784fd47334f4

### Fix

https://github.com/user-attachments/assets/da04c29b-d2f7-4f1b-922e-91624e2fd2e0

## Related Issue

https://openmrs.atlassian.net/browse/O3-4294

### DIFF
--- a/packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx
@@ -94,8 +94,6 @@ const VisitDateTimeField: React.FC<VisitDateTimeFieldProps> = ({
                     <TimePickerSelect
                       aria-label={t('timeFormat ', 'Time Format')}
                       id={`${timeFormatFieldName}Input`}
-                      invalid={Boolean(errors[timeFormatFieldName])}
-                      invalidText={errors[timeFormatFieldName]?.message}
                       onChange={(event) => onChange(event.target.value as amPm)}
                       value={value}
                     >
--- a/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.test.tsx
+++ b/packages/esm-patient-chart-app/src/visit/visit-form/visit-form.test.tsx
@@ -271,21 +271,6 @@ describe('Visit form', () => {
     await user.click(screen.getByLabelText(/Outpatient visit/i));
   });
 
-  it('displays an error message when the visit start date is in the future', async () => {
-    const user = userEvent.setup();
-
-    renderVisitForm();
-
-    const dateInput = screen.getByRole('textbox', { name: /date/i });
-    const futureDate = dayjs().add(1, 'month').format('DD/MM/YYYY');
-
-    await user.clear(dateInput);
-    await user.type(dateInput, futureDate);
-    await user.tab();
-
-    expect(screen.getByText(/start date needs to be on or before/i)).toBeInTheDocument();
-  });
-
   // TODO: Figure out why this test is failing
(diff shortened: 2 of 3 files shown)


packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -157,7 +156,8 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
           (value) => {
             const today = dayjs();
             const startDate = dayjs(value);
-            return displayVisitStopDateTimeFields ? true : startDate.isSameOrBefore(today, 'day');
```
denniskigen, line 160: This is the actual fix. The rest of the diff just reorganizes imports.

brandones (dismissed):
Looks good! Can we also fix the logic that sets `maxVisitStartDateTime`, so that future dates are not selectable in the datepicker?

denniskigen:
> Looks good! Can we also fix the logic that sets `maxVisitStartDateTime`, so that future dates are not selectable in the datepicker?

AFAICT, there are already multiple safeguards that prevent users from selecting future visit start dates:

1. The visit form schema has a specific validation for the start date:

```tsx
visitStartDate: z.date().refine(
  (value) => {
    const today = dayjs();
    const startDate = dayjs(value);

    return startDate.isSameOrBefore(today, 'day');
  },
  t('invalidVisitStartDate', 'Start date needs to be on or before {{firstEncounterDatetime}}', {
    firstEncounterDatetime: formatDatetime(new Date()),
    interpolation: {
      escapeValue: false,
    },
  }),
),
```

2. There's also a validation function for the start time:

```tsx
const validateStartTime = (data: z.infer<typeof visitFormSchema>) => {
  const [visitStartHours, visitStartMinutes] = convertTime12to24(data.visitStartTime, data.visitStartTimeFormat);
  const visitStartDatetime = new Date(data.visitStartDate).setHours(visitStartHours, visitStartMinutes);
  return new Date(visitStartDatetime) <= new Date();
};
```

Regarding the specific logic for `maxVisitStartDatetime`:

- It's only used when editign an existing visit with encounters.
- It takes the earliest encounter dateteim `Math.min(...allEncounterDatetimes)` to ensure that the visit start date cannot be after the first encounter date.
- This is an additional constraint on top of the `no future dates` validation.

So the logic as-is should not allow selecting a future visit start date.

ibacher:
So, I two things:

* `maxVisitStartDatetime` came out of the need to edit retrospective visits and ensure that the visit date / time wasn't set to a point _after_ any encounter that was part of the visit started (since the backend would reject that). It's a poorly named variable, though.
* While I get that the validation logic catches the issue, I think just _not_ being able to select invalid dates is much better UX, since it prevents accidental errors. I don't think that the current definition of the `maxVisitStartDatetime` is the right way to do it, but maybe we change the `VisitDateTimeField` so that the max value is either `maxVisitStartDatetime` or `new Date()`

Otherwise, LGTM.

ibacher (approved):

PR #2175, opened 2025-01-10
(fix) O3-4286: Fix error when starting a visit immediately after ending one

Muppasanipraneeth:
## Requirements Checklist
- [x] PR title includes ticket number and follows conventional commit format
- [x] Code follows OpenMRS 3.0 Styleguide and design documentation
- [x] Changes are covered by existing or new tests
# Summary
Fixed an error that occurs when starting a new visit immediately after ending another one.

## Issue Description
The problem occurs when two visits are created within the same minute. Here's the scenario:

1. **First Visit (Visit A)**
   - Start time: `2025-01-11T02:33:00.000Z`
   - End time: `2025-01-11T02:33:33.497Z`

2. **Second Visit (Visit B)**
   - Start time: `2025-01-11T02:33:00.000Z`

The system was throwing a "visitCannotOverlapAnotherVisitOfTheSamePatient" error because:
- The payload only included hours and minutes in the datetime
- Even though the visits had different seconds, this precision was lost
- This made the system think the visits were overlapping

### Solution
Include seconds in the datetime payload. Example of working times:

1. **First Visit (Visit A)**
   - Start time: `2025-01-11T02:43:26.000Z`
   - End time: `2025-01-11T02:43:45.330Z`

2. **Second Visit (Visit B)**
   - Start time: `2025-01-11T02:43:51.000Z`

## Before the fix
https://github.com/user-attachments/assets/335d00e3-c74a-48bd-a273-f6f0a61b1764
## After the fix
https://github.com/user-attachments/assets/5a0a1de1-6f20-4281-a1ae-3b87f10696ea

## Related Items
- Closed PR: [openmrs/openmrs-esm-patient-chart#2147](https://github.com/openmrs/openmrs-esm-patient-chart/pull/2147)


packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -145,8 +145,9 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
     // Validates that the start time is not in the future
     const validateStartTime = (data: z.infer<typeof visitFormSchema>) => {
       const [visitStartHours, visitStartMinutes] = convertTime12to24(data.visitStartTime, data.visitStartTimeFormat);
-      const visitStartDatetime = new Date(data.visitStartDate).setHours(visitStartHours, visitStartMinutes);
-      return new Date(visitStartDatetime) <= new Date();
+      const visitStartDatetime = new Date(data.visitStartDate);
+      visitStartDatetime.setHours(visitStartHours, visitStartMinutes, 0, 0);
+      return visitStartDatetime <= new Date();
```
ibacher, line 150: Doesn't this just make things worse, since we've zeroed out the seconds and milliseconds?

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.component.tsx
```diff
@@ -436,6 +437,8 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
               dayjs(visitStartDate).date(),
               hours,
               minutes,
+              dayjs(visitStartDate).second(),
+              dayjs(visitStartDate).millisecond(),
```
ibacher, line 441: With this change, why do we need the whole `new Date()` construct at all?

packages/esm-patient-chart-app/src/visit/visit-prompt/end-visit-dialog.component.tsx
```diff
@@ -16,13 +16,29 @@ const EndVisitDialog: React.FC<EndVisitDialogProps> = ({ patientUuid, closeModal
   const { currentVisit, currentVisitIsRetrospective, mutate } = useVisit(patientUuid);
   const { queueEntry } = useVisitQueueEntry(patientUuid, currentVisit?.uuid);
 
+  const calculateStopDatetime = (): Date => {
+    const MinVisitdurMs = 1000;
```
ibacher, line 20: * Variables should be something like `minVisitDurationMs`
* We shouldn't be hard-coding this at all
* I don't think this helps solve the problem in the ticket which is starting a new visit after ending the current one. This makes an adjustment to the visit stop date based on the _current_ visit start date, but the visit start date we care about is the _new_ visit. The form itself should already enforce the rule that `stopTime >= startTime` for a visit.

ibacher (changes requested):
Thanks @Muppasanipraneeth. I'm not sure I understand from the code or the explanation in the PR how this addresses the issue in the ticket. This seems to add unnecessary complexity.

The issue occurs when I stop Visit 1 and then immediately try to start Visit 2, Visit 2 gets rejected for overlapping with Visit 1.

NethmiRodrigo:
@Muppasanipraneeth please update your branch and resolve the conflicts

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.workspace.tsx
```diff
@@ -876,4 +879,4 @@ const VisitFormExtensionSlot: React.FC<VisitFormExtensionSlotProps> = React.memo
   },
 );
 
-export default StartVisitForm;
+export default StartVisitForm;
```
ibacher, line 882:
```suggestion
export default StartVisitForm;

```

packages/esm-patient-chart-app/src/index.ts
```diff
@@ -24,6 +24,7 @@ import patientChartPageComponent from './root.component';
 import patientDetailsTileComponent from './patient-details-tile/patient-details-tile.component';
 import startVisitActionButtonComponent from './actions-buttons/start-visit.component';
 import startVisitActionButtonOnPatientSearch from './visit/start-visit-button.component';
+import startVisitFormComponent from './visit/visit-form/visit-form.workspace';
```
ibacher, line 27: Why are we adding this import?

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.workspace.tsx
```diff
@@ -142,8 +142,9 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
     // Validates that the start time is not in the future
     const validateStartTime = (data: z.infer<typeof visitFormSchema>) => {
       const [visitStartHours, visitStartMinutes] = convertTime12to24(data.visitStartTime, data.visitStartTimeFormat);
-      const visitStartDatetime = new Date(data.visitStartDate).setHours(visitStartHours, visitStartMinutes);
-      return new Date(visitStartDatetime) <= new Date();
+      const visitStartDatetime = new Date(data.visitStartDate);
+      visitStartDatetime.setHours(visitStartHours, visitStartMinutes, 0, 0);
```
ibacher, line 146: I'm not sure that zeroing out seconds is always the right thing to do here. Can you explain why we're doing so?

ibacher: I see. So the limiting factor here is the time picker component? I guess that makes sense.

ibacher: But, then, shouldn't we either zero out the seconds component when sending this to the backend or use the seconds (and milliseconds) from the current timestamp? Otherwise, we're effectively sending data the user didn't supply to the backend and that's what gets stored (not to mention, we have a similar validation check on the backend that might fail with a quick enough connection).

Muppasanipraneeth: @ibacher sir any update

ibacher: I actually think the better thing to do is just to drop this validation rule altogether.

Muppasanipraneeth: > I actually think the better thing to do is just to drop this validation rule altogether.

ok sir

Muppasanipraneeth:
updated @ibacher sir

NethmiRodrigo (approved):
LGTM, thanks @Muppasanipraneeth! @denniskigen @ibacher is this good to go?

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.workspace.tsx
```diff
@@ -439,6 +439,7 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
               dayjs(visitStartDate).date(),
               hours,
               minutes,
+              currentSeconds,
```
NethmiRodrigo, line 442: @Muppasanipraneeth why do we do `new Date().getSeconds()` for the seconds in the start date and `dayjs(visitStopDate).second()` for the seconds in the stop date? What I mean is shouldn't both use the same way, either both using the javascript date function or dayjs to get seconds?

Muppasanipraneeth: image
 I did first only @NethmiRodrigo  but sir told use current seconds

Muppasanipraneeth: i will update

NethmiRodrigo: Does that mean that one of the methods doesn't give the current second?

NethmiRodrigo (approved):
LGTM, thanks @Muppasanipraneeth!

denniskigen (approved):

Muppasanipraneeth:
thanks @denniskigen  @NethmiRodrigo  @ibacher  😊

PR #2183, opened 2025-01-10
(fix) visit-form: Fix visit stop datetime field validations

denniskigen:
## Requirements

## Summary

This PR fixes validations in the visit form's stop date and stop time field in edit mode. Presently, when the user selects a stop date, that value does not get included in the payload sent to the backend. That's because logic exists that ensures that the stop datetime only gets included in the payload if the user selects a stop date, stop time, and stop time format. All of these fields are required to be filled out for the backend to accept the payload. This PR adds validation logic that ensures that if a stop date is selected but a stop time and stop time format are not, the user gets shown validation errors. This makes it clear to the user that they need to fill out all the required fields.

## Screenshots

https://github.com/user-attachments/assets/aad5820b-26c0-481d-9c5c-5d76abf61db5

## Related Issue

https://openmrs.atlassian.net/browse/O3-2720


packages/esm-patient-chart-app/translations/en.json
```diff
@@ -16,7 +16,7 @@
   "checkFilters": "Check the filters above",
   "close": "Close",
   "confirm": "Confirm",
-  "confirmDeletingVisitTextWithStartAndEndDate": "Are you sure you want to delete {{visit}} which started {{visitStartDate}} and ended {{visitEndDate}}?",
+  "confirmDeletePastVisitText": "Clicking confirm will delete this {{visit}} and all of its associated encounters.",
```
denniskigen, line 19: This fixes the text of the delete visit modal, which looks very awkward at the moment:

CleanShot 2025-01-10 at 1  25 53@2x

packages/esm-patient-chart-app/src/visit/visits-widget/visit-detail-overview.component.tsx
```diff
@@ -80,7 +80,9 @@ function VisitDetailOverviewComponent({ patientUuid }: VisitOverviewComponentPro
                 ))}
               </>
             ) : (
-              <EmptyState headerTitle={t('visits', 'visits')} displayText={t('Visits', 'Visits')} />
+              <div className={styles.emptyStateContainer}>
```
denniskigen, line 83: This fixes the spacing around the empty state, adding a consistent 1rem vertical margin around the container.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.workspace.tsx
```diff
@@ -739,22 +751,22 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
...
+                        role="alert"
+                        style={{ margin: '0', minWidth: '100%' }}
+                        kind="error"
+                        lowContrast={true}
+                        title={t('missingVisitType', 'Missing visit type')}
+                        subtitle={t('selectVisitType', 'Please select a Visit Type')}
+                      />
+                    </div>
+                  </section>
+                )}
```
denniskigen, line 769: Moves the inline notification higher up so it looks more closely connected to the LocationSelector component above it:

CleanShot 2025-01-11 at 2  02 56@2x

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.workspace.tsx
```diff
@@ -659,17 +669,19 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
...
-              <div className={styles.sectionField}>
-                <VisitFormExtensionSlot
-                  name="visit-form-top-slot"
-                  patientUuid={patientUuid}
-                  visitFormOpenedFrom={openedFrom}
-                  setVisitFormCallbacks={setVisitFormCallbacks}
-                />
-              </div>
-            </section>
+            {config.showUpcomingAppointments && (
```
denniskigen, line 672: I assume this should only be shown when the `config.showUpcomingAppointments` is truthy per the comment above, right, @chibongho?

packages/esm-patient-chart-app/src/visit/hooks/useDeleteVisit.hook.tsx
```diff
@@ -42,9 +42,9 @@ export function useDeleteVisit(patientUuid: string, visit: Visit, onVisitDelete
       .then(() => {
         mutateVisits();
         mutateCurrentVisit();
-        // TODO: Needs to be replaced with Actionable Snackbar when Actionable
+
         if (!isCurrentVisitDeleted) {
-          showActionableNotification({
```
denniskigen, line 47: Replaces the old `ActionableNotification` component with a `Snackbar`.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.workspace.tsx
chibongho: Yes. I actually wanted to get rid of the `showUpcomingAppointments` config value from patient-chart, and move it to the appointments app instead, since it is appointments specific logic.  (same with `showServiceQueueFields` to the queues app). We could do it if the `<section> ... </section>` is part of the extension, but that might also mess up the CSS as the extension is wrapped in a `<div>` when its gets slotted in.

That said, I hope this will be moot and we can get rid of this extension soon. The upcoming RDE mockup strongly implies that the start visit UX should be separate from UX to create appointments / queues etc...

chibongho (approved):
LGTM!

PR #2206, opened 2025-01-20
(fix): Order reasons not displaying in dropdown list

makombe:
## Requirements

## Screenshots

 Before:
Screenshot from 2025-01-20 12-10-35

After:
Screenshot from 2025-01-20 12-11-22

### DIFF
--- a/packages/esm-patient-tests-app/src/test-orders/add-test-order/test-order-form.component.tsx
+++ b/packages/esm-patient-tests-app/src/test-orders/add-test-order/test-order-form.component.tsx
@@ -118,7 +118,7 @@ export function LabOrderForm({
   const { orderReasons } = useOrderReasons(orderReasonUuids);
 
   const filterItemsByName = useCallback((menu) => {
-    return menu?.item?.value?.toLowerCase().includes(menu?.inputValue?.toLowerCase());
+    return menu?.item?.display?.toLowerCase().includes(menu?.inputValue?.toLowerCase());
   }, []);
 
   const handleFormSubmission = useCallback(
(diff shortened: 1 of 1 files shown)


ojwanganto (approved):

packages/esm-patient-tests-app/src/test-orders/add-test-order/test-order-form.component.tsx
```diff
@@ -118,7 +118,7 @@ export function LabOrderForm({
   const { orderReasons } = useOrderReasons(orderReasonUuids);
 
   const filterItemsByName = useCallback((menu) => {
-    return menu?.item?.value?.toLowerCase().includes(menu?.inputValue?.toLowerCase());
+    return menu?.item?.display?.toLowerCase().includes(menu?.inputValue?.toLowerCase());
```
vasharma05, line 121: Shouldn't we have a check that the string matching should only happen when the inputValue is a valid and non-empty string?

makombe: Thanks @vasharma05. Updated

denniskigen:
Thanks for the fix, @makombe.

Generally speaking, we shouldn't rely on the [shouldFilterItem](https://react.carbondesignsystem.com/?path=/docs/components-combobox--overview#shouldfilteritem) prop to filter items in a ComboBox based on user input. While setting this prop does filter items as users type (showing only matching items in the dropdown), a problem occurs when clicking the ComboBox again to select a different item - the dropdown will only contain the selected item. Other items are excluded from the list unless the user explicitly clears the ComboBox selection by clicking the 'x' button.

This behavior is not a bug; it's by design. This can be verified by observing both the default example in the [shouldFilterItem](https://react.carbondesignsystem.com/?path=/docs/components-combobox--overview#shouldfilteritem) docs in Carbon's Storybook page for the ComboBox component and the underlying `Downshift` library's [useComboBox](https://www.downshift-js.com/use-combobox) hook implementation.

Until this behavior changes upstream, it's preferable to remove the prop altogether and rely on the default filtering behavior of the ComboBox, where a blue outline is automatically applied to exact matches as the user types into the field.

donaldkibet:
Merging this in. Thanks everyone for the reviews

PR #2208, opened 2025-01-21
(fix): format observation datetime to accurate local time to prevent inaccurate timestamps

amosmachora:
## Requirements

## Summary
This PR formats the test timestamp in the results panel using the openmrs framework `formatDatetime` function that accurately handles the users locale. Since the date we had was already formatted to a certain format we first convert it to an ISO string _(which was the original format)_ then format it correctly to the users local time removing the inconsistency.

### Results
image

### Visits
image

## Related Issue

https://openmrs.atlassian.net/browse/O3-4297


packages/esm-patient-tests-app/src/test-results/filter/filter-context.tsx
```diff
@@ -78,7 +80,14 @@ const FilterProvider = ({ roots, children }: FilterProviderProps) => {
     const allTimes = [
       ...new Set(
         Object.values(tests)
-          .map((test: ReducerState['tests']) => test?.obs?.map((entry) => entry.obsDatetime))
+          .map(
+            (test: ReducerState['tests']) =>
+              test?.obs?.map((entry) => {
+                const isoFormattedString = entry.obsDatetime.replace(' ', 'T').replace('.0', '.000').concat('Z');
+                const date = new Date(isoFormattedString);
+                return format(date, 'yyyy-MM-dd HH:mm:ss');
```
donaldkibet, line 88: Why not use the framework `formatDatetime` here?

packages/esm-patient-tests-app/src/test-results/trendline/trendline.component.tsx
```diff
@@ -117,6 +117,11 @@ const Trendline: React.FC<TrendlineProps> = ({
 
   const dataset = chartTitle;
 
+  const formatObsDatetime = (obsDatetime: string) => {
+    const isoFormattedString = obsDatetime.replace(' ', 'T').replace('.0', '.000').concat('Z');
+    return isoFormattedString;
+  };
```
donaldkibet, line 123:
```suggestion
  const formatObsDatetime = (obsDatetime: string) => {
  return dayjs(obsDatetime).toISOString();
};
```
dayjs already provides this conversion to ISO format.

donaldkibet (changes requested):

packages/esm-patient-tests-app/src/test-results/filter/filter-context.tsx
amosmachora: I did not want to break the existing functionality. In my investigation using `formatDatetime` returns dates like `Today 06:50 am` which was not my intention. I wanted to return a string like this `2024-09-04 06:23:23.0` so that wherever the value is being used it doesn't break anything. In short I wanted to return the original date but formatted correctly,

ojwanganto: I vote for consistency across the system and I vote for the formatters provided by the framework. Is there a strong reason not to go with the formatted date from the framework's function?

amosmachora: Yes @ojwanganto that date object at that point is not for display but rather to make sure they are in the correct timezone and that approach is what made it possible. using `formatDateTime` or any other format function from the framework did not quite fit the use case. 

I am open to suggestions though.

packages/esm-patient-tests-app/src/test-results/trendline/trendline.component.tsx
amosmachora: The `obsDatetime` parameter is weird. It's an already formatted date string like this: `2025-01-22 05:28:00.0` Following your suggestion rolls back the date by 3 hours (probably because I am on EAT). 

```
obsDatetime - 2025-01-22 05:28:00.0
isoFormattedString - 2025-01-22T05:28:00.000Z 
formattedWithDayJS - 2025-01-22T02:28:00.000Z
```
thanks for highlighting though I have seen a use case for it on the tooltip for accuracy purposes

packages/esm-patient-tests-app/src/test-results/filter/filter-context.tsx
ibacher: I think it's fine to operate on time objects, but please use `dayjs()`, which supports everything you're trying to do here in a much cleaner way and is already part of the framework.

packages/esm-patient-tests-app/src/test-results/filter/filter-context.tsx
```diff
@@ -10,6 +10,8 @@ import {
 } from './filter-types';
 import reducer from './filter-reducer';
 import { type MappedObservation, type TestResult, type GroupedObservation, type Observation } from '../../types';
+import { formatDatetime } from '@openmrs/esm-framework';
```
ibacher, line 13: Please remove unused imports.

packages/esm-patient-tests-app/src/test-results/trendline/trendline.component.tsx
ibacher: So, yes, because `dayjs` interprets it as a local timestamp (which is what that format claims it is). Instead, you're adding an adjustment so this is treated as if it were in UTC, which is not something we should be deciding here, but on the backend.

denniskigen:
@amosmachora did you see Ian's latest comments on the ticket?

packages/esm-patient-tests-app/src/test-results/trendline/trendline.component.tsx
mogoodrich: Agree with @ibacher ... I'm forgetting, when does the server return something like "2025-01-22 05:28:00.0" (ie without specifying time zone).

amosmachora: @mogoodrich highly likely that it does not. This is a modification that is most certainly done at the front end, somewhere up the "callstack".

amosmachora:
> @amosmachora did you see Ian's latest comments on the ticket?

Yes I did Dennis, not gonna lie worked out perfectly for me because I had no clue how to rewrite the code with the dayjs implementation that Ian was suggesting : )

packages/esm-patient-tests-app/src/test-results/trendline/trendline.component.tsx
denniskigen: The obstree resource @mogoodrich

amosmachora: didn't know that @denniskigen

mogoodrich: We should certainly consider fixing that @amosmachora  @denniskigen , where does the obstree resource live?

denniskigen: I'm referring to [this guy](https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/ObsTreeResource1_9.java#L45) in the REST-WS module.

mogoodrich: thanks @denniskigen  !

@ibacher  @dkayiwa  I'm thinking we want to fix this?  specifically, when generating the value_datetime and obs_datetime here:

https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/ObsTreeResource1_9.java#L153

... just format in the standard ISO string format instead?

ibacher: @mogoodrich Yep. We do need to make allowance for, e.g., implementations that have been storing time in UTC rather than local time. But, yeah, ideally, all conversations through the API are done in ISO timestamps and the ambiguity should mostly go away.

mogoodrich: Thanks @ibacher ... isn't that even more of a reason to include an ISO string here?  Stripping off the timezone seems to assume that the server time zone is the facility timezone, and we just want to display that time to the user, regardless of what timezone their client is in.  (We've done that in the past with things like this: https://github.com/openmrs/openmrs-module-uicommons/blob/master/omod/src/main/webapp/resources/scripts/filters/serverDate.js but this has become problematic when the server is in UTC or some timezone that don't correspond to the clinic timezone)

denniskigen:
@amosmachora, it looks like @mogoodrich's [change](https://github.com/openmrs/openmrs-module-webservices.rest/commit/644c6f09d91eebb967a86ca1e90d11c7070cad11) to the REST Web services module got checked in. Do you wanna have another look at this?

denniskigen:
Based on my testing, this seems to have been resolved fully by @mogoodrich's changes.

amosmachora:
@denniskigen if @mogoodrich changes correctly, fix this issue. Do you think we will need this PR?

denniskigen:
No, we don't. We can close it out. Thanks again, everyone!

PR #2210, opened 2025-01-23
(fix) O3-4375: Make useVisits conditional

NethmiRodrigo:
## Requirements

## Summary

This PR makes the `useVisits` hook conditional so that the API call does not happen if `patientUUID` is null, and introduces a new function `invalidateUseVisits` so that mutate can be called for `useVisits` without having to call the useVisits hook (which would call the API).

## Related Issue

https://issues.openmrs.org/browse/O3-4375


ibacher (approved):
Thanks!

PR #2217, opened 2025-01-28
(fix) O3-4382: Fixing the reset button functionality

UNCANNY69:
## Requirements

## Summary
In the initial implementation, the FilterProvider component was being re-created whenever the parent component (TreeViewWrapper) re-rendered. This caused the state within the FilterProvider to reset, even when no changes were made to its inputs. The observed flow of console.log statements showed that the FilterProvider was being re-rendered unnecessarily, leading to unintended side effects like resetting the checkbox values stored in the filter state.

The root cause was that the FilterProvider component's re-creation was linked to the lifecycle of TreeViewWrapper, which was itself re-rendering due to changes in its own state or props.

**Solution:**
Initially it led me to think roots were needed to be memoized the roots , but for the life of me I am not sure why it did not work thus I used _React.memo_ as it memoizes the rendered output of the FilterProvider and skips re-renders unless its props change. By memoizing it, the FilterProvider remains the same instance across renders unless the roots prop is updated.

## Screenshots

https://github.com/user-attachments/assets/2a8f0739-53eb-452e-81a3-4b2ec64f59b4


packages/esm-patient-tests-app/src/test-results/tree-view/tree-view-wrapper.component.tsx
```diff
@@ -24,12 +24,12 @@ const TreeViewWrapper: React.FC<TreeViewWrapperProps> = (props) => {
   const { roots, isLoading, error } = useGetManyObstreeData(conceptUuids);
 
   if (error) return <ErrorState error={error} headerTitle={t('dataLoadError', 'Data load error')} />;
-
+  const MemoizedFilterProvider = React.memo(FilterProvider);
```
denniskigen, line 27: Because `MemoizedFilterProvider` is declared within the `TreeViewWrapper` component:

- A new component type is created on every render of `TreeViewWrapper`
- React will unmount the older provider and mount a new one each time `TreeViewWrapper` re-renders
- Any state in the subtree of `FilterProvider` will reset on re-render

We should move the memoized component outside the parent to preserve its identity as a stable reference:

```tsx
const MemoizedFilterProvider = React.memo(FilterProvider); 

const TreeViewWrapper: React.FC<TreeViewWrapperProps> = (props) => {
  // ... rest of the code
  return <MemoizedFilterProvider roots={...} />;
};
```

denniskigen (commented):
Thanks for the attempt @UNCANNY69. While the memoization suggestion helps, the core issue appears deeper in the `FilterContext` flow. The `RESET_TREE` action in the `FilterContext` reducer fires correctly but fails to clear the checkboxes. There seems to be an issue with the `FilterProvider` state initialization that prevents re-initialization when the roots change.

To test this hypothesis out, try adding this force reset to `FilterProvider` that fires when `roots` changes:

```tsx
   useEffect(() => {
     dispatch({ type: ReducerActionType.RESET_TREE });
   }, [roots]);
```

You'll see that this resets the filters. There's still some exploration to be done, but I think it's better to focus there for the moment.

UNCANNY69:
> Thanks for taking a stab at this, @UNCANNY69. While the memoization suggestion helps, the core issue appears to be deeper within the `FilterContext` flow. The `RESET_TREE` action in the `FilterContext` reducer fires correctly but fails to clear the checkboxes. There seems to be an issue with the `FilterProvider` state initialization that prevents re-initialization when the roots change.
> 
> To test this hypothesis out, try adding this force reset to `FilterProvider` that fires when `roots` changes:
> 
> ```tsx
>    useEffect(() => {
>      dispatch({ type: ReducerActionType.RESET_TREE });
>    }, [roots]);
> ```
> 
> You'll see that this resets the filters. There's still some exploration to be done, but I think it's better to focus there for the moment.

This did not cause the reset to work @denniskigen cause initially I tried to memorize the roots which also led to similar behaviour

packages/esm-patient-tests-app/src/test-results/tree-view/tree-view-wrapper.component.tsx
UNCANNY69: I am trying to exactly use the above mentioned behaviour that would happen on putting MemoizedFilterProvider  within the TreeViewWrapper , otherwise if I put it outside the reset wont work again this is not a elegant solution and will look more into . Can I possible shift the location of the reset button to TreeView that makes the button work with correct flow of events?

UNCANNY69:
@denniskigen  I made the roots stable , thus this code below stopped working but the `FilterProvider` still retriggers

 ```
  useEffect(() => {
     dispatch({ type: ReducerActionType.RESET_TREE });
   }, [roots]);
```

can I try to change the location of the reset button of move the state to an upper level?

PR #640, opened 2022-04-06
(enhancement) use `showModal` function to display visit-modals

donaldkibet:
## Requirements

- [x] This PR has a title that briefly describes the work done, including the ticket number if there is a ticket.
- [ ] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide).
- [ ] My work includes tests or is validated by existing tests.

## Summary
1. Update the current visit-modal i.e `start-visit` `end-visit` and `cancel-visit` to use `showModal` to be displayed by `showModal` functionality. This removes the need to add an event listener and `useVisitDialog` hook to determine when the modals appear.

PR #2222, opened 2025-01-30
(feat) O3-4363 - Add Visit context header to Order, Visit Note, and F…

chibongho:
…orms workspaces

## Requirements

## Summary

This PR adds RDE features: a new visit context header to the patient chart workspaces (order / visit note / clinical forms), and a new visit switcher modal accessible from the visit context header. 

The concept of a "Visit Context" when viewing a patient chart is not new; previous RDE work has that when the `rde` feature flag is turned on). Screenshots from **before** this change:

Untitled-1
Untitled

Instead of rendering in the top nav, the visit context header is rendered at the top of the workspace with this change, See the [mockups here](https://app.zeplin.io/project/60d59321e8100b0324762e05/dashboard?sid=67975feb441fd9cac835648a). 

The new changes are behind the existing `rde` feature flag, so they should not be noticeable for most people.

Notes:
- I'm keeping the old visit context rendered (red in screenshot above) at the top nav for now. This is because the End visit / Cancel visit buttons in the patient banner (yellow in screenshot) modifies the visit in context. With our new mockups, Since we're moving towards having the visit context be within the workspace (instead of the entire patient chart), this will probably change. [Discussion here](https://openmrs.atlassian.net/browse/O3-4361?focusedCommentId=146002).
- RDE doesn't actually work yet in the affected workspaces. That said, they were no more broken that before. Here's what I tested with the visit context set to a past visit:
  - Visit note workspace. The form actually has a date picker (but no time picker?) to select the visit note date (`encounterDatetime`), but it doesn't work. See: https://openmrs.atlassian.net/browse/O3-4420
  - Orders workspace. I was able to place an order to a past visit. The created encounter has its `encounterDateTime` set to the visit's start time. However the order's `dateActivated` field is set to `now`. Is that expected?
  - Clinical forms. I tried to fill out 2 forms that come with the refapp:
    - "Structured SOAP Note" form. This form already has fields to ask for the encounter datetime. The created encounter has the specified encounter datetime, but is not tied to a visit.
    - "Ward admission" form. This form does not ask for the encounter datetime. The created encounter defaults to `now`, and is tied to the active visit, instead of the past visit in the visit context.

## Screenshots

The visit context header in a workspace:
image

The modal that opens when clicking on "change" in the visit context header. The same modal also shows when the user tries to open the order / visit notes / forms workspace without a visit (active or past) in context.
image

## Related Issue

https://openmrs.atlassian.net/browse/O3-4363


packages/esm-patient-chart-app/src/visit/visit-prompt/start-visit-dialog.component.tsx
```diff
@@ -1,37 +1,17 @@
...
 import { useTranslation } from 'react-i18next';
-import { Button, ModalBody, ModalHeader, ModalFooter } from '@carbon/react';
-import { launchPatientWorkspace, launchPatientChartWithWorkspaceOpen } from '@openmrs/esm-patient-common-lib';
-import { useFeatureFlag } from '@openmrs/esm-framework';
 import styles from './start-visit-dialog.scss';
 
 interface StartVisitDialogProps {
   patientUuid: string;
   closeModal: () => void;
-  visitType: string;
```
chibongho, line 11: Removing this as it doesn't seem to be used. This was brought up before: 
https://github.com/openmrs/openmrs-esm-patient-chart/pull/640/files#diff-7e621046e4d227da306880633891cb2d35e4f168f95967ff73f00aa973bc6d06R10

vasharma05:
Hi @chibongho!
Thanks for working on this. I have a suggestion to re-modal all this. Can we implement the header context in the workspace system itself? We can pass a property in `additionalProps` when calling the `launchWorkspace` whether to show the visit context or not. If yes, we can render the visit context as an extension.

OR

We can create an extension slot based on the prop mentioned above (rendered by the workspace system) and attach the RDE visit context extension (from the patient chart) or I believe it'd be better to handle all RDE stuff from a separate app (@openmrs/esm-patient-rde-app), since these changes are not only limited to the patient chart but also to ward patient workspace as well.

TLDR, 2 singular points for having the extension slot being rendered (workspace renderer) and extension (from patient chart or the patient rde app, whichever suits better)

Do let me know your thoughts on this.
Thanks!

packages/esm-patient-orders-app/src/order-basket/order-basket.workspace.tsx
```diff
@@ -110,8 +118,11 @@ const OrderBasket: React.FC<DefaultPatientWorkspaceProps> = ({
     closeWorkspace({ onWorkspaceClose: clearOrders });
   }, [clearOrders, closeWorkspace]);
 
+  const isRdeEnabled = useFeatureFlag('rde');
+
   return (
     <>
+      {isRdeEnabled && <ExtensionSlot name="visit-context-header-slot" state={{ patientUuid }} />}
```
vasharma05, line 125: This is what I am talking about, why should we go manually into every patient chart workspace and render the extension slot, when we can use the workspace system to do the same?

packages/esm-patient-forms-app/src/forms/forms-dashboard.workspace.tsx
```diff
@@ -2,10 +2,15 @@ import React from 'react';
...
 import { type DefaultPatientWorkspaceProps } from '@openmrs/esm-patient-common-lib';
+import { ExtensionSlot, useFeatureFlag } from '@openmrs/esm-framework';
 
 export default function FormsWorkspace(props: DefaultPatientWorkspaceProps) {
+  const isRdeEnabled = useFeatureFlag('rde');
+  const { patientUuid } = props;
+
   return (
     <div className={styles.container}>
+      {isRdeEnabled && <ExtensionSlot name="visit-context-header-slot" state={{ patientUuid }} />}
```
vasharma05, line 13: Same here

packages/esm-patient-chart-app/src/visit/visits-widget/visit.resource.tsx
```diff
@@ -1,56 +1,28 @@
...
   useConfig,
+  useOpenmrsInfinite,
   type OpenmrsResource,
   type Privilege,
   type Visit,
 } from '@openmrs/esm-framework';
+import useSWR, { mutate } from 'swr';
 import { type ChartConfig } from '../../config-schema';
 
 export function useInfiniteVisits(patientUuid: string) {
```
vasharma05, line 14: Why re-working on what is already implemented in the `useVisit` hook

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context.ts
```diff
@@ -0,0 +1,12 @@
+import { getVisitStore, useStoreWithActions, type Visit } from '@openmrs/esm-framework';
+
+export function useVisitContextStore() {
+  return useStoreWithActions(getVisitStore(), {
```
vasharma05, line 4: Shouldn't this be moved into the react utils? Also, isn't this indirectly `useVisit`?

vasharma05 (commented):
I am seeing a lot of duplicated work which can be done in the `useVisit` hook.
Alongside, we should work on the workspace system to render an extension slot if we pass the slot names. Hence we won't have to duplicate our effort in every patient workspace and directly use the workspace system to do out work. 

A sample example will be:

```
import {launchWorkspace} from '@openmrs/esm-framework';

launchPatientWorkspace = (workspaceName, additionalProps) => launchPatientWorkspace(workspaceName, {...additionalProps, renderExtensionSlots: ['visit-context-header-slot']}
```

vasharma05 (dismissed):
I’ve noticed a lot of repetitive work that could be handled using the `useVisit` hook. Additionally, we should improve the workspace system to render an extension slot when we provide the slot names. This way, we won’t have to duplicate our efforts in each patient workspace and can directly use the workspace system to accomplish our tasks.

Here’s a sample example:
```
import {launchWorkspace} from '@openmrs/esm-framework';

launchPatientWorkspace = (workspaceName, additionalProps) => launchPatientWorkspace(workspaceName, {...additionalProps, renderExtensionSlots: ['visit-context-header-slot']}
```

packages/esm-patient-chart-app/src/visit-header/retrospective-visit-label.component.tsx
```diff
@@ -10,6 +10,9 @@ interface RetrospectiveVisitLabelProps {
 
 const RetrospectiveVisitLabel: React.FC<RetrospectiveVisitLabelProps> = ({ currentVisit }) => {
   const { t } = useTranslation();
+  if (!currentVisit) {
+    return <></>;
+  }
```
mogoodrich, line 15: We only render the retrospective entry toggle if it's *not* the current visit?    (I'm probably just not understanding this correctly and it's legacy functionality anyway?)

mogoodrich: ah, current visit means the visit in the current context, not the active visit?

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-header.component.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+interface VisitContextHeaderProps {
+  patientUuid: string;
+}
+
+const VisitContextHeader: React.FC<VisitContextHeaderProps> = ({ patientUuid }) => {
+  const { t } = useTranslation();
+  const { currentVisit, isLoading } = useVisit(patientUuid);
+  const { manuallySetVisitUuid, setVisitContext } = useVisitContextStore();
+  const isActiveVisit = !Boolean(currentVisit && currentVisit.stopDatetime);
```
mogoodrich, line 18: It occurs to me that there's likely to be confusion where people think current visit = active visit (see my confusion above) though a better name hasn't come to mind for me yet.

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-header.component.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  // This hook is needed because the `useVisit` hook sometimes
+  // returns a currentVisit that isn't actually the one in the visit context.
+  // TODO: move this into the useVisit hook
+  useEffect(() => {
```
mogoodrich, line 23: Okay... :)  Hopefully this will make more sense to me when we get to the the visit context hook... "manuallySetVisit" isn't super intuitive...

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-info.component.tsx
```diff
@@ -0,0 +1,35 @@
...
+const VisitContextInfo: React.FC<VisitContextInfoProps> = ({ visit }) => {
+  const { t } = useTranslation();
+
+  if (visit) {
+    const isActive = !Boolean(visit.stopDatetime);
+
+    return (
+      <>
+        {isActive
+          ? t('currentActiveVisit', 'Current active visit')
```
mogoodrich, line 20: I guess this is end-user-facing, but this does continue to propagate the confusion about what "current' means.

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context.ts
```diff
@@ -0,0 +1,12 @@
...
+    setVisitContext(_, newSelectedVisit: Visit) {
+      return {
+        manuallySetVisitUuid: newSelectedVisit.uuid,
```
mogoodrich, line 7: Can you explain how this is different from other ways the visit context visit is set?

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-header.component.tsx
mogoodrich: Okay, I still don't understand after reviewing the code... what is the case when a currentVisit isn't actually on the one in the visit context?

mogoodrich (commented):
Thanks @chibongho , generally looks good, though admittedly this was a lot to review and not something I'm able to dive super deeply into.

I asked some questions, but mainly to better understand naming conventions.

At the end of the day, it's probably best to get this in sooner rather than later and start testing/playing around with it, but it would be to good to confirm with @ibacher  and @denniskigen  that there aren'y any releases imminent, because this definitely risks introducing instabiliity.

packages/esm-patient-chart-app/src/visit-header/retrospective-visit-label.component.tsx
chibongho: Yes, the "current visit" is the visit in context. It's a term that's from prior RDE work. I added this if statement to prevent an NPE. This component is actually going away soon anyway, as we have [a ticket](https://openmrs.atlassian.net/browse/O3-4361) to remove the modified visit header in the top nav.

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-header.component.tsx
chibongho: `currentVisit` and the visit in context really should be the exact same thing. This code here is to get around a bug in `useVisit` where it defaults the `currentVisit` to the `activeVisit` if the visit store's `manuallySetVisitUuid` is null, but it doesn't update the store's `manuallySetVisitUuid` value accordingly.

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-info.component.tsx
chibongho: Yeah, this is from the mockup and I think the wording actually makes pretty good sense to the end user. I propose we keep this as is, but rename the `currentVisit` in the code base.

chibongho:
> Thanks @chibongho , generally looks good, though admittedly this was a lot to review and not something I'm able to dive super deeply into.
> 
> I asked some questions, but mainly to better understand naming conventions.
> 
> At the end of the day, it's probably best to get this in sooner rather than later and start testing/playing around with it, but it would be to good to confirm with @ibacher and @denniskigen that there aren'y any releases imminent, because this definitely risks introducing instabiliity.

Yeah, the naming convention right now is confusing. I tried to keep backwards compatibility by not modifying the hooks / stores (and their names) from previous RDE work. Looking at the code more, I think it's probably safe to change them, as the prior RDE work seemed incomplete, and is likely not used by anyone. (@ibacher and @denniskigen does that sound fair?)

So we can rename `currentVisit` to `visitInContext` and `manuallySetVisitUuid` to `visitInContextUuid`.

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context.ts
chibongho: There is already a `setCurrentVisit()` function from core to directly set the store's current visit. However, that function does not cause react updates. This hook is made to trigger react updates on setting the visit context.

mogoodrich:
> So we can rename `currentVisit` to `visitInContext` and `manuallySetVisitUuid` to `visitInContextUuid`.

This sounds great, thanks @chibongho !

mogoodrich (approved):
Thanks for the clarifications @chibongho !

chibongho:
> > So we can rename `currentVisit` to `visitInContext` and `manuallySetVisitUuid` to `visitInContextUuid`.
> 
> This sounds great, thanks @chibongho !

@mogoodrich OK, I went down the rabbit hole of renaming things and it's more intrusive than I expected. I did push my draft changes to my fork [here](https://github.com/openmrs/openmrs-esm-patient-chart/commit/e76e49405453c0e1baa21f798c7c1d29a8e1ed95) without any testing. It's straightforward enough, but it's big enough that I don't think it should be part of this PR.

mogoodrich:
> @mogoodrich OK, I went down the rabbit hole of renaming things and it's more intrusive than I expected. I did push my draft changes to my fork [here](https://github.com/openmrs/openmrs-esm-patient-chart/commit/e76e49405453c0e1baa21f798c7c1d29a8e1ed95) without any testing. It's straightforward enough, but it's big enough that I don't think it should be part of this PR.

Fine with me, thanks @chibongho !

mogoodrich:
> So what should we do about this "current visit" stuff, along with things in this PR that uses it more? If we want to preverse backends compatiblity, then I think we should stick with the approach of this PR (leaving useVisit() in core untouched and add more visit store functionality in patient chart). If we feel comfortable with breaking backward compatiblity, I think we can change esm-core's `useVisit()` hook to only return the active visit, and move the visit store and stuff related to the "current visit" into esm-patient-chart.

This generally makes sense to me @chibongho though definitely want to hear what others think.  My gut would be to go with the option that breaks backwards compatibility (but which seems like the best long-term solution) but of course easy for me to say since we don't currently use this functionality at PIH... :)

ibacher:
> It seems like the concept of a "current visit" should only be relevant within the patient chart.

So, that's not _entirely_ true. For example, if we want to allow the retrospective entry of lab results (we do) in the Lab app, we seem to need the same concept or, at the very least, we need a way to refer to the concept of "the visit being edited". (I think similar considerations apply to dispensing and potentially billing). To that end, I think the issue is actually the `useVisit()` hook attempting to return two "types" of visits (active or current) rather than only using the active visit if there is no other "current" visit.

Maybe another slight concept breaking thing I want to plug-in here, though, is that the Chart is intended as a view for clinicians. It's not necessarily the only app or place to collect data for an individual patient and its at those various other points that we'll always need some kind of retrospective entry

That's probably a more breaking change, but, fundamentally, I'd imagine we'd have hooks like:

* `useActiveVisits(patientUuid)` - Returns the patient's active visit(s) if any
* `useVisit(patientUuid)` - Return the patient's "current" visit or most recent active visit if no "current" has been set

ibacher:
> but does our UI currently support this pattern?

No, and we may need to adapt something like the visit header pattern elsewhere.

> for the "visit in context", is the patient even relevant?

Definitely a rabbit-hole...

Short answer is "no". But it's more convenient to have a single hook that can be called to get the "current" visit and if no visit has been selected, the current visit will be the active visit, for which we do need the patient UUID.

So, if we had an app that supported the workflow Select Patient -> Select Visit -> Do Stuff, we wouldn't need a `useVisit()` hook that took a patient UUID (or maybe any argument), but that app would not be the patient chart (since the "select visit" part adds a hurdle that should not be there at point of care).

mogoodrich:
> For example, if we want to allow the retrospective entry of lab results (we do) in the Lab app, we seem to need the same concept or, at the very least, we need a way to refer to the concept of "the visit being edited".

Taking with @chibongho today I think I'm more convinced that the "current" (maybe "selected" is a better term?) visit should be local to the patient chart, if I'm understanding correctly.  I would think that if the user jumped out of the patient chart entirely and, say, entered the lab app, they wouldn't expect (nor should) be in the same visit context? Is that what we are talking about?   It's possible that other apps may want to follow the same pattern to create a visit context, but they shouldn't be sharing a single one?

chibongho:
I think it makes sense for apps other than patient chart to have "current visit" (or "selected visit", or "visit in context") for a patient and display it in the visit context header in workspaces / forms when adding data that supports RDE. But how sticky / persistent should the "current visit" be? I think we want some persistence. (When we need to enter 10 RDE visit notes for patient A visit X, it'd be nice to not have to re-select X every time we open the visit form workspace. Ditto when filling out 10 RDE lab results.) However, right now, the current visit is stored in a global Visit store, which means the "current visit" selected in the patient chart is also the same "current visit" when we navigate out of the patient chart into the labs app. This results in the following behaviors.

- Go to patient chart of patient A, open patient notes workspace and select visit X as "current visit". Exit patient chart and go to lab app. Open workspace to add lab result for patient A. Visit X is auto selected as "current visit" in the visit context header.
- Go to patient chart of patient A, open patient notes workspace and select visit X as "current visit". Exit patient chart and go to lab app. Open workspace to add lab result for a different patient B. Visit context header defaults to active visit of B (or null if it doesn't exists).
  - Go back to patient chart of patient A, open visit notes workspace, visit A is no longer selected in visit context header.

The above could be unexpected / confusing until the user understands the "current visit" persistence. I think it would be less confusing if we have the visit store stickiness be on a per-app basis instead.

Somewhat related, the visit store is backed by localStorage right now. Should that be the case? It implemented manually (instead of using the built-in zustand localStorage support) and adds some complexity. If we want to keep it, and we want the visit store stickiness to be per-app, then we need to namespace the values within the visit store by app name.

mogoodrich:
> The above could be unexpected / confusing until the user understands the "current visit" persistence. I think it would be less confusing if we have the visit store stickiness be on a per-app basis instead.

+1 to per app.  In both the above use cases, I would expect the user to have to reselect the "current visit" for the patient.  Although there is some utility, of course, I really feel like it would be more confusing than helpful.  

Tangentially (and it also occurs to me that this is relevant for my "one-one-current-visit-so-patient-isnt-relevant-comment") regardless of what we do we should be very careful about getting into a state where the "current" patient and "current" visit don't coincide, and risk displaying health information for one patient and attribute to another.   @chibongho  @ibacher  @mseaton

ibacher:
>  I would think that if the user jumped out of the patient chart entirely and, say, entered the lab app, they wouldn't expect (nor should) be in the same visit context?

Yeah, sorry. I might've added to the confusion here. Current visit context should be local to the chart, I just wanted to be clear that the _concept_ is itself is applicable to non-chart cases, but probably differently (in the chart, the "current" visit is state shared across a large number of extensions and whatnot; outside of the chart, the "current" visit may just be context local to form entry).

> Somewhat related, the visit store is backed by localStorage right now. Should that be the case?

Ugh... This is tricky. On the one hand, I don't think localStorage is the right place to store this. On the other hand, if I opened a second tab _in the patient chart of the same patient_, I would probably expect to be in the same visit context as I was before.

Maybe we need to consider a URL change when the user selects a visit context so that the URL is now `/openmrs/spa/patient/<patient uuid>/visit/<visit uuid>/chart/<navigation tab>`? Pushing the visit state into the URL seems to me, at a first glance, like the "right" sort of context persistence. It's not (like localStorage) effectively a global variable; if I open a new tab in the chart, the URL would contain the visit context, so I'd remain in the right visit context, but, because it's part of the chart URL, outside of the chart, nothing changes.

> what we do we should be very careful about getting into a state where the "current" patient and "current" visit don't coincide, and risk displaying health information for one patient and attribute to another

Yeah, and I'd actually be happier if we decide to retain a global `useVisit()` hook, for that to actually perform that check rather than just assuming whatever was pushed into the store was appropriate. However, I don't know that a global `useVisit()` hook really is what we should be supporting.

chibongho:
Storing in URL could be tricky because:
- In the patient chart, every left nav item (like "Patient Summary", "Vitals and Biometrics") is its own URL. We'd expect the current visit to persist across navigation those left nav items.
- URL changes affect the browser history. So the user can go from `/userA/visitX` to `/userA/visitY`, press back and inadvertently change the "current visit" back to `visitX`. (If no workspace is opened, the user might not even see visual changes). 

There are ways around them, but I think they could be weird edge cases. If we want to have the current visit persist across tabs, I think keeping it in localStorage is easiest to implement.

> Yeah, and I'd actually be happier if we decide to retain a global useVisit() hook, for that to actually perform that check rather than just assuming whatever was pushed into the store was appropriate. However, I don't know that a global useVisit() hook really is what we should be supporting.
I think it does that already (you won't get the current visit back if it doesn't belong to the input patient). Although I also want the hook to have a side effect of setting the current visit to the active visit if it's not defined / belong to a different patient, something it doesn't do right now. I think it makes sense to have a global hook to get the current visit that works easily with whatever per-app "current visit" storage we decide on, so we can use it in places outside the patient chart. I think it's enough of a deviation of the current `useVisit()` hook behavior that it warrants deprecation / creating a new hook.

ibacher:
> There are ways around them, but I think there could be weird edge cases

I think we don't avoid weird edge-cases by using localStorage, though. My worry is that with that, the context can change without a visual prompt to the user of any sort. But I also think that clicking the back button and not being back in the context I was in before (since the visit didn't change in localStorage) is a worse sort of error along those lines.

But also, what happens here if I open two tabs, both in the chart and then in one tab, I navigate to a different app?

> I think it does that already (you won't get the current visit back if it doesn't belong to the input patient)

Good point. It does.

> Although I also want the hook to have a side effect of setting the current visit to the active visit if it's not defined / belong to a different patient, something it doesn't do right now

Is there a reason we _wouldn't_ want that behaviour elsewhere? The problem here isn't that this is some new behaviour, it's that the `useVisit()` hooks concept of "current visit" is backed by a global store, but the "current visit" (and this is, I think, the upshot of this discussion) should be something app-local.

packages/esm-patient-chart-app/src/visit-header/retrospective-visit-label.component.tsx
```diff
@@ -10,6 +10,9 @@ interface RetrospectiveVisitLabelProps {
...
   return (
     <Toggletip align="bottom">
       <ToggletipButton label={t('retrospectiveEntry', 'Retrospective Entry')}>
```
denniskigen, line 18: Minor, but we probably don't need the optional chaining checks on the `currentVisit` object in this return block because we're sure it exists, right?

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-info.component.tsx
```diff
@@ -0,0 +1,35 @@
...
+          : t('fromDateToDate', '{{fromDate}} - {{toDate}}', {
+              fromDate: formatDate(parseDate(visit.startDatetime), { time: false }),
+              toDate: formatDate(parseDate(visit.stopDatetime), { time: false }),
+            })}
+        <span className={styles.separator}>&middot;</span>
+        <Building />
+        <span className={styles.visitLocation}>{visit.location.display}</span>
+      </>
+    );
+  } else {
```
denniskigen, line 30: Can we prefer the early return if there's no visit?

```tsx
  if (!visit) {
    return <></>;
  }

  const isActive = !Boolean(visit.stopDatetime);

  return (
    <>
      {isActive
        ? t('currentActiveVisit', 'Current active visit')
        : t('fromDateToDate', '{{fromDate}} - {{toDate}}', {
            fromDate: formatDate(parseDate(visit.startDatetime), { time: false }),
            toDate: formatDate(parseDate(visit.stopDatetime), { time: false }),
          })}
      <span className={styles.separator}>&middot;</span>
      <Building />
      <span className={styles.visitLocation}>{visit.location.display}</span>
    </>
  );
```

Also, isn't it preferable to return `null` instead of an empty fragment?

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-switcher.modal.tsx
```diff
@@ -0,0 +1,124 @@
+import { Button, InlineLoading, ModalBody, ModalFooter, ModalHeader, RadioButton } from '@carbon/react';
+import { ErrorState, type Visit } from '@openmrs/esm-framework';
+import { launchPatientWorkspace } from '@openmrs/esm-patient-common-lib/src';
```
denniskigen, line 3:
```suggestion
import { launchPatientWorkspace } from '@openmrs/esm-patient-common-lib';
```

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-switcher.modal.tsx
```diff
@@ -0,0 +1,124 @@
...
+        </Button>
+      </ModalBody>
+      <ModalFooter>
+        <Button kind="secondary" onClick={closeModal}>
+          {t('cancel', 'Cancel')}
+        </Button>
+        <Button
+          disabled={selectedVisit == null || isLoading}
+          onClick={() => {
+            setVisitContext(visits.find((v) => v.uuid == selectedVisit));
```
denniskigen, line 70:
```suggestion
            setVisitContext(visits?.find((v) => v.uuid === selectedVisit));
```

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-switcher.modal.tsx
```diff
@@ -0,0 +1,124 @@
...
+const VisitContextSwitcherModal: React.FC<VisitContextSwitcherProps> = ({
+  patientUuid,
+  closeModal,
+  onAfterVisitSelected,
+}) => {
+  const { t } = useTranslation();
+  const { visits, isLoading, error } = useInfiniteVisits(patientUuid);
+  const { patientUuid: selectedVisitPatientUuid, manuallySetVisitUuid, setVisitContext } = useVisitContextStore();
+  const [selectedVisit, setSelectedVisit] = useState<string>(
+    selectedVisitPatientUuid == patientUuid ? manuallySetVisitUuid : null,
```
denniskigen, line 27:
```suggestion
    selectedVisitPatientUuid === patientUuid ? manuallySetVisitUuid : null,
```

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-switcher.scss
```diff
@@ -0,0 +1,87 @@
+@use '@carbon/colors';
+@use '@carbon/layout';
+@use '@openmrs/esm-styleguide/src/vars' as *;
+
+.visitstList {
```
denniskigen, line 5:
```suggestion
.visitsList {
```

Typo?

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-header.component.tsx
```diff
@@ -0,0 +1,62 @@
...
+      className={classNames(styles.visitContextHeader, isActiveVisit ? styles.activeVisit : styles.retroactiveVisit)}
+    >
+      <div className={styles.addingTo}>{t('addingToVisit', 'Adding to:')}</div>
+      <div className={styles.visitType}>{currentVisit.visitType.display}</div>
+      <div className={styles.changeVisitButton}>
+        <Button kind="ghost" size="sm" onClick={openVisitSwitcherModal}>
+          {t('change', 'Change')}
+        </Button>
+      </div>
+      <div className={styles.visitInfo}>
```
denniskigen, line 55: Ditto here. Can we ensure that the content of this fragment is centered both vertically and horizontally with rules such as:

```scss
display: flex;
align-items: center;
justify-items: center;
```

packages/esm-patient-chart-app/src/visit-header/retrospective-visit-label.component.tsx
chibongho: I did in fact get an NPE without it. I suspect it's a double rendering issue where the currentVisit is null on the first render. 

A bit moot why, as this component is going away in a [separate PR](https://github.com/openmrs/openmrs-esm-patient-chart/pull/2262).

chibongho:
> I think we don't avoid weird edge-cases by using localStorage, though. My worry is that with that, the context can change without a visual prompt to the user of any sort. But I also think that clicking the back button and not being back in the context I was in before (since the visit didn't change in localStorage) is a worse sort of error along those lines.
> 
> But also, what happens here if I open two tabs, both in the chart and then in one tab, I navigate to a different app?

How about sessionStorage instead? If you open the patient chart in 2 tabs, each one will have its own visit context, and we make it the only way to change the context is by the user actually doing it via UI within that tab.

>> Although I also want the hook to have a side effect of setting the current visit to the active visit if it's not defined / belong to a different patient, something it doesn't do right now
>
> Is there a reason we wouldn't want that behaviour elsewhere? The problem here isn't that this is some new behaviour, it's that the useVisit() hooks concept of "current visit" is backed by a global store, but the "current visit" (and this is, I think, the upshot of this discussion) should be something app-local.

Yeah, I think we'll want that behavior everywhere. So I guess just change `useVisit()` to use app-local storage for the visit context?

jayasanka-sack (commented):
Looks good to me, @chibongho! Let's throw in some unit tests and get it merged.

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-header.component.tsx
```diff
@@ -1,30 +1,35 @@
 import { Button, Loading } from '@carbon/react';
-import { showModal, useVisit } from '@openmrs/esm-framework';
+import { showModal, useFeatureFlag, useVisit } from '@openmrs/esm-framework';
 import classNames from 'classnames';
 import React, { useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 import styles from './visit-context-header.scss';
 import VisitContextInfo from './visit-context-info.component';
 import { useVisitContextStore } from './visit-context';
+import { useSystemVisitSetting } from '@openmrs/esm-patient-common-lib/src';
```
ibacher, line 9: Shouldn't this import just be from `@openmrs/esm-patient-common-lib`? Also, can we order this import with the framework.

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-header.component.tsx
```diff
@@ -34,6 +39,9 @@ const VisitContextHeader: React.FC<VisitContextHeaderProps> = ({ patientUuid })
     });
   };
 
+  if (!showVisitContextHeader) {
+    return <></>;
```
ibacher, line 43: nit: `null` is also a valid React render result.

packages/esm-patient-orders-app/src/order-basket/order-basket.workspace.tsx
```diff
@@ -178,13 +187,13 @@ const OrderBasket: React.FC<DefaultPatientWorkspaceProps> = ({
...
       </div>
-      {activeVisitRequired && !activeVisit && (
+      {visitRequired && !currentVisit && (
         <ActionableNotification
           kind="error"
           actionButtonLabel={t('startVisit', 'Start visit')}
           onActionButtonClick={openStartVisitDialog}
           title={t('startAVisitToRecordOrders', 'Start a visit to order')}
-          subtitle={t('activeVisitRequired', 'An active visit is required to make orders')}
+          subtitle={t('visitRequired', 'A visit is required to make orders')}
```
ibacher, line 196: I'm on board with dropping the "active" stuff, but I think this might not be as easily understood? Does "A current visit is required to make orders" or "You must be in a visit context to place orders" or something along those lines?

ibacher (approved):
Mostly minor quibbles. This seems basically fine.

packages/esm-patient-orders-app/src/order-basket/order-basket.workspace.tsx
chibongho: I'll change it to "You must select a visit to place orders". I find the term "current visit" confusing, and want to remove it from users (even though we're probably stuck with it in the code base).

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-header.component.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  const { currentVisit, isLoading } = useVisit(showVisitContextHeader ? patientUuid : null);
+  const { manuallySetVisitUuid, setVisitContext } = useVisitContextStore();
+  const isActiveVisit = !Boolean(currentVisit && currentVisit.stopDatetime);
+
+  // This hook is needed because the `useVisit` hook sometimes
+  // returns a currentVisit that isn't actually the one in the visit context.
+  // TODO: move this into the useVisit hook
+  useEffect(() => {
+    if (showVisitContextHeader && !isLoading && currentVisit.uuid != manuallySetVisitUuid) {
```
denniskigen, line 29:
```suggestion
    if (showVisitContextHeader && !isLoading && currentVisit.uuid !== manuallySetVisitUuid) {
```

Yeah, can we be consistent with the strict equality checks?

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-header.component.tsx
denniskigen: With the added benefit that `return null`:
- Is more performant as it doesn't create any objects or elements as opposed to `return <></>` which creates a React fragment even though it doesn't get render anything in the DOM.
- Uses less memory because `null` is a primitive value whereas `<></>` creates a React fragment.
- Communicates more intent as it's the idiomatic way of saying "don't render anything" in React.

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-info.component.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+const VisitContextInfo: React.FC<VisitContextInfoProps> = ({ visit }) => {
+  const { t } = useTranslation();
+
+  if (!visit) {
+    return null;
+  }
+
+  const isActive = !Boolean(visit.stopDatetime);
```
denniskigen, line 18: Is `const isActive = !visit.stopDatetime` better here?

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-switcher.modal.tsx
```diff
@@ -0,0 +1,125 @@
...
+  setSelectedVisit(visitUuid: string);
+}
+
+/**
+ * A clickable row within the the visit context switcher to select a visit. This
+ * has slightly different UX than a regular radio button, as the entire card
+ * (not just the radio button and the label) is clickable
+ */
+const VisitCardRow: React.FC<VisitCardRowProps> = ({ visit, setSelectedVisit: setSelected, isSelected }) => {
+  const isActive = !Boolean(visit.stopDatetime);
```
denniskigen, line 95: nit: ditto here.

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-switcher.scss
```diff
@@ -0,0 +1,79 @@
...
+    cursor: pointer;
+  }
+}
+
+.visitCardRowRadioButton {
+  display: flex;
+  align-items: center;
+}
+
+.wardPatientCardButton {
```
denniskigen, line 57: I wouldn't expect to see ward related stuff in here. Also, it doesn't seem like we're using this anywhere.

denniskigen (approved):
Thanks, @chibongho!

denniskigen (approved):
Thanks, @chibongho!

packages/esm-patient-orders-app/src/order-basket/order-basket.workspace.tsx
ibacher: I like that better.

PR #2223, opened 2025-01-31
(fix) O3-4417: Fix interpolation handling in react-i18next test mock

harshthakkr:
## Requirements

## Summary
This PR updates the French translation by setting `escapeValue: false` to ensure correct rendering of special characters.

### Before
Screenshot 2025-01-31 at 11 16 44 AM

### After
Screenshot 2025-02-04 at 1 21 03 PM

## Related Issue
[O3-4417](https://openmrs.atlassian.net/jira/software/c/projects/O3/issues/O3-4417?jql=project%20%3D%20%22O3%22%20ORDER%20BY%20created%20DESC)

[O3-4417]: https://openmrs.atlassian.net/browse/O3-4417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/esm-patient-tests-app/translations/fr.json
+++ b/packages/esm-patient-tests-app/translations/fr.json
@@ -57,7 +57,7 @@
   "priority": "Priorité",
   "priorityRequired": "Priority is required",
   "recentResults": "Résultats récents",
-  "recentTestResults": "Résultats d'examens récents",
+  "recentTestResults": "Résultats d’examens récents",
   "referenceNumber": "Numéro de référence",
   "referenceRange": "Plage de référence",
   "refreshData": "Refresh data",
--- a/packages/esm-patient-tests-app/translations/fr.json
+++ b/packages/esm-patient-tests-app/translations/fr.json
@@ -57,7 +57,7 @@
   "priority": "Priorité",
   "priorityRequired": "Priority is required",
   "recentResults": "Résultats récents",
-  "recentTestResults": "Résultats d’examens récents",
+  "recentTestResults": "Résultats d'examens récents",
   "referenceNumber": "Numéro de référence",
   "referenceRange": "Plage de référence",
   "refreshData": "Refresh data",
--- a/__mocks__/react-i18next.js
+++ b/__mocks__/react-i18next.js
@@ -31,9 +31,9 @@ const renderNodes = (reactNodes) => {
 
 const useMock = [(key) => key, {}];
 useMock.t = (key, defaultValue, options = {}) => {
-  let translatedString = defaultValue ? defaultValue : key;
-  Object.keys(options).forEach((key) => {
-    translatedString = defaultValue.replace(`{{${key}}}`, `${options[key]}`);
+  let translatedString = defaultValue || key;
+  Object.entries(options).forEach(([k, v]) => {
+    translatedString = translatedString.replace(new RegExp(`{{${k}}}`, 'g'), v);
   });
 
   return translatedString;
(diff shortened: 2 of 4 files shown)


packages/esm-patient-tests-app/translations/fr.json
```diff
@@ -57,7 +57,7 @@
   "priority": "Priorité",
   "priorityRequired": "Priority is required",
   "recentResults": "Résultats récents",
-  "recentTestResults": "Résultats d'examens récents",
+  "recentTestResults": "Résultats d’examens récents",
```
denniskigen, line 60: I think that change ought to be made in Transifex instead - because that's the source of truth.

denniskigen: Ignore my previous suggestion. I think the fix involves telling `i18next` not to escape the `displayText` value in the `EmptyState` when interpolating it into the provided string [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-common-lib/src/empty-state/empty-state.component.tsx#L27) by adding an `interpolation block` with `escapeValue` set to true:

```tsx
<p className={styles.content}>
  {t('emptyStateText', 'There are no {{displayText}} to display for this patient', {
    displayText: props.displayText.toLowerCase(),
    interpolation: {
      escapeValue: false,
    },
  })}
</p>
```

Explanation:

- The `interpolation: { escapeValue: false }` option tells i18next to skip HTML entity encoding for this specific interpolation value. 
- This preserves special characters while maintaining React's default XSS protection for other parts of the application.
- The change is scoped to only affect this specific translation string's `displayText` parameter

harshthakkr: I also came up with this solution, but I wasn’t aware that it could be applied to a specific string as well. My main concern was how this change might affect other characters in the French language.

Thanks for helping, @denniskigen!

denniskigen: Should be fine, I think.

denniskigen: @harshthakkr, looks like the string is getting picked up correctly in the UI following a rebase to integrate translation updates from the latest automated Transifex PR:

CleanShot 2025-02-05 at 11  58 09@2x

I've taken the liberty of pushing a commit undoing the HTML entity decoding step in the `EmptyState` component. As we discussed yesterday, in as much as escaping HTML entity decoding using `escapeValue` solved the problem - the proper solution was to change the string by hand upstream in Transifex. The major downside of escaping interpolated values is that it opens us up to the risk of XSS attacks due to unescaped values.

denniskigen:
**TL;DR**:

The French translation string for recent test results in Transifex was updated from using a straight apostrophe (U+0027) in "d'examens" to a right single quotation mark (U+2019). This change:

1. Matches French typographic standards for contractions
2. Resolves rendering issues caused by automatic HTML escaping
3. Maintains visual consistency with surrounding text

The update propagated through our automated translation synchronization system and now displays correctly across supported browsers.

The PR's scope expanded to address an issue with interpolation handling in react-i18next test mocks. Previously, the mock failed to process interpolation replacements correctly, resulting in unprocessed placeholders like {{duration}} appearing in test outputs.

This fix ensures interpolated values substitute properly during testing, aligning test environment behavior with actual user-facing UI rendering.

denniskigen (approved):
Thanks, @harshthakkr!

harshthakkr:
Thank you @denniskigen!

PR #2225, opened 2025-02-03
(fix) O3-4398: Invalidate lab orders after submitting test results form

amosmachora:
## Requirements

## Summary

This PR adds a new optional prop `invalidateLabOrders` that enables the invalidation of lab orders in the Laboratory app after entering test results using the Test Results form.

## Screenshots

This is an example of a use case where this will help in the labs ESM.

https://github.com/user-attachments/assets/b85c1fd1-c090-479c-b170-00f6760e9792

## Related Issue

https://openmrs.atlassian.net/browse/O3-4398
## Other

An accompanying PR on the labs app can be found [here](https://github.com/openmrs/openmrs-esm-laboratory-app/pull/117).

### DIFF
--- a/packages/esm-patient-orders-app/src/lab-results/lab-results-form.component.tsx
+++ b/packages/esm-patient-orders-app/src/lab-results/lab-results-form.component.tsx
@@ -21,9 +21,11 @@ import {
 import { useLabResultsFormSchema } from './useLabResultsFormSchema';
 import ResultFormField from './lab-results-form-field.component';
 import styles from './lab-results-form.scss';
+import { cache } from 'swr/_internal';
 
 export interface LabResultsFormProps extends DefaultPatientWorkspaceProps {
   order: Order;
+  mutateLabOrders?: () => void;
 }
 
 const LabResultsForm: React.FC<LabResultsFormProps> = ({
@@ -31,6 +33,7 @@ const LabResultsForm: React.FC<LabResultsFormProps> = ({
   closeWorkspaceWithSavedChanges,
   order,
   promptBeforeClosing,
+  mutateLabOrders,
 }) => {
   const { t } = useTranslation();
   const abortController = useAbortController();
@@ -179,8 +182,9 @@ const LabResultsForm: React.FC<LabResultsFormProps> = ({
         abortController,
       );
       closeWorkspaceWithSavedChanges();
-      mutateResults();
       mutateOrderData();
+      mutateResults();
+      mutateLabOrders();
       showNotification(
         'success',
         t('successfullySavedLabResults', 'Lab results for {{orderNumber}} have been successfully updated', {
--- a/packages/esm-patient-orders-app/src/lab-results/lab-results-form.component.tsx
+++ b/packages/esm-patient-orders-app/src/lab-results/lab-results-form.component.tsx
@@ -21,7 +21,6 @@ import {
 import { useLabResultsFormSchema } from './useLabResultsFormSchema';
 import ResultFormField from './lab-results-form-field.component';
 import styles from './lab-results-form.scss';
-import { cache } from 'swr/_internal';
 
 export interface LabResultsFormProps extends DefaultPatientWorkspaceProps {
   order: Order;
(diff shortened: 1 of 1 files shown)


packages/esm-patient-orders-app/src/lab-results/lab-results-form.component.tsx
```diff
@@ -24,13 +24,15 @@ import styles from './lab-results-form.scss';
 
 export interface LabResultsFormProps extends DefaultPatientWorkspaceProps {
   order: Order;
+  mutateLabOrders?: () => void;
```
denniskigen, line 27: The current implementation of the [mutateLabOrders](https://github.com/openmrs/openmrs-esm-laboratory-app/pull/117) callback appears to serve two purposes simultaneously:

1. Launching the `test-results-form-workspace` workspace.
2. Triggering revalidation of lab orders

Is this dual responsibility intentional? Given that we're [already operating within the workspace context](https://github.com/amosmachora/openmrs-esm-patient-chart/blob/main/packages/esm-patient-orders-app/src/routes.json#L75) when invoking this form, should the workspace launch be considered a separate concern from the data revalidation?

If so, consider separating these concerns by:

1. Keeping the workspace launch as a distinct action.
2. Passing only the revalidation logic via props.

This would better align with the Single Responsibility Principle and prevent workspace re-launching from within an already active workspace context.

denniskigen: I've left this [additional comment](https://github.com/openmrs/openmrs-esm-laboratory-app/pull/117/files#r1939986045) that should clarify what I meant above.

denniskigen: I've also tested this out locally and I don't see how the `Add lab results` button in the Laboratory app is related to the `Add results` button in the Orders app. 

Your [other PR](https://github.com/openmrs/openmrs-esm-laboratory-app/pull/117/files) adds a mutate function that gets passed down to the workspace when you launch the Lab results form by clicking the [Add lab results button](https://github.com/openmrs/openmrs-esm-laboratory-app/blob/main/src/lab-tabs/actions/add-lab-request-results-action.component.tsx#L15) in the Laboratory app.

On the other hand, the [Add results button](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-orders-app/src/components/orders-details-table.component.tsx#L679) in the Orders table in Patient Chart has a [different handler](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-orders-app/src/components/orders-details-table.component.tsx#L633) altogether. 

Could you help shed some light on that, @amosmachora?

amosmachora: The two buttons are not related at all. They all launch the same workspace on their click handlers but from different places in the app. 

That's why I wouldn't want to couple the `mutateLabOrders` functionality to the workspace itself and would like it to remain agnostic and have what it needs be passed down as an optional prop. I am not sure I am answering your question.

Do you think they should use the same handler?

amosmachora:
Hello @denniskigen yesterday during the coffee break you said you had ideas where we could achieve this functionality without passing the prop, whenever you find a way just highlight it and I will make the change.

packages/esm-patient-orders-app/src/lab-results/lab-results-form.component.tsx
denniskigen: So the thing is, AFAICT, even if you pass `mutateLabOrders` down as a prop to the workspace, it only gets passed to the workspace that gets launched from clicking the `Add lab results` button in the Lab app. 

Adding it as a prop to the orders app doesn't work because the logic that triggers launching that workspace has no knowledge of the `mutateLabOrders`  prop. Does that make sense?

amosmachora: Isn't that what we want? I think that was essentially what I was trying to achieve to find a way for whatever caller to pass a mutate function and have it just work. This was essential in the lab since it allowed for the mutation of lab orders rendered as rows. if for example in the orders app, a need would arise to mutate something (highly unlikely) they would pass a prop and it would just work.

Now that I think of it maybe the prop could be named more generically like `mutate`.

> Adding it as a prop to the orders app doesn't work because the logic that triggers launching that workspace has no knowledge of the mutateLabOrders prop.

that's why I made it optional since it is very specific to the lab workspace and no other workspace instance. 

Does this feel like a hack? Any better ideas

denniskigen:
As mentioned in the other PR, I've tested this out and it looks like it works optimally. Thanks, @amosmachora!

denniskigen (approved):
LGTM. Thanks, @amosmachora!

PR #2240, opened 2025-02-12
(feat) O3-4443 Add ability to create a visit that ended in the past

chibongho:
## Requirements

## Summary

Currently, when creating a visit, the visit form only has fields for start date / time, and it is not possible to create a RDE visit with an end time. (We can’t work around it by first creating an RDE visit with no end time and edit it later, because it overlaps with any visits that's after it). This PR adds the ability to create a ended RDE visit.

Additional Refactoring:
- We have zod validation for the visit form, but also additional form validation `onSubmit` on the start / end time fields. This PR combines the `onSubmit` validation logic into the zod form schema
- The zod schema and default values of the visit form are moved from the visit form to a hook in `visit-form.resources.ts`
- We use a combination of `DatePicker`, `TimePicker` and a AM/PM Dropdown for inputting the datetime of the visit start and visit end. We store the value of those inputs respectively as `Date` (with hours and minutes set to 0), `hh:mm` string, and `"AM" / "PM"` string. I found that confusing and changed the `DatePicker` value to a `YYYY-MM-DD` string instead. In hindsight, I don't feel as good about it as I hoped. It does make some things (like zod validation) simpler, but it also adds a transformation layer to get the value from / to the carbon `DatePicker`
- I spent a lot of time trying to figure out why the unit tests are not getting form validation errors strings for `DatePicker` and `TimePicker`. One of the existing tests was disabled because of that. I managed to get that one test working, but when I tried adding more tests around date / time field validation I couldn't get those to work. I gave up and added unit tests around the zod form schema instead.

## Screenshots

When creating a visit:
image
image
image

When editing a visit:
image

Video:
https://github.com/user-attachments/assets/96fe2dfd-71a2-499b-ad16-e91784f0e522

## Related Issue

https://openmrs.atlassian.net/browse/O3-4443


packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx
```diff
@@ -1,123 +1,207 @@
...
+      'visitStartDate',
+      'visitStartTime',
+      'visitStartTimeFormat',
+      'visitStopDate',
+      'visitStopTime',
+      'visitStopTimeFormat',
+    ],
+  });
+
+  const hasStartTime = ['ongoing', 'past'].includes(visitStatus);
```
mogoodrich, line 49: A "new" visit status is just an ongoing visit that has not be saved?

packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx
```diff
@@ -1,123 +1,207 @@
...
+          dateField={{ name: 'visitStopDate', label: t('endDate', 'End date') }}
+          timeField={{ name: 'visitStopTime', label: t('endTime', 'End time') }}
+          timeFormatField={{ name: 'visitStopTimeFormat', label: t('endTimeFormat', 'End time format') }}
+          minDate={Math.max(lastEncounterDateTime, selectedVisitStartDateTime?.getTime())}
+          maxDate={Date.now()}
+        />
+      )}
+    </section>
+  );
+};
```
mogoodrich, line 84: Tangential rabbit-hole... are we validating start and end times so that you can't have overlapping visitings at the same visit location?  (We need to figure this out but this could be a separate ticket).

packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx
```diff
@@ -1,123 +1,207 @@
...
+
+interface DateTimeFieldProps {
+  dateField: Field;
+  timeField: Field;
+  timeFormatField: Field;
```
mogoodrich, line 89: what is this used for?

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.workspace.tsx
```diff
@@ -21,72 +12,77 @@ import {
...
 } from './visit-form.resource';
-import BaseVisitType from './base-visit-type.component';
-import LocationSelector from './location-selector.component';
-import VisitAttributeTypeFields from './visit-attribute-type.component';
-import VisitDateTimeField from './visit-date-time.component';
 import styles from './visit-form.scss';
+dayjs.extend(isSameOrBefore);
 
-interface StartVisitFormProps extends DefaultPatientWorkspaceProps {
+interface VisitFormProps extends DefaultPatientWorkspaceProps {
```
mogoodrich, line 67: +1 to this rename!

mogoodrich (commented):
Added some comments... a lot to review here, at the end of the day we will likely need to just get this tested thoroughly...

packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx
chibongho: No, we are not validating that currently. Looks like zod support asynchronous validation, so it shouldn't be difficult. Filed ticket: https://openmrs.atlassian.net/browse/O3-4447.

packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx
chibongho: It's for the `VisitDateTimeField` component below. I'll rename it `VisitDateTimeFieldProps`.

packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx
chibongho: A new visit is one where the start time defaults to `now` onSubmit, and we don't need to show the DatePicker / TimePicker for the start time.

I guess I don't actually need this boolean because we return an empty component when `visitStatus == 'new'`

packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx
mogoodrich: Great, thanks @chibongho !

packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx
mogoodrich: Not a big deal, re: the name, I was more just wondering how it worked!

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.workspace.tsx
```diff
@@ -426,59 +228,27 @@ const StartVisitForm: React.FC<StartVisitFormProps> = ({
...
-        const { handleCreateExtraVisitInfo, attributes } = extraVisitInfo ?? {};
-        if (!payload.attributes) {
-          payload.attributes = [];
-        }
-        payload.attributes.push(...attributes);
-        handleCreateExtraVisitInfo && handleCreateExtraVisitInfo();
+        handleCreateExtraVisitInfo?.();
       }
 
+      const abortController = new AbortController();
```
samuelmale, line 251: Are we dispatching an abort signal incase the component unmounts?

chibongho: I'm just keeping the existing implementation. I believe the proper way to implement this is to use `useAbortController()` instead of creating a new AbortController(). This is a pattern that's in a lot of places, and we should probably address it in a separate PR.

chibongho:
Fixed the conflict.

One [E2E test](https://github.com/openmrs/openmrs-esm-patient-chart/actions/runs/14062488075/job/39376556011?pr=2240) (`allergies.spec.ts`) is failing. It's also failing in [other branches](https://github.com/openmrs/openmrs-esm-patient-chart/actions/runs/14061649125/job/39373708011) too

denniskigen:
Yeah, the allergies failure spec is unrelated to your changes - I'm trying to figure out why it's happening.

packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx
```diff
@@ -1,115 +1,192 @@
...
+      'visitStartDate',
+      'visitStartTime',
+      'visitStartTimeFormat',
+      'visitStopDate',
+      'visitStopTime',
+      'visitStopTimeFormat',
+    ],
+  });
+
+  const hasStopTime = 'past' == visitStatus;
```
ibacher, line 49:
```suggestion
  const hasStopTime = 'past' === visitStatus;
```

e2e/specs/edit-existing-visit.spec.ts
```diff
@@ -26,13 +33,18 @@ test('Edit an existing visit', async ({ page }) => {
...
 
     const startDateInput = chartPage.page.getByTestId('visitStartDateInput');
     const startDateDayInput = startDateInput.getByRole('spinbutton', { name: /day/i });
     const startDateMonthInput = startDateInput.getByRole('spinbutton', { name: /month/i });
     const startDateYearInput = startDateInput.getByRole('spinbutton', { name: /year/i });
-    const startTimeInput = chartPage.page.locator('input#visitStartTime');
+
+    // await expect(chartPage.page.getByRole('textbox', { name: /start date/i })).toBeVisible();
+    // await expect(chartPage.page.getByRole('textbox', { name: /start time/i })).toBeVisible();
+    // await expect(chartPage.page.getByLabel(/start time format/i)).toBeVisible();
```
ibacher, line 45: We should probably just remove these?

packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx
ibacher: And in general a few other places.

ibacher (commented):
I'm not sure I love the "New" / "On-going" / "Past" distinction here; it's just that I don't necessarily think it's terribly intuitive. Since we already display a date and time, I think we can just collapse the "New" / "On-going" thing into one.

denniskigen:
I tried starting an ongoing visit with a past date and the submission failed (with `Visit.visitCannotOverlapAnotherVisitOfTheSamePatient`). Should this be flagged by the validation logic beforehand?

CleanShot 2025-03-27 at 3  39 04@2x

ibacher:
> Should this be flagged by the validation logic beforehand?

That might be tricky because whether or not overlapping visits are allowed is controlled by a GP, and we'd have to have all of the patient's visits (or at least their start and end dates) in context to be able to do the check (on the backend, we just create a SQL query to look for any overlap). That said, we definitely need to handle this error more gracefully than that!

chibongho:
> I'm not sure I love the "New" / "On-going" / "Past" distinction here; it's just that I don't necessarily think it's terribly intuitive. Since we already display a date and time, I think we can just collapse the "New" / "On-going" thing into one.

"New" hides the start date/time pickers altogether, and leaves the `startDatetime` field `null` in its payload on submit (letting the server default it to `now`). I think it's the cleanest way to fix [this issue](https://openmrs.atlassian.net/browse/O3-4286), because we have a more precise start time, (even though an alternate fix has since been merged in). I also had Ciaran's input on this. Do you think some of the confusion can be fixed with better wording? @ibacher 

> > That might be tricky because whether or not overlapping visits are allowed is controlled by a GP, and we'd have to have all of the patient's visits (or at least their start and end dates) in context to be able to do the check (on the backend, we just create a SQL query to look for any overlap). That said, we definitely need to handle this error more gracefully than that!
> 
> FWIW, here's what we do in O2:
> Screencast.from.2025-03-27.10-17-19.webm

`zod` allows for async / Promise-based validations, so we can add a debounced check on visit overlap when user changes the start / end date times. It does mean sending more requests though. Either that or we can just show / parse the 400 error nicer. Thoughts?

chibongho:
New snackbar to show error message better:

image

ibacher (approved):

PR #2253, opened 2025-02-19
SAVE AND DISCARD BUTTON FIXED

EDSONZ-WASSWA:
Made Changes in the visit-notes-form.scss
This solves the problem of "In the Visit Note form the Save and Discard buttons should always be visible" for the ticket O3-4376.
fixed the buttons and now there is no need to scroll down to find the buttons.
Ticket link: https://openmrs.atlassian.net/issues/O3-4376?filter=-1.


EDSONZ-WASSWA:
@ibacher

NethmiRodrigo:
@EDSONZ-WASSWA thank you for picking this up. Please stick to the PR template and our [guidelines](https://o3-docs.openmrs.org/docs/frontend-modules/contributing#pull-requests) on naming the PR title, specifically:
> The PR title should begin with the conventional commit label that relates to your PR, followed by a ticket number if applicable. Examples of good PR titles:
(fix) O3-2657: Show inline errors when saving orders fails

EDSONZ-WASSWA:
Thank you point notted

________________________________

ibacher:
Please don't delete the PR template. It is there because it provides guidance and helps us more quickly and accurately review your PR. Specifically, we ask for a summary of what your changes are, what problems it solves and screenshots / videos / etc. showing that this change does, in fact, solve the issue in the ticket.

ibacher:
FWIW, this doesn't look like it really solves the issue per se. What I immediately see is this:

Screenshot

Which now makes it look like the buttons are too far from the bottom.

If I shrink the height of my browser a bit, I now get this:

Screenshot 2

Notice that the control in the "Image" section is not visible, nor can I scroll the "rest" of it into view. Finally, if I scroll up a little bit, you can see the buttons however "above" the content in a way that looks kind of weird:

Screenshot 3

In all of these cases, the buttons should be flush with the bottom of the screen and fully allow all form controls to be visible. They should also continue to occupy the full width of the parent element.

PR #2255, opened 2025-02-21
(fix) O3-4461: Remove view timeline button from individual test panels

Muppasanipraneeth:
## Requirements

## Summary

- After discussion in  [slack](https://openmrs.slack.com/archives/CKS32D55G/p1741341149194779) about this ticket 
-  suggested to  remove the button (view timeline)

## Screenshots
Screenshot 2025-03-10 at 5 27 22 PM

## Related Issue
https://openmrs.atlassian.net/browse/O3-4461

### DIFF
--- a/packages/esm-patient-tests-app/src/test-results/individual-results-table/individual-results-table.component.tsx
+++ b/packages/esm-patient-tests-app/src/test-results/individual-results-table/individual-results-table.component.tsx
@@ -132,7 +132,7 @@ const IndividualResultsTable: React.FC<IndividualResultsTableProps> = ({ isLoadi
                   iconDescription="view timeline"
                   kind="ghost"
                   renderIcon={(props: ComponentProps<typeof ArrowRightIcon>) => <ArrowRightIcon size={16} {...props} />}
-                  onClick={() => launchResultsDialog(headerTitle, subRows[0]?.conceptUuid)}
+                  onClick={() => launchResultsDialog(headerTitle, subRows?.entries[0]?.conceptUuid)}
                   size="sm"
                 >
                   {t('viewTimeline', 'View timeline')}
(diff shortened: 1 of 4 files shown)


Muppasanipraneeth:
@denniskigen @NethmiRodrigo can you review this

Muppasanipraneeth:
thanks @vasharma05 😊

NethmiRodrigo (commented):
Although this fix works, I don’t think it fxes the intended functionality. If I’m not mistaken, it currently shows the trendline of only the first test in the list, which a user would already see by clicking on that test. It seems redundant to have a button on the top of the table that only shows the first entry of the table and does the same as clicking the first entry of the table.

Unfortunately this isn’t working in o3 or test3 either (to see what the original behavior was before it was broken), but based on [a comment in the Zeplin designs](https://zpl.io/jppMjAp), there’s a comment clarifying that this should display the overtime tab.

@paulsonder @ciaranduffy Can you confirm if this is still the expected behavior?

cduffy2:
Hi @NethmiRodrigo 

Clicking view timeline should show all test results in a modal window. The user can specify the time window using the tabs in the modal. 

Clicking the `Over time` content switcher at the top of the screen puts the entire page into a timeline view like this: https://zpl.io/AOmnj6B

I hope that clears it up?

NethmiRodrigo:
Thanks @cduffy2! @Muppasanipraneeth is to clear to you on how to proceed?

denniskigen:
@cduffy2 mind sharing a screenshot of what that modal window looks like?

paulsonder:
@denniskigen Here's a short video...

https://www.loom.com/share/cbd7e2e8c1e746218e9997ace0bc392e?sid=d8983390-fc6a-47db-a67a-fff1eae0c057

denniskigen:
Ah, thanks Paul. What about clicking the `View timeline` button in the top right hand corner? Does that take you to the `Over time` view?

paulsonder:
Looks good!

denniskigen:
@Muppasanipraneeth you can track [this discussion](https://openmrs.slack.com/archives/CKS32D55G/p1741341149194779) on #ux-design-advisory.

denniskigen:
For posterity, the conclusion here was to remove the `View timeline` button from individual test panels:

> I think originally it was meant to take the user to the timeline with it already scrolled to that test, but given that the view switchers are sticky now, I think it might actually be redundant. I would suggest we remove the button

denniskigen (approved):

PR #2262, opened 2025-02-24
(feat) O3-4361: Remove visit header and RDE context from top nav in patient chart

chibongho:
## Requirements

## Summary

This PR is rebased upon pending PRs https://github.com/openmrs/openmrs-esm-patient-chart/pull/2240 and https://github.com/openmrs/openmrs-esm-patient-chart/pull/2222, so marking this as draft for now. (The actual changes for this PR is confined to the latest commit, f0bdcd34, so you can diff accordingly). I also still need to fix some more wording changes (start visit -> add visit), but I think the PR is mostly in a reviewable state.

This PR removes the visit header from the top nav of the patient chart.
image

The removal also affects quite a few things. Other changes:
- Patient banner's actions overflow menu:
image

  - The "Start visit" button has been renamed to "Add visit". I think this is now the only button to add a visit, since with the removal of the visit header, there is no "Start visit" button at the top nav.
    - Previously this button only appears when the patient does not have an active visit. It's changed to always appear, as it is possible to add a (RDE) visit even when there is an active visit.
  - The "Add past visit" button has been removed, as the "Add visit" button trigger the add visit workspace, where the user can add either a new, ongoing, or past visit.
  - The "End visit" button previously did different things depending on whether the "current visit" is an active visit or a retroactive visit. Confusingly, if it's a retroactive visit, the button does not modify the visit, but instead sets the "current visit" on the frontend to null. This button has been changed to always end the active visit, and relabeled as "End active visit".
  - The "Cancel visit" button, similar to "End visit", previously did different things depending on whether "current visit" is an active visit or a retroactive visit. This has been changed to always void the active visit, and relabelld as "Delete active visit".
- We were using "Cancel visit" wording for active visits and "Delete visit" wording for retro visits. They both mean the same thing: voiding a visit. The wording has been standardized now to "delete" regardless of whether visit is active/retro. There was a cancel-visit-dialog and a delete-visit-dialog; I deleted one of them.
- With the visit header removed, we still want to retain the "X" button, which lets user navigate out of the patient chart. (It does so by looping through the browser history until it finds a URL that doesn't container the patient uuid). The button is retained, but added as an extension, slotted onto the top nav.
- Almost everything related to service queues is removed from the patient chart. This should be ready when we upgrade the backend queues module to 2.6.0-SNAPSHOT, which provides support to auto-end and auto-void queue entries when their corresponding visit ends / voids.

## Related Issue

https://openmrs.atlassian.net/browse/O3-4361

## Other

Known issues:
- After clicking on the the patient banner's actions overflow menu items (like "Add visit"), the overflow menu does not dismiss.
- The browser's back and forward button gives inconsistent navigation cross the home app and the patient chart.

Open questions:
- With these changes, it's possible to open the "Start visit" workspace even when there is an active visit. In that case, should we make the form not offer an option to create a "new" visit ("ongoing" and "in the past" would still be valid), since it overlaps with the existing active visit? There had been discussions about how it's possible to have multiple active visits for a patient, but we don't really support it right now.
- There are 2 ways to delete an active visit. There is the "Cancel visit" button (renamed to "Delete active visit" in this PR) in the actions overflow menu, and there is the "Delete visit" button in visit summaries tab. The delete button in visit summaries hide where the visit has an encounter, but the one in the overflow menu does not. We should probably keep the behavior consistent, but which way should we do it? (Show both  regardless of any encounters, or hide both if visit has > 1 encounter?)


mogoodrich (approved):
Awesome, thanks @chibongho !  I gave this a quick review but things look correct.... if there are issues we will likely need to hunt for them during testing.

To you question about whether it should be possible to start a visit when a patient already has an activie visit:  technically, a patient can have two or more active visits at the same time as long as they are different visit locations, but, I think, at least for now, having the UI restrict to one active visit at a time makes the most sense to me. Thanks!

chibongho:
@gracepotma A portion of this change made its way in [in here](https://github.com/openmrs/openmrs-esm-patient-chart/pull/2289) , The rest of this PR should probably wait for these other PRs to get merged in first: https://github.com/openmrs/openmrs-esm-patient-chart/pull/2240 and https://github.com/openmrs/openmrs-esm-patient-chart/pull/2222,.

PR #2271, opened 2025-02-27
(fix) O3-4481 - mutate visits data whenever a patient's visit is updated

chibongho:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspiration.
- [ ] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://om.rs/o3ui).
- [x] My work includes tests or is validated by existing tests.

## Summary

In the patient chart, we use several different hooks to get the visits of a patient: (`useVisit`, `useVisits`, `usePastVisits`, `useInfiniteVisits`). When we add/edit/remove visits, all of those hooks need to mutate their data, but in several places, we are only calling mutate on some of them. This PR provides a unified hook to do that, and applies it where needed.

This bug is not triaged, but it's been here for a while, and likely don't need to be included in 3.3 release.

## Screenshots

Original bug prompting this change, note that visit summary is not updated when the active visit ends:
https://github.com/user-attachments/assets/795ce313-f9bb-40e5-9462-17fb9cd11a00

With PR changes:
https://github.com/user-attachments/assets/d5f5b227-6beb-48ea-83f6-9b6e4bcc861c

## Related Issue

https://openmrs.atlassian.net/browse/O3-4481

PR #2278, opened 2025-03-08
(fix) Enable more automatic visit data revalidations

denniskigen:
## Requirements

## Summary

This change fixes an issue where deleting or restoring a visit would not properly invalidate all visit data. This was particularly noticeable when using the Visit summary dashboard.

It does so by adding calls to mutateInfiniteVisits in the useDeleteVisit hook, ensuring that all visit data views are properly revalidated when a visit is deleted or restored.

Additionally, this change:
- Removes an unused useVisits import from the Visit Form workspace component
- Improves code clarity in visit.resource.tsx by extracting the API URL construction

This diff intentionally factors out the visit data revalidation logic into a smaller changeset before tackling the more complex task of unifying all visit revalidation logic into a single hook. This incremental approach will make the subsequent unification of visit revalidation logic more manageable.

### DIFF
--- a/packages/esm-patient-chart-app/src/visit/visits-widget/visit.resource.tsx
+++ b/packages/esm-patient-chart-app/src/visit/visits-widget/visit.resource.tsx
@@ -57,15 +57,15 @@ export function useVisits(patientUuid: string) {
   const customRepresentation =
     'custom:(uuid,encounters:(uuid,diagnoses:(uuid,display,rank,diagnosis),form:(uuid,display),encounterDatetime,orders:full,obs:(uuid,concept:(uuid,display,conceptClass:(uuid,display)),display,groupMembers:(uuid,concept:(uuid,display),value:(uuid,display),display),value,obsDatetime),encounterType:(uuid,display,viewPrivilege,editPrivilege),encounterProviders:(uuid,display,encounterRole:(uuid,display),provider:(uuid,person:(uuid,display)))),visitType:(uuid,name,display),startDatetime,stopDatetime,patient,attributes:(attributeType:ref,display,uuid,value)';
 
+  const apiUrl = patientUuid ? `${restBaseUrl}/visit?patient=${patientUuid}&v=${customRepresentation}` : null;
+
   const {
     data,
     error,
     isLoading,
     isValidating,
     mutate: localMutate,
-  } = useSWR(patientUuid ? ['visits', patientUuid] : null, () =>
-    openmrsFetch(`${restBaseUrl}/visit?patient=${patientUuid}&v=${customRepresentation}`),
-  );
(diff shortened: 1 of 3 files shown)


packages/esm-patient-chart-app/src/visit/visits-widget/visit.resource.tsx
```diff
@@ -57,15 +57,15 @@ export function useVisits(patientUuid: string) {
...
   const {
     data,
     error,
     isLoading,
     isValidating,
     mutate: localMutate,
-  } = useSWR(patientUuid ? ['visits', patientUuid] : null, () =>
-    openmrsFetch(`${restBaseUrl}/visit?patient=${patientUuid}&v=${customRepresentation}`),
-  );
+  } = useSWR(patientUuid ? ['visits', patientUuid] : null, () => openmrsFetch(apiUrl));
```
denniskigen, line 68: @NethmiRodrigo what's the thinking behind removing the custom array cache key in [this commit](https://github.com/openmrs/openmrs-esm-patient-chart/commit/ad545bbef7347b695ab30d93390d3c58c9d1d441)? Should that change be made here too?

NethmiRodrigo: I don't think we should make the change here. 
This custom key was introduced purely so that the global mutate function could be used for `useVisits` hook. The purpose was, certain components that wanted to mutate `useVisits` would do so by getting its mutate function like `const { mutateVisits } = useVisits(patientUuid);` which would trigger the `useVisits` API call. Having the function [`invalidateUseVisits`](https://github.com/openmrs/openmrs-esm-patient-chart/blob/705fb733416976692e1e80b1bd636dd9f920ae01/packages/esm-patient-chart-app/src/visit/visits-widget/visit.resource.tsx#L79) meant that those components didn't have to use the `useVisist` hook just to mutate it. The `useVisit` hook was given a custom key to pass it in to the global mutate function in the `invalidateUseVisits` function. Since Chi Bong's work would do this globally, the `invalidateUseVisits` function (and therefore, the custom key) and would no longer be required. But until then, we'd need this, unless we are removing `invalidateUseVisits`, too.

NethmiRodrigo (approved):

chibongho (approved):
This [PR](https://github.com/openmrs/openmrs-esm-patient-chart/pull/2271) should have fixed this too, but I still can't figure out the failing E2E tests. Let's merge this in for now.

denniskigen:
Yeah, the whole goal here was to try and make sense of why exactly things fail in your PR, @chibongho. Thanks for the reviews. Merging this in so we can update the other one and see what that brings.

PR #2280, opened 2025-03-10
(chore) Automatically pull updates to O3 framework

ibacher:
## Requirements

## Summary

This adds a CI job that runs every hour to update the O3 framework if it is out-of-date.

More precisely, this job will update any libraries from the `@openmrs` NPM org or the `openmrs` package itself to the `next` version if a newer version is detected.

## Other

This also moves us to using Node 20 on CI and attempts to only run the Transifex jobs on this repo and not on forks.


ibacher:
@denniskigen I noticed that the @openmrs/ngx-file-uploader in this repo seems to be behind. Is a safe library to update?

denniskigen:
Yes, it should be safe to update, @ibacher. It gets used in the Angular form engine.

denniskigen (approved):
LGTM. Thanks, @ibacher

.github/workflows/update-openmrs-deps.yml
```diff
@@ -0,0 +1,55 @@
+on:
+  workflow_dispatch:
+  schedule:
+    # every hour
+    - cron: '30 */1 * * *'
```
jayasanka-sack, line 5: Shall we change this to daily or something?

NethmiRodrigo (approved):
Thanks Ian, LGTM! 

> Long term, the goal is to support auto-updating the framework and the patient-common-lib across our repos.

Does this mean that currently we can't auto update the framework in a similar way across the other repos?

.github/workflows/update-openmrs-deps.yml
ibacher: I figured this is relatively low-cost and it's probably better to get these updates in faster. Ideally, we'd just trigger this workflow from esm-core or something.

PR #2285, opened 2025-03-13
(fix) Fix RangeError when selecting a date in the visit form

denniskigen:
## Requirements

## Summary

While the [OpenmrsDatePicker](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/datepicker/index.tsx#L107) component's type definitions suggest it accepts [strings](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/datepicker/index.tsx#L79) for `minDate` and `maxDate` props, the implementation appears to have issues with formatted date strings in "DD/MM/YYYY" format. This is currently manifesting in a regression in the Start Visit form where clicking the datepicker yields a `RangeError: date value is not finite in DateTimeFormat.formatRange()` error. 

[Repro of the bug when starting a visit](https://github.com/user-attachments/assets/d1d13731-a7de-422f-9b89-e7333e92382f)

This error specifically occurs when starting a new visit, but not when editing existing visits, likely because existing visits already have properly formatted date objects. Removing the explicit date formatting from the `minDate` and `maxDate` object constructions resolves the RangeError by passing dayjs objects directly instead of formatted strings.

### RangeError fixed
https://github.com/user-attachments/assets/ba50db5e-2304-43bb-9c61-5d3d9e65cc06

## Related Issue
https://openmrs.atlassian.net/browse/O3-4538

### DIFF
--- a/packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx
@@ -38,7 +38,7 @@ const VisitDateTimeField: React.FC<VisitDateTimeFieldProps> = ({
   // submission, hence just using the min date with hours/ minutes/ seconds set to 0 and max date set to
   // last second of the day. We want to just compare dates and not time.
   const minDateObj = minDate ? dayjs(new Date(minDate).setHours(0, 0, 0, 0)) : null;
-  const maxDateObj = maxDate ? dayjs(new Date(maxDate).setHours(23, 59, 59, 59)).format('DD/MM/YYYY') : null;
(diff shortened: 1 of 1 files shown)


ibacher (approved):
In any case, `dayjs(new Date(maxDate).setHours(23, 59, 59, 59)).format('DD/MM/YYYY')` is kind of non-sensical.

ibacher:
This is a good point, though. With Carbon, you could specify the date format and then I think the string would be parsed using that format. With OpenmrsDatePicker, we just use `dayjs()` to parse strings, which means it needs to be formatted as a ISO 8601 date (e.g., `YYYY-MM-DDThh:mm:ss.SSSZ` or partial versions of that), which is the format we should be using for anything _not_ displayed to the user.

PR #2289, opened 2025-03-14
(feat) O3-4361 Remove visit header from top nav in patient chart

chibongho:
## Requirements

## Summary

This PR is a lighter version of https://github.com/openmrs/openmrs-esm-patient-chart/pull/2262. It removes the visit header from the top nav, but without also changing of various buttons in patient banner to not behave differently when we're in a past visit context. 

The primary purpose is to unblock [this PR](https://github.com/openmrs/openmrs-esm-core/pull/1279/files). When tested against it, the patient chart's left nav has styling issues. Rather than fixing it, I'm just removing it as it's going away anyway.

With this lighter change, buttons like "End visit" does something different / surprising when you're in a past visit context, with no UI indication. That said, I don't think anyone expects RDE to be functional at the moment, and those buttons should behave as expected when that other PR is merged.

## Screenshots

ltr:
https://github.com/user-attachments/assets/a73ddac0-fdca-4012-a341-9855e8a125d3

rtl: (it looks ugly, but it's like this on dev3 already)
https://github.com/user-attachments/assets/5082a0dc-ef1c-46c5-876b-e0ad6ce3fbb3

## Other

The patient chart implemented (correctly, I think) the left nav to become right nav on rtl languages. I'm keeping this behavior here. The home app and other apps (like bed management), however, do not do that. 

The way the left nav works now, it expects the patient chart app / home app to provide their own left margin so its main content doesn't overlap with the left nav in desktop mode, and that they handle rtl language correctly. I think we should re-consider that.


denniskigen (approved):
LGTM!

denniskigen (approved):
LGTM! Could you fix up the failing e2e test?

PR #2295, opened 2025-03-19
(feat) O3-4480 Add visit history table to patient chart

chibongho:
## Requirements

## Summary

This is initial work on the visit history table, with no configuration on the columns. (That will be added in subsequent PR).

## Screenshots

image

https://github.com/user-attachments/assets/92b892f1-6623-4f05-b990-427cfcaffe20

## Other

The "Summary Cards" tab shows an infinite list view of visit summary cards, with "load more" button to load more visits. The same button also appears when we are in viewing the encounters table (because the encounter table only lists encounters from those visits loaded in the "Summary Cards" tab). This UI is a bit strange, but I'm leaving is as for now; there is a chance that the Summary Cards view might go away altogether.


gracepotma:
A few things to clarify: (none of these are blockers for this PR; just FYI/feedback)
* The list should default to showing _20_ visits instead of 10 (must-have but not a blocker for merging this PR)
* The duration (days) of the visit should be shown

ibacher:
> The list should default to showing 20 visits instead of 10 (must-have but not a blocker for merging this PR)

I think we should be _very_ careful about specifying this. I'm not sure we can guarantee loading 20 rows at once performantly, depending on the data that needs to be conveyed in the table. There's a trade-off between flexibility and performance here that needs to be thought through carefully.

chibongho:
> A few things to clarify: (none of these are blockers for this PR; just FYI/feedback)
> 
>     * The list should default to showing _20_ visits instead of 10 (must-have but not a blocker for merging this PR)
> 
>     * The duration (days) of the visit should be shown

Not finalized, but I'm thinking adding configuration for both of those

jayasanka-sack:
@chibongho since we are still on the process of adding columns to this table, how about we make summary card tab the default and merge this PR?

cc: @ibacher @denniskigen

chibongho:
I think we can remove the infinite scroll summary cards view altogether if there is no objection to that (proposed in [talk post](https://talk.openmrs.org/t/proposed-changes-to-the-patient-chart-visits-page/45688)). We'll first need to add the expanded row in the table to see the card view before we remove that.

jayasanka-sack:
Yeah, I get your point. What I meant is since we currently can’t show the detailed view with the table until the expanded rows are in place, my suggestion is to make the summary cards view the default for now. Users can still switch to the table by clicking the table tab. We can go ahead and merge this PR, and once the table supports the same level of detail as the cards, we can remove the summary cards view altogether.

jayasanka-sack (approved):
Thanks for getting this done and thanks for your patience, @chibongho! 🙏 This took a bit longer than usual, but I really appreciate you sticking with it. Merging now 🚀❤️

PR #2296, opened 2025-03-20
(fix) O3-4361 Re-add close button for patient chart in top nav

chibongho:
## Requirements

### DIFF
--- a/e2e/specs/close-patient-chart.spec.ts
+++ b/e2e/specs/close-patient-chart.spec.ts
@@ -1,0 +1,42 @@
+import { expect } from '@playwright/test';
+import { deletePatient, generateRandomPatient, type Patient } from '../commands';
+import { test } from '../core';
+import { ChartPage } from '../pages';
+
+let patient: Patient;
+
+test.beforeEach(async ({ api }) => {
+  patient = await generateRandomPatient(api);
+});
+
+test('Close patient chart', async ({ page }) => {
+  const chartPage = new ChartPage(page);
+  const topNav = page.getByRole('banner', { name: 'OpenMRS' });
+  const closeButton = topNav.getByRole('button', { name: /close patient chart/i });
+
+  await test.step('When I visit the patient chart', async () => {
+    await chartPage.goTo(patient.uuid);
+  });
+
+  await test.step('Then clicking on it should take me home, making the close button disappear', async () => {
+    await closeButton.click();
+    await expect(page.url()).toMatch(/.*\/spa\/home/);
+    await expect(closeButton).toBeHidden();
+  });
+
+  await test.step('When I visit the patient chart again and navigate to several pages within it', async () => {
+    await chartPage.goTo(patient.uuid);
+    await page.getByRole('link', { name: /medications/i }).click();
+    await page.getByRole('link', { name: /orders/i }).click();
+    await page.getByRole('link', { name: /results/i }).click();
+  });
+
+  await test.step('Then clicking on close button should take me home again', async () => {
+    await closeButton.click();
+    await expect(page.url()).toMatch(/.*\/spa\/home/);
+  });
+});
+
+test.afterEach(async ({ api }) => {
+  await deletePatient(api, patient.uuid);
+});
--- a/packages/esm-patient-chart-app/src/routes.json
+++ b/packages/esm-patient-chart-app/src/routes.json
@@ -199,6 +199,14 @@
       "online": true,
       "offline": true,
       "order": 1
+    },
+    {
+      "name": "patient-chart-close-button",
+      "slot": "top-nav-actions-slot",
+      "component": "closeButton",
+      "online": true,
+      "offline": true,
+      "order": -1
     }
   ],
   "modals": [
--- a/packages/esm-patient-chart-app/src/visit-header/close-button.extension.tsx
+++ b/packages/esm-patient-chart-app/src/visit-header/close-button.extension.tsx
@@ -1,0 +1,47 @@
+import { HeaderGlobalAction } from '@carbon/react';
+import { CloseFilledIcon, getHistory, goBackInHistory, navigate } from '@openmrs/esm-framework';
+import { usePatientChartStore } from '@openmrs/esm-patient-common-lib';
+import React, { useCallback } from 'react';
+import { useTranslation } from 'react-i18next';
+import styles from './close-button.scss';
+
+function CloseButton() {
+  const { t } = useTranslation();
+  const { patientUuid } = usePatientChartStore();
+
+  const onClosePatientChart = useCallback(() => {
+    const history = getHistory();
+    // Get the last page the user was on before opening the patient chart by going backward
+    // through the history until a URL does not include patientUuid
+    let onCloseTarget = '';
+    for (let i = history.length - 1; i >= 0; i--) {
+      if (!history[i].includes(patientUuid)) {
+        onCloseTarget = history[i];
+        break;
+      }
+    }
+
+    if (onCloseTarget) {
+      goBackInHistory({ toUrl: onCloseTarget });
+    } else {
+      navigate({ to: '${openmrsSpaBase}/home' });
+    }
+  }, [patientUuid]);
+
+  const inPatientChart = window.location.pathname.includes(`${window.openmrsBase}/spa/patient/${patientUuid}`);
+  if (!patientUuid || !inPatientChart) {
+    return null;
+  }
+
+  return (
+    <HeaderGlobalAction
+      aria-label={t('closePatientChart', 'Close patient chart')}
+      className={styles.headerGlobalBarCloseButton}
+      onClick={onClosePatientChart}
+    >
+      <CloseFilledIcon size={20} />
+    </HeaderGlobalAction>
+  );
+}
+
+export default CloseButton;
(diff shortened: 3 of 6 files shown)


packages/esm-patient-chart-app/src/routes.json
```diff
@@ -199,6 +199,14 @@
...
       "offline": true,
       "order": 1
+    },
+    {
+      "name": "patient-chart-close-button",
+      "slot": "top-nav-actions-slot",
+      "component": "closeButton",
+      "online": true,
+      "offline": true,
+      "order": -1
```
chibongho, line 209: This gives a warning now because it doesn't like the negative number. I can set this to 0 after this is merged: https://github.com/openmrs/openmrs-esm-patient-management/pull/1530

packages/esm-patient-chart-app/src/visit-header/close-button.extension.tsx
```diff
@@ -0,0 +1,47 @@
...
+import styles from './close-button.scss';
+
+function CloseButton() {
+  const { t } = useTranslation();
+  const { patientUuid } = usePatientChartStore();
+
+  const onClosePatientChart = useCallback(() => {
+    const history = getHistory();
+    // Get the last page the user was on before opening the patient chart by going backward
+    // through the history until a URL does not include patientUuid
```
SarangVT, line 15: The root cause behind getting the close button in spa/home is not resetting patientUuid when closing the patient chart.
Update this using
const patientChartStore = getPatientChartStore();

packages/esm-patient-chart-app/src/visit-header/close-button.extension.tsx
```diff
@@ -0,0 +1,47 @@
...
+    let onCloseTarget = '';
+    for (let i = history.length - 1; i >= 0; i--) {
+      if (!history[i].includes(patientUuid)) {
+        onCloseTarget = history[i];
+        break;
+      }
+    }
+
```
SarangVT, line 23: Add this: patientChartStore.setState({ patientUuid: undefined });
 Updating the patientUuid

packages/esm-patient-chart-app/src/visit-header/close-button.extension.tsx
```diff
@@ -0,0 +1,47 @@
+import { HeaderGlobalAction } from '@carbon/react';
+import { CloseFilledIcon, getHistory, goBackInHistory, navigate } from '@openmrs/esm-framework';
+import { usePatientChartStore } from '@openmrs/esm-patient-common-lib';
```
SarangVT, line 3: Plz update the import for getPatientChartStore
import { getPatientChartStore, usePatientChartStore } from '@openmrs/esm-patient-common-lib';

packages/esm-patient-chart-app/src/routes.json
SarangVT: You may set it to 0 now also, is there any other misconfiguration arising from it, elsewhere?

e2e/specs/close-patient-chart.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  await test.step('Then clicking on close button should take me home again', async () => {
+    await closeButton.click();
+    await expect(page.url()).toMatch(/.*\/spa\/home/);
+  });
+});
+
+test.afterEach(async ({ api }) => {
+  await deletePatient(api, patient.uuid);
+});
```
SarangVT, line 42: Please also try to check if the patient is actually deleted or not
This may help:
const patientExists = await api.get(`/patient/${patient.uuid}`).catch(() => null);
  expect(patientExists).toBeNull();

e2e/specs/close-patient-chart.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  await test.step('When I visit the patient chart', async () => {
+    await chartPage.goTo(patient.uuid);
+  });
+
+  await test.step('Then clicking on it should take me home, making the close button disappear', async () => {
+    await closeButton.click();
+    await expect(page.url()).toMatch(/.*\/spa\/home/);
+    await expect(closeButton).toBeHidden();
+  });
```
SarangVT, line 25: Plz you may also check that if patientUUid is cleared in the global store
This may help:
const patientUuid = await page.evaluate(() => {
  return window.openmrs.getGlobalStore('patient-chart-global-store').getState().patientUuid;
});
expect(patientUuid).toBeUndefined();

SarangVT (commented):
This code review addresses the issue where the Close Patient Chart button remains visible when it should be hidden. Additionally, it includes improvements to the E2E tests to enhance reliability and prevent flaky failures.

Here’s a Loom video demonstrating the fix in action on my local system:
[🔗 Loom Link](https://www.loom.com/share/05e92d2e9de7408db1399dfd8cbd55ad?sid=d2609254-a673-448c-a57c-3c04c19d8e43)

I am Sarang Thakare, and this is my first code review in OpenMRS. I am also applying for GSoC 2025, specifically for the Immunization and Vaccinations App Project.

Thank you for this opportunity!

packages/esm-patient-chart-app/src/visit-header/close-button.extension.tsx
chibongho: The unsetting of the patient should have been handled here when we exit the patient chart:

https://github.com/openmrs/openmrs-esm-patient-chart/blob/c7bcaff703c507cd9db9d8557ec190a6ba5f0d49/packages/esm-patient-chart-app/src/patient-chart/patient-chart.component.tsx#L40-L42

I suspect running `yarn start` is not behaving the same as running the frontend from the docker image. While adding code to unset it works, I think it will cause a double rendering.

ibacher:
> When I ran this via yarn start --sources packages/esm-patient-chart-app

Because of how [start is defined in this repo](https://github.com/openmrs/openmrs-esm-patient-chart/blob/f23f6e9a85a0fcf7f98eac2853696baf2fda41f6/package.json#L9), I think that actually results in duplicating the chart. You can just run `yarn start` to test this out.

packages/esm-patient-chart-app/src/visit-header/close-button.extension.tsx
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    if (onCloseTarget) {
+      goBackInHistory({ toUrl: onCloseTarget });
+    } else {
+      navigate({ to: '${openmrsSpaBase}/home' });
+    }
+  }, [patientUuid]);
+
+  const inPatientChart = window.location.pathname.includes(`${window.openmrsBase}/spa/patient/${patientUuid}`);
```
ibacher, line 31: The routing here is more complex because `${window.openmrsBase}/spa/patient/${patientUuid}/edit` is "in" the registration app.

Also, is preferable to use `window.getOpenmrsSpaBase()` as this ensures that the final `/` is always present. `window.openmrsBase` is a user supplied value that may or may not have a slash depedning on what someone entered. We also support the `spaBase` URL having no relation to `openmrsBase`.

e2e/specs/close-patient-chart.spec.ts
ibacher: Putting expectations into `afterEach` is not something we should be generally recommending. In any case, this is clean-up. Whether or not it is successful is irrelevant to whether or not the test works.

ibacher (commented):
I'm overall unsure about this feature; we've used a close button for the patient chart before, but that was when the patient chart was itself an overlay over the rest of the app. This now introduces a navigation feature that we don't have for any other application, which feels weird.

I'd prefer we had some sort of consensus developed that this is, indeed, a pattern we want to support before implementing this. It may be that this is something we want _generically_ (i.e., if I'm in the dispensing app, should I see a close button to take me back to the home page?) or not at all (i.e., the home page is something I can return to be navigating in the left menu, which is what the Zeplin designs imply to me).

chibongho:
> Because of how [start is defined in this repo](https://github.com/openmrs/openmrs-esm-patient-chart/blob/f23f6e9a85a0fcf7f98eac2853696baf2fda41f6/package.json#L9), I think that actually results in duplicating the chart. You can just run yarn start to test this out.

That's exactly it! The close button works fine by just doing `yarn start`.

> I'm overall unsure about this feature; we've used a close button for the patient chart before, but that was when the patient chart was itself an overlay over the rest of the app. This now introduces a navigation feature that we don't have for any other application, which feels weird.
> 
> I'd prefer we had some sort of consensus developed that this is, indeed, a pattern we want to support before implementing this. It may be that this is something we want generically (i.e., if I'm in the dispensing app, should I see a close button to take me back to the home page?) or not at all (i.e., the home page is something I can return to be navigating in the left menu, which is what the Zeplin designs imply to me).

I can ask on slack to confirm, but I think we need something like this. The close button navigates the user back to the last non-patient-chart page they were in. This could be any app that links to the patient chart, like the queues app. It's reasonable for the user to navigate to the patient chart from the queues app, go to a few tabs within the chart, then "close" it to get back to the queues app. That navigation pattern is less common between other apps, so I think the patient chart is privilege in this sense.

ibacher:
One other question here: why did we land on having the close button the furthest left rather than the furthest right (that is, in the same place it is in the current patient chart with the overlay bar)?

chibongho:
Ahh, right. It used to be like this:
image

I put it to the left because that's the place that doesn't shift the other existing buttons in the `top-nav-actions-slots`. I'll ask UI advisory on this one.

PR #2312, opened 2025-03-27
(fix) O3-4570 : fix order basket failing to load in ward app due to missing patientUuid

donaldkibet:
## Requirements

## Summary
See https://openmrs.atlassian.net/browse/O3-4570

## Screenshots
[Kapture 2025-03-27 at 15.04.13.webm](https://github.com/user-attachments/assets/83a7a120-b709-45cc-9231-635a51ef8065)

### DIFF
--- a/packages/esm-patient-orders-app/src/order-basket/order-basket.workspace.tsx
+++ b/packages/esm-patient-orders-app/src/order-basket/order-basket.workspace.tsx
@@ -9,6 +9,7 @@ import {
   postOrders,
   postOrdersOnNewEncounter,
   useOrderBasket,
+  usePatientChartStore,
   useVisitOrOfflineVisit,
 } from '@openmrs/esm-patient-common-lib';
 import { type ConfigObject } from '../config-schema';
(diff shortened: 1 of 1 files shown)


denniskigen:
@ibacher I can't remember the suggestion you offered for tackling this on the squad call earlier. It makes sense to me that the `patientUuid` isn't available as a prop as we're not rendering this within the context of the patient chart.

denniskigen:
Also, tangential, but why are there two order basket icons in the Ward app siderail?

ibacher:
I don't fully remember, but I think it was to just push it as a prop [here](https://github.com/openmrs/openmrs-esm-patient-management/blob/9c59a05675f595588821f109a8d42461db6a3b24/packages/esm-ward-app/src/ward-patient-card/ward-patient-card.component.tsx#L22). The ward app shouldn't touch the patient chart store, I don't think.

donaldkibet:
> Maybe:
> 
> ```ts
> state: {
>   wardPatient,
>   patient: wardPatient.patient,
>   patientUuid: wardPatient.patient.uuid
> }
> ```

This is works, will add this fix on the ward app

PR #2318, opened 2025-03-31
(fix) O3-4600: Remove leading zero in the dose input in immunization form 

dhruvchandak30:
## Requirements

## Summary

This PR improves the dose input field by fixing validation issues and enhancing the user experience:

Validation Fix: Previously, users could enter 0 as a dose number, which didn’t make sense. Now, only positive integers are allowed.
Dose Limit: To keep things accurate, a sequence can have a maximum of 4 doses.
UI Improvement: The dose input is now a dropdown instead of a free-text field. This fixes an issue where backspacing didn’t work properly—before, if you tried to enter "2," it would show "02." Now, selecting a dose is much smoother.

## Screenshots

Before:
image
image

After:
image

## Related Issue

https://openmrs.atlassian.net/browse/O3-4600


dhruvchandak30:
image
image

@elishabantana  It's running perfectly fine; I didn't get any errors.

elishabantana:
Why does it show , build Expected?

dhruvchandak30:
It is saying `requires an approval from a maintainer` and `At least 1 approving review is required by reviewers with write access.` then only it can proceed with running the build and E2E tests

kilas-Aniso (commented):
@dhruvchandak30 Good job in removing zero as a valid dose input as it doesnt make sense having zero as a dose number like you explained in Jira. Also doses shouldnt be more than a certain number, most vaccines have 2 to 4 doses max so that is a valid point too.

NethmiRodrigo (changes requested):
Hey @dhruvchandak30, thank you for pointing out this bug and working on it, but changing the `NumberInput` to a Dropdown, isn't the ideal fix. We should look into how to get rid of the leading zero, and why it persists. Another place that we use the `NumberInput` forms is in the form-engine-lib for clinical forms, and we don't seem to get this bug in the clinical forms, here's the code for it - https://github.com/openmrs/openmrs-esm-form-engine-lib/blob/main/src/components/inputs/number/number.component.tsx so you can look into it for guidance on how to go forward with this. 
Seems like setting `allowEmpty` to true would do the trick - https://github.com/openmrs/openmrs-esm-form-engine-lib/blob/68966f9d6bb9ebb5f34deaa5f068cb8ec22f3449/src/components/inputs/number/number.component.tsx#L62.
We should also probably set `disableWheel` to true, since if you scroll with the mouse within the input, it would case the value to increase/decrease.

packages/esm-patient-immunizations-app/src/immunizations/components/dose-input.component.tsx
```diff
@@ -34,12 +34,19 @@ export const DoseInput: React.FC<{
         <NumberInput
           id="doseNumber"
           label={t('doseNumberWithinSeries', 'Dose number within series')}
-          min={0}
-          onChange={(event) => field.onChange(parseInt(event.target.value || 0))}
+          min={1}
+          onChange={(event) => {
+            const value = event.target.value;
+            field.onChange(value === '' ? undefined : parseInt(value, 10));
+          }}
```
NethmiRodrigo, line 41: We should avoid having inline JSX functions. Please extract this into its own function definition. Read here - https://o3-docs.openmrs.org/docs/coding-conventions/performance
```
// Bad - new function created every render
<Button onClick={() => handleClick(id)} />
 
// Good - stable function reference
const handleButtonClick = useCallback(() => {
  handleClick(id);
}, [id, handleClick]);
 
<Button onClick={handleButtonClick} />
```
I see in form engine, we've used the second parameter in the onChange function as the value - https://github.com/openmrs/openmrs-esm-form-engine-lib/blob/36f51c8b87e48a114b67e3d9bf42fd18901776f0/src/components/inputs/number/number.component.tsx#L26, we should probably do the same here.

NethmiRodrigo (changes requested):

packages/esm-patient-immunizations-app/src/immunizations/components/dose-input.component.tsx
dhruvchandak30: Made the changes!

NethmiRodrigo (approved):
Thanks @dhruvchandak30!

PR #2329, opened 2025-04-02
(fix) Allergies form should not change value to undefined

ibacher:
## Requirements

## Summary

In the E2E tests, when Playwright clicks on the allergen, two __item_clicked__ state changes are fired by Downshift (the library Carbon uses to implement the ComboBox). The first time this is fired, the selectedItem property is properly set to a value, but the second time it's set to `undefined`. Here we catch any attempts to set the value to `undefined` in the ComboBox onChange() handler and don't update the form state. However, clicking the button to clear the entry appears to send `null`, so this still works. This does lose the functionality where clicking on an already-selected item deselects that item (since that sets the current value to `undefined`), but that feels like a reasonable tradeoff.

## Other

There's a lot of small refactorings here and there. For one things, I missed the extension slot here when pushing the patient "everywhere". Mostly, though, I've just added a lot of memoization, especially where functions call other functions or arrays are passed down as props.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -3,5 +3,5 @@
 
 set -e  # die on error
 
-npx lint-staged
+yarn lint-staged
 yarn turbo run extract-translations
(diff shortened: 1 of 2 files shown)


denniskigen (approved):
LGTM. Thanks, @ibacher!

denniskigen:
FYI @NethmiRodrigo @chibongho

PR #2333, opened 2025-04-02
(feat) Centralize allergy form validation logic

denniskigen:
## Requirements

## Summary

This PR centralizes validation logic in the allergies form workspace so that the zod schema is the source of truth for validation rules. Presently, some validation logic exists in useEffect hooks that determines whether the submit button should be disabled. This logic isn't tied to any display logic, which means that validation errors are not shown to the user when they are encountered.

With this PR, fields are validated when the user attempts to save or update an allergy. If validation fails, the user is presented with the appropriate error message.

## Screenshots

https://github.com/user-attachments/assets/a30d0885-15b0-4bb3-8e4a-a3b0338c064e


NethmiRodrigo (approved):
Thanks @denniskigen, LGTM!

__mocks__/allergies.mock.ts
```diff
@@ -1,9 +1,9 @@
...
-  FOOD = 'FOOD',
-  ENVIRONMENT = 'ENVIRONMENT',
-  OTHER = 'OTHER',
-}
+const ALLERGEN_TYPES = {
+  DRUG: 'DRUG',
+  FOOD: 'FOOD',
+  ENVIRONMENT: 'ENVIRONMENT',
+  OTHER: 'OTHER',
+} as const;
```
denniskigen, line 6: So we're not charlatans who don't practice what we preach 😄

CleanShot 2025-04-04 at 1  38 52@2x

ibacher (approved):
Much cleaner

PR #2347, opened 2025-04-07
(refactor) Improve allergy form initialization and loading states

denniskigen:
## Requirements

## Summary

This commit addresses a critical issue where the allergy form would initialize with incorrect values when editing existing allergies. The problem occurred because:

- Form initialization happened before allergic reactions were loaded from the server
- Default values were only set once during form initialization
- No loading state was shown while data was being fetched

Key changes in this PR include:

- Adding a loading state to the form that waits for allergic reactions to be loaded before initializing default values and rendering the form.
- Adding a ref to track initialization state to prevent the form from being rendered with incorrect values.
- Sorting the allergic reactions alphabetically by display name for improved readability and consistency.
- Adding a `CheckboxSkeleton` loading state to the form when allergic reactions are being loaded to prevent large layout shifts.
- Cleaning up the default value initialization logic for improved readability and maintainability.


jayasanka-sack (approved):
Nice catch, and thanks for fixing it Dennis. much appreciated!

Just a quick note: I noticed the E2E test for editing an allergy doesn’t validate updates to the reaction field. Do you think it makes sense to include that as part of this PR?

NethmiRodrigo:
> Just a quick note: I noticed the E2E test for editing an allergy doesn’t validate updates to the reaction field. Do you think it makes sense to include that as part of this PR?
+1. Was just about to make a ticket for it.

packages/esm-patient-allergies-app/src/allergies/allergies-form/allergy-form.test.tsx
```diff
@@ -126,23 +126,6 @@ describe('AllergyForm', () => {
       // Test severity validation
       await user.click(screen.getByRole('radio', { name: /moderate/i }));
       expect(screen.queryByText(/severity is required/i)).not.toBeInTheDocument();
-
-      // Test successful submission with all required fields
```
NethmiRodrigo, line 130: Question, should this be removed?

NethmiRodrigo (approved):
Thanks @denniskigen! LGTM!

packages/esm-patient-allergies-app/src/allergies/allergies-form/allergy-form.test.tsx
denniskigen: I couldn't get this to work after making the changes to the form initialization logic on edit. So, I elected to remove it for the time being. FWIW, we already have some other tests that test the submission flow, so it felt "safe" to remove these assertions.

denniskigen:
Updated in latest commit!

PR #2363, opened 2025-04-09
(fix) O3-4363 - Add pagination to visit context switcher

chibongho:
## Requirements

## Summary

Currently, the visit context switcher only shows a set number of visits, and there is no way to select older visits pass that maximum number. This PR adds pagination support to the visit list.

## Screenshots

image


ibacher:
This is a widget where I think infinite scroll makes more sense than pagination, no? Even better would probably be to be able to search by date, though.

chibongho:
Yeah, I wish we have better UX guidance on this one. 

I still think infinite scroll is bad UX, if you know exactly what which visit you're looking for and it's near the end of the list. That said, even with pagination alone, the user needs to fish for the right page containing the visit they want. How about pagination + date picker?

ibacher:
So, my only reason for arguing for infinite scroll here is to keep things consistent with the Location Picker, since that's  what this most closely resembles. I definitely agree with a date picker or better yet, at date range picker.

denniskigen:
@ibacher @chibongho do you reckon this is good to go for this release cycle?

ibacher:
Yes, I think so.

chibongho:
Updated visit context switcher to use infinite scroll instead of pagination. Also added a date picker to limit visits started on or prior to selected date.

https://github.com/user-attachments/assets/736ed87f-c1f6-4069-8d98-3eb0892d2577

Based on (somewhat outdated) mockup from: 
https://app.zeplin.io/project/60d59321e8100b0324762e05/screen/679761bdffaf2ed81745f257

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-switcher.modal.tsx
```diff
@@ -21,13 +23,27 @@ const VisitContextSwitcherModal: React.FC<VisitContextSwitcherProps> = ({
...
   const { t } = useTranslation();
-  // TODO: add pagination for the visits list.
-  const { visits, isLoading, error } = useInfiniteVisits(patientUuid);
+  const [maxStartDate, setMaxStartDate] = useState(new Date());
+  const maxStartDateDebounced = useDebounce(maxStartDate);
+
+  const rep = 'custom:(uuid,display,visitType,startDatetime,stopDatetime,location)';
+  const { visits, isLoading, error, hasMore, loadMore } = useInfiniteVisits(
+    patientUuid,
+    { toStartDate: dayjs(maxStartDateDebounced).endOf('day').toISOString() },
```
ibacher, line 32: Shouldn't the filtering for startDate be up to the beginning of the day or an I misunderstanding the logic here?

packages/esm-patient-chart-app/src/visit/visits-widget/visit.resource.tsx
```diff
@@ -5,28 +5,45 @@ import {
...
 
-export function usePaginatedVisits(patientUuid: string, pageSize: number) {
+export function usePaginatedVisits(
+  patientUuid: string,
+  pageSize: number,
+  params: Record<string, number | string> = {},
+) {
   const url = new URL(
-    makeUrl(`${restBaseUrl}/visit?patient=${patientUuid}&v=${customRepresentation}`),
+    `${window.openmrsBase}/${restBaseUrl}/visit?patient=${patientUuid}&v=${customRepresentation}`,
```
ibacher, line 41: It's preferential to use `getOpenmrsSpaBase()`, which just guarantees that there is always a `/` at the end of the `spaBase`. The `spaBase` variable itself is whatever was defined at build time, which may or may not have a trailing slash.

```suggestion
    `${window.getOpenmrsSpaBase()}${restBaseUrl}/visit?patient=${patientUuid}&v=${customRepresentation}`,
```

packages/esm-patient-common-lib/src/pagination/pagination.component.tsx
```diff
@@ -31,7 +34,13 @@ export const PatientChartPagination: React.FC<PatientChartPaginationProps> = ({
   return (
     <>
       {totalItems > 0 && (
-        <div className={isTablet ? styles.tablet : styles.desktop}>
+        <div
+          className={classNames(
+            isTablet ? styles.tablet : styles.desktop,
+            styles.paginationContainer,
+            grey ? styles.greyBackground : '',
+          )}
```
ibacher, line 42: One reason to use `classnames` is because it supports this ternary thing kind of natively, e.g.

```suggestion
          className={classNames(styles.paginationContainer, {
          	[styles.tablet]: isTablet,
            [styles.desktop]: !isTablet ,
            [styles.greyBackground]: grey
          })}
```

ibacher (commented):
Couple of small suggestions and a question, but basically LGTM

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-switcher.modal.tsx
chibongho: The filter is for start times on or prior to that date, so if "2025-05-01" is selected, it should include all times prior to "2025-05-01 11:59:59 PM" in local time.

ibacher: Got it!

ibacher (approved):

PR #2371, opened 2025-04-11
(feat) O3-4615: Fetch paginated encounters directly in the encounters table

chibongho:
## Requirements

## Summary

Currently, the “All encounters” table doesn’t actually fetch encounter data directly. Instead, it loops through the visits data loaded in “Visit summaries” and gets the encounters from each visit.

This PR rewrites the "All encounters" table to fetch paginated encounters directly. 

Notes:
- Confusingly, the component displaying "All encounters" was named "Visits Table", and there was also an "EncountersTable" component that's not being used. The unused one is now removed and the "Visit Table" is renamed to "EncountersTable".
- The search filtering is refactored to be done on the server side instead of browser side, and is changed to only support filtering by `encounterType`. (free text search filtering is removed)
- Previously, the same table component is used for both in "All encounters" tab (which displays all encounters from all visits) and  encounters section in the visits summary card (which only displays encounters from one single visit). This is now separated into 2 separate components (`AllEncountersTable` and `VisitEncountersTable`).

## Screenshots

image

image

## Related Issue

https://openmrs.atlassian.net/browse/O3-4615


denniskigen:
Tangential, but can we change the order of the items in the Content Switcher? If we're defaulting to the Summary cards view, I think it makes sense to have that as the first Switch.

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/all-encounters-table.component.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+interface AllEncountersTableProps {
+  patientUuid: string;
+}
+
+/**
+ * This component shows a table of all encounters (across all visits) of a patient
+ */
+const AllEncountersTable: React.FC<AllEncountersTableProps> = ({ patientUuid }) => {
+  const pageSize = 20;
```
ibacher, line 14: Presumably at some point this will become somehow configurable? In general, doesn't it make sense to have this be a property controlled by the datatable itself, i.e., via something like the "Items per page:" bit [here](https://carbondesignsystem.com/components/data-table/usage/#anatomy)?

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/all-encounters-table.component.tsx
```diff
@@ -0,0 +1,46 @@
...
+    currentPage,
+    isLoading,
+    totalCount,
+    goTo,
+    mutate: mutateEncounters,
+  } = usePaginatedEncounters(patientUuid, encounterTypeToFilter?.uuid, pageSize);
+
+  const paginatedEncounterRows = useMemo(() => {
+    return paginatedEncounters?.map(mapEncounter);
+  }, [paginatedEncounters]);
```
ibacher, line 28: Is there a reason we need to do this here instead of in `usePaginatedEncounters`?

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/encounters-table.component.tsx
```diff
@@ -0,0 +1,330 @@
...
+                      <TableHeader className={styles.tableHeader} key={i} {...getHeaderProps({ header })}>
+                        {header.header}
+                      </TableHeader>
+                    ))}
+                    <TableExpandHeader />
+                  </TableRow>
+                </TableHead>
+                <TableBody>
+                  {rows?.map((row, i) => {
+                    const encounter = paginatedMappedEncounters[i];
```
ibacher, line 202: I think the remaining properties of the encounter are already available on the `row` object. See [here](https://github.com/carbon-design-system/carbon/blob/8382893451a000b31255022110df69ec80f75ff7/packages/react/src/components/DataTable/DataTable.tsx#L108)

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/encounters-table.component.tsx
```diff
@@ -0,0 +1,330 @@
...
+            <TableContainer className={styles.tableContainer}>
+              <TableToolbar {...getToolbarProps()}>
+                {encounterTypeToFilter !== undefined && (
+                  <TableToolbarContent>
+                    <div className={styles.filterContainer}>
+                      <ComboBox
+                        className={styles.substitutionType}
+                        id="encounterTypeFilter"
+                        aria-label={t('filterByEncounterType', 'Filter by encounter type')}
+                        size={'sm'}
```
denniskigen, line 175:
```suggestion
                        size='sm'
```

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/encounters-table.resource.tsx
denniskigen, line 1: Can we change this to a `.ts` extension since it doesn't contain JSX?

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/all-encounters-table.component.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const encountersTableProp: EncountersTableProps = {
```
denniskigen, line 30: nit: shouldn't this variable name be pluralized too?

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/encounters-table.component.tsx
denniskigen: We usually do something like this elsewhere to determine the corresponding encounter:

```tsx
  const encounter = paginatedMappedEncounters.find(enc => enc.id === row.id);
```

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/encounters-table.test.tsx
```diff
@@ -1,41 +1,58 @@
...
-  visits: mockMappedEncounters,
-};
+import { screen, within } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { mockEncountersAlice, mockEncounterTypes, mockPatientAlice } from '__mocks__';
+import { renderWithSwr } from 'tools';
+import EncountersTable from './encounters-table.component';
+import { type EncountersTableProps, mapEncounter, useEncounterTypes } from './encounters-table.resource';
+
+const defaultProps: () => EncountersTableProps = () => ({
```
denniskigen, line 10:
```suggestion
const buildTestProps: () => EncountersTableProps = () => ({
```

nit: Naming

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/encounters-table.component.tsx
chibongho: Yeah, the `row` object contains the cells and not the original object itself. Since we have the index `i`, I think we should just do `paginatedMappedEncounters[i]`.

image

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/all-encounters-table.component.tsx
chibongho: I think the `MappedEncounter` type is really specific to how this component transforms `Encounters` into Carbon DataTable rows, and want to keep it out of the data fetching hook (even though the hook is defined in `encounter-table.resource.ts` and is meant to be only used here). It's also why I refactored the mocks encounters to be of the standard `Encounter` type instead of `MappedEncounter`.

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/encounters-table.component.tsx
ibacher: Reasonable. Yeah, if it doesn't have the props, then doing the look-up by index is definitely better than running `find()`.

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/all-encounters-table.component.tsx
ibacher: From that perspective, wouldn't it make sense to push the mapping down into the `EncounterTable` instead? That way we don't have to call `mapEncounter()` both here and in the `VisitEncountersTable` component and the `EncountersTable` thing can just take standard encounters.

chibongho: yeah, that makes sense.

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/encounters-table.component.tsx
```diff
@@ -0,0 +1,330 @@
...
+    [onEncountersUpdated, t],
+  );
+
+  if (isLoadingEncounterTypes) {
+    return <DataTableSkeleton role="progressbar" zebra />;
+  }
+
+  return (
+    <div className={styles.container}>
+      <DataTable
```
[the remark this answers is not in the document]
chibongho: This is done at the `<TableContainer>`, is that fine?

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/encounters-table.component.tsx
chibongho: I think I missed conditionally sizing this depending on the desktop / tablet. I'll change.

chibongho:
> Tangential, but can we change the order of the items in the Content Switcher? If we're defaulting to the Summary cards view, I think it makes sense to have that as the first Switch.

I had another PR almost ready to just remove the Summary Cards view altogether along with updates to unit / e2e tests.

ibacher (approved):

denniskigen (approved):
Looks good to me. Thanks!

PR #2374, opened 2025-04-13
(fix) O3-4631: Make vitals overdue tag threshold configurable

dhruvchandak30:
## Requirements

## Summary

There is an issue with how our system flags vitals as “out of date.” When I recorded vitals at 11:42 PM last night (April 9th), they showed up as completely current. However, just 21 minutes later after midnight (at 00:03 on April 10th), these same vitals suddenly appeared with an “out of date” warning tag. This seems misleading since these measurements are still very recent!

The problem appears to be in our code that uses calendar dates rather than actual elapsed time. I considered changing this to use a time window (like 12 hours) instead of strictly checking if vitals were taken “today”.

## Related Issue

https://openmrs.atlassian.net/browse/O3-4631


ibacher (approved):

ibacher (changes requested):
Can we make this time window configurable?

packages/esm-patient-vitals-app/src/config-schema.ts
```diff
@@ -58,6 +58,11 @@ export const configSchema = {
       _type: Type.UUID,
       _default: '67a71486-1a54-468f-ac3e-7091a9a79584',
     },
+    thresholdHours: {
```
NethmiRodrigo, line 61: Nit:
Maybe we could rename this to be slightly more descriptive like - `vitalsOverdueThresholdHours`

packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.component.tsx
```diff
@@ -61,14 +61,22 @@ const VitalsHeader: React.FC<VitalsHeaderProps> = ({ patientUuid, hideLinks = fa
 
   if (latestVitals && Object.keys(latestVitals)?.length && conceptMetadata?.length) {
     const hasActiveVisit = Boolean(currentVisit?.uuid);
-    const vitalsTakenToday = Boolean(dayjs(latestVitals?.date).isToday());
-    const vitalsOverdue = hasActiveVisit && !vitalsTakenToday;
     const now = dayjs();
-    const vitalsOverdueDayCount = Math.round(dayjs.duration(now.diff(latestVitals?.date)).asDays());
+    const vitalsTakenTimeAgo = dayjs.duration(now.diff(latestVitals?.date));
+    const VITALS_THRESHOLD_HOURS = config.vitals.thresholdHours;
```
ibacher, line 66:
```suggestion
    const vitalsThresholdHours = config.vitals.thresholdHours;
```

packages/esm-patient-vitals-app/translations/en.json
```diff
@@ -32,6 +32,8 @@
   "goToSummary": "Go to Summary",
   "heartRate": "Heart rate",
   "height": "Height",
+  "hoursOldVitals_one": "<0>These vitals are <1>{Math.round(vitalsTakenTimeAgo.asHours())} hours old</1></0>",
```
NethmiRodrigo, line 35:
```suggestion
  "hoursOldVitals_one": "<0>These vitals are <1>{Math.round(vitalsTakenTimeAgo.asHours())} hour old</1></0>",
```

packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.component.tsx
```diff
@@ -61,14 +61,22 @@ const VitalsHeader: React.FC<VitalsHeaderProps> = ({ patientUuid, hideLinks = fa
 
   if (latestVitals && Object.keys(latestVitals)?.length && conceptMetadata?.length) {
     const hasActiveVisit = Boolean(currentVisit?.uuid);
-    const vitalsTakenToday = Boolean(dayjs(latestVitals?.date).isToday());
-    const vitalsOverdue = hasActiveVisit && !vitalsTakenToday;
     const now = dayjs();
-    const vitalsOverdueDayCount = Math.round(dayjs.duration(now.diff(latestVitals?.date)).asDays());
+    const vitalsTakenTimeAgo = dayjs.duration(now.diff(latestVitals?.date));
+    const vitalsThresholdHours = config.vitals.vitalsOverdueThresholdHours;
+    const vitalsOverdue = hasActiveVisit && vitalsTakenTimeAgo.asHours() > vitalsThresholdHours;
```
NethmiRodrigo, line 67: If the number of hours that the vitals were last taken is exactly the number of hours in the threshold, it won't be flagged as overdue. @ibacher @denniskigen Is that expected, or should we change this condition to equal and greater than? If the latter, we should probably change the description of the config too.

ibacher: Equal to or greater than makes sense.

denniskigen: Could you effect this change, @dhruvchandak30?

NethmiRodrigo (approved):

PR #2378, opened 2025-04-15
(fix) Support multiple units in BMI calculation

VaishSiddharth:
## Requirements

## Summary
Right now if weight and height are not in `kg` and `m` respectively the BMI is wrongly calculated.
With this PR multiple units `weight 'lb', 'lbs', 'g'`, `height: 'm', 'cm', 'in'` will be supported.

## Related Issue
[O3-1781](https://openmrs.atlassian.net/browse/O3-1781)

## Other

Slack conversation: https://openmrs.slack.com/archives/C02UNMKFH8V/p1744360774246389

[O3-1781]: https://openmrs.atlassian.net/browse/O3-1781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
+++ b/packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
@@ -1,7 +1,19 @@
 import isNumber from 'lodash/isNumber';
 import { type ConceptMetadata } from '../common';
 
-export function calculateBodyMassIndex(weight: number, height: number): number {
+export function calculateBodyMassIndex(weight: number, height: number, weightUnit: string, heightUnit: string): number {
+  if (weightUnit == 'lb') {
+    weight = weight * 0.45359237;
+  }
+  if (weightUnit == 'gm') {
+    weight = weight / 1000;
+  }
+  if (heightUnit == 'm') {
+    height = height * 100;
+  }
+  if (heightUnit == 'in') {
+    height = height * 0.0254;
+  }
   if (!weight || !height) return;
 
   if (weight > 0 && height > 0) {
--- a/packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
+++ b/packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
@@ -2,7 +2,7 @@ import isNumber from 'lodash/isNumber';
 import { type ConceptMetadata } from '../common';
 
 export function calculateBodyMassIndex(weight: number, height: number, weightUnit: string, heightUnit: string): number {
-  if (weightUnit == 'lb') {
+  if (weightUnit == 'lb' || weightUnit == 'lbs') {
     weight = weight * 0.45359237;
   }
   if (weightUnit == 'g') {
--- a/packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
+++ b/packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
@@ -5,7 +5,7 @@ export function calculateBodyMassIndex(weight: number, height: number, weightUni
   if (weightUnit == 'lb') {
     weight = weight * 0.45359237;
   }
-  if (weightUnit == 'gm') {
+  if (weightUnit == 'g') {
     weight = weight / 1000;
   }
   if (heightUnit == 'm') {
--- a/packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
+++ b/packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
@@ -1,23 +1,28 @@
 import isNumber from 'lodash/isNumber';
 import { type ConceptMetadata } from '../common';
 
-export function calculateBodyMassIndex(weight: number, height: number, weightUnit: string, heightUnit: string): number {
-  if (weightUnit == 'lb' || weightUnit == 'lbs') {
-    weight = weight * 0.45359237;
-  }
-  if (weightUnit == 'g') {
-    weight = weight / 1000;
-  }
-  if (heightUnit == 'm') {
-    height = height * 100;
-  }
-  if (heightUnit == 'in') {
-    height = height * 0.0254;
-  }
+export function calculateBodyMassIndex(
+  weight: number,
+  height: number,
+  weightUnit: 'lb' | 'lbs' | 'g',
+  heightUnit: 'm' | 'cm' | 'in',
+): number {
   if (!weight || !height) return;
 
   if (weight > 0 && height > 0) {
-    return Number((weight / (height / 100) ** 2).toFixed(1));
+    if (weightUnit === 'lb' || weightUnit === 'lbs') {
+      weight = weight * 0.45359237;
+    }
+    if (weightUnit === 'g') {
+      weight = weight / 1000;
+    }
+    if (heightUnit === 'cm') {
+      height = height / 100;
+    }
+    if (heightUnit === 'in') {
+      height = height * 0.0254;
+    }
+    return Number((weight / height ** 2).toFixed(1));
   }
 }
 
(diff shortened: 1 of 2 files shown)


packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
```diff
@@ -1,7 +1,19 @@
 import isNumber from 'lodash/isNumber';
 import { type ConceptMetadata } from '../common';
 
-export function calculateBodyMassIndex(weight: number, height: number): number {
+export function calculateBodyMassIndex(weight: number, height: number, weightUnit: string, heightUnit: string): number {
+  if (weightUnit == 'lb') {
+    weight = weight * 0.45359237;
+  }
+  if (weightUnit == 'gm') {
```
rbuisson, line 8: `gm` for grams I assume. IU label for gram is `g`.

denniskigen:
Can we add a PR summary for posterity?

packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
```diff
@@ -1,7 +1,19 @@
 import isNumber from 'lodash/isNumber';
 import { type ConceptMetadata } from '../common';
 
-export function calculateBodyMassIndex(weight: number, height: number): number {
+export function calculateBodyMassIndex(weight: number, height: number, weightUnit: string, heightUnit: string): number {
```
ibacher, line 4:
```suggestion
export function calculateBodyMassIndex(weight: number, height: number, weightUnit: 'lb' | 'lbs' | 'g', heightUnit: 'm' | 'cm' | 'in'): number {
```

packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
```diff
@@ -1,7 +1,19 @@
...
+  if (weightUnit == 'g') {
+    weight = weight / 1000;
+  }
+  if (heightUnit == 'm') {
+    height = height * 100;
+  }
+  if (heightUnit == 'in') {
+    height = height * 0.0254;
+  }
   if (!weight || !height) return;
```
ibacher, line 17: This should still be the first line. No point in trying to "convert" an undefined height or weight. We should probably also do the conversion after the `> 0` guards.

packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
```diff
@@ -1,7 +1,19 @@
 import isNumber from 'lodash/isNumber';
 import { type ConceptMetadata } from '../common';
 
-export function calculateBodyMassIndex(weight: number, height: number): number {
+export function calculateBodyMassIndex(weight: number, height: number, weightUnit: string, heightUnit: string): number {
+  if (weightUnit == 'lb' || weightUnit == 'lbs') {
```
ibacher, line 5: For comparisons in JS, we always want to use `===`.

packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
```diff
@@ -1,7 +1,19 @@
...
+    weight = weight * 0.45359237;
+  }
+  if (weightUnit == 'g') {
+    weight = weight / 1000;
+  }
+  if (heightUnit == 'm') {
```
ibacher, line 11: This branch should be for cm, right? BMI is kg / m ** 2, so we don't need to change a value given in meters.

ibacher (changes requested):
Thanks! We could use a slightly better commit message ("Support multiple units in BMI calculation") and a description of why we're making this change in the ticket.

packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
VaishSiddharth: Did this because the [existing function](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts#L8) expects height in centimetre.

Okay so for ease of understanding, refactoring the code. WDYT?

```
export function calculateBodyMassIndex(
  weight: number,
  height: number,
  weightUnit: 'lb' | 'lbs' | 'g',
  heightUnit: 'm' | 'cm' | 'in',
): number {
  if (!weight || !height) return;

  if (weight > 0 && height > 0) {
    if (weightUnit === 'lb' || weightUnit === 'lbs') {
      weight = weight * 0.45359237;
    }
    if (weightUnit === 'g') {
      weight = weight / 1000;
    }
    if (heightUnit === 'cm') {
      height = height / 100;
    }
    if (heightUnit === 'in') {
      height = height * 0.0254;
    }
    return Number((weight / (height ** 2)).toFixed(1));
  }
```

packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
VaishSiddharth: Not sure if this is the best way because then we have to call the function like this

```
const computedBodyMassIndex = calculateBodyMassIndex(
        weight,
        height,
        conceptUnits.get(config.concepts.weightUuid) as 'lb' | 'lbs' | 'g',
        conceptUnits.get(config.concepts.heightUuid) as 'm' | 'cm' | 'in',
      );
```

VaishSiddharth:
@ibacher can you please have a final look.

packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
ibacher: Oh. That's fair. I hadn't realized we were depeding on the input being cm instead of m, but, I guess that makes sense.

ibacher (approved):

denniskigen:
Thanks for working on this, @VaishSiddharth. I came back to it because [Anish asked on Slack](https://openmrs.slack.com/archives/C02UNMKFH8V/p1775811396896769) whether imperial BMI was fully supported in the RefApp. Digging in, I found there's an older Jira ticket for the same problem ([O3-1781](https://openmrs.atlassian.net/browse/O3-1781), filed by Grace) that wasn't linked from this PR. It has a few extra acceptance criteria worth surfacing for follow-up work, since they're not addressed yet:

1. **Case-insensitive unit matching, with trailing `s` and `.` stripped.** The current `===` check in `vitals-biometrics-form.utils.ts` only matches the exact literals `lb`, `lbs`, `g`, `m`, `cm`, `in`. So `Lb`, `LB`, `Lbs`, `IN`, `lb.`, `cms` etc. silently fall through and get treated as kg/m, which would bite any deployment whose dictionary uses non-canonical casing.

2. **Error on unrecognized units.** Right now if the unit string doesn't match any case, the function returns `weight / height ** 2` on the raw values rather than throwing. That produces a wrong BMI with no signal anything is off, which is the same failure mode as the original bug. The ticket explicitly asked the calculator to raise on unknown units.

3. **Display-side calculator is unchanged.** There's a second `calculateBodyMassIndex` in `packages/esm-patient-vitals-app/src/common/helpers.ts` that still hardcodes kg + cm. It's used as a fallback in `common/data.resource.ts` for historical encounters without a stored BMI obs, so the vitals overview table can still misreport BMI for older imperial-unit data.

None of this is a critique of the PR, just wanted to leave a trail for whoever picks up the next pass. Happy to file a follow-up ticket or open a PR for the above if that's useful.

[O3-1781]: https://openmrs.atlassian.net/browse/O3-1781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2395, opened 2025-04-18
(fix) O3-4615 fix encounters table filtering and pagination

chibongho:
## Requirements

## Summary

Fixes a couples things in the encounters table.
- Pagination page size change was straight up unimplemented. That was an oversight by me.
- The encounter filter is supposed to show up in the "All Encounters" table, but not in the Visit Card's encounters table. The filter was conditionally rendered with `encounterTypeToFilter !== undefined`, which worked in dev but not in prod. I made an explicit boolean param for conditionally rendering it instead.

### DIFF
--- a/packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/encounters-table.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/encounters-table.component.tsx
@@ -22,7 +22,8 @@ import {
   ComboBox,
   InlineLoading,
   DataTableSkeleton,
- Pagination } from '@carbon/react';
+  Pagination,
+} from '@carbon/react';
 import {
   EditIcon,
   isDesktop,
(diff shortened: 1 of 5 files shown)


packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/encounters-table.component.tsx
```diff
@@ -324,12 +327,12 @@ const EncountersTable: React.FC<EncountersTableProps> = ({
           forwardText={t('nextPage', 'Next page')}
```
denniskigen, line 327: Any reason why we're not reusing [PatientChartPagination](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-common-lib/src/pagination/pagination.component.tsx#L18) here? I think it makes sense to for consistency with other widgets in the Chart.

denniskigen (commented):
LGTM, save for one concern about the need for rolling our own custom pagination solution instead of reusing the `PatientChartPagination` component from common-lib.

denniskigen (approved):
LGTM, save for one concern about the need for rolling our own custom pagination solution instead of reusing the `PatientChartPagination` component from common-lib.

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/encounters-table.component.tsx
chibongho: `PatientCharPagination` is a more stripped down component that does not allow for changing pageSize. I think it's more appropriate for smaller widgets with limited height.

image

denniskigen: Yeah, that makes sense.

PR #2401, opened 2025-04-21
(perf) O3-4656: Optimized Order Detail's Table with client-side filtering to reduce redundant API calls

dhruvchandak30:
## Requirements

## Summary

Optimized the `OrderDetailsTable` by avoiding repeated API calls when switching order types. Now, all orders are fetched once initially, and we use `useMemo` to filter them on the client side. This makes the UI faster and reduces unnecessary load on the backend, without changing any user-facing features.

## Screenshots

Before:
https://github.com/user-attachments/assets/f9b75561-3a58-428d-8e68-c6d25e113d19

After:
https://github.com/user-attachments/assets/43462d0f-709b-4f1b-a949-121951859382

## Related Issue

https://openmrs.atlassian.net/browse/O3-4656

### DIFF
--- a/packages/esm-patient-orders-app/src/components/order-details-table.component.tsx
+++ b/packages/esm-patient-orders-app/src/components/order-details-table.component.tsx
@@ -129,11 +129,17 @@ const OrderDetailsTable: React.FC<OrderDetailsProps> = ({
   const [selectedToDate, setSelectedToDate] = useState(null);
   const selectedOrderName = orderTypes?.find((x) => x.uuid === selectedOrderTypeUuid)?.name;
   const {
-    data: allOrders,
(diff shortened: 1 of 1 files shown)


dhruvchandak30:
@denniskigen  Can you please review this?
Thank you!

denniskigen:
Thanks for taking this on, @dhruvchandak30. I don't think this is the correct approach. Leaving the filtering to the search handler on the backend is what we think is the best approach here.

dhruvchandak30:
> Thanks for taking this on, @dhruvchandak30. I don't think this is the correct approach. Leaving the filtering to the search handler on the backend is what we think is the best approach here.

You're right, thank you for clarifying!

PR #2408, opened 2025-04-28
(fix) O3-4481: Visit table should update after changes to visit

chibongho:
## Requirements

## Summary

In the patient chart, we have various places that require mutating visit data (ex: on creating visit, on deleting visit). Currently, the mutation functions are obtained by calling `useVisit` and `useInfiniteVisit`. This is an incomplete solution, and is causing the visits table to not update when needed. This PR makes changes to store the mutate functions in the visit store and call them where appropriate.

Note:
- The visit context store has been moved to core. This PR deletes the one implemented in this repo in favor of the core one.
- The `numberOfVisitsToLoad` config option is removed in this PR. it was limiting the number of rows returned by the `useInfiniteVisits` hook. This was done for performance reasons for the Visit Summary Cards infinite scroll view, which has been removed. We still use `useInfiniteVisits` for the infinite scroll view in the [visit context switcher](https://github.com/openmrs/openmrs-esm-patient-chart/pull/2363), but it's loading substantially less fields that I don't think we need to limit it with this config.

## Screenshots

Video capture of manual testing. Note the updates to the visit table and the "Active Visit" pill in patient banner as visits are added / modified / deleted.

https://github.com/user-attachments/assets/0fbb9f65-c706-4ebd-99d1-a27415bf3d0f


makombe:
@chibongho can we make `Diagnoses` column configurable. Maybe not all visits will have this info. It will also allow implementations be able to show it if needed.

chibongho:
@makombe Yes, we plan on making the columns configurable. Ticket [here](https://openmrs.atlassian.net/browse/O3-4480).

NethmiRodrigo (approved):

denniskigen (approved):
Excellent work, @chibongho! Thanks. This definitely warrants a patch of Patient Chart for inclusion in 3.4.0.

PR #2167, opened 2025-01-06
(fix) O3-4302: Hide voided diagnoses in patient chart visits widget

CynthiaKamau:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspiration.
- [ ] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://om.rs/o3ui).
- [ ] My work includes tests or is validated by existing tests.

## Summary

Hide voided diagnosis in visits widget on patient chart

## Screenshots

Screenshot 2025-01-06 at 13 54 42

PR #2412, opened 2025-04-29
(feat) O3-4038: Add edit capability to visit notes form

denniskigen:
## Requirements

## Summary

This PR adds support for editing existing visit notes. It also updates the visit note form test suite to cover the new functionality.

## Screenshots

https://github.com/user-attachments/assets/103e5ee6-c77e-4ed1-9d86-b763cafcfd8a

## Related Issue

https://openmrs.atlassian.net/browse/O3-4038

### DIFF
--- a/packages/esm-patient-notes-app/src/notes/visit-notes.resource.ts
+++ b/packages/esm-patient-notes-app/src/notes/visit-notes.resource.ts
@@ -1,13 +1,6 @@
 import useSWR from 'swr';
 import useSWRInfinite from 'swr/infinite';
-import {
-  type Diagnosis,
-  type Obs,
-  openmrsFetch,
-  type OpenmrsResource,
-  restBaseUrl,
-  useConfig,
-} from '@openmrs/esm-framework';
+import { openmrsFetch, restBaseUrl, useConfig } from '@openmrs/esm-framework';
 import { type ConfigObject } from '../config-schema';
 import type {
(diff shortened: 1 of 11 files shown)


packages/esm-patient-notes-app/src/notes/visit-notes.resource.ts
```diff
@@ -11,6 +18,23 @@ import type {
   VisitNotePayload,
 } from '../types';
 
+export interface MappedEncounter {
```
chibongho, line 21: I'd prefer using the standard `Encounter` type and keeping the mapped type confined to the table itself. (That way there is no type duplication, and the mockEncounters in unit tests can be of the standard type.) This can be future work.

chibongho (approved):
Looks good! Couple of other comments:

- Should we have the backend not return voided diagnoses by default? (I think this has come up [before](https://github.com/openmrs/openmrs-esm-patient-chart/pull/2167#issuecomment-2588142459)). I think this is a problem specific to diagnoses and not other resources.
- When we edit an existing encounter, should the RDE visit context switcher disable changing of the visit context? I can't think of a scenario where it makes sense to change the visit the encounter is associated with.

denniskigen:
Ok, I've filed https://openmrs.atlassian.net/browse/O3-4677 to keep track of the issues raised from the code reviews.

PR #2419, opened 2025-05-01
(feat) O3-4675: Support passing patient, visit, and representation in…

mseaton:
…to customFormsUrl in patient-forms-app

## Related Issue
https://issues.openmrs.org/browse/O3-4675


packages/esm-patient-forms-app/src/hooks/use-forms.ts
```diff
@@ -10,17 +10,28 @@ import {
...
+      : customFormsUrl
     : showHtmlFormEntryForms
-    ? formEncounterUrl
-    : formEncounterUrlPoc;
+      ? formEncounterUrl
+      : formEncounterUrlPoc;
+  url = url
+    .replace('{patientUuid}', patientUuid)
+    .replace('{visitUuid}', visitUuid)
+    .replace('{representation}', customFormRepresentation);
```
ibacher, line 34: We actually have a function to do this, called `interpolateUrl()` in the framework, which works with `${patientUuid}`-style syntax. I think it would be nicer to consistently use that here.

mogoodrich (approved):
LGTM though defer to @ibacher  on proper utility methods to use

packages/esm-patient-forms-app/src/hooks/use-forms.ts
mseaton: Oh nice, thanks @ibacher - I looked for something like that but didn't find it.  I've updated the code to use this.  Let me know how that looks.

PR #2434, opened 2025-05-08
(test) 03-4687: Add Automated e2e test for cancelling a drug Order

gitcliff:
## Requirements

## Summary
This PR is adding automated test for cancelling a drug Order

## Related Issue
https://openmrs.atlassian.net/browse/O3-4687

### DIFF
--- a/e2e/specs/drug-orders.spec.ts
+++ b/e2e/specs/drug-orders.spec.ts
@@ -2,7 +2,7 @@ import { expect } from '@playwright/test';
 import { type Visit } from '@openmrs/esm-framework';
 import { generateRandomPatient, deletePatient, type Patient, startVisit, endVisit } from '../commands';
 import { test } from '../core';
-import { MedicationsPage } from '../pages';
+import { MedicationsPage, OrdersPage } from '../pages';
 
 let patient: Patient;
 let visit: Visit;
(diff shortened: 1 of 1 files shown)


e2e/specs/drug-orders.spec.ts
```diff
@@ -222,6 +222,140 @@ test('Record, edit and discontinue a drug order', async ({ page }) => {
...
+    await page.getByRole('button', { name: /sign and close/i }).click();
+  });
+
+  await test.step('Then I should see a success notification', async () => {
+    await expect(page.getByText(/placed order for aspirin/i)).toBeVisible();
+  });
+
+  await test.step('And I should see the newly added order in the active medications list', async () => {
+    const headerRow = medicationsPage.medicationsTable().locator('thead > tr');
+    const dataRow = medicationsPage.medicationsTable().locator('tbody > tr');
```
ibacher, line 318: Why are these rules defined using `medicationsPage`. I think we can just use `page` here.

e2e/specs/drug-orders.spec.ts
```diff
@@ -222,6 +222,140 @@ test('Record, edit and discontinue a drug order', async ({ page }) => {
...
+    await expect(dataRow).toContainText(/aspirin 81mg/i);
+    await expect(dataRow).toContainText(/1 tablet/i);
+    await expect(dataRow).toContainText(/oral/i);
+    await expect(dataRow).toContainText(/once daily/i);
+    await expect(dataRow).toContainText(/3 days/i);
+    await expect(dataRow).toContainText(/indication headache/i);
+  });
+
+  await test.step('When I click on the Orders section', async () => {
+    await ordersPage.page.getByRole('link', { name: 'Orders' }).click();
```
ibacher, line 331: Ditto here. I don't think we need to create the `OrdersPage` instance unless we actually need some functionality there, but everything looks like it could be re-written in terms of `page` with no functionality lost?

ibacher (commented):
Thanks Cliff. Mostly looks good, but a couple of small notes.

One question: do you think it might be worth creating a new API-style method like `generateRandomPatient()` to actually do the drug ordering part? That way we could test the order cancellation workflow independently of the drug ordering flow, which makes sense because they are separate workflows (i.e., there's very little real-world use case for ordering and then immediately cancelling a drug order).

gitcliff:
> One question: do you think it might be worth creating a new API-style method like `generateRandomPatient()` to actually do the drug ordering part? That way we could test the order cancellation workflow independently of the drug ordering flow, which makes sense because they are separate workflows (i.e., there's very little real-world use case for ordering and then immediately cancelling a drug order).

yes @ibacher that makes sense

e2e/specs/drug-orders.spec.ts
```diff
@@ -222,7 +143,38 @@ test('Record, edit and discontinue a drug order', async ({ page }) => {
   });
 });
 
+test('Cancel a existing drug order', async ({ page, api }) => {
+  url = process.env.E2E_BASE_URL;
+  await test.step('When I click on the Orders section', async () => {
+    await page.goto(url + `/spa/patient/${patient.uuid}/chart/Orders`);
```
ibacher, line 149: This is actually a good place to use a custom page, so the we don't have to do this here.

e2e/specs/drug-orders.spec.ts
```diff
@@ -1,120 +1,44 @@
...
   patient = await generateRandomPatient(api);
   visit = await startVisit(api, patient.uuid);
+  orderer = await getProvider(api);
+  encounter = await createEncounter(api, patient.uuid, orderer.uuid);
+  drugOrder = await generateRandomDrugOrder(api, patient.uuid, encounter.uuid, orderer.uuid);
 });
 
-test('Record, edit and discontinue a drug order', async ({ page }) => {
-  const medicationsPage = new MedicationsPage(page);
+test('Edit and discontinue a drug order', async ({ page }) => {
```
ibacher, line 35: We do still need a test that covers recording a drug order, even if editing and deleting are done separately.

gitcliff: sure done @ibacher

e2e/specs/drug-orders.spec.ts
```diff
@@ -1,228 +1,285 @@
...
 import { test } from '../core';
-import { MedicationsPage } from '../pages';
+import { OrdersPage } from '../pages';
 
 let patient: Patient;
 let visit: Visit;
+let drugOrder: Order;
+let encounter: Encounter;
+let orderer: Provider;
+let url: String;
```
denniskigen, line 25:
```suggestion
const url = process.env.E2E_BASE_URL;
```

It looks like the value doesn't changes throughout the test execution.

gitcliff:
> Looks like you're reassigning to a const variable somewhere per the e2e test log.

yeah i had skipped that bit, looks fine now

e2e/commands/drug-order-operations.ts
```diff
@@ -0,0 +1,43 @@
...
+  providerUuid: string,
+): Promise<Order> => {
+  const order = await api.post('order', {
+    data: {
+      orderType: '131168f4-15f5-102d-96e4-000c29c2a5d7',
+      type: 'drugorder',
+      action: 'new',
+      drug: '09e58895-e7f0-4649-b7c0-e665c5c08e93',
+      urgency: 'ROUTINE',
+      dateActivated: dayjs().format(),
```
ibacher, line 18: I think rather than just passing in the UUID of the encounter, we should pass in the created encounter and ensure the `dateActivated` is on or after the Encounter datetime.

e2e/specs/drug-orders.spec.ts
```diff
@@ -1,228 +1,283 @@
...
-    await page.getByRole('button', { name: /sign and close/i }).click();
-  });
-
-  await test.step('Then I should see a success notification', async () => {
-    await expect(page.getByText(/discontinued aspirin 81mg/i)).toBeVisible();
-  });
-
-  await test.step('And the medications table should be empty', async () => {
-    await expect(page.getByText(/there are no active medications to display for this patient/i)).toBeVisible();
+test.describe.serial('Drug Order Tests', () => {
```
ibacher, line 35: I think `serial` does exactly what we wanted to avoid. That is, if the "Record a drug test" fails, we should still run the "Edit a drug order". I'm also unsure about grouping the tests like this, as it now makes them rely on a single patient, whereas we should have a patient-per-test.

```suggestion
test.describe('Drug Order Tests', () => {
```

e2e/specs/drug-orders.spec.ts
```diff
@@ -1,228 +1,283 @@
...
+    });
+
+    await test.step('And I search for "Aspirin" in the search bar', async () => {
+      await page.getByRole('searchbox', { name: /search for a drug or orderset/i }).fill('aspirin');
+    });
+
+    await test.step('And I add "Aspirin 81mg" to the basket', async () => {
+      await page
+        .getByRole('listitem')
+        .filter({ hasText: /aspirin 325mg — 325mg — tablet/i })
```
ibacher, line 54: Can't _just_ change the description. The description should match what we're looking for, so if this is meant to be Aspirin 81mg, then we should find that, etc.

e2e/commands/encounter-operations.ts
```diff
@@ -0,0 +1,57 @@
...
+  value: any;
+  obsDatetime: string;
+}
+
+export const createEncounter = async (
+  api: APIRequestContext,
+  patientId: string,
+  providerId: string,
+): Promise<Encounter> => {
+  const encounterRes = await api.post('encounter', {
```
ibacher, line 36: Shouldn't the Encounter here be associated with the visit?

e2e/types/index.ts
ibacher, line 1: I'd move this to the `commands` folder unless we need the types elsewhere. Keeping the folder structure as close to the Playwright standard as possible will help others on-board to these tests and figure things out.

ibacher (changes requested):

e2e/specs/drug-orders.spec.ts
denniskigen: It looks like there was a mismatch between the test assertions and actions and the test description, which is what @gitcliff fixed here.

e2e/types/index.ts
gitcliff: thanks @bacher,, Done

e2e/commands/encounter-operations.ts
```diff
@@ -0,0 +1,60 @@
...
+  patientId: string,
+  providerId: string,
+  visit: Visit,
+): Promise<Encounter> => {
+  const encounterRes = await api.post('encounter', {
+    data: {
+      encounterDatetime: dayjs().format(),
+      form: 'c75f120a-04ec-11e3-8780-2b40bef9a44b',
+      patient: patientId,
+      visit: visit,
```
ibacher, line 43: I think we can just use the visitUuid, but we should make sure that the encounterDatetime is slightly > the visit date time.

gitcliff: thanks @ibacher ,, updated it

ibacher (approved):

PR #2442, opened 2025-05-15
(feat) O3-4702 : Add visit uuid to custom form Url

donaldkibet:
## Requirements

## Summary
This PR allows us to pass `visitUuid` to the forms custom endpoint.  This does not break form endpoints

## Screenshots
[Kapture 2025-05-15 at 14.02.13.webm](https://github.com/user-attachments/assets/c13c9b6c-20b3-4951-8213-2c54a27af613)

### DIFF
--- a/packages/esm-patient-forms-app/src/hooks/use-forms.ts
+++ b/packages/esm-patient-forms-app/src/hooks/use-forms.ts
@@ -23,13 +23,7 @@ function useCustomFormsUrl(patientUuid: string, visitUuid: string) {
   const { customFormsUrl, showHtmlFormEntryForms } = useConfig<ConfigObject>();
   const hasCustomFormsUrl = Boolean(customFormsUrl);
 
-  let baseUrl = hasCustomFormsUrl
-    ? customFormsUrl.indexOf('?') === -1
-      ? `${customFormsUrl}?patientUuid=\${patientUuid}&visitUuid=\${visitUuid}`
-      : customFormsUrl
-    : showHtmlFormEntryForms
-      ? formEncounterUrl
-      : formEncounterUrlPoc;
+  const baseUrl = hasCustomFormsUrl ? customFormsUrl : showHtmlFormEntryForms ? formEncounterUrl : formEncounterUrlPoc;
 
   const url = interpolateUrl(baseUrl, {
     patientUuid: patientUuid,
--- a/packages/esm-patient-forms-app/src/hooks/use-forms.ts
+++ b/packages/esm-patient-forms-app/src/hooks/use-forms.ts
@@ -34,7 +34,6 @@ function useCustomFormsUrl(patientUuid: string, visitUuid: string) {
   return {
     url,
     hasCustomFormsUrl,
-    showHtmlFormEntryForms,
   };
 }
 
--- a/packages/esm-patient-forms-app/src/hooks/use-forms.ts
+++ b/packages/esm-patient-forms-app/src/hooks/use-forms.ts
@@ -19,21 +19,33 @@ import {
 } from '../constants';
 import { isValidOfflineFormEncounter } from '../offline-forms/offline-form-helpers';
 
-export function useFormEncounters(cachedOfflineFormsOnly = false, patientUuid: string = '', visitUuid: string = '') {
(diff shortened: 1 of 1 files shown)


makombe (approved):
Thanks @donaldkibet

packages/esm-patient-forms-app/src/hooks/use-forms.ts
```diff
@@ -19,22 +19,28 @@ import {
...
-  let url = hasCustomFormsUrl
-    ? customFormsUrl.indexOf('?') === -1
-      ? `${customFormsUrl}?patientUuid=${patientUuid}`
-      : customFormsUrl
-    : showHtmlFormEntryForms
-      ? formEncounterUrl
-      : formEncounterUrlPoc;
-  url = interpolateUrl(url, {
+
+  let baseUrl = hasCustomFormsUrl ? customFormsUrl : showHtmlFormEntryForms ? formEncounterUrl : formEncounterUrlPoc;
```
mseaton, line 26: Shouldn't `baseUrl` be a `const` now?

Does anyone have any clue what this whole `formEncounterUrlPoc` thing is, and whether we need to support it still?  What's the deal here?

packages/esm-patient-forms-app/src/hooks/use-forms.ts
```diff
@@ -19,22 +19,28 @@ import {
...
   const { customFormsUrl, showHtmlFormEntryForms } = useConfig<ConfigObject>();
   const hasCustomFormsUrl = Boolean(customFormsUrl);
-  let url = hasCustomFormsUrl
-    ? customFormsUrl.indexOf('?') === -1
-      ? `${customFormsUrl}?patientUuid=${patientUuid}`
-      : customFormsUrl
-    : showHtmlFormEntryForms
-      ? formEncounterUrl
-      : formEncounterUrlPoc;
-  url = interpolateUrl(url, {
```
mseaton, line 32: Defer to others on whether we can remove compatibility for custom urls with no parameters.  I do think it's nicer to remove this, but might need to make this a (breaking) change?

packages/esm-patient-forms-app/src/hooks/use-forms.ts
donaldkibet: If i recall this correctly, `formEncounterUrlPoc` was added to support Ampath forms, The Angular form engine developed by Ampath had a naming convention, where `POC` was prefix to actual form and in cases the form name missed the prefix, it was treated as a reference form, this enabled them to only fetch correct forms for users to work with.

donaldkibet: I think we can retired `formEncounterUrlPoc` and have implementations use custom url e.g `/mymodule/patientforms?patient=${patientUuid}&visit=${visitUuid}&q=poc` cc @Alfred-Mutai @Kipkemoii

packages/esm-patient-forms-app/src/hooks/use-forms.ts
```diff
@@ -19,22 +19,28 @@ import {
...
+  const url = interpolateUrl(baseUrl, {
     patientUuid: patientUuid,
     visitUuid: visitUuid,
     representation: customFormRepresentation,
   });
 
+  return {
+    url,
+    hasCustomFormsUrl,
+    showHtmlFormEntryForms,
```
ibacher, line 37: Why are we returning this?

ibacher (changes requested):
Please respond to the comments before merging.

packages/esm-patient-forms-app/src/hooks/use-forms.ts
donaldkibet: We need **hasCustomFormsUrl** to prevent filtering **customFormUrl results**. If we remove `formEncounterUrlPoc`, we can also eliminate the filter that excludes forms with "component" in their name (AMPATH use case). This would leave just one filter for published forms. I'm removing showHtmlFormEntryForms as it's unused.

ibacher: Sorry, I was talking very specifically about `showHtmlFormEntryForms` since we don't consume it in the other hook.

ibacher (approved):

PR #2459, opened 2025-05-20
(fix) Use correct x-axis scale for biometrics and vitals charts

denniskigen:
## Requirements

## Summary

This PR fixes an issue where the x-axis scale was not being used correctly for the biometrics and vitals charts. Currently, the x-axis scale is set to `ScaleTypes.LABELS`, which spaces each data point evenly regardless of the actual time difference between measurements. This scale type is best for named groups, categories, or arbitrary labels.

For date/time data, `ScaleTypes.TIME` should be used. This scale spaces points according to the actual time difference between measurements—so if two measurements are a day apart, they appear close together; if they're months apart, they appear far apart. This diff switches the x-axis scale to `ScaleTypes.TIME` for the biometrics and vitals charts.

## Screenshots

Source pulse values:

CleanShot 2025-05-20 at 14 22 08@2x

CleanShot 2025-05-20 at 14 23 10@2x

### Before
CleanShot 2025-05-20 at 14 19 55@2x

### After
CleanShot 2025-05-20 at 14 19 05@2x

## Related Issue
https://openmrs.atlassian.net/browse/O3-4714

### DIFF
--- a/packages/esm-patient-vitals-app/src/biometrics/biometrics-chart.component.tsx
+++ b/packages/esm-patient-vitals-app/src/biometrics/biometrics-chart.component.tsx
@@ -8,6 +8,8 @@ import { type ConfigObject } from '../config-schema';
 import { type PatientVitalsAndBiometrics } from '../common';
 import styles from './biometrics-chart.scss';
 
+type BiometricType = 'weight' | 'height' | 'bmi';
+
 interface BiometricsChartProps {
   conceptUnits: Map<string, string>;
   config: ConfigObject;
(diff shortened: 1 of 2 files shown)


ibacher:
I think we need a bigger rethink here than just let's make this a "linear scale". The problem is, while the "trendline" with this change now reflects a sense of time, it's too easy to end up with too many points squashed together in away that makes the actual trends harder to read (imagine if this graph were reversed so that the data points in 2023 were the most recent; what sense am I supposed to make of the values there?

I think that this needs to be paired with the ability to "zoom in" on a specific time period. We would also probably need to specify a default time period to view, rather than trying to plot all datapoints.

denniskigen:
@ibacher might something like [this](https://charts.carbondesignsystem.com/zoombar) help address the concern?

ibacher:
@denniskigen Very much so! Actually probably much better than trying to push this all into configuration.

rbuisson:
I wonder how that works for touch screens, but I trust Carbon has thought about it :)

PR #2462, opened 2025-05-21
(feat) O3-3626: Allergen fields should be read-only when editing existing allergies 

Muppasanipraneeth:
## Requirements

## Summary

This PR prevents duplicate allergen selection in the Allergy Form.

- Added optional `disabled?: boolean` property to the Allergen interface to mark allergens that are already recorded for the patient. In `allergy-form.resource.ts`
-  Updated `useAllergies` hook to correctly retrieve allergies by `patient.id` instead of `patientUuid`.

## Screenshots
https://github.com/user-attachments/assets/5045d6c6-a7c6-464b-a299-10773e2142d3

## Related Issue
https://openmrs.atlassian.net/browse/O3-3626

### DIFF
--- a/packages/esm-patient-allergies-app/src/allergies/allergies-form/allergy-form.scss
+++ b/packages/esm-patient-allergies-app/src/allergies/allergies-form/allergy-form.scss
@@ -111,8 +111,3 @@
   align-items: center;
   min-height: 50%;
 }
-
-.disabledAllergen {
-  background-color: $grey-2;
-  cursor: not-allowed;
-}


ibacher:
I agree with both points. When editing an allergy, it shouldn't be possible to change the allergen, etc., and, as such, it doesn't make sense to warn the user about data they can not change.

@dkayiwa What would you think about adding a validation rule to enforce the above (for an existing allergy, it should not be possible to change the allergen).

dkayiwa:
> What would you think about adding a validation rule to enforce the above (for an existing allergy, it should not be possible to change the allergen).

Just like for Obs, we automatically void an allergy and create a new one whenever one attempts to edit it: https://github.com/openmrs/openmrs-core/blob/2.7.4/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java#L1433

On another note, it also looks like the [AllergyValidator](https://github.com/openmrs/openmrs-core/blob/2.7.4/api/src/main/java/org/openmrs/validator/AllergyValidator.java) is never called at any one time for PatientService.setAllergies() because of the assumptions made here: https://github.com/openmrs/openmrs-core/blob/2.7.4/api/src/main/java/org/openmrs/aop/RequiredDataAdvice.java#L110-L130

denniskigen:
Thanks!

denniskigen (approved):

PR #2468, opened 2025-05-22
fix/03-4454: Clicking on the Order Basket should not do anything when…

jwnasambu:
… the order basket is open

## Requirements

### DIFF
--- a/packages/esm-patient-orders-app/src/order-basket-action-button/order-basket-action-button.extension.tsx
+++ b/packages/esm-patient-orders-app/src/order-basket-action-button/order-basket-action-button.extension.tsx
@@ -1,21 +1,42 @@
 import React, { type ComponentProps } from 'react';
 import { useTranslation } from 'react-i18next';
-import { ActionMenuButton, ShoppingCartIcon } from '@openmrs/esm-framework';
+import { ActionMenuButton, ShoppingCartIcon, useWorkspaceContext } from '@openmrs/esm-framework';
 import { useLaunchWorkspaceRequiringVisit, useOrderBasket } from '@openmrs/esm-patient-common-lib';
 
 const OrderBasketActionButton: React.FC = () => {
   const { t } = useTranslation();
   const { orders } = useOrderBasket();
+  const { activeWorkspace } = useWorkspaceContext();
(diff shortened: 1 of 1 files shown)


packages/esm-patient-orders-app/src/order-basket-action-button/order-basket-action-button.extension.tsx
```diff
@@ -1,21 +1,42 @@
...
+      launchOrderBasket();
+    }
+  };
+
+  const getIcon = (props: ComponentProps<typeof ShoppingCartIcon>) => (
+    <span
+      onClick={(e) => {
+        if (isWorkspaceOpen) {
+          e.stopPropagation();
+        }
```
ibacher, line 25: Do we need this kind of guard here? Isn't it enough to just change the `handleClick()` stuff?

packages/esm-patient-orders-app/src/order-basket-action-button/order-basket-action-button.extension.tsx
```diff
@@ -1,21 +1,42 @@
 import React, { type ComponentProps } from 'react';
 import { useTranslation } from 'react-i18next';
-import { ActionMenuButton, ShoppingCartIcon } from '@openmrs/esm-framework';
+import { ActionMenuButton, ShoppingCartIcon, useWorkspaceContext } from '@openmrs/esm-framework';
 import { useLaunchWorkspaceRequiringVisit, useOrderBasket } from '@openmrs/esm-patient-common-lib';
 
 const OrderBasketActionButton: React.FC = () => {
   const { t } = useTranslation();
   const { orders } = useOrderBasket();
+  const { activeWorkspace } = useWorkspaceContext();
```
ibacher, line 9: I don't think this exists? There is a `useWorkspaces()`, but that may require additional details. I don't think it tells you what workspace is active.

ibacher (commented):
Maybe it is better to make this more general, i.e., by moving this into the `launchWorkspace()` function in core and just make it do nothing if the requested workspace is the same as the current workspace.

The tricky bit here is that the order basket uses multiple workspaces, and we want the button to do nothing (other than maybe collapse the workspace) for all of them.

PR #2495, opened 2025-06-11
(fix) O3-4701: Results Viewer should only display a record-per-obs

mozzy11:
## Requirements

#### Before :
When no test is selected under Individual Tests
image

When no test is selected under Over Time
image

When a test is selected under Individual Tests
image

When a test is selected under Over Time
image

#### After :

When no test is selected under Individual Tests
image

When no test is selected under Over Time
image

When a test is selected under Individual Tests , it auto selectes all the check boxes with a simillar test
image

When a test is selected under Over Time ,it auto selectes all the check boxes with a simillar test
image


ibacher (commented):
Please fill out the PR template and especially pay attention to the stuff about PR names.

mozzy11:
@ibacher but the problem is the left-hand menu directly controls the results displayed on the right.

For example what should hapen ie if a user checks Heamatology tests at the root and un checks the Heamatology tests under Blood work ?? 

do we display the results or not ??

ibacher:
> For example what should hapen ie if a user checks Heamatology tests at the root and un checks the Heamatology tests under Blood work ??

Ideally, if those sets refer to the same concept, then checking or unchecking one box should check or uncheck the other. Basically, we also need to remove the possibility of getting into that kind of inconsistent state.

ibacher:
> But i also dont see any advantage of having duplicated items in the left-hand menu

The idea is that we're letting implementations build up their left menu here. It's not so much that it's a good or bad idea. It's just something we should make sure we support.

mozzy11:
@ibacher , i have worked on the review, the data is now deduplicated while the left hand menu still shows the actual tests structure.
The chack boxes are now synchrozied

ibacher:
We're going to talk about this on the O3 Squad call. Thanks for the work here @mozzy11!

PR #2506, opened 2025-06-16
O3-4780: Individual Test Results Table shows repeated values

jwnasambu:
## Requirements

## Related Issue
https://openmrs.atlassian.net/browse/O3-4780

### DIFF
--- a/packages/esm-patient-tests-app/src/test-results/individual-results-table/individual-results-table.component.tsx
+++ b/packages/esm-patient-tests-app/src/test-results/individual-results-table/individual-results-table.component.tsx
@@ -81,45 +81,52 @@ const IndividualResultsTable: React.FC<IndividualResultsTableProps> = ({ isLoadi
     ];
   }, [t]);
 
-  const tableRows = useMemo(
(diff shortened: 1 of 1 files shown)


ibacher:
> The backend is likely returning the same test results in multiple concept trees, and the frontend is faithfully processing all of them without deduplication

Indeed. This was, IIRC, [part of the design of the backend endpoint](https://talk.openmrs.org/t/o3-test-result-trees-driven-by-concept-sets/35828). The idea, funnily enough, was to _minimize_ the amount of processing the frontend would need to do by ensuring that for each category (concept set) we returned all results in that category. Since the same concept can appear in multiple categories, there are duplicated obs.

Would it make sense to fix the way we return data from the backend? The `obstree` endpoint basically exists to power this view, so we can adjust it in anyway that makes sense. Would it be better if, for example, we returned the obs UUID with the obs results, since then we can completely deduplicate them without issue.

jwnasambu:
Thanks @denniskigen and @ibacher I am on the issue now

PR #2531, opened 2025-06-25
(feat) O3-4228: RDE support for the order basket

amosmachora:
## Requirements

## Summary
This PR adds RDE support to the order basket. When RDE is enabled, the user will now have the option to select a point in time within the bounds of a past visit in which the user can add retrospective data.

## Screenshots
https://github.com/user-attachments/assets/e0b4933d-e37b-4fde-af38-27952e7f98a5

## Related Issue
https://openmrs.atlassian.net/browse/O3-4228

### DIFF
--- a/packages/esm-form-engine-app/src/config-schema.ts
+++ b/packages/esm-form-engine-app/src/config-schema.ts
@@ -1,3 +1,14 @@
-export const configSchema = {};
+import { Type } from '@openmrs/esm-framework';
 
-export interface ConfigObject {}
+export const configSchema = {
+  hideUnansweredQuestionsInReadonlyForms: {
+    _type: Type.Boolean,
+    _description:
+      'Controls whether empty fields are hidden in embedded readonly forms. When true, empty non-transient fields are hidden when forms are displayed in embedded-view mode (e.g., in the Visit dashboard Encounters widget).',
+    _default: false,
+  },
+};
+
+export interface ConfigObject {
+  hideUnansweredQuestionsInReadonlyForms: boolean;
+}
--- a/packages/esm-form-entry-app/package.json
+++ b/packages/esm-form-entry-app/package.json
@@ -51,7 +51,7 @@
     "@ngx-translate/core": "^15.0.0",
     "@openmrs/ngx-file-uploader": "next",
     "@openmrs/ngx-formentry": "next",
-    "jspdf": "^3.0.1",
+    "jspdf": "^3.0.2",
     "moment": "^2.29.4",
     "ngx-bootstrap": "^12.0.0",
     "ngx-webcam": "^0.4.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1932,10 +1932,10 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@babel/runtime@npm:^7.1.2, @babel/runtime@npm:^7.11.2, @babel/runtime@npm:^7.12.5, @babel/runtime@npm:^7.14.5, @babel/runtime@npm:^7.15.4, @babel/runtime@npm:^7.16.3, @babel/runtime@npm:^7.17.2, @babel/runtime@npm:^7.19.0, @babel/runtime@npm:^7.21.0, @babel/runtime@npm:^7.24.5, @babel/runtime@npm:^7.26.7, @babel/runtime@npm:^7.27.3, @babel/runtime@npm:^7.7.2, @babel/runtime@npm:^7.8.4, @babel/runtime@npm:^7.9.2":
-  version: 7.27.6
-  resolution: "@babel/runtime@npm:7.27.6"
-  checksum: 10/cc957a12ba3781241b83d528eb69ddeb86ca5ac43179a825e83aa81263a6b3eb88c57bed8a937cdeacfc3192e07ec24c73acdfea4507d0c0428c8e23d6322bfe
+"@babel/runtime@npm:^7.1.2, @babel/runtime@npm:^7.11.2, @babel/runtime@npm:^7.12.5, @babel/runtime@npm:^7.14.5, @babel/runtime@npm:^7.15.4, @babel/runtime@npm:^7.16.3, @babel/runtime@npm:^7.17.2, @babel/runtime@npm:^7.19.0, @babel/runtime@npm:^7.21.0, @babel/runtime@npm:^7.24.5, @babel/runtime@npm:^7.26.9, @babel/runtime@npm:^7.27.3, @babel/runtime@npm:^7.7.2, @babel/runtime@npm:^7.8.4, @babel/runtime@npm:^7.9.2":
+  version: 7.28.3
+  resolution: "@babel/runtime@npm:7.28.3"
+  checksum: 10/f2415e4dbface7496f6fc561d640b44be203071fb0dfb63fbe338c7d2d2047419cb054ef13d1ebb8fc11e35d2b55aa3045def4b985e8b82aea5d7e58e1133e52
   languageName: node
   linkType: hard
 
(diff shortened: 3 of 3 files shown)


packages/esm-patient-orders-app/src/order-basket/order-basket.workspace.tsx
```diff
@@ -60,20 +64,90 @@ const OrderBasket: React.FC<DefaultPatientWorkspaceProps> = ({
...
+      if (!currentVisit) {
+        return;
+      }
+
+      // check if the date is within the bounds of the current visit
+      const isWithinBounds = isWithinInterval(completeDate, {
+        start: new Date(currentVisit.startDatetime),
+        end: currentVisit.stopDatetime ? new Date(currentVisit.stopDatetime) : new Date(),
+      });
+
```
makombe, line 104: Since this is an RDE, I suggest restricting the date selection directly in the UI by setting appropriate boundaries. This will eliminate the need for an additional validation check elsewhere. Specifically, we can set
`minDate = startDatetime or visitStartDate`

`maxDate = stopDatetime or visitStopDate`

packages/esm-patient-orders-app/src/order-basket/order-basket.workspace.tsx
```diff
@@ -60,20 +64,90 @@ const OrderBasket: React.FC<DefaultPatientWorkspaceProps> = ({
...
+        title: t('rdeDateOutOfBounds', 'Retrospective date is out of bounds'),
+        subtitle: t(
+          'rdeDateOutOfBoundsMessage',
+          `The retrospective date must be within {{startDate}} and {{endDate}}.`,
+          {
+            startDate: format(currentVisit.startDatetime, 'PPP hh:mm a'),
+            endDate: currentVisit.stopDatetime
+              ? format(currentVisit.stopDatetime, 'PPP hh:mm a')
+              : t('currentDate', 'current date'),
+          },
```
makombe, line 134: With setting the maxDate and minDate we will not need this now.

packages/esm-patient-orders-app/src/order-basket/order-basket.workspace.tsx
amosmachora: I have added a restriction from the date picker itself. The issue is that sometimes, as much as a user will click on a certain date, the time itself might be out of range. I am not sure the time picker has a way to limit the times natively like the date picker, so that check is for making sure that the user doesn't go out of the visit's bounds.

packages/esm-patient-orders-app/src/order-basket/order-basket.workspace.tsx
```diff
@@ -60,20 +66,90 @@ const OrderBasket: React.FC<DefaultPatientWorkspaceProps> = ({
...
+    (dateTime: { retrospectiveDate: Date; retrospectiveTime: string; retrospectiveTimeFormat: string }) => {
+      if (!dateTime.retrospectiveDate || !dateTime.retrospectiveTime || !dateTime.retrospectiveTimeFormat) {
+        setRdeDate(undefined);
+        return;
+      }
+
+      let [rawHour, minute] = dateTime.retrospectiveTime.split(':').map(Number);
+
+      // Adjust hour for AM/PM
+      let hour = rawHour;
```
chibongho, line 79: use `convertTime12to24` instead of reimplementing its logic

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/retrospective-data-date-time-picker/retrospective-date-time-picker.component.tsx
```diff
@@ -55,7 +54,30 @@ const RetrospectiveDateTimePicker = ({
...
     }
-  }, [onChange, retrospectiveDate, retrospectiveTime, retrospectiveTimeFormat]);
+
+    // if the user passes a time that is not the selected format, we reset the time format
+    if (retrospectiveTime && retrospectiveTimeFormat) {
+      const [hours, minutes] = retrospectiveTime.split(':');
+      const isAm = retrospectiveTimeFormat === 'AM';
+      const isPm = retrospectiveTimeFormat === 'PM';
+
+      if (isAm && parseInt(hours, 10) >= 12) {
```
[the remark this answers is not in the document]
amosmachora: > The <TimePicker> below has a regex pattern passed to it. Does it correctly prevent dates that are not in the 12-hour format?

I referred to other places the `<TimePicker />` component was used , they all seemed to have that pattern to prevent users from entering 24 hour format times but in my UI it seems to be still possible, for other places the time picker is used I haven't verified that. I have added logic to make sure users don't enter invalid times. Thanks for catching that!

packages/esm-patient-orders-app/src/order-basket/order-basket.workspace.tsx
amosmachora: Done!. Simplifies things a lot. I did not know that function existed. Thanks.

packages/esm-patient-orders-app/src/order-basket/order-basket.workspace.tsx
```diff
@@ -46,6 +52,8 @@ const OrderBasket: React.FC<DefaultPatientWorkspaceProps> = ({
   } = useOrderEncounter(patientUuid);
   const [isSavingOrders, setIsSavingOrders] = useState(false);
   const [creatingEncounterError, setCreatingEncounterError] = useState('');
+  const [hasRdeDateBoundsError, setHasRdeDateBoundsError] = useState(false);
+  const [rdeDate, setRdeDate] = useState<Date>();
```
chibongho, line 56: Right now, when we select a visit, we have to manually put in the date. Maybe we can default it to the visit start time?

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/retrospective-data-date-time-picker/retrospective-date-time-picker.component.tsx
```diff
@@ -104,7 +122,24 @@ const RetrospectiveDateTimePicker = ({
                 <TimePicker
                   id={'retrospective-time-picker-input'}
                   labelText={t('time', 'Time')}
-                  onBlur={onBlur}
+                  onBlur={(event) => {
+                    const timeValue = event.target.value;
+                    if (timeValue) {
+                      const pattern = /^(0[1-9]|1[0-2]):([0-5][0-9])$/;
+                      if (!pattern.test(timeValue)) {
+                        showSnackbar({
```
chibongho, line 130: I know we're not consistent about this, but try to show the error message inline with the form element (see screenshot). Avoid snackbars and notifications because:
- They are sometimes shown further away from the input element
- They do not self-dismiss when the user corrects the input value. Also, If the user manually dismiss them, they no longer can tell which element has a problem (especially if they get distracted and then come back to the form).

image

chibongho (approved):
I'm approving it for now, since it's behind a feature flag, and it kind of works. That said, I have some comments about functional and UX issues.

============================
This one is not your fault, but if we set the encounter date in the order basket, that open a Drug order form, and then go back to the order basket, the date picker gets reset. For this to not happen, we need to properly support child workspaces.


Also, the date / time picker values get reset if I uncheck and re-check the "Enable" checkbox. It would be nice it the value stayed across toggling the checkbox.


=======================
image
We should keep the styling for "Encounter time" consistent with other workspace headers. Here's how the Visit note form looks:
image

Regarding the header, while "encounter" is technically a user-facing term, I think we can make it clearer by saying "Order time" or similar. (We might need to adjust `retrospective-date-time-picker.component.tsx` to take in the header.

For the checkbox, "Enable" is confusing to the user. I suggest changing the wording to "In the past" or similar.

===================
image
- Use inline error messages (keep them close to the input element). . 
- "Retrospective date" is not a user-facing term. Call it "Order date" or something similar.

jwnasambu:
@chibongho, @makombe  Kindly are there any other changes needed for this PR to be merged?

packages/esm-patient-orders-app/src/order-basket/order-basket.workspace.tsx
amosmachora: Not a bad default. Let's go with that.

amosmachora:
Hey @chibongho I have made most of the UI improvements you requested earlier except the out of bounds error UI as that will require changing a lot of the design. Will make that change in a follow up PR.

cc @makombe .

makombe (approved):
Thanks @amosmachora LGTM. Will merge this for testing

makombe:
@amosmachora please  resolve conflict. Thanks

amosmachora:
@makombe, the conflict has been fixed. Build fails for another reason I dont really fully understand.

PR #2241, opened 2025-02-12
(feat) O3-4444: Map 'Start date' in drug order form to `dateActivated`

denniskigen:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspiration.
- [ ] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://om.rs/o3ui).
- [x] My work includes tests or is validated by existing tests.

## Summary

This PR makes the following changes to the medications app:

- Maps the `startDate` field to the `dateActivated` property in the drug order form. Previously, the payload sent to the Order resource in the backend when creating a new drug order or editing an existing one was missing the `dateActivated` property. This meant that the backend wound up setting the `dateActivated` property to the current date and time. This is especially problematic when editing an existing order, as the current date was being used as the `dateActivated` property. This PR fixes this issue by mapping the `startDate` field to the `dateActivated` property and having its value default to the current date and time for new orders, and using the existing value for existing orders.
- Amends the "Start date" field in the drug order form to read "Activation date" instead of "Start date".
- Amends the "Start date" column header in the medications details table to read "Activation date" instead of "Start date".

PR #2552, opened 2025-06-30
(fix) O3-4433: Allow future date selection in the drug order form's start date field

ssebayigga-sharif:
## Requirements

## Summary
  The clinician was not able to select a future date , I was able to solve that problem and now the clinician can select any future date of his or her choosing. 

## Screenshots

Before making the changes and the future date was not selectable.
image
After making the changes and the future date  became selectable.
image
   

## Related Issue
https://openmrs.atlassian.net/jira/software/c/projects/O3/issues/O3-4433 

## Other
 most has been covered


ssebayigga-sharif:
Screenshot (6)

ssebayigga-sharif:
Screenshot (9)

ibacher:
Thanks @ssebayigga-sharif! Have we confirmed that we're able to support orders entered this way? I think the backend will require that future dates be entered in to the `dateScheduled` field rather than the `dateActivated` field, which is probably what we are using.

ssebayigga-sharif:
Thanks @ibacher did I miss out anything sir

denniskigen:
The other important context to be aware of that makes this proposed solution a non-starter is the comment on line 601: `TODO: This input does nothing`. Note how in the [prepMedicationOrderPostData](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-medications-app/src/api/api.ts#L124-L125) function, neither the  `dateActivated` nor `scheduledDate` properties get set from `order.startDate`. The Order Entry API definitely supports these two fields, but the `prepMedicationOrderPostData` function doesn't map `startDate` to either of them. 

So that should be the first port of call in terms of fixing this issue. FWIW, I tried fixing this a while back in https://github.com/openmrs/openmrs-esm-patient-chart/pull/2241 but got scuppered by some backend validation errors. Might be worth having a look at the approach.

PR #2559, opened 2025-07-04
(fix) Restructure encounter tile with grid layout and add visit validation config

NethmiRodrigo:
## Requirements

## Summary

This PR fixes a couple of bugs related to the encounter tile:
1. Updates the tile to no longer group columns by their encounter type. This caused a weird layout with the grouped columns being closer to each other. 
2. Updates the encounter tile to render in a Grid using the Carbon `Grid` component.
3. Updates the formatDate function to not include time as it would previously result in wrong dates.
4. When parsing the values to the encounter tiles, the parameter `isDate` passed to the function `getObsFromEncounter` is now set to false to get the proper date string. Previously it was the formatted string (e.g `17 -- 07 -- 2025`) that would get passed when `isDate` is true, since this isn't a valid date string, it would result in returning `NaN` when calculating date differences.
5. Prompts the start a visit dialog if there is no active visit when trying to launch a form from the encounter list.

## Screenshots

Before:
Screenshot 2025-07-08 at 1 59 12 PM

After:
Screenshot 2025-07-08 at 1 58 06 PM


packages/esm-patient-chart-app/src/clinical-views/encounter-tile/encounter-tile.component.tsx
```diff
@@ -1,84 +1,79 @@
...
-                </span>
-              </>
-            )}
-          </div>
-        );
-      })}
-    </div>
+    <>
+      <span className={styles.tileTitle}>{t(column.header)}</span>
+      {!(obsValue === '--' && summaryValue !== '--' && summaryValue !== null) && (
```
ibacher, line 66: Maybe?

```suggestion
      {!(obsValue === '--' && summaryValue !== '--' && isNil(summaryValue)) && (
```

`isNil(x)` is basically `typeof x === 'undefined' || x === null`

packages/esm-patient-chart-app/src/clinical-views/encounter-tile/tile.scss
```diff
@@ -18,17 +18,11 @@
...
-  flex-direction: row;
-  gap: layout.$spacing-05;
-  justify-content: space-around;
-}
-
-.tileSubTitle {
-  @include type.type-style('body-compact-01');
+  font-size: var(--cds-label-01-font-size, 0.75rem);
+  line-height: var(--cds-label-01-line-height, 1.33333);
+  letter-spacing: var(--cds-label-01-letter-spacing, 0.32px);
```
ibacher, line 23: I think this is equivalent right?

```suggestion
  @include type.$label-01;
```

packages/esm-patient-chart-app/src/clinical-views/utils/helpers.ts
```diff
@@ -20,21 +21,24 @@ export function launchEncounterForm(
...
-    formInfo: {
-      encounterUuid,
-      formUuid: form?.uuid,
-      patientUuid: patientUuid,
-      visit: visit,
-      additionalProps: {
-        mode: action === 'add' ? 'enter' : action,
-        formSessionIntent: intent,
-        openClinicalFormsWorkspaceOnFormClose: false,
+  if (!visit) {
```
ibacher, line 24: Could we put this behind a configuration option so OHRI can continue to use these without visits? (The configuration option should default to using visits).

packages/esm-patient-chart-app/src/clinical-views/utils/index.ts
```diff
@@ -3,7 +3,7 @@ import type { Encounter, ColumnDefinition, ConfigConcepts, EncounterTileColumn,
 import dayjs from 'dayjs';
 
 const calculateDateDifferenceInDate = (givenDate: string): string => {
-  return `${dayjs().diff(dayjs(givenDate), 'days')} days`;
+  return `${Math.abs(dayjs().diff(dayjs(givenDate), 'days'))} days`;
```
ibacher, line 6: While we have a helper for `age()` that wouldn't quite work here, I don't love this hard-coding of `days`. However, I think we can at least leverage [Intl.DurationFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat) like:

```ts
return new Intl.DurationFormat(getLocale()).format({ days: Math.abs(dayjs().diff(dayjs(givenDate), 'days')) });
```

That should get us it formatted and localized correctly.

ibacher (commented):
Thanks @NethmiRodrigo! A couple of persnickety comments, but generally LGTM.

packages/esm-patient-chart-app/src/clinical-views/utils/index.ts
NethmiRodrigo: @ibacher I get a type error saying `DurationFormat` does not exist, do we need to update something?

packages/esm-patient-chart-app/src/clinical-views/encounter-tile/tile.scss
NethmiRodrigo: Yep! Minus the font color

packages/esm-patient-chart-app/src/clinical-views/utils/index.ts
ibacher: Oh... I see. Well, ok. What happens if we call `age()` here instead?

denniskigen:
Thanks, Nethmi!

PR #2577, opened 2025-07-15
(fix) Filter out undefined encounters in EncountersTable

denniskigen:
## Requirements

## Summary

This fix ensures `paginatedMappedEncounters` only contains valid encounter objects by filtering out any undefined or null values after mapping. This prevents 'Cannot read properties of undefined (reading "id")' errors in the UI and E2E tests when the encounters array is empty or contains invalid data. This is a defensive and canonical fix that improves robustness without affecting valid data rendering.


denniskigen:
@gitcliff it looks like the test failures in the distro refapp might be owing to this issue.

gitcliff (commented):
thanks @denniskigen LGTM ..may be just for are interested in only filtering `null` and `undefined` values and keep other falsy values like 0 or '' ?

PR #2584, opened 2025-07-17
(feat): Add support to render visit details in encounter list

NethmiRodrigo:
## Requirements

## Summary

This PR does the following:
1. Adds support to configure encounter lists to display visit details like the visit date, if there is a visit date attached to the encounter, if not it will render the default of `--`. 
2. Fixes a bug in an if condition where previously, the helper function would return a `--` if there was no `obsConcept` which shouldn't be the case as there would be no `obsConcept` configured for columns that intend to display encounter/visit details.
3. Previously the type `visitType` actually displayed the `encounterType`. This was adjusted so that the type `visitType` actually renders the `visitType` by adding support for it.

## Screenshots

Before:
Screenshot 2025-07-17 at 2 42 07 PM

After (support for visit date):

Screenshot 2025-07-17 at 2 27 12 PM

Support for visit type:
Screenshot 2025-07-17 at 2 30 04 PM


denniskigen (approved):
LGTM. Thanks!

PR #2588, opened 2025-07-18
(feat): Add support to render encounterDate in the encounter tiles

NethmiRodrigo:
## Requirements

## Summary

This PR adds support to render the `encounterDatetime` in an encounter tile

## Screenshots

Screenshot 2025-07-18 at 1 05 33 PM

### DIFF
--- a/packages/esm-patient-chart-app/src/clinical-views/types.ts
+++ b/packages/esm-patient-chart-app/src/clinical-views/types.ts
@@ -7,7 +7,7 @@ export type TableHeaderType = {
 };
 
 export interface Encounter extends OpenmrsResource {
-  encounterDatetime: Date;
+  encounterDatetime: string;
   encounterType: { uuid: string; name: string };
   patient: {
     uuid: string;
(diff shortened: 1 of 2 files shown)


packages/esm-patient-chart-app/src/clinical-views/types.ts
```diff
@@ -7,7 +7,7 @@ export type TableHeaderType = {
 };
 
 export interface Encounter extends OpenmrsResource {
-  encounterDatetime: Date;
+  encounterDatetime: string;
```
NethmiRodrigo, line 10: Because this gets actually set to an ISO string and not a Date object

denniskigen: Should we extend the `EncounterPropertyType` enum to include `encounterDatetime'?

NethmiRodrigo: Yeah, thank you for catching that!

denniskigen (approved):
Thanks!

PR #2607, opened 2025-07-25
(fix): Encounter tile should be able to render lab results

NethmiRodrigo:
## Requirements

## Summary

Previously, if the value of an observation wasn't a string, it was assumed that it would always be an object. This meant that lab test results would not be rendered. This PR fixes it by checking if the value is a string or a number.

## Screenshots

Before:
Screenshot 2025-07-25 at 6 34 45 PM

After:
Screenshot 2025-07-25 at 6 34 25 PM

### DIFF
--- a/packages/esm-patient-chart-app/src/clinical-views/utils/index.ts
+++ b/packages/esm-patient-chart-app/src/clinical-views/utils/index.ts
@@ -35,7 +35,7 @@ export const getEncounterTileColumns = (tileDefinition: MenuCardProps, config: C
           config: config,
         });
       }
-      return typeof obsValue === 'string' || typeof obsValue === 'number' ? obsValue : obsValue?.name?.name ?? '--';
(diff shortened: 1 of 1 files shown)


packages/esm-patient-chart-app/src/clinical-views/utils/index.ts
```diff
@@ -35,7 +35,7 @@ export const getEncounterTileColumns = (tileDefinition: MenuCardProps, config: C
           config: config,
         });
       }
-      return typeof obsValue === 'string' ? obsValue : obsValue?.name?.name ?? '--';
+      return typeof obsValue === 'string' || typeof obsValue === 'number' ? obsValue : obsValue?.name?.name ?? '--';
```
denniskigen, line 38:
```suggestion
      return typeof obsValue === 'string' || (typeof obsValue === 'number' && !isNaN(obsValue)) 
        ? obsValue 
        : obsValue?.name?.name ?? '--';
```

Would this edit work a touch better?

denniskigen (approved):
LGTM

PR #2620, opened 2025-07-28
(fix) Pickup visit from the form-entry workspace props

samuelmale:
## Requirements

## Summary

Addresses an issue where encounters aren't associated with the current visit. This was happening because the form-engine-app was not supporting the `visitUuid` prop [exposed by the form-entry workspace](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-forms-app/src/forms/form-entry.workspace.tsx#L87).


packages/esm-form-engine-app/src/form-renderer/form-renderer.component.tsx
```diff
@@ -38,6 +40,15 @@ const FormRenderer: React.FC<FormRendererProps> = ({
     !encounterUuid && openClinicalFormsWorkspaceOnFormClose && launchWorkspace(clinicalFormsWorkspaceName);
   }, [closeWorkspace, encounterUuid, openClinicalFormsWorkspaceOnFormClose, clinicalFormsWorkspaceName]);
 
+  const visit = useMemo(() => {
+    if (visitRaw) {
+      return visitRaw;
+    }
+    if (visitUuid) {
+      return { uuid: visitUuid } as Visit;
```
NethmiRodrigo, line 48: @samuelmale @ibacher I think this is the reason the e2e test is failing, this makes the visit object have only the uuid property and it doesn't have the startDatetime

samuelmale: It is less likely to be an issue because the FE doesn't infer the encounter's date from the Visit object at any point.

samuelmale: @ibacher @NethmiRodrigo The issue is actually with the test itself! So the form submission fails because of a validation error of the encounter datetime which apparently is [before the visit start date](https://github.com/openmrs/openmrs-core/blob/4dc3ff70009aedda7b1ed642227b0ce12e05b506/api/src/main/java/org/openmrs/validator/EncounterValidator.java#L95). How does that happen? So we apply Playwright's fake timers [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/e2e/specs/clinical-forms.spec.ts#L114) when we attempt to travel in time. This rewinds the clock back to Epoch, hence ending up with an encounter date that looks like: `1970-01-01T00:01:01.325Z` yet the visit start date isn't relative to the fake timers because it was created during the setup phase.

samuelmale:
Okay, so I've pushed a commit that puts the failing test in a more useful state. So the test is failing because the RFE attempts to set the encounter's date instead of letting the server handle that. We need to disable that for the RFE, just like it was [disabled in the AFE](https://github.com/openmrs/openmrs-ngx-formentry/pull/131).

samuelmale:
> We need to disable that for the RFE, just like it was https://github.com/openmrs/openmrs-ngx-formentry/pull/131.

Wired PR: https://github.com/openmrs/openmrs-esm-form-engine-lib/pull/601

PR #2657, opened 2025-08-07
(Fix) Program  Completion Date should not allow selection of a date Equal or less than Enrollment date

mozzy11:
## Requirements

### DIFF
--- a/packages/esm-patient-programs-app/src/programs/programs-form.workspace.tsx
+++ b/packages/esm-patient-programs-app/src/programs/programs-form.workspace.tsx
@@ -225,8 +225,17 @@ const ProgramsForm: React.FC<ProgramsFormProps> = ({
           {...field}
           id="completionDate"
           data-testid="completionDate"
-          minDate={new Date(watch('enrollmentDate'))}
+          minDate={(() => {
+            const date = new Date(watch('enrollmentDate'));
+            date.setDate(date.getDate() + 1);
+            return date;
+          })()}
           maxDate={new Date()}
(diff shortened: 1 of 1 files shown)


packages/esm-patient-programs-app/src/programs/programs-form.workspace.tsx
```diff
@@ -225,8 +233,17 @@ const ProgramsForm: React.FC<ProgramsFormProps> = ({
...
+          minDate={(() => {
+            const date = new Date(watch('enrollmentDate'));
+            date.setDate(date.getDate() + 1);
+            return date;
+          })()}
           maxDate={new Date()}
+          isDisabled={(() => {
+            const d = new Date(watch('enrollmentDate'));
+            const today = new Date();
+            return d.setHours(0, 0, 0, 0) === today.setHours(0, 0, 0, 0);
```
ibacher, line 245: For this, it's probably actually better to use `dayjs` (or, alternatively, I guess `setUTCTime()`, but I always prefer libraries for date manipulation; this implementation can run into some weird edge-cases around DST shifts.

ibacher (changes requested):

packages/esm-patient-programs-app/src/programs/programs-form.workspace.tsx
mozzy11: Thanks @ibacher . Switched to using dayjs

ibacher (approved):

PR #2554, opened 2025-07-01
(feat) Optimize visit data revalidation to avoid SWR cache cascades

denniskigen:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspiration.
- [ ] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://om.rs/o3ui).
- [x] My work includes tests or is validated by existing tests.

## Summary

This PR addresses performance bottlenecks in visit and encounter data management by replacing global revalidation with more targeted cache invalidation.

Presently, the global `mutateVisit()` revalidation function triggers 13+ unnecessary network requests per operation, and all visit-related components are revalidated even when only specific data changed.

This approach:

- Replaces the global `mutateVisit()` with targeted SWR cache invalidation to reduce 13+ revalidation requests to 1 per operation.
- Moves revalidation utilities from the chart app to `@openmrs/esm-patient-common-lib` for better reusability.
- Creates multiple utility functions for surgical cache invalidation:
  - `invalidateVisitHistory()` - Only invalidates visit history table data
  - `invalidatePatientEncounters()` - Only invalidates encounter data
  - `invalidateVisitAndEncounterData()` - Combines both for operations affecting both
  - `useOptimisticVisitMutations()` - Hook for optimistic updates without network requests
- Fixes the encounter table cache invalidation regex to handle flexible URL parameter order.
- Updates all visit/encounter operations to use surgical cache targeting:
  - Visit creation/editing (`visit-form.workspace.tsx`)
  - Visit deletion/restoration (`useDeleteVisit.hook.tsx`)
  - Encounter deletion (`encounters-table.component.tsx`)
  - Form submissions (`form-entry.workspace.tsx`)
  - Visit note creation (`visit-notes-form.workspace.tsx`)
  - Order basket operations (`order-basket.workspace.tsx`)
  - Vitals/biometrics forms (`vitals-biometrics-form.workspace.tsx`)
  - End visit dialog (`end-visit-dialog.component.tsx`)
- Preserves current visit component performance by avoiding unnecessary revalidation.
- Ensures the `AllEncountersTable` updates correctly after form submissions and other encounter operations.

[...]

PR #2669, opened 2025-08-13
(BREAKING) O3-4962 store visitContext in patient chart store

chibongho:
## Requirements

## Summary

__This PR:__
- Adds the `visitContext` (which I renamed from `currenVisit`) into the patient chart store
- Updates places that uses the `currenVisit` from the `useVisit()` hook to use the store's `visitContext` instead.
- Removes the useVisitOrOfflineVisit hook. The patient chart store is now the source of truth for `visitContext`. On creating a visit with the visit form workspace, regardless of whether the visit is an offline visit, the visit is committed to the patient chart store as the new `visitContext`. 
- Updates all workspaces and extensions to **NOT** use the patient chart store as much as possible for the `patient` and `visitContext`, but instead get them via props. This is because we generally have no control over how these components are used outside of the patient chart, where using the store does not make sense. (For example: the start visit form is accessible in the service queues app through the patient search result. The clinical forms workspaces are accessible in the ward app.) This also makes the existing workspaces closer to how the new workspace system is supposed to work.
- As a safety feature, the `usePatientChartStore()` hook now requires the `patientUuid` as a param. The actual values in the store are only returned when the store is in fact storing data for the specified patient. Note that this is not a deviation from existing behavior when relying on `useVisit()` to gee the `currenctVisit` (which is null if the visit store is storing visit of another patient). 

__Next steps after this PR:__
- As we no longer rely on `currenvVisit` from the `useVisit` hook, we should make a breaking change to the `useVisit` hook in `esm-core` to remove it, along with the visit store. Thus, `useVisit(patientUuid)` should only return the (latest) active visit of the patient.
- We should be persist the `visitContext` across refreshes. To do so, we can make a change to `createGlobalStore` in `esm-core` to allow for creating stores backed by localStorage or sessionStorage.
- The `useLaunchWorkspaceRequiringVisit` should go away as we migrate to the new workspace system, where the `visitContext` should be part of the workspace group's prop.

__Longer term changes:__
- We are still using `useVisit()` in quite a few places in the patient chart to get the active visit. The causes the same performance issue as calling `usePatient()` in multiple places (which was fixed in the great props driling PR). We should be able to store the active visit in the patient chart store as well.


packages/esm-patient-common-lib/src/store/patient-chart-store.ts
```diff
@@ -1,40 +1,38 @@
-import { createGlobalStore, useStore } from '@openmrs/esm-framework';
+import { type Actions, createGlobalStore, useStoreWithActions, type Visit } from '@openmrs/esm-framework';
 
 export interface PatientChartStore {
-  patientUuid?: string;
-  patient?: fhir.Patient;
+  patientUuid: string;
+  patient: fhir.Patient;
+  visitContext: Visit;
+  mutateVisitContext: () => void;
```
samuelmale, line 7: The visit object and its mutator seem to be conceptually linked; in that case, do you think it makes sense to:

```ts

patient: fhir.Patient;
visitContext: {
    visit: Visit;
    mutate: () => void;
}
```

chibongho: I don't have a strong preference either way, even anticipating that `activeVisit` and `mutateActiveVisit` might be added to the store in the future. I do worry slightly that people would do this:

```
const {visitContext: {visit: visitContext, mutate: mutateVisitContext}} = state;
```

packages/esm-patient-chart-app/src/clinical-views/encounter-list/encounter-list-tabs.extension.tsx
```diff
@@ -1,22 +1,26 @@
...
 import { useTranslation } from 'react-i18next';
 import { Tabs, Tab, TabList, TabPanels, TabPanel } from '@carbon/react';
-import { useConfig, useVisit } from '@openmrs/esm-framework';
+import { useConfig } from '@openmrs/esm-framework';
 import { EncounterList } from './encounter-list.component';
 import { getMenuItemTabsConfiguration } from '../utils/encounter-list-config-builder';
 import styles from './encounter-list-tabs.scss';
 import { filter } from '../utils/helpers';
 import { type Encounter } from '../types';
+import { usePatientChartStore } from '@openmrs/esm-patient-common-lib/src';
```
denniskigen:
```suggestion
import { usePatientChartStore } from '@openmrs/esm-patient-common-lib';
```

For this and all other occurrences

packages/esm-patient-chart-app/src/patient-banner-tags/visit-attribute-tags.extension.tsx
```diff
@@ -26,13 +26,21 @@ const getAttributeValue = (attributeType, value) => {
   }
 };
 
+/**
+ * This extension slots to the patient-banner-tags-slot by default.
+ * TODO: This component is used outside the patient chart, but incorrectly
```
denniskigen: What should we do instead in those instances? Would we rely on injecting the visit by passing it down via props/state from the parent banner context and optionally fall back to fetching visit data by calling `useVisit` with the supplied `patientUuid` if it's not provided?

packages/esm-patient-chart-app/src/visit/hooks/useDeleteVisit.test.tsx
```diff
@@ -56,14 +58,13 @@ describe('useDeleteVisit', () => {
   beforeEach(() => {
     jest.clearAllMocks();
```
denniskigen: Unrelated to this diff, but we're already clearing all mocks in the top-level jest config, so this call and the one on line 141 are redundant.

packages/esm-patient-chart-app/src/patient-chart/patient-chart.resources.ts
```diff
@@ -0,0 +1,20 @@
...
+  'form:(uuid,name),location:ref,' +
+  'encounterType:ref,' +
+  'encounterProviders:(uuid,display,' +
+  'provider:(uuid,display))),' +
+  'patient:(uuid,display),' +
+  'visitType:(uuid,name,display),' +
+  'attributes:(uuid,display,attributeType:(name,datatypeClassname,uuid),value),' +
+  'location:(uuid,name,display))';
+
+export function useVisitByUuId(visitUuid: string, representation: string = defaultVisitCustomRepresentation) {
```
denniskigen:
```suggestion
export function useVisitByUuid(visitUuid: string | null, representation: string = defaultVisitCustomRepresentation) {
```

packages/esm-patient-chart-app/src/patient-chart/patient-chart.resources.ts
```diff
@@ -0,0 +1,20 @@
...
+  const url = `${restBaseUrl}/visit/${visitUuid}?v=${representation}`;
+
+  return useSWR<{ data: Visit }>(visitUuid ? url : null, openmrsFetch);
```
denniskigen:
```suggestion
  const { data, error, isLoading, isValidating, mutate } = useSWR<{ data: Visit }>(key, openmrsFetch);

  return {
    visit: data?.data,
    error,
    isLoading,
    isValidating,
    mutate,
  };
```

Nitpicky, but it's more canonical to expose destructured, domain-specific fields instead of handing that responsibility to callers.

packages/esm-patient-chart-app/src/patient-chart/patient-chart.component.tsx
```diff
@@ -16,31 +16,85 @@ import ChartReview from '../patient-chart/chart-review/chart-review.component';
...
-  }, [patientUuid]);
+    if (!isValidatingVisitContext && !isValidatingActiveVisit && storePatientUuid) {
+      if (activeVisit) {
+        setVisitContext(activeVisit, mutateActiveVisit);
+      } else if (newVisitContext) {
+        setVisitContext(newVisitContext.data, newMutateVisitContext);
+      } else {
+        setVisitContext(null, null);
+      }
+    }
```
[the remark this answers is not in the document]
chibongho: good idea.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.workspace.tsx
```diff
@@ -299,12 +302,15 @@ const VisitForm: React.FC<VisitFormProps> = ({
             // 1. Current visit data (for critical components like visit summary, action buttons)
             // 2. Visit history table (for the paginated visit list)
 
-            // Update current visit data for critical components (useVisit hook)
-            mutateCurrentVisit();
+            // Update patient's visit data for critical components
+            const mutateSavedOrUpdatedVisit = () => invalidateVisitByUuid(globalMutate, visit.uuid);
+            mutateActiveVisit();
+            setVisitContext?.(visit, mutateSavedOrUpdatedVisit);
```
denniskigen:
```suggestion
            if (visit?.patient?.uuid === patientUuid) {
              setVisitContext?.(visit, mutateSavedOrUpdatedVisit);
            }
```

Might we want to add this guard against stale context updates?

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.workspace.tsx
```diff
@@ -343,9 +349,12 @@ const VisitForm: React.FC<VisitFormProps> = ({
           config.offlineVisitTypeUuid,
           payload.startDatetime,
         ).then(
-          () => {
+          (visit) => {
             // Use same targeted approach for offline visits for consistency
-            mutateCurrentVisit();
+            const mutateSavedOrUpdatedVisit = () => invalidateVisitByUuid(globalMutate, visit.uuid);
+            mutateActiveVisit();
+            setVisitContext?.(visit, mutateSavedOrUpdatedVisit);
```
denniskigen:
```suggestion
            if (visit?.patient?.uuid === patientUuid) {
              setVisitContext?.(visit, mutateSavedOrUpdatedVisit);
            }
```

Ditto here.

packages/esm-patient-chart-app/src/visit/hooks/useDeleteVisit.tsx
```diff
@@ -47,20 +52,23 @@ export function useDeleteVisit(visit: Visit, onVisitDelete = () => {}, onVisitRe
...
     setIsDeletingVisit(true);
 
-    deleteVisit(visit?.uuid)
+    deleteVisit(activeVisit?.uuid)
       .then(() => {
-        // Update current visit data for critical components (useVisit hook)
-        mutateCurrentVisit();
+        // Update active visit data
+        mutateActiveVisit();
+        if (activeVisit.uuid == visitContext?.uuid) {
```
denniskigen:
```suggestion
        if (activeVisit.uuid == visitContext?.uuid && activeVisit.patient?.uuid === patientUuid) {
```

Might an additional patient UUID guard make this more robust?

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/retrospective-data-date-time-picker/retrospective-date-time-picker.component.tsx
```diff
@@ -14,22 +13,22 @@ type FormValues = {
 
 type RetrospectiveDateTimePickerProps = {
   patientUuid: string;
```
denniskigen:
```suggestion
```

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-header.extension.tsx
```diff
@@ -31,26 +40,19 @@ const VisitContextHeader: React.FC<VisitContextHeaderProps> = ({ patientUuid })
...
-      </div>
-    );
-  }
   return (
     <div
       className={classNames(styles.visitContextHeader, isActiveVisit ? styles.activeVisit : styles.retroactiveVisit)}
     >
       <div className={styles.addingTo}>{t('addingToVisit', 'Adding to:')}</div>
-      <div className={styles.visitType}>{currentVisit.visitType.display}</div>
+      <div className={styles.visitType}>{visitContext.visitType.display}</div>
```
denniskigen:
```suggestion
      <div className={styles.visitType}>{visitContext?.visitType?.display}</div>
```

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-header.test.tsx
```diff
@@ -1,33 +1,21 @@
...
 import React from 'react';
-import VisitContextHeader from './visit-context-header.component';
+import VisitContextHeader from './visit-context-header.extension';
+import { mockPatient } from 'tools';
 
 const mockUseSystemVisitSetting = jest.fn(useSystemVisitSetting).mockReturnValue({
   systemVisitEnabled: true,
   errorFetchingSystemVisitSetting: null,
   isLoadingSystemVisitSetting: false,
 });
```
denniskigen:
```suggestion

const mockUseSystemVisitSetting = jest.mocked(useSystemVisitSetting);
```

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-header.test.tsx
```diff
@@ -1,33 +1,21 @@
...
```
denniskigen:
```suggestion
```

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-switcher.test.tsx
```diff
@@ -40,7 +32,20 @@ jest.mock('../visit.resource', () => ({
   useInfiniteVisits: () => mockUseInfiniteVisits('some-uuid'),
 }));
 
+const mockSetVisitContext = jest.fn();
+jest.mock('@openmrs/esm-patient-common-lib', () => ({
+  ...jest.requireActual('@openmrs/esm-patient-common-lib'),
+  usePatientChartStore: jest.fn(() => ({
```
denniskigen:
```suggestion
  ...jest.requireActual('@openmrs/esm-patient-common-lib'),
  useSystemVisitSetting: jest.fn(),
  usePatientChartStore: jest.fn(() => ({
```

packages/esm-patient-chart-app/src/visit/visits-widget/visit-context/visit-context-switcher.test.tsx
```diff
@@ -40,7 +32,20 @@ jest.mock('../visit.resource', () => ({
   useInfiniteVisits: () => mockUseInfiniteVisits('some-uuid'),
 }));
```
denniskigen:
```suggestion
```

packages/esm-patient-chart-app/src/visit/visits-widget/current-visit-summary.test.tsx
```diff
@@ -1,34 +1,38 @@
...
+import { getConfig } from '@openmrs/esm-framework';
+import { mockPatient, waitForLoadingToFinish } from 'tools';
+import { usePatientChartStore } from '@openmrs/esm-patient-common-lib';
+import CurrentVisitSummary from './current-visit-summary.extension';
 
 const mockGetConfig = jest.mocked(getConfig);
-const mockUseVisits = jest.mocked(useVisit);
+const mockUsePatientChartStore = jest.mocked(usePatientChartStore);
+
+jest.mock('@openmrs/esm-patient-common-lib/src/store/patient-chart-store', () => ({
```
denniskigen:
```suggestion
jest.mock('@openmrs/esm-patient-common-lib', () => ({
```

packages/esm-patient-chart-app/src/visit/visits-widget/current-visit-summary.extension.tsx
```diff
@@ -1,34 +1,26 @@
...
-
-  if (!!error) {
-    return <ErrorState headerTitle={t('failedToLoadCurrentVisit', 'Failed loading current visit')} error={error} />;
-  }
-
-  if (!currentVisit) {
+  if (!visitContext) {
     return (
       <EmptyState
         headerTitle={t('currentVisit', 'Current visit')}
```
denniskigen: Unrelated to the diff but the `displayText` here should probably be: 

```
displayText={t('activeVisits__lower', 'active visits')}
```

so the full empty state text reads as "There are no active visits to display for this patient"

packages/esm-patient-chart-app/src/visit/visits-widget/current-visit-summary.test.tsx
```diff
@@ -1,34 +1,38 @@
...
+      visitContext: null,
+      mutateVisitContext: null,
+      setPatient: jest.fn(),
+      setVisitContext: jest.fn(),
     });
-
-    render(<CurrentVisitSummary patientUuid="some-uuid" />);
+    render(<CurrentVisitSummary patientUuid={mockPatient.id} />);
     expect(screen.getByText(/current visit/i)).toBeInTheDocument();
     expect(screen.getByText('There are no active visit to display for this patient')).toBeInTheDocument();
```
denniskigen:
```suggestion
    expect(screen.getByText('There are no active visits to display for this patient')).toBeInTheDocument();
```

packages/esm-patient-chart-app/src/visit/visits-widget/current-visit-summary.test.tsx
```diff
@@ -1,34 +1,38 @@
...
   });
```
denniskigen:
```suggestion

  test('returns null when patientUuid does not match store patientUuid', () => {
    mockUsePatientChartStore.mockReturnValue({
      patientUuid: 'different-patient-id',
      patient: mockPatient,
      visitContext: null,
      mutateVisitContext: null,
      setPatient: jest.fn(),
      setVisitContext: jest.fn(),
    });
    render(<CurrentVisitSummary patientUuid={mockPatient.id} />);
    expect(screen.queryByText(/current visit/i)).not.toBeInTheDocument();
  });
 
```

This might be a useful test.

packages/esm-patient-common-lib/src/visit/revalidation-utils.ts
```diff
@@ -100,3 +100,12 @@ export function invalidateVisitAndEncounterData(mutate: KeyedMutator<unknown>, p
...
   invalidatePatientEncounters(mutate, patientUuid);
 }
+
+/**
+ * Invalides a visit fetched by URL /visit/<uuid>
+ * @param mutate - SWR mutate function from useSWRConfig()
+ * @param visitUuid
+ */
+export function invalidateVisitByUuid(mutate: KeyedMutator<unknown>, visitUuid: string) {
+  mutate(new RegExp(`${restBaseUrl}/visit${visitUuid}`));
```
denniskigen:
```suggestion
  mutate(new RegExp(`${restBaseUrl}/visit/${visitUuid}`));
```

packages/esm-patient-chart-app/src/visit/visits-widget/current-visit-summary.test.tsx
```diff
@@ -43,14 +47,12 @@ describe('CurrentVisitSummary', () => {
         },
         encounters: [],
```
denniskigen:
```suggestion
        encounters: [],
        patient: {
          uuid: mockPatient.id,
          display: 'Test Patient',
        },
```

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.workspace.tsx
```diff
@@ -285,46 +290,12 @@ const ImmunizationsForm: React.FC<DefaultPatientWorkspaceProps> = ({
...
-              render={({ field, fieldState }) => (
-                <OpenmrsDatePicker
-                  {...field}
-                  className={styles.datePicker}
-                  id="nextDoseDate"
-                  invalid={Boolean(fieldState?.error?.message)}
-                  invalidText={fieldState?.error?.message}
-                  labelText={t('nextDoseDate', 'Next dose date')}
-                  minDate={vaccinationDate}
+                  minDate={immunizationToEditMeta ? null : new Date()}
```
denniskigen: Why did this line change?

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.workspace.tsx
chibongho: `usePatientChartStore` should already be guarding against he case when the patient we start the visit for is not the same as the patient in the store. In that case, the `setVisitContext` is just a dummy function that does nothing.

That said, I'm slowing leaning towards the idea of having 2 versions of the visit form workspace (albeit they still share a lot of code), one meant for use within the patient chart and one outside. Same with all other workspaces in the patient chart that are used outside, of which there are surprisingly many (at least 12).

packages/esm-patient-chart-app/src/patient-chart/patient-chart.component.tsx
ibacher: Does it maybe make sense to refactor all of this into it's own hook rather than doing it as part of this component?

ibacher (commented):
I've been nowhere near as thorough as Dennis, but a few things that stuck out at me.

denniskigen:
Note to whoever is able to run this locally, check that we don't run into this regression again https://github.com/openmrs/openmrs-esm-patient-chart/pull/2554. We shouldn't, by the very nature of the approach, but it'd be nice to be absolutely certain.

chibongho:
> Note to whoever is able to run this locally, check that we don't run into this regression again https://github.com/openmrs/openmrs-esm-patient-chart/pull/2554. We shouldn't, by the very nature of the approach, but it'd be nice to be absolutely certain.

Confirmed. The e2e test to start a visit only triggers one request for the mutate:

https://github.com/user-attachments/assets/ecfd00b5-e3a7-4141-a2d1-acd5180ab090

ibacher (approved):

PR #2694, opened 2025-08-20
(feat) O3-4930: Add all orderable tests to the results viewer

NethmiRodrigo:
## Requirements

## Summary

This PR adds the `All orderable tests` to the default `resultsViewerConcepts` array in the config for it to appear in the results viewer.

## Screenshots

Screenshot 2025-08-20 at 11 57 29 PM

## Related Issue

https://openmrs.atlassian.net/browse/O3-4930


ibacher (commented):
Can we drop the other categories from the default configuration please?

ibacher (approved):

ibacher:
FWIW, I'm not merging this for now to unblock #2506, which is, I think, slightly more critical

ibacher:
Although it does appear that not all tests were correctly loaded:

Screenshot of TestOrderability CS

donaldkibet:
Thanks @NethmiRodrigo not closely related to the changes you have made, how do we plan to support results for things like imaginig orders i.e Ultra-sound,x-rays etc

rbuisson:
@ibacher  @NethmiRodrigo  is there still something blocking this?

ibacher:
> is there still something blocking this?

The metadata for "All Orderable Tests" doesn't get loaded correctly, so this fix would probably just cut down on the number of visible tests in the system.

rbuisson:
Thanks @ibacher . @NethmiRodrigo , need any help with this?

rbuisson:
Hi @gracepotma , were you able to look at this?

denniskigen:
Hey @NethmiRodrigo — circling back on this. The code change itself is straightforward and @ibacher approved the direction. The blocker was the upstream metadata issue with "All Orderable Tests" not loading correctly on dev3.

Has that been resolved since the content PR was merged? If so, this may just need a rebase and E2E re-run to verify. If the metadata issue persists, it's still blocking.

One note: with "All Orderable Tests" as the single root concept, the `Bloodwork` special-case in `augmentObstreeData` (`useObstreeData.ts`, line 46) becomes dead code. That's fine — it was a workaround for the overlapping Hematology/Bloodwork trees — but worth noting for cleanup.

denniskigen:
One more thing beyond the metadata blocker: the config swap isn't safe on its own even once the metadata is fixed.

With the current dev3 shape (`All Orderable Tests` → `Tests Orderability` → `Complete blood count` → `Haemoglobin`), `augmentObstreeData` would build `flatName` values like `All Orderable Tests-Tests Orderability-Complete blood count-Haemoglobin`. The panel grouping logic in `filter-context.tsx` (line 166) assumes the actual panel name is always `flatNameParts[1]`, which would be `Tests Orderability` — not the real panel. So every test would collapse under a single "Tests Orderability" heading instead of their actual panels (Complete blood count, Lipid panel, etc.).

The test mocks also still carry the old shallow `Hematology-...` flatNames while just renaming the root to "All Orderable Tests", so they don't exercise the real code path.

This means the PR needs to update the downstream grouping logic to handle the deeper hierarchy before the config default can be swapped.

VeronicaMuthee:
> Amylase is not among them, and @ibacher noted in the last round of reviews that the expected count was ~300 from OCL.

@denniskigen I can see that v25 has 523 concepts, and Amylyse is part of the collection. I'm not sure why it isn't appearing. Let me ask the OCL team what they think could be the issue.

denniskigen:
Closing for now. Feel free to reopen once we have everything setup correctly in OCL.

anish-mekom:
Hi @denniskigen, just checking in on this PR. Have all the required items been configured correctly in OCL?

PR #2705, opened 2025-08-23
(feat) Flag past medications in visit history table with tags

denniskigen:
## Requirements

## Summary

This PR adds a tag to past medications in the visit history table to indicate that the medication has been discontinued if:

- The order has a `DISCONTINUE` action
- The order has a `dateStopped` that is in the past
- The order has an `autoExpireDate` that is in the past

## Screenshots

https://github.com/user-attachments/assets/a1cfc538-20ff-4d99-8fcc-1272c10b6724

### DIFF
--- a/packages/esm-patient-chart-app/src/config-schema.ts
+++ b/packages/esm-patient-chart-app/src/config-schema.ts
@@ -16,6 +16,11 @@ export const esmPatientChartSchema = {
     _default: false,
     _description: 'Disable notes/tests/medications/encounters tabs when empty',
   },
+  drugOrderTypeConceptUuid: {
+    _type: Type.UUID,
+    _description: "Concept UUID for the 'Drug' order type used to fetch medications",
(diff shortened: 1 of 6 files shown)


packages/esm-patient-chart-app/src/config-schema.ts
```diff
@@ -16,6 +16,11 @@ export const esmPatientChartSchema = {
     _default: false,
     _description: 'Disable notes/tests/medications/encounters tabs when empty',
   },
+  drugOrderTypeConceptUuid: {
```
samuelmale, line 19: Can we use the config from the medications app?

```ts
const { drugOrderTypeUUID } = useConfig({ externalModuleName: '@openmrs/esm-patient-medications-app' })

```

denniskigen: Yeah, makes sense to have a single source of truth

PR #2752, opened 2025-09-23
(fix) O3-5002: Display units of obs value on encounter tile

RajPrakash681:
## Description
This PR implements utility functions to extract units directly from encounter observation data, eliminating the need for separate API calls to fetch concept units in the Patient Chart vitals and biometrics display.

## Requirements
- [x] This change is covered by automated tests (or does not need tests)
- [x] This change has been tested manually
- [x] This change uses only public APIs
- [x] This change is backwards compatible

## Screenshot
Screenshot 2025-09-23 224403

## Related Issue
Fixes missing units display in Patient Chart vitals and biometrics section
[5002](https://openmrs.atlassian.net/browse/O3-5002)

## Others
NA


packages/esm-patient-chart-app/src/clinical-views/utils/concept-utils.ts
```diff
@@ -0,0 +1,39 @@
...
+ * This eliminates separate API calls for concept units
+ */
+export function useConceptUnits(conceptUuid: string) {
+  return {
+    units: '', // Default empty value that will be overridden in the component
+    error: null,
+    isLoading: false,
+  };
+}
+
```
NethmiRodrigo, line 16: I don't see a need for this function, it isn't used

packages/esm-patient-chart-app/src/routes.json
```diff
@@ -38,6 +38,17 @@
...
+      "component": "encounterTile",
+      "order": 2,
+      "meta": {
+        "columnSpan": 4
+      },
+      "online": true,
+      "offline": true
+    },
+    {
+      "name": "clinical-views-details",
```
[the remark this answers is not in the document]
NethmiRodrigo: issue: This is actually incorrect, the component `encounterTile` isn't even registered in the index, so this change does not work. Please revert this change.

packages/esm-patient-chart-app/src/clinical-views/encounter-tile/encounter-tile.component.tsx
```diff
@@ -65,13 +68,13 @@ const EncounterData: React.FC<{
...
         <span className={styles.tileValue}>
-          <p>{obsValue}</p>
+          <p>{withUnit(obsValue, units)}</p>
         </span>
       )}
 
       {!isNil(summaryValue) && summaryValue !== '--' && (
         <span className={styles.tileValue}>
-          <p>{summaryValue}</p>
+          <p>{withUnit(summaryValue, units)}</p>
```
NethmiRodrigo, line 77: Better to leave the units out of the summary value because most times, its used to show the date of the encounter etc

packages/esm-patient-chart-app/src/clinical-views/utils/concept-utils.ts
```diff
@@ -0,0 +1,39 @@
...
+  if (!encounter || !encounter.obs || !conceptUuid) {
+    return '';
+  }
+
+  // Find the observation for this concept
+  const obs = encounter.obs.find(o => o.concept?.uuid === conceptUuid);
+  
+  // Return the units if found
+  // Note: The updated representation in useLastEncounter includes the 'units' field
+  return (obs?.concept as any)?.units || '';
```
NethmiRodrigo, line 33: suggestion: Without casting this to any, update the type [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/9925188560aa816ff92688eb59bce655320e6e87/packages/esm-patient-chart-app/src/clinical-views/types.ts#L41) to have units

NethmiRodrigo (changes requested):

NethmiRodrigo:
> ok first let me update all the things you suggested then i will update and sorry for inconvenience happened to you due to this i am just a newbie i will do my best to fix this all

Not an inconvenience at all, pretty impressive work for a newbie, good job!

packages/esm-patient-chart-app/src/clinical-views/encounter-tile/encounter-tile.component.tsx
```diff
@@ -65,7 +67,7 @@ const EncounterData: React.FC<{
       <span className={styles.tileTitle}>{t(column.header)}</span>
       {!(obsValue === '--' && summaryValue !== '--' && !isNil(summaryValue)) && (
         <span className={styles.tileValue}>
-          <p>{obsValue}</p>
+          <p>{withUnit(obsValue, units)}</p>
         </span>
       )}
```
denniskigen, line 73: Should `summaryValue` on line 77 also be wrapped with `withUnit`?

packages/esm-patient-chart-app/src/clinical-views/encounter-tile/encounter-tile.component.tsx
```diff
@@ -65,7 +67,7 @@ const EncounterData: React.FC<{
       <span className={styles.tileTitle}>{t(column.header)}</span>
       {!(obsValue === '--' && summaryValue !== '--' && !isNil(summaryValue)) && (
         <span className={styles.tileValue}>
-          <p>{obsValue}</p>
+          <p>{withUnit(obsValue, units)}</p>
         </span>
```
denniskigen, line 71: This and all other instances where we have `<p>` tags nested in `<span>` elements is invalid HTML. `<span>`s are inline elements and therefore cannot contain block-level elements like `<p>` tags.

RajPrakash681: Thanks for pointing that out . I’ve updated the code to replace the invalid <p> inside <span> with <div> wrappers around the values. This should resolve the HTML validity issue.

packages/esm-patient-chart-app/src/clinical-views/encounter-tile/encounter-tile.component.tsx
NethmiRodrigo: @denniskigen the gist is that this was done, but I asked to remove it because the summary value usually is used to show things like the `encounterDatetime`, but now that I've had some sleep (sorry, lol), I realize this was wrong. 
You're right, it should be wrapped in withUnits. @RajPrakash681 the `summaryObsValue` can have units, but different to the units of of the `obsValue`. So you need to wrap the `summaryObsValue` with `withUnits` but you need to get the unit of the summary obs. You can get its concept from `column.summaryConcept.primaryConcept`

RajPrakash681:
@NethmiRodrigo good evening, any updates on this any more changes needs to be done?

RajPrakash681:
@NethmiRodrigo @denniskigen please look at this and give feedback what changes needs to be done in this

NethmiRodrigo:
FYI @rbuisson and @VaishSiddharth, tagging since you were interested in having this :)

PR #2943, opened 2026-01-13
(feat) Add Browse tab alongside search for drug selection

olewandowski1:
## Requirements

## Summary

This PR introduces a new way to browse drugs by category while preserving the existing search‑based workflow.

**1. Drug Categories via Concept Sets**
Drugs can be grouped into configurable categories using **Concept Sets**, which are managed via **Legacy Admin**. These Concept Sets define how drugs are grouped and can be adapted to local needs (for example, by program, formulary, or service). Each category corresponds to a Concept Set containing the relevant drug concepts.

image

**2. Browse Tab for Browsing Drugs**
The drug selection UI has been enhanced with a new **Browse** tab, available **alongside the existing Search tab (which remains the default).** The Browse tab allows users to browse drugs by selecting a category, making it easier to discover medications **without knowing their exact names**.

https://github.com/user-attachments/assets/c53b74e9-d7e5-4bc6-9457-451feba26fc5

**3. Configuration via Implementer Tools**
Implementers can control which categories appear in the Browse tab by configuring Concept Set UUIDs under the `drugCategoryConceptSets` setting in `@openmrs/esm-patient-medications-app` via Implementer Tools. Only the configured Concept Sets are exposed to users.

https://github.com/user-attachments/assets/67fea1f8-930f-45e2-8439-ef2b64737015

### UI Workflow 
**Preconditions:** Concept Sets Predefined Meds 1 and Predefined Meds 2 are configured and available.
image
image

https://github.com/user-attachments/assets/09556778-4f6a-4623-8068-faac428b0490

## Other

Sample config:

```jsonc
{
	"@openmrs/esm-patient-medications-app": {
		"drugCategoryConceptSets": [
			"162305AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA", # Heparins
			"162170AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" # Sulfonamides
		]
	}
}
```


packages/esm-patient-medications-app/src/add-drug-order/drug-search/drug-search.resource.tsx
```diff
@@ -54,6 +80,74 @@ export function useDrugSearch(query: string) {
...
+
+  const drugs = useMemo(() => drugsResponse?.data?.results ?? [], [drugsResponse?.data?.results]);
+
+  return {
+    drugs,
+    isLoading: isLoadingConceptSet || isLoadingDrugs,
+    error: conceptSetError ?? drugError,
+    memberConceptNames,
+  };
+}
```
denniskigen: The original `useDrugListByConceptSet` hook does this:                               
              
```tsx                                                                                           
const memberConceptNames = conceptSetResp?.data?.setMembers?.map((m) => m.name.display);  
                                                                
const drugsResponse = useSWRImmutable(                                            
  `${restBaseUrl}/drug?q=${memberConceptNames.join(',')}&v=...`                   
);           
```                                                                     
                                                                                  
And that sends a request like: `/drug?q=Amoxicillin,Aspirin,Acetaminophen`.
                                                                   
This isn't quite aligned with the backend which has this flow:                                                      
                                                                                  
1. `DrugSearchHandler1_10.java` passes the `q` parameter directly to                  
`conceptService.getDrugs(query, ...)`
2. `HibernateConceptDAO.tokenizeName()` splits search terms by whitespace only: `escapedName.trim().split(" ")`                                                     
3. Commas are treated literally - searching for "Amoxicillin,Aspirin" looks for a 
drug literally named that, not two separate drugs.                         
                                                                                  
Result: The original code returned empty results because no drug is named         
"Amoxicillin,Aspirin,Acetaminophen".                                              
                                                                                  
We could remedy that using something like this:                                                                      

```tsx                                                  
async function fetchDrugsByConceptSet(conceptSetUuid: string): Promise<DrugListFetchResult> {
  // First, fetch the concept set members
  const conceptSetResponse = await openmrsFetch<ConceptSetMembersResponse>(
    `${restBaseUrl}/concept/${conceptSetUuid}?v=custom:(uuid,setMembers:(uuid,name))`,
  );

  const memberNames =
    conceptSetResponse.data?.setMembers?.map((member) => member.name?.display).filter(Boolean) ?? [];

  if (memberNames.length === 0) {
    return { drugs: [], hasFailures: false };
  }

  // Fetch drugs for each concept name with a small concurrency cap.
  const drugs: Array<DrugSearchResult> = [];
  let hasFailures = false;
  for (let start = 0; start < memberNames.length; start += maxConcurrentDrugSearches) {
    const batch = memberNames.slice(start, start + maxConcurrentDrugSearches);
    const requests = batch.map((name) =>
      openmrsFetch<{ results: Array<DrugSearchResult> }>(
        `${restBaseUrl}/drug?q=${encodeURIComponent(name)}&v=${drugSearchRepresentation}`,
      ),
    );
    const settled = await Promise.allSettled(requests);
    if (settled.some((result) => result.status === 'rejected')) {
      hasFailures = true;
    }
    const batchDrugs = settled
      .filter((result): result is PromiseFulfilledResult<FetchResponse<{ results: Array<DrugSearchResult> }>> =>
        result.status === 'fulfilled',
      )
      .flatMap((result) => result.value?.data?.results ?? []);
    drugs.push(...batchDrugs);
  }

  // Dedupe by drug uuid
  const deduped = Array.from(new Map(drugs.map((drug) => [drug.uuid, drug])).values());

  return { drugs: deduped, hasFailures };
}
```                           

This function batches requests (5 at a time)  which prevents overwhelming server with 50+ parallel requests for large concept sets.                                                      
                                                                                  
Adding comma-separated support to `/drug?q=` would require making changes to the backend. The frontend fix proposed works with existing backends - no server changes needed.

olewandowski1: Great catch, Denis. Thanks for pointing that out.

ibacher:
Thanks @olewandowski1! I think this feature is clearly something the community can benefit from but I want to suggest a few wrinkles and things to hopefully improve things a fair bit.

1. From the screenshots, it seems that the "Browse" tab is rendered regardless of if its configured. If there are no configured concept sets, it seems like there's no value in rendering the tab (and it's concomitant network calls).
2. I'm a little concerned about the overall amount of network traffic that this will potentially generate. It looks like there's: 1) a call to load the concepts sets via /conceptreferences, 2) a call to load the set members for a specific concept set when it's selected and then 3) a call to load the drugs matching the concept name. We can already pre-load the set members as part of the /conceptreferences call and that feels like it should reduce the amount of network calls we need to make (these network calls are a point where the application traditionally hits waterfalls, so the fewer we can make the better).
3. I see that we're querying drugs by name, because that's what the REST API supports, but the ConceptService itself supports the ability to query drugs by their related concepts directly ([see here](https://github.com/openmrs/openmrs-core/blob/a56c82b310dc7c3062a4802fe1fcea8dfcb5395b/api/src/main/java/org/openmrs/api/db/ConceptDAO.java#L460C20-L460C28)) and it seems like it might be better to do that. This is because: 1) (most critically) drugs in the drug table might have brandnames specified (I'd still want to see all "Advil" entries if the concept set included "ibuprofen", so as-is this will miss valid drugs and 2) matching on name might accidentally include drugs that were not intended to be there, i.e., if the concept set includes just "ibuprofen" I may not want "ibuprofen + ..." drugs to appear.
4. Nested categorization: Currently, as implemented, this only really supports one-level of categorization with a single concept set-per-category, but it seems like it would be more useful to be able to support nested categories. To take an example common in OpenMRS-land, it would be more useful (I think) to be able to select maybe a broad class of drugs (ARVs for example) and then a narrower set of drugs within that (first-line ARV regimens). We added support for something similar in the results viewer using nested concept sets to build nested categories like this, which required the addition of a new endpoint (we called it "obstree" because it was ultimately fetching obs), but I can guess that something like a "concepttree" endpoint that could fetch all the nested concepts within a set of concept sets is useful not just for this, but also for other use-cases, like potentially classifying laboratory test orders.

druchniewicz:
@ibacher @denniskigen @dkayiwa bumping

packages/esm-patient-medications-app/src/add-drug-order/drug-search/drug-browse-results.component.tsx
```diff
@@ -0,0 +1,94 @@
...
+}: DrugBrowseResultsProps) {
+  const { t } = useTranslation();
+
+  if (isLoading) {
+    return <DrugBrowseSkeleton />;
+  }
+
+  if (isError) {
+    return <Tile>{t('errorFetchingDrugCategory', 'Error fetching drugs')}</Tile>;
+  }
```
ibacher: So `isError` here is just `hasFailures` from the `useDrugsByConcepts` hook, but it's not immediately clear that an error fetching one page of results means we should throw away other possible results. It seems like we could get away with showing partial results with some kind of inline notification.

druchniewicz:
@denniskigen all comments have been resolved. Can you please take a look?

druchniewicz:
@denniskigen can you take a look on this as well as it is close to be merged I believe? Thanks in advance

druchniewicz:
@ibacher @denniskigen can we merge this PR?

denniskigen:
Is the branch up to date? It has likely accrued conflicts. Do you mind rebasing against `main`?

druchniewicz:
> Is the branch up to date? It has likely accrued conflicts. Do you mind rebasing against `main`?

@olewandowski1 could you take a look on it? If you don't have time please let me know

packages/esm-patient-medications-app/src/add-drug-order/drug-search/drug-search.resource.tsx
```diff
@@ -1,12 +1,12 @@
...
-import useSWRImmutable from 'swr/immutable';
 import { type FetchResponse, openmrsFetch, restBaseUrl, useFeatureFlag, type Visit } from '@openmrs/esm-framework';
 import {
   type Drug,
   type DrugOrderBasketItem,
   type DrugOrderTemplate,
   type OrderTemplate,
 } from '@openmrs/esm-patient-common-lib';
+import { useMemo } from 'react';
+import useSWRImmutable from 'swr/immutable';
```
denniskigen: Ian asked for these to be put back where they were. React/library imports (`useMemo`, `useSWRImmutable`) should come before framework imports per the [import conventions](https://o3-docs.openmrs.org/docs/coding-conventions/code-organization):

```tsx
import { useMemo } from 'react';
import useSWRImmutable from 'swr/immutable';
import { type FetchResponse, openmrsFetch, restBaseUrl, useFeatureFlag, type Visit } from '@openmrs/esm-framework';
```

packages/esm-patient-medications-app/src/add-drug-order/drug-search/drug-search.resource.tsx
```diff
@@ -54,6 +77,94 @@ export function useDrugSearch(query: string) {
...
+  const { data, error, isLoading } = useSWRImmutable<ConceptTreeNode, Error>(
+    conceptUuid ? `${restBaseUrl}/concepttree?concept=${conceptUuid}` : null,
+    fetcher,
+  );
+
+  return useMemo(
+    () => ({
+      tree: data,
+      isLoading,
+      isError: error,
```
denniskigen: Minor naming inconsistency: this returns `isError: error` where the value is an actual `Error | undefined`, not a boolean. `useConceptSets` (line 160) returns `error` directly, which is clearer. Consider renaming to `error` for consistency:

```tsx
return useMemo(
  () => ({
    tree: data,
    isLoading,
    error,
  }),
  [data, isLoading, error],
);
```

packages/esm-patient-medications-app/src/add-drug-order/drug-search/drug-browse.component.tsx
```diff
@@ -0,0 +1,181 @@
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
```
denniskigen, line 1: The new components (`DrugBrowse`, `DrugBrowseResults`, `DrugBrowseEmptyState`) and hooks (`useConceptTree`, `useDrugsByConcepts`, `useConceptSets`) don't have any unit tests. Would be good to add at least basic coverage for the hooks and the key rendering states (loading, error, empty, populated).

denniskigen (commented):
Nice work iterating on this, @olewandowski1. The Browse tab works well end-to-end, the config-driven approach is clean, and the previous feedback from Ian and me has been largely addressed. A few remaining items below.

Also, the PR description still references "List tab" in several places, but the UI now says "Browse". Could you update the description to match?

olewandowski1:
Thanks for the valuable feedback @denniskigen . Could you please take a look and review again?

packages/esm-patient-medications-app/src/add-drug-order/drug-search/drug-browse.component.tsx
```diff
@@ -0,0 +1,181 @@
...
+  closeWorkspace,
+  patient,
+  visit,
+}: DrugBrowseProps) {
+  const { t } = useTranslation();
+  const isTablet = useLayoutType() === 'tablet';
+
+  const [rootConceptSet, setRootConceptSet] = useState<ConceptSet | null>(null);
+  const [breadcrumbs, setBreadcrumbs] = useState<ConceptTreeNode[]>([]);
+  const { tree, isLoading: isLoadingTree, error: isTreeError } = useConceptTree(rootConceptSet?.uuid);
```
denniskigen: Nit: `isTreeError` is destructured from the `error` field of `useConceptTree`, which returns `Error | undefined`, not a boolean. The `is*` prefix reads as boolean and the truthiness check on line 118 works regardless. Consider renaming to `treeError` for consistency with `conceptSetsError`.

packages/esm-patient-medications-app/translations/en.json
```diff
@@ -36,9 +41,12 @@
   "editFrequencyComboBoxTitle": "Frequency",
   "editRouteComboBoxTitle": "Route",
```
denniskigen, line 42: Nit: key/value mismatch. The key is category-scoped (`errorFetchingDrugCategory`) but the value is generic (`Error fetching drugs`). Either rename the key to `errorFetchingDrugs`, or tighten the value to match the key.

denniskigen (approved):
LGTM. Thanks for working through the iterations on this, @olewandowski1 and @druchniewicz - the feature feels polished. Two tiny nits inline that we should sort out before this lands:

olewandowski1:
@denniskigen Fixed. Ready for the review.

denniskigen (approved):
Both nits from the last round are addressed cleanly in [ad7823a](https://github.com/openmrs/openmrs-esm-patient-chart/commit/ad7823a) — the `treeError` rename and the `errorFetchingDrugs` key fix landed exactly where I expected. Thanks for the quick turnaround, @olewandowski1 and @druchniewicz, and for sticking with this through multiple review rounds. Ready to ship.

PR #2955, opened 2026-01-19
(feat) O3-5078: Add sticky notes to the patient chart

Muppasanipraneeth:
## Requirements

## Summary

- Adds a sticky note to the patient banner — one persistent note per patient, editable in place and voided on delete (matches the O2 coreapps behaviour per @ibacher's feedback).
- CRUD is backed by the REST `obs` resource (`/ws/rest/v1/obs`), keyed on a configurable concept UUID. Using REST rather than FHIR gives us `auditInfo.creator.display` for the provider name shown in the designs.
- Fetched obs are filtered to those with no `encounter` and no `formNamespaceAndPath`, so if a distro's `stickyNoteConceptUuid` overlaps with a concept the visit-note form also writes against, those form-generated obs don't surface here.
- Clicking the header button opens the popover in any non-empty state (skeleton while loading, inline error + retry on failure, or the note). When no note exists, it opens the create modal directly.
- Panel dismissal uses `@openmrs/esm-framework`'s `useOnClickOutside` plus an ESC handler; clicks inside an open edit/delete modal are ignored so they don't collapse the panel underneath.
- If `stickyNoteConceptUuid` is cleared (empty string), the header button is hidden entirely — distros that don't want the feature see nothing.
- Character limit, modal layout, and provider-name footer follow the Zeplin designs.
- Includes unit tests for the panel, header button, create/edit modal, delete modal, and the `isStandaloneStickyNote` filter.

## Screenshots

https://github.com/user-attachments/assets/5f0027e1-496f-4694-bfe3-65beb4f1c663

## Related Issue

[jira ticket](https://openmrs.atlassian.net/browse/O3-5078)
[project Detailed Description](https://openmrs.atlassian.net/wiki/spaces/projects/pages/592150547/Sticky+Note+in+O3)

## Other
[Related PR](https://github.com/openmrs/openmrs-esm-patient-chart/pull/2745)


packages/esm-patient-notes-app/src/sticky-notes/sticky-note-panel.tsx
```diff
@@ -0,0 +1,156 @@
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { Button, InlineLoading, SkeletonText, TextArea } from '@carbon/react';
+import { formatDate, getCoreTranslation, showSnackbar, useConfig } from '@openmrs/esm-framework';
+import { type ConfigObject } from '../config-schema';
+import AddStickyNote from './add-sticky-note-button';
+import DeleteStickyNote from './delete-sticky-note-button';
+import EditStickyNote from './edit-sticky-note-button';
+import { createStickyNote, updateStickyNote, useStickyNotes } from './resources';
+import styles from './sticky-note-panel.scss';
```
denniskigen:
```suggestion
import React, { useCallback, useEffect, useMemo, useState } from 'react';
import { useTranslation } from 'react-i18next';
import { Button, InlineLoading, SkeletonText, TextArea } from '@carbon/react';
import { formatDate, getCoreTranslation, showSnackbar, useConfig } from '@openmrs/esm-framework';
import { type ConfigObject } from '../config-schema';
import { createStickyNote, updateStickyNote, useStickyNotes } from './resources';
import AddStickyNote from './add-sticky-note-button';
import DeleteStickyNote from './delete-sticky-note-button';
import EditStickyNote from './edit-sticky-note-button';
import styles from './sticky-note-panel.scss';
```

nit: Import order

denniskigen (commented):
Thanks, @Muppasanipraneeth! A few thoughts from reading through this implementation:

**1. Failed reads are handled as "no sticky note exists"**

In `sticky-note-panel.tsx`, the panel only looks at `notes` and `isLoading`:

```tsx
const { notes: stickyNotes, isLoading, mutate } = useStickyNotes(patientUuid);
```

But `useStickyNotes()` returns `notes: []` on fetch failure because of the fallback in `resources.ts`. Since `isError` is ignored, a transient GET failure is treated the same as "this patient has no sticky note", and the panel flips into create mode.

That means a clinician could be prompted to create a new sticky note even when one already exists, simply because the initial read failed. This seems worth fixing with an explicit error/retry state.

**2. New sticky-note strings won't be extracted for translation**

The package's `extract-translations` script only scans:
- `*.component.tsx`
- `*.modal.tsx`
- `*.extension.tsx`
- `*.workspace.tsx`
- `*.hook.tsx`

The new sticky-note files are plain `*.tsx` files (`sticky-note-panel.tsx`, `sticky-note-header-button.tsx`, `delete-sticky-note-modal.tsx`, etc.), so the `t(...)` keys introduced here won't be picked up for catalogs.

That means the feature will silently fall back to English in localized installs. The fix is to rename the files to follow the established convention: `sticky-note-panel.component.tsx`, `delete-sticky-note-modal.modal.tsx`, `sticky-note-header-button.extension.tsx`, etc.

denniskigen:
Latest changes following rework to align with O2 implementation: https://github.com/user-attachments/assets/ff835f91-3931-49ed-a06e-1e68a6e6eb8f @ibacher @VeronicaMuthee

packages/esm-patient-notes-app/src/sticky-notes/sticky-note-header-button.component.tsx
```diff
@@ -0,0 +1,63 @@
...
+  const { t } = useTranslation();
+  const [showPanel, setShowPanel] = useState(false);
+  const { note, isLoading, error, mutate } = useStickyNote(patientUuid);
+
+  const handleClose = useCallback(() => {
+    setShowPanel(false);
+  }, []);
+
+  const handleClick = useCallback(() => {
+    if (note) {
```
ibacher: This should probably be a no-op when `isLoading` or there is an error?

packages/esm-patient-notes-app/src/sticky-notes/delete-sticky-note.modal.tsx
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }, [noteUuid, mutate, close, onClose, t]);
+  return (
+    <>
+      <ModalHeader closeModal={close}>{t('confirmDeleteStickyNote', 'Confirm delete sticky note')}</ModalHeader>
+      <ModalBody>
+        <p>{t('confirmDeleteStickyNoteMessage', 'Are you sure you want to delete this sticky note?')}</p>
+      </ModalBody>
+      <ModalFooter>
+        <Button kind="secondary" onClick={close}>
```
ibacher: Theoretical race-condition (though we definitely have this elsewhere), since this isn't disabled when deleting, and since it just immediately calls `close()`, it's possible for this to fire before the DELETE operation has returned, meaning that someone could submit the delete, and the closing the modal losing the `onClose()` callback (whatever that might do).

packages/esm-patient-notes-app/src/config-schema.ts
```diff
@@ -12,11 +12,17 @@ export const configSchema = {
     _default: true,
     _description: 'Indicates whether a primary diagnosis is required when submitting a visit note',
   },
+  stickyNoteConceptUuid: {
+    _type: Type.UUID,
```
ibacher: Even though we aren't consistent about this and don't have this implemented, this is theoretically important:

```suggestion
    _type: Type.ConceptUuid,
```

packages/esm-patient-notes-app/src/sticky-notes/delete-sticky-note-button.component.tsx
```diff
@@ -0,0 +1,39 @@
...
+      onClose,
+    });
+  };
+  return (
+    <IconButton
+      label={t('deleteStickyNote', 'Delete sticky note')}
+      onClick={handleDelete}
+      kind="ghost"
+      size="sm"
+      align="top-right"
```
ibacher: Should be "top-end". Carbon has deprecated "top-right" and so on since a lot of our components were written.

packages/esm-patient-notes-app/src/sticky-notes/delete-sticky-note-button.component.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { IconButton } from '@carbon/react';
+import { useTranslation } from 'react-i18next';
+import { showModal, TrashCanIcon } from '@openmrs/esm-framework';
```
ibacher, line 3: Nit: this should be before 'react-i18next'

packages/esm-patient-notes-app/src/sticky-notes/sticky-note.modal.tsx
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }, [close, existingNote, isEditMode, isUnchanged, mutate, patientUuid, stickyNoteConceptUuid, t, trimmed]);
+
+  return (
+    <>
+      <ModalHeader closeModal={close} title={t('stickyNote', 'Sticky note')} />
+      <ModalBody>
+        <TextArea
+          id="sticky-note-text"
+          labelText=""
```
ibacher: This should have an `aria-label` even if we don't render a visible label.

packages/esm-patient-notes-app/src/sticky-notes/sticky-note.resource.ts
ibacher, line 1: `sticky-not.resource.ts` is more consistent with our naming standards

ibacher (changes requested):
The provider's name and the date the obs was made _must_ be included. I guess we're no longer using the FHIR API, but those properties should be there.

Technical stuff:

* We could use a test for empty input or a blank string
* We need a way of identifying "genuine" sticky notes from other concepts using the same UUID. O2 didn't do this, but O2 didn't have the O3 visit note using the same concept.

ibacher:
I forgot to say this is overall excellent work and I'm excited to get it in!

denniskigen:
Following the latest commit, nothing gets rendered when the config property is not set:

CleanShot 2026-04-23 at 15 44 26@2x

ibacher (approved):

PR #3002, opened 2026-02-06
(fix) O3-5041: Heading semantics across patient chart

UjjawalPrabhat:
## Requirements

## Summary

Changes widget section headings from `<h4>` to `<h2>` so the heading hierarchy (`<h1>` → `<h2>`) no longer skips levels. The dashboard title renders as `<h1>`, but all widget headers were `<h4>`, skipping `<h2>` and `<h3>` — a WCAG accessibility violation.

SCSS selectors updated in tandem to preserve existing visual styles.

## Related Issue

[O3-5041](https://openmrs.atlassian.net/browse/O3-5041)

## Other

Related [PR](https://github.com/openmrs/openmrs-esm-patient-management/pull/2251)

[O3-5041]: https://openmrs.atlassian.net/browse/O3-5041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


denniskigen (changes requested):
Thanks for tackling heading semantics, @UjjawalPrabhat. This is a real accessibility concern, but I don't think the current sweep is safe to merge as-is.

The main issue is that this PR applies the same heading-level change across many different contexts, and that rule does not hold everywhere. For example, in the drug-order workspace, the workspace title comes from `Workspace2`, which renders it in the shell header rather than as a semantic heading element. That means the in-form "Order Form" heading is still the only actual page heading there, so changing it from `<h1>` to `<h2>` leaves that workspace with no `<h1>`.

Some of the changes do look good, especially replacing heading tags inside tooltip content with non-heading text. But this example shows that the heading levels need to be reviewed in context — page by page / workspace by workspace — rather than changed with one blanket rule.

Given that, and given how stale and conflicted this branch is as of today, I'd prefer closing this and reopening a narrower pass that audits each page's heading structure explicitly.

PR #3035, opened 2026-02-17
(fix) O3-5432: Validate completion date against enrollment date in program enrollment form

UjjawalPrabhat:
## Requirements

## Summary

The program enrollment form let you pick a completion date before the enrollment date and submitted it anyway, which the API rejected with a generic 400 error and no actionable guidance. It also blocked valid same-day completion: the completion picker started at enrollment + 1 day and was disabled outright when the enrollment date was today, even though the backend `PatientProgramValidator` only rejects completion dates strictly before the enrollment date.

The form now validates the date order up front with a Zod `superRefine` check that shows an inline error on the completion date field ("Completion date cannot be before the enrollment date"), and the picker constraints are relaxed so the completion date can equal the enrollment date.

Same-day completion needs one more nuance: the date picker emits midnight timestamps, which the backend would reject as being before a same-day enrollment. The submitted completion timestamp is therefore aligned to the enrollment timestamp when it falls earlier on the same day, while stored completion times that already fall later that day are preserved on resave.

Tests cover the inline validation error, same-day completion in create mode, and preservation of stored same-day completion timestamps in edit mode.

## Screenshots

image

## Related Issue

[O3-5432](https://openmrs.atlassian.net/browse/O3-5432)

## Other

[O3-5432]: https://openmrs.atlassian.net/browse/O3-5432?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


denniskigen (commented):
Thanks for adding the cross-field validation!

VeronicaMuthee:
@denniskigen, please check out this PR.

packages/esm-patient-programs-app/src/programs/programs-form.workspace.tsx
```diff
@@ -46,13 +46,26 @@ export interface ProgramsFormProps {
...
+  z
+    .object({
+      selectedProgram: z.string().refine((value) => !!value, t('programRequired', 'Program is required')),
+      enrollmentDate: z.date(),
+      completionDate: z.date().optional().nullable(),
+      enrollmentLocation: z.string(),
+      selectedProgramStatus: z.string(),
+    })
+    .superRefine((data, ctx) => {
+      if (data.completionDate && data.enrollmentDate && data.completionDate < data.enrollmentDate) {
```
denniskigen: Thanks for the fixup, @UjjawalPrabhat. The operator and the picker constraints now line up nicely with the backend. There's one wrinkle left before same-day completion actually works: this compares full timestamps, but the two fields disagree on time-of-day. The date picker emits midnight local time, while `enrollmentDate` defaults to `new Date()` in create mode and `parseDate(dateEnrolled)` in edit mode, both of which carry a time-of-day. So enrolling today and completing today still shows "Completion date cannot be before the enrollment date". I reproduced this in both modes: in create mode by picking today as the completion date, and in edit mode with an enrollment stored as `2020-01-16T15:00:00.000+0000` (the same shape as the fixtures in `__mocks__/programs.mock.ts`), where setting the completion date to `2020-01-16` blocks the save. A day-granularity comparison fixes it:

```suggestion
      if (
        data.completionDate &&
        data.enrollmentDate &&
        dayjs(data.completionDate).isBefore(data.enrollmentDate, 'day')
      ) {
```

Relatedly, the backend validator compares raw timestamps too (`OpenmrsUtil.compareWithNullAsLatest` in [`PatientProgramValidator`](https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/validator/PatientProgramValidator.java#L108-L112)), so once this check allows same-day dates, the save can still fail. I confirmed against dev3: posting `dateCompleted` at midnight with `dateEnrolled` at noon on the same day returns a 400 with `enrolledDateShouldBeBeforecompletionDate`, while equal timestamps save fine. To make same-day work end-to-end, the payload in `onSubmit` needs to align the two when they fall on the same day, something like:

```ts
dateCompleted: completionDate
  ? dayjs(completionDate).isSame(enrollmentDate, 'day')
    ? dayjs(enrollmentDate).format()
    : dayjs(completionDate).format()
  : null,
```

With both changes in place, your existing tests pass and same-day completion saves successfully. It'd be nice to add a test covering the same-day case in create mode (completion date set to today) so this doesn't regress.

packages/esm-patient-programs-app/src/programs/programs-form.test.tsx
```diff
@@ -166,6 +166,24 @@ describe('ProgramsForm', () => {
...
+
+    const completionDateInput = screen.getByRole('textbox', { name: /date completed/i });
+    const saveButton = screen.getByRole('button', { name: /save and close/i });
+
+    await user.click(completionDateInput);
+    await user.paste('2019-12-01');
+    await user.tab();
+
+    await user.click(saveButton);
+
```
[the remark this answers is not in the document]
denniskigen: Yeah, I see the mock in Core still destructures `isInvalid` while the real component and this form pass `invalid`, so the assertion is blocked for now. Let me try to handle that upstream.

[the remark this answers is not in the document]
denniskigen: Fixed in https://github.com/openmrs/openmrs-esm-core/pull/1814

packages/esm-patient-programs-app/src/programs/programs-form.workspace.tsx
```diff
@@ -122,7 +139,11 @@ const ProgramsForm: React.FC<PatientWorkspace2DefinitionProps<ProgramsFormProps,
         patient: patientUuid,
         program: selectedProgram,
         dateEnrolled: enrollmentDate ? dayjs(enrollmentDate).format() : null,
-        dateCompleted: completionDate ? dayjs(completionDate).format() : null,
+        dateCompleted: completionDate
+          ? dayjs(completionDate).isSame(enrollmentDate, 'day')
+            ? dayjs(enrollmentDate).format()
+            : dayjs(completionDate).format()
+          : null,
```
denniskigen, line 149: This one's on me, since the same-day alignment came from my earlier suggestion. There's one wrinkle in edit mode: `completionDate` initializes from the stored API value, which can validly be later than enrollment on the same day, since the backend only rejects completion strictly before enrollment. As written, resaving an enrollment with `dateEnrolled` at 09:00 and `dateCompleted` at 17:00 rewrites the completion to 09:00, even when the user only changed something unrelated like the location. Substituting the enrollment timestamp only when the completion timestamp actually falls before it keeps the create-mode midnight case working while preserving valid stored times:

```suggestion
        dateCompleted: completionDate
          ? dayjs(completionDate).isSame(enrollmentDate, 'day') && dayjs(completionDate).isBefore(enrollmentDate)
            ? dayjs(enrollmentDate).format()
            : dayjs(completionDate).format()
          : null,
```

VeronicaMuthee:
LGTM @UjjawalPrabhat @denniskigen. Here is a small suggestion that Claude is making: →The only substantive follow-up I'd request is a short code comment documenting why dateCompleted is coerced to the enrollment timestamp on the same day, to protect the fix from future refactors.

packages/esm-patient-programs-app/src/programs/programs-form.workspace.tsx
denniskigen: Applied in 01ea17332 on @UjjawalPrabhat's behalf (he's without his computer at the moment), together with an edit-mode test asserting that a stored same-day completion time later than the enrollment time survives a resave, and a short code comment documenting the alignment, per @VeronicaMuthee's note above.

denniskigen (approved):
LGTM. Thanks, @UjjawalPrabhat!

PR #3139, opened 2026-03-19
(fix) O3-5515: Remove hardcoded 10-point data limit from vitals and biometrics charts

sohamdhande:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title  includes a conventional commit label. See existing PR titles for inspiration.
- [ ] My work is based on designs, which are linked or shown either in the Jira ticket or the description below. (See also: Styleguide)

## Summary
The vitals and biometrics chart components previously hardcoded .slice(0, 10) in their useMemo blocks, silently discarding all observations beyond the 10 most recent. For a pediatric patient seen monthly over 2 years, this means 14 of 24 weight measurements are never shown.

This PR replaces that arbitrary count-based truncation with a 12-month default time window, keeping charts focused on clinically relevant recent data while no longer silently dropping older observations. The existing Carbon Charts zoom bar `(zoomBar: { top: { enabled: true } })` allows users to navigate the full history when needed.

What changed

Removed `.slice(0, 10)` from vitals-chart.component.tsx and biometrics-chart.component.tsx
Added a cutoff of 12 months (via useMemo) so charts default to recent data
cutoff is stable across renders and included in the chartData dependency array
What this is not This is a standalone fix and is not a prerequisite for or dependency of any growth chart work. The growth chart is a separate app and will manage its own data range logic independently.

Testing No dedicated unit tests exist for these chart components (confirmed via grep). TypeScript `(tsc --noEmit) `and ESLint both pass clean after this change.

## Screenshots

N/A — no UI layout changes. The chart now renders all data points instead of silently capping at 10. The zoom bar handles navigation across the full history.

## Related Issue

https://issues.openmrs.org/browse/O3-3

## Other

No tests exist for `vitals-chart.component.tsx` or 
`biometrics-chart.component.tsx` (confirmed via grep). TypeScript  (`tsc --noEmit`) and ESLint both pass clean after this change.
Note on performance: Both charts already have Carbon Charts' built-in  zoom bar enabled (`zoomBar: { top: { enabled: true } }`). This gives users a native pan/zoom control for navigating large datasets without  rendering all points at once. For patients with extremely large observation histories, the zoom bar provides the appropriate UX — the hardcoded limit was not the right solution for that concern.

### DIFF
--- a/packages/esm-patient-vitals-app/src/biometrics/biometrics-chart.component.tsx
+++ b/packages/esm-patient-vitals-app/src/biometrics/biometrics-chart.component.tsx
@@ -1,5 +1,6 @@
 import React, { useId, useMemo, useState } from 'react';
 import classNames from 'classnames';
+import dayjs from 'dayjs';
 import { useTranslation } from 'react-i18next';
 import { Tab, TabListVertical, TabPanel, TabPanels, TabsVertical } from '@carbon/react';
 import { LineChart, ScaleTypes } from '@carbon/charts-react';
@@ -76,11 +77,17 @@ const BiometricsChart: React.FC<BiometricsChartProps> = ({ patientBiometrics, co
     value: BiometricType;
   }>;
 
+  const cutoff = useMemo(() => dayjs().subtract(12, 'month'), []);
+
   const chartData = useMemo(
     () =>
       patientBiometrics
         .filter((biometrics) => biometrics[selectedBiometrics.value])
-        .sort((biometricA, biometricB) => new Date(biometricA.date).getTime() - new Date(biometricB.date).getTime())
+        .filter((biometrics) => dayjs(biometrics.date).isAfter(cutoff))
+        .sort(
+          (biometricA, biometricB) =>
+            new Date(biometricA.date).getTime() - new Date(biometricB.date).getTime(),
+        )
         .map(
           (biometrics) =>
             biometrics[selectedBiometrics.value] && {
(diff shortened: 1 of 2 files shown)


denniskigen (changes requested):
Thanks for the follow-up on ibacher's feedback, @sohamdhande — a time-based window is a better approach than the arbitrary 10-point cap.

Two things to consider:

1. **Make the time window configurable.** 12 months is a reasonable default, but different implementations have different needs (pediatric clinics tracking growth over years vs. acute care). A config schema option (e.g., `chartWindowMonths` defaulting to 12) would let implementations tune it.

2. **Handle the empty-window case.** If a patient hasn't been seen in over 12 months but has older data, the chart renders blank. A fallback — e.g., showing the N most recent observations when the time filter yields zero — would keep the chart useful.

sohamdhande:
Thanks for the feedback @denniskigen!

Updated the implementation the time window is now configurable via chartWindowMonths (defaults to 12), so different widget configs can adjust it as needed.

Also handled the empty-window edge case: if no observations fall within the configured range, it falls back to the most recent ones so the chart always renders something meaningful rather than going blank.

Let me know if this looks good or if you want anything changed.

Josephkagimu1:
Thanks for this effort @sohamd1567 , it would have been better to check on the E2E tests , OpenMRS CI  problem plus other few failing tests .

Josephkagimu1 (commented):
kindly work on the failing tests , also verify that the build was successful

PR #3155, opened 2026-03-22
(fix) O3-4296: Improve en translation strings in form entry app

NitinKumar1-1:
## Requirements

## Summary

Fixes grammar and casing inconsistencies in `esm-form-entry-app/translations/en.json`.

* Corrected "a hour" → "an hour" in `addHour`
* Changed "Minus a hour/minute/second" → "Subtract an/a hour/minute/second"
* Fixed sentence casing on `previousValue`, `selectWeeks`, `useValue`, `maxDate`, `minDate`, `maxLength`, `minLength`, and `loadingComponent`
* Changed "Kindly notify" → "Please notify" in `formSubmissionFailed`
* Added ellipsis to `submitting` for consistency with other loading states

## Related Issue

https://issues.openmrs.org/browse/O3-4296

## Other

Companion commit in openmrs-ngx-formentry library https://github.com/openmrs/openmrs-ngx-formentry/commit/45e3ef0e93b74d9551ac1eba58b1433649be482c


denniskigen (changes requested):
Thanks for working on this! The grammar fixes are good catches — "a hour" is a genuine bug. I have some feedback on both files before this can be merged.

### `en.json`

The grammar and casing corrections are welcome ("a hour" → "an hour", "Minus a hour" → "Subtract an hour", "Component Loading failed..." → "Component loading failed.", title case → sentence case, etc.).

However, the `{count}` changes to the relative time strings (`daysAgo`, `hoursAgo`, `minutesAgo`, `monthsAgo`, `yearsAgo`) will break at runtime. These strings are consumed by the [`TimeAgoPipe` in `openmrs-ngx-formentry`](https://github.com/openmrs/openmrs-ngx-formentry/blob/master/projects/ngx-formentry/src/form-entry/pipes/time-ago.pipe.ts#L49-L63), which concatenates the number before the translated string:

```ts
return days + ' ' + this.translate.instant('daysAgo');
```

The pipe calls `translate.instant('daysAgo')` with no interpolation params, so the current format `" days ago"` (with leading space) is intentional. Changing it to `"{count} days ago"` would render `"5 {count} days ago"` at runtime.

If you'd like to move to proper i18next-style interpolation, that's a great improvement — but it requires a coordinated change to the pipe in [`openmrs-ngx-formentry`](https://github.com/openmrs/openmrs-ngx-formentry) first. Please revert the `{count}` changes in this PR.

Two formatting issues as well:
- Please remove the blank lines between entries — no other translation file in this repo uses them and they'll create noise in future Transifex sync diffs.
- The trailing newline at the end of the file was removed — please preserve it.

### `print.component.tsx`

This change needs to be reverted. `patientDetails.age` is the return value of the framework's [`age()`](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-utils/src/age-helpers.ts#L83) function, which returns a pre-formatted, locale-aware string via `Intl.DurationFormat` — for example `'4 yrs, 5 mths'` for a child, `'24 yrs'` for an adult, or `'3 days'` for a neonate.

The PR wraps this in:

```ts
t('ageYears', { defaultValue: '{{count}} years', count: Number(patientDetails?.age) })
```

`Number('4 yrs, 5 mths')` returns `NaN`, so this would render **"NaN years"** on the print page. It also discards the rich formatting that displays appropriate units for different age ranges. The existing code that renders `patientDetails.age` directly is correct.

The `.join(', ')` change to `identifiers` is harmless but unnecessary — `Array.toString()` already produces comma-separated output.

denniskigen (commented):
Thanks for the updates, @NitinKumar1-1 — the reverts look good and the grammar/casing fixes are clean.

One small thing: `en.json` is still missing its trailing newline (you can see the `\ No newline at end of file` marker in the diff). Could you add that back?

denniskigen (commented):
Correction to my last comment — I missed that `en.json` in `esm-form-entry-app` is managed by Transifex and synced from `openmrs-ngx-formentry`. Editing it here will get overwritten on the next sync. The grammar fixes are all good catches, but they need to be made in the source strings in [`openmrs-ngx-formentry`](https://github.com/openmrs/openmrs-ngx-formentry) instead. That leaves only the `print.component.tsx` trailing space removal in this PR, which is correct but probably too small to warrant a standalone PR.

denniskigen:
Ignore my last two comments about Transifex — I was wrong. `en.json` here is the Transifex **source** file (it gets pushed to Transifex, not pulled from it), so your edits won't get overwritten. The grammar fixes in this PR are in the right place.

The same bugs (`addHour`, `minusHour`, etc.) also exist in [`openmrs-ngx-formentry/src/translations/en.json`](https://github.com/openmrs/openmrs-ngx-formentry/blob/master/src/translations/en.json), which maintains its own copy independently. That can be a follow-up PR if you're interested, but it's not blocking this one.

So back to where we were: could you restore the trailing newline at the end of `en.json`? After that, this is good to go.

NitinKumar1-1:
@denniskigen Thanks for the clarification 

I’ll add the trailing newline to fix the formatting issue here.

For the translation updates managed via Transifex, I understand they need to be applied upstream. I’ll open a separate PR in openmrs-ng-formentry with the relevant fixes.

Thanks again for the guidance!

denniskigen (approved):
LGTM. Thanks, @NitinKumar1-1!

denniskigen (approved):

PR #3166, opened 2026-03-24
(feat) O3-5541: Gate visit note datepicker behind RDE feature flag

omsrivastava512:
## Requirements

## Summary

The visit note form currently always shows a date picker, but selecting a past date only makes sense when retrospective data entry (RDE) is enabled. Without RDE, the date picker is misleading — users can pick a past date, but the rest of the form (e.g. visit selection) doesn't support backdating.

This PR gates the date picker behind the `rde` feature flag. When RDE is off, the date picker is hidden and `noteDate` defaults to `new Date()` at form mount time. Because this always falls within the existing 30-minute threshold in `onSubmit`, `encounterDatetime` is omitted from the payload and the server assigns the timestamp itself.

## Changes

- Make `noteDate` optional in the Zod schema when RDE is disabled
- Conditionally render the date picker based on the `rde` feature flag
- Add tests for both flag states

## Screenshots

image

## Related Issue

https://openmrs.atlassian.net/browse/O3-5541

## Other

[Slack Discussion](https://openmrs.slack.com/archives/C02UNMKFH8V/p1774268513702999)


denniskigen (commented):
Hey @omsrivastava512 — thanks for picking this up! Nice, focused change. A couple of things to consider:

### `noteDate` still required in the schema

The Zod schema has `noteDate: z.date()` (required), and the default value of `new Date()` means validation passes even when the datepicker is hidden. This works today, but it's a bit fragile — if someone later removes or changes that default, the form would break with no visible datepicker to fix it.

Worth making `noteDate` optional in the schema when RDE is disabled, or at minimum adding a code comment explaining why the default value is load-bearing here. Something like:

```ts
noteDate: isRetrospectiveDataEntryEnabled ? z.date() : z.date().optional(),
```

You'd also want to adjust the `onSubmit` handler to handle `noteDate` being `undefined` — though in practice the existing 30-minute guard (`finalNoteDate.diff(now, 'minute') <= 30 → null`) would need a small tweak since `dayjs(undefined)` returns the current time, which would still null out. So it works, but being explicit would be cleaner.

### Missing test coverage

There are no tests for the new conditional behavior. The existing test file (`visit-notes-form.test.tsx`) doesn't mock `useFeatureFlag` at all. It would be good to add at least:

1. A test that verifies the datepicker is **not** rendered when `useFeatureFlag('rde')` returns `false`
2. A test that verifies the datepicker **is** rendered when `useFeatureFlag('rde')` returns `true`

You can follow the pattern used in `order-basket-action-button.test.tsx` for mocking the feature flag.

### Minor: submission behavior when RDE is off

When RDE is disabled, `noteDate` defaults to `new Date()` and the user can't change it. In `onSubmit`, this always falls within the 30-minute window, so `encounterDatetime` ends up `undefined` in the payload, letting the server set it. That's the intended behavior per the PR description — just flagging that it's worth a brief comment in the code to make this implicit contract visible to future readers.

Overall the approach is solid — gating behind the feature flag is the right call.

denniskigen (approved):
LGTM. Thanks, @omsrivastava512!

omsrivastava512:
Hey @denniskigen, now that the datepicker is gated behind RDE, do you think [this](https://openmrs.atlassian.net/browse/O3-5518) is something that can be picked up? More precisely, having more controlled frontend instead of relying on the backend completely.

PR #2973, opened 2026-01-27
(feat) O3-5374: Add support for future medication start dates

Sadashii:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspiration.
- [ ] My work is based on designs, which are linked or shown either in the Jira ticket or the description below. (See also: [Styleguide](http://om.rs/o3ui))
- [x] My work includes tests or is validated by existing tests.

## Summary

As per discussions held on:
1. Old Jira ticket: https://openmrs.atlassian.net/browse/O3-4433
2. Connected Talk thread: https://talk.openmrs.org/t/guidance-needed-should-openmrs-allow-providers-to-schedule-medications-to-start-at-a-future-date/45848
3. Expanded scope: https://openmrs.atlassian.net/browse/O3-5374

This PR adds support for medication orders that start in the future, while preserving existing past and same-day start-date behavior.

The implementation:

1. Adds "Upcoming medications" sections to the Medications page and patient Summary page.
2. Reworks medication-order fetching so one hook buckets orders into upcoming, active, and past states.
3. Allows future start dates for new medication orders and for editing already-upcoming orders.
4. Keeps active-medication revisions limited to today or earlier because the backend stops the previous order from the replacement order's `dateActivated`.
5. Posts past and same-day starts using `dateActivated`.
6. Posts future starts using `scheduledDate` with `urgency=ON_SCHEDULED_DATE`, matching the OpenMRS order contract.
7. Keeps duplicate active/upcoming medication search results visible with an "Already prescribed" cue, while disabling duplicate actions to avoid backend overlap errors.
8. Updates tests for the Vitest setup and adds coverage for bucketing, future-start payloads, duplicate blocking, same-day date snapping, and revision behavior.

### Upcoming Aspirin 162.5mg order scheduled for 28th May

Screenshot 2026-05-22 at 01 46 55

### Search result duplicate guard

Screenshot 2026-05-22 at 01 47 20

## Related Issue
https://openmrs.atlassian.net/browse/O3-5374
https://openmrs.atlassian.net/browse/O3-5375
https://openmrs.atlassian.net/browse/O3-5376

## Other

PR #3198, opened 2026-04-02
(fix) O3-4444: Persist medication order start date

aditya123000:
## Requirements

## Summary

Persists the user-selected Start date for NEW, RENEW, and REVISE medication orders by sending it as `dateActivated` and aligning the encounter handling with backend semantics.

- **NEW**: the chosen `startDate` is sent as `dateActivated`. The encounter created in the same POST is backdated to the earliest selected date so the backend's `dateActivated >= encounterDatetime` constraint holds.
- **RENEW**: per backend semantics a renewal is a new prescription event, so it now lands in a fresh encounter on the visit with `encounterDatetime` tracking the chosen start date. The form does not inherit `dateActivated` from the previous order.
- **REVISE**: posts against the previous order's existing encounter with the user-selected `dateActivated`. The date picker enforces a minimum of `previous.dateActivated + 1s` so the backend's `previous.dateStopped = new.dateActivated - 1s` computation produces a valid order chain.
- **DISCONTINUE**: unchanged.

Pure helpers extracted for testability: `getEarliestStartDate` (basket → encounter-date derivation) and `getStartDateMinimum` (date picker floor).

### Out of scope (deferred to follow-ups)

- Future-dated orders, handled by `scheduledDate` + `urgency: ON_SCHEDULED_DATE` in #2973.
- Datetime precision on date-picker floors. The picker enforces minimums at calendar-day precision while the backend enforces them at datetime precision. Two instances: a NEW order on the visit's start day yields a `Date` at 00:00 local that can fall before `visit.startDatetime` (triggering `Encounter.datetimeShouldBeInVisitDatesRange`); a same-day REVISE can submit 00:00 for the day, which falls before `previous.dateActivated + 1s`. Cleanest fix is a clamp-on-submit (or a schema check against `getStartDateMinimum`).
- Backdating with a reused encounter on visit-disabled installs. When `orderEncounterUuid` is preset (only on installs that don't support visits and where the user already has an order encounter today), `order-basket.component.tsx` posts orders against that existing encounter without backdating it. A backdated NEW/RENEW will fail the backend's `dateActivated >= encounterDatetime` check. Fix shape: fetch the existing encounter's `encounterDatetime`, and either backdate (if permissible) or create a fresh encounter when the earliest basket `startDate` precedes it.
- `orderEncounterType` is currently duplicated between `esm-patient-medications-app` and `esm-patient-orders-app` config schemas; consolidating into `esm-patient-common-lib` is a follow-up.

### NEW order with backdated start date

Placed Paracetamol with `Start date` sent to yesterday -> order lands with Start date 05-May-2026 + an encounter dated 05/04/2025 (POST body shows `dateActivated: 2026-05-04T21:00:00Z`).

https://github.com/user-attachments/assets/f340e76f-2123-424c-8c59-44afd5fc12db

### RENEW with backdated start date

Renewed Zyrtec with `Start date` set to yesterday -> renewal lands in a fresh encounter (not the original order's encounter), with `dateActivated: 2026-05-04T21:00:00Z`. The original keeps running until it is discontinued or auto-expires.

https://github.com/user-attachments/assets/6e3e3f67-e256-4a90-bd54-4a1535a2755b

### RENEW with default (today)

Renewed Paracetamol without changing `Start date` -> renewal lands today in a separate fresh encounter (proves the new-encounter behaviour irrespective of backdating). 

https://github.com/user-attachments/assets/e01f6308-15f1-4b2e-b606-1533fc3cf916

### REVISE picker constraint

On REVISE of the active Amoxicillin, the date picker rejects dates earlier than `previous.dateActivated + 1s`; valid revise saves cleanly with the original dose superseded.

https://github.com/user-attachments/assets/e0396e99-b252-43c1-af72-2c760b5797bb

### DISCONTINUE preserves original

Discontinued Amoxicillin lands in Past Medications with the `DISCONTINUED tag and the original `Start date` intact; payload includes no `dateActivated` field.

https://github.com/user-attachments/assets/6ef1af1e-5319-4417-bd1b-6a1cb74753d8

## Related Issue

[O3-4444](https://openmrs.atlassian.net/browse/O3-4444)

## Other

[O3-4444]: https://openmrs.atlassian.net/browse/O3-4444


omsrivastava512 (dismissed):
The implementation is clean and the tests are thorough, good work on both. But this touches medication order payload logic and there's no linked ticket or prior discussion anywhere. Who validated that `dateActivated` should be sent here? In a clinical system this kind of payload change needs sign-off before implementation, not after. I'd suggest opening a Jira ticket, getting the right people to confirm this is the intended behaviour, and linking it here before this gets merged.

denniskigen (commented):
Thanks for working on this @aditya123000 — the start date field not being persisted is a real gap, and the test coverage you've added is thorough.

This same change was previously attempted in #2241 (ticket [O3-4444](https://openmrs.atlassian.net/browse/O3-4444)), which was closed without merging. That ticket can be reused here. There's also overlap with #2973, which adds past and future start date support.

A couple of things to flag about how the backend handles dateActivated:

1. **dateActivated cannot be set to a future date**. The backend [rejects future values](https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/validator/OrderValidator.java#L112-L113) during validation. If omitted, the backend [defaults it to the current date](https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java#L232). Future clinical starts are modeled with `scheduledDate` + `urgency: ON_SCHEDULED_DATE`, not by pushing a future value into dateActivated. So mapping the form's "Start date" directly to dateActivated works for today/past dates, but won't compose with future-start support in #2973.

2. **For REVISE, backdating dateActivated changes order-chain sequencing**. The backend [stops the previous order one second before](https://github.com/openmrs/openmrs-core/blob/f661d0638/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java#L187) the new order's dateActivated, and [marks orders with past dates as retrospective](https://github.com/openmrs/openmrs-core/blob/f661d0638/api/src/main/java/org/openmrs/api/impl/OrderServiceImpl.java#L175-L179). With this PR, `buildMedicationOrder` now defaults `startDate` to the original order's `dateActivated` instead of `new Date()`. So a clinician revising an order without touching the date field would silently backdate the revision, trigger retrospective handling, and resequence the order chain. For `RENEW` (posted as action: 'NEW' with `previousOrder`), the resequencing concern is weaker, but it still silently backdates the new order.

**Suggestion**: keep buildMedicationOrder's existing logic (`action === 'DISCONTINUE' ? order.dateActivated : new Date()`) and only add `dateActivated` to the POST payload for NEW orders where the user has explicitly selected a date.

Also, the `{/* TODO: This input does nothing */}` comment on line 588 of drug-order-form.component.tsx should be removed since this PR makes the field functional.

packages/esm-patient-medications-app/src/api/api.ts
```diff
@@ -283,7 +286,7 @@ export function buildMedicationOrder(order: Order, action: OrderAction): DrugOrd
     patientInstructions: order.dosingType !== 'org.openmrs.FreeTextDosingInstructions' ? order.dosingInstructions : '',
```
denniskigen, line 292: Suggest reverting this to the original:

```ts
startDate: action === 'DISCONTINUE' ? order.dateActivated : new Date(),
```

Defaulting `startDate` to `order.dateActivated` for RENEW and REVISE pre-fills the form with the original activation date, which can be weeks or months in the past. Combined with sending `dateActivated` in the payload, a clinician revising an order without touching the date field would silently backdate the new order, and the backend would treat that as retrospective and resequence the order chain.

packages/esm-patient-orders-app/src/utils/index.ts
```diff
@@ -71,7 +71,7 @@ export function buildMedicationOrder(order: Order, action?: OrderAction): DrugOr
     patientInstructions: order.dosingType !== 'org.openmrs.FreeTextDosingInstructions' ? order.dosingInstructions : '',
     asNeeded: order.asNeeded,
     asNeededCondition: order.asNeededCondition ?? null,
-    startDate: action === 'DISCONTINUE' ? order.dateActivated : new Date(),
+    startDate: order.dateActivated,
```
denniskigen: Same suggestion as in the medications-app copy. Reverting keeps RENEW/REVISE behavior unchanged.

packages/esm-patient-medications-app/src/api/api.ts
```diff
@@ -165,6 +165,7 @@ export const prepMedicationOrderPostData: PostDataPrepFunction = (
         : 'org.openmrs.SimpleDosingInstructions',
       dosingInstructions: order.isFreeTextDosage ? order.freeTextDosage : order.patientInstructions,
       concept: order.drug.concept.uuid,
+      ...(order.startDate ? { dateActivated: toOmrsIsoString(new Date(order.startDate)) } : {}),
```
denniskigen: Nice approach. The conditional spread keeps the payload backward-compatible when the user hasn't explicitly picked a date.

packages/esm-patient-medications-app/src/api/api.ts
```diff
@@ -193,6 +194,7 @@ export const prepMedicationOrderPostData: PostDataPrepFunction = (
         : 'org.openmrs.SimpleDosingInstructions',
       dosingInstructions: order.isFreeTextDosage ? order.freeTextDosage : order.patientInstructions,
       concept: order.drug.concept.uuid,
```
denniskigen, line 202: Correcting myself: please keep these spreads on RENEW and REVISE. Once `buildMedicationOrder` is reverted to default to `new Date()` for non-DISCONTINUE actions, the form's start date defaults to today for RENEW/REVISE, and forwarding it as `dateActivated` lets clinicians set an explicit effective date when they need to. RENEW is essentially a new prescription with a pointer to the previous order, and REVISE genuinely is a modification whose `dateActivated` defines when the revision takes effect. Sorry for the churn here.

denniskigen (commented):
Thanks for iterating on this @aditya123000, and for chasing down Jira access. The conditional-spread pattern is a clean approach and the test coverage is solid.

**Editing this review to correct an earlier suggestion:** I had said to drop the `dateActivated` spread from RENEW and REVISE. On reflection that was overcautious. RENEW is functionally a new prescription with a pointer to the previous order, and REVISE genuinely is a modification whose `dateActivated` defines when the revision takes effect. Keeping the spread on all three branches gives clinicians an explicit start date for any of them. The silent-backdating concern is fully addressed by reverting `buildMedicationOrder`.

For context on the relationship with #2973: future-dated prescriptions still belong there (`scheduledDate` + `urgency: ON_SCHEDULED_DATE`). One additional REVISE constraint that should live in #2973 or a sibling ticket: the backend computes `previous.dateStopped = new.dateActivated - 1s`, so the date picker on REVISE should eventually enforce `min={previous.dateActivated + 1s}` to avoid stopping the previous order before its own activation.

Concrete shape for this PR:

1. Revert `buildMedicationOrder` in both `packages/esm-patient-medications-app/src/api/api.ts` and `packages/esm-patient-orders-app/src/utils/index.ts` to `action === 'DISCONTINUE' ? order.dateActivated : new Date()`.
2. Keep the `dateActivated` spread on all three branches (`NEW`, `RENEW`, `REVISE`) of `prepMedicationOrderPostData`.
3. Update the tests to assert: `NEW`/`RENEW`/`REVISE` send the user-selected `dateActivated`, `DISCONTINUE` preserves `order.dateActivated`, and `buildMedicationOrder` does **not** preserve the previous activation date for `RENEW`/`REVISE` (locks the regression shut).
4. Remove the `TODO: This input does nothing` comment at `drug-order-form.component.tsx:588` since the field now persists for all non-DISCONTINUE actions.
5. Add `O3-4444` to the PR description (the ticket key alone is fine, no Jira access needed for that).

Happy to re-review as soon as that's pushed.

denniskigen:
Pushed an update aligning the implementation with Ian's guidance and the backend's semantics. Four commits on top:

- `ea6ce34f` reverts `buildMedicationOrder` so RENEW/REVISE no longer pre-fill the form's `startDate` from the previous order's `dateActivated`.
- `8eb8b334` adds `postOrderOnNewEncounter` (single-order new-encounter helper) in `esm-patient-common-lib` and a `previousOrderDateActivated` field on `DrugOrderBasketItem`.
- `fbd2d758` derives the encounter datetime from the earliest basket `startDate` when signing the order basket, so `dateActivated >= encounterDatetime` holds for backdated NEW orders.
- `7bde5a4e` wires the start date through for NEW/RENEW/REVISE: routes RENEW through a fresh encounter, constrains the REVISE picker to `previous.dateActivated + 1s`, and adds a zod refine rejecting future start dates.

Verified manually against a local OpenMRS backend for the four core cases (NEW with backdate, RENEW with backdate in a fresh encounter, REVISE picker constraint and valid revise, DISCONTINUE unchanged). Tests added for the pure helpers and `buildMedicationOrder`/`prepMedicationOrderPostData` behavior. Description and [O3-4444](https://openmrs.atlassian.net/browse/O3-4444) updated to reflect the broader scope.

Two known follow-ups, intentionally deferred:

1. **Time-of-day picker bug**: `minDate` uses the visit's exact `startDatetime`, but the picker greys out at day-level and the picked `Date` lands at 00:00 local. So if a visit started at e.g. 09:33, the picker visually allows that day but the picked `Date` falls before the visit start, triggering `Encounter.datetimeShouldBeInVisitDatesRange`. Cleanest fix is clamp-on-submit. Worth its own ticket.
2. **`orderEncounterType` config duplication** between medications-app and orders-app, with a TODO in the code pointing at consolidating into `esm-patient-common-lib`.

Happy to address either here or split out.


[O3-4444]: https://openmrs.atlassian.net/browse/O3-4444?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

aditya123000:
Thanks @denniskigen for the clarification and for revisiting the earlier suggestion — the distinction between preventing silent backdating vs still allowing explicit clinician-selected `dateActivated` values makes sense now.

I’ll update the PR accordingly by:

1. Reverting `buildMedicationOrder` in both locations to default non-DISCONTINUE actions to `new Date()`
2. Keeping the conditional `dateActivated` spread for NEW / RENEW / REVISE payloads
3. Updating the tests to lock the regression cases you outlined
4. Removing the obsolete TODO comment
5. Adding O3-4444 to the PR description

I’ll push the updated changes shortly for re-review.

denniskigen:
@aditya123000 sorry about the miscommunication. I'd already pushed all those changes on top of your branch after I'd been asked to take over this PR and bring it to completion following the scope change.

aditya123000:
@denniskigen No problem at all — thanks for the clarification and for taking it through the broader scope changes.
I appreciate all the detailed review guidance throughout this PR. I learned a lot about the OpenMRS order semantics and the renew/revise flow handling from the discussion.
Thanks again for helping bring this to completion.

packages/esm-patient-medications-app/src/add-drug-order/add-drug-order.component.tsx
```diff
@@ -115,41 +118,75 @@ const AddDrugOrder: React.FC<AddDrugOrderProps> = ({
...
+    [
+      clearOrders,
+      closeWorkspace,
+      mutateOrders,
+      patientUuid,
+      t,
+      orderToEditOrdererUuid,
+      orderEncounterType,
+      visitContext,
+      session,
```
NethmiRodrigo: The entire session object is in the dependency array, so whenever any session field changes, the entire callback is recreated. Only `session?.sessionLocation?.uuid` is actually read — narrow the dep:

```tsx
const sessionLocationUuid = session?.sessionLocation?.uuid;
// ...
}, [..., orderEncounterType, visitContext, sessionLocationUuid]);
```

packages/esm-patient-medications-app/src/add-drug-order/add-drug-order.component.tsx
```diff
@@ -115,41 +118,75 @@ const AddDrugOrder: React.FC<AddDrugOrderProps> = ({
...
+          kind: 'error',
+          title: t('errorSavingDrugOrder', 'Error saving drug order'),
+          subtitle: error.message,
+        });
+      };
+
+      if (finalizedOrder.action === 'RENEW') {
+        const orderPost = prepMedicationOrderPostData(finalizedOrder, patientUuid, null, orderToEditOrdererUuid);
+        const encounterDate = finalizedOrder.startDate
+          ? new Date(finalizedOrder.startDate as string | Date)
```
NethmiRodrigo: The `as string | Date` cast seems unnecessary here since the `startDate` is already typed as `Date | string` and `new Date()` accepts both.

```tsx
const encounterDate = finalizedOrder.startDate ? new Date(finalizedOrder.startDate) : new Date();
```

packages/esm-patient-medications-app/src/add-drug-order/add-drug-order.component.tsx
```diff
@@ -115,41 +118,75 @@ const AddDrugOrder: React.FC<AddDrugOrderProps> = ({
...
+      const handleError = (error: { message?: string }) => {
+        showSnackbar({
+          isLowContrast: false,
+          kind: 'error',
+          title: t('errorSavingDrugOrder', 'Error saving drug order'),
+          subtitle: error.message,
+        });
+      };
+
+      if (finalizedOrder.action === 'RENEW') {
```
NethmiRodrigo: Would be good if we can add a test here. The RENEW dispatch branch in `submitDrugOrderToServer` has no component-level coverage — only the helper has unit tests.

```tsx
it('routes RENEW orders through postOrderOnNewEncounter with the selected start date as encounterDate', async () => {
  const startDate = new Date('2026-05-04T00:00:00.000Z');
  // render with a RENEW initialOrder whose startDate = startDate, submit
  expect(postOrderOnNewEncounter).toHaveBeenCalledWith(
    expect.objectContaining({ action: 'NEW', previousOrder: expect.any(String) }),
    expect.any(String),
    expect.any(String),
    expect.anything(),
    expect.any(String),
    undefined,
    startDate,
  );
});
```

NethmiRodrigo (commented):
Just a couple of nits:

denniskigen:
Thanks for the close review @NethmiRodrigo. Your nits surfaced something bigger: chasing the dep-array narrowing made me re-trace the dispatch, and the entire RENEW branch turns out to be unreachable.

- `submitDrugOrderToServer` has one call site (`add-drug-order.component.tsx:264`), and that dispatch gates strictly on `currentOrder?.action === 'REVISE'`.
- `handleFormSubmission` in the form preserves `initialOrderBasketItem.action` unchanged, so action can't mutate from RENEW to REVISE between form open and save.
- The "Renew" entry point on the medications list (`handleRenewClick`) doesn't open `AddDrugOrder` at all. It pushes the RENEW item into the basket and opens the `order-basket` workspace. RENEW orders are then signed via the basket's `postOrdersOnNewEncounter` flow, with `encounterDate` derived from `getEarliestStartDate(orders)`.

So the `if (finalizedOrder.action === 'RENEW')` branch couldn't fire. Te medications-app didn't need its own single-order RENEW dispatch because the basket already handles it correctly. This also matches the documented renewal flow (Renew → basket → save basket; basket POSTs as `action: 'NEW'` with `previousOrder` set).

I've pushed [a commit](https://github.com/openmrs/openmrs-esm-patient-chart/pull/3198/changes/a68ab1bdc961433a34a339e96656a35d570d7d8a) removing the unreachable branch, the now-orphaned `postOrderOnNewEncounter` helper in `esm-patient-common-lib` (no other consumers) and its tests, the `useSession` import and `sessionLocationUuid` (only the dead branch read it), and the `orderEncounterType` config field added for this branch alone.

packages/esm-patient-medications-app/src/add-drug-order/add-drug-order.component.tsx
denniskigen: Good catch — but stepping back, the whole branch this dep was supporting turned out to be unreachable. Removed it in `a68ab1bd`. Details in the [top-level reply](https://github.com/openmrs/openmrs-esm-patient-chart/pull/3198#issuecomment-4406509275).

packages/esm-patient-medications-app/src/add-drug-order/add-drug-order.component.tsx
denniskigen: Same as above — the line is gone in `a68ab1bd`. Thanks for spotting the unnecessary cast; tracing it is what surfaced the dead-code issue. Details in the [top-level reply](https://github.com/openmrs/openmrs-esm-patient-chart/pull/3198#issuecomment-4406509275).

packages/esm-patient-medications-app/src/add-drug-order/add-drug-order.component.tsx
denniskigen: Mooted by `a68ab1bd` — removed the RENEW dispatch branch entirely since it was unreachable. The basket flow handles RENEW (already covered by `order-basket.utils.test.ts`'s `getEarliestStartDate` tests). Details in the [top-level reply](https://github.com/openmrs/openmrs-esm-patient-chart/pull/3198#issuecomment-4406509275).

ibacher (approved):
I think I'm happy with this in this state. Note, I've only read the code though.

PR #3202, opened 2026-04-02
(feat) Add patient task list app to patient chart app

vasharma05:
## Requirements

## Summary

- Migrated the standalone `openmrs-esm-task-list` app into `openmrs-esm-patient-chart/packages` as a new package: `packages/esm-patient-task-list-app`.
- Renamed the app package and runtime module ID from `@openmrs/esm-task-list-app` to `@openmrs/esm-patient-task-list-app`.
- Aligned the new package with patient-chart workspace conventions (`rspack`, shared `jest`/`tsconfig` patterns, repo metadata, scripts, and dependency structure).

## Migration Process

- Created the new package directory under patient-chart workspaces.
- Copied app source and assets from task-list:
  - `src/**` (workspace components, resources, config, routes, declarations)
  - `translations/en.json`, `translations/fr.json`, `translations/ar.json`
  - `README.md`
- Replaced task-list identity references with patient-task-list identity and patient-chart repository metadata.
- Added package-level `rspack.config.js`.

## Package Changes

Updated config files to match patient-chart standards:

- `jest.config.js` now extends root `jest.config.js`.
- `tsconfig.json` now extends root `tsconfig.json`.

Key package updates in `packages/esm-patient-task-list-app/package.json`:

- `name` -> `@openmrs/esm-patient-task-list-app`
- `browser` -> `dist/openmrs-esm-patient-task-list-app.js`
- Build/dev scripts migrated from `webpack` to `rspack`.
- `repository` / `homepage` / `bugs` URLs point to `openmrs-esm-patient-chart`.
- Version aligned to patient-chart package versioning (`12.1.0`).
- Script and dependency structure aligned with neighboring patient-chart packages.

## Additional Fix Included

- Updated `task-details-view` test helper to avoid Testing Library node-access lint violations.
- Adjusted date assertion to support multiple matching date strings in rendered output (`getAllByText`), fixing failing test cases.


ibacher (commented):
Thanks @vasharma05! Few remarks on some small tweaks to make.

denniskigen (commented):
Thanks for this @vasharma05 — the CarePlan mapping, system task templates via PlanDefinition, and the three due-date modes are well structured. Solid test coverage too.

@ibacher already flagged translation files, import order, and unused styles. Two correctness issues to add:

1. **Editing visit-based tasks can re-anchor them to the wrong visit**

`extractDueDate` in `task-list.resource.ts` parses the `encounter-associatedEncounter` extension into a local `visitUuid` variable but never returns it — the function only returns `{ dueDate, dueDateType }`. So every fetched `THIS_VISIT` / `NEXT_VISIT` task loses its original reference visit in frontend state.

When the user edits and saves, `AddTaskForm` reconstructs `referenceVisitUuid` from the current active visit or latest visit instead of preserving the original. On update, the backend clears `dueDateReferenceVisit` ([CarePlanMapper.java lines 298–300](https://github.com/openmrs/openmrs-module-tasks/blob/main/api/src/main/java/org/openmrs/module/tasks/api/fhir/CarePlanMapper.java#L298-L300)) and only restores it from the incoming `encounter-associatedEncounter` extension ([lines 384–394](https://github.com/openmrs/openmrs-module-tasks/blob/main/api/src/main/java/org/openmrs/module/tasks/api/fhir/CarePlanMapper.java#L384-L394)), so dropping that value in frontend state can silently re-anchor the task to a different visit.

Fix: return `visitUuid` from `extractDueDate`, populate it on `task.dueDate.referenceVisitUuid` in `createTaskFromCarePlan`, and preserve it through the edit flow in `AddTaskForm`.

2. **New tasks can save with no due schedule**

The `ContentSwitcher` falls back to index 0 when `dueDateType` is `undefined` (`add-task-form.component.tsx` line 365), so the UI looks like "Next visit" is selected even though the form state still has no due-date type. Submitting the form sends no due date type. The user sees a selected option but the form doesn't capture it unless they click a different tab and click back.

3. **Cleanup (non-blocking):**

- `useConfig()` called without type parameter in `add-task-form.component.tsx` and `task-list.resource.ts` — should be `useConfig<Config>()`
- `parseAssignment` returns `null` on the fallthrough path but the return type is `... | undefined`
- `(ext as any)` casts in `extractDueDate` and `extractPriority` — consider a typed extension interface
- `deleteTask` is a soft delete (sets status to `cancelled`) but the name and modal copy say "delete"

ibacher:
> Prior to merging this in,

No, because we also don't magically pick this up just because it's here. We can add both the frontend and backend in one go (which gives me time to rework somethings in the backend).

packages/esm-patient-task-list-app/src/workspace/task-list.resource.ts
```diff
@@ -0,0 +1,560 @@
...
+    const validTasks = parsedTasks.filter((task) => Boolean(task.uuid));
+    const activeTasks = validTasks.filter((task) => task.status !== 'cancelled');
+
+    return activeTasks.sort((a, b) => {
+      if (a.completed !== b.completed) {
+        return a.completed ? 1 : -1;
+      }
+
+      const aDue = a.dueDate?.date?.getTime() ?? 0;
+      const bDue = b.dueDate?.date?.getTime() ?? 0;
```
denniskigen: Thanks for iterating on this, @vasharma05. The mapping code lines up cleanly with the backend — I verified the extension values, the visit re-anchoring fix, and the cancelled-task filtering against `openmrs-module-tasks` and everything checks out.

One small thing: tasks without a due date get `?? 0` here, which sorts them *before* all dated tasks. In a clinical task list, tasks with explicit due dates (especially overdue ones) are probably more urgent. Could we flip the default so dateless tasks sort to the bottom?

```suggestion
      const aDue = a.dueDate?.date?.getTime() ?? Infinity;
      const bDue = b.dueDate?.date?.getTime() ?? Infinity;
```

packages/esm-patient-task-list-app/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "url": "https://github.com/openmrs/openmrs-esm-patient-chart/issues"
+  },
+  "dependencies": {
+    "@carbon/react": "^1.83.0",
+    "@hookform/resolvers": "^5.2.2",
+    "react-hook-form": "^7.66.0",
+    "zod": "^4.1.12"
+  },
+  "peerDependencies": {
+    "@openmrs/esm-framework": "9.x",
```
denniskigen, line 47: nit: This range pulls yarn's monorepo-wide `react-hook-form` resolution from 7.52.2 to 7.72.0, which also affects `esm-patient-immunizations-app` (the only other consumer, at `^7.46.2`). To keep the lockfile change scoped to this PR, could we match the existing range?

```suggestion
    "react-hook-form": "^7.52.0",
```

packages/esm-patient-task-list-app/src/declarations.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module '*.css';
+declare module '*.scss';
+declare module '*.png';
+
+declare type SideNavProps = object;
```
denniskigen: nit: This looks like a leftover from the standalone task-list app. Nothing in this package references it, and no other patient-chart package declares it either. We should remove it.

```suggestion
```

denniskigen:
A couple more things:

1. In `AddTaskForm`, system-task defaults are only set, not cleared. Since the backend makes `priority`, `rationale`, and `defaultAssigneeProviderRoleId` independently optional, a user can select a system task with defaults, then switch to a custom task or another system task without one of those defaults, and the stale values will still be submitted. We should clear those dependent fields before applying the newly selected system task’s defaults, and also clear them when switching to custom input.

2. `routes.json` currently allows `webservices.rest >=2.2.0`, but provider-role assignment is enabled by default and calls `/ws/rest/v1/providerrole`. I checked the local `openmrs-module-webservices.rest` repo and that resource appears to be available only from the 3.1.0-tagged line. So either the frontend should require `webservices.rest >=3.1.0`, or `allowAssigningProviderRole` should default to `false` / be feature-gated when that endpoint is unavailable.

I don’t think we need a separate `fhir2` dependency here after checking the tasks module, since `openmrs-module-tasks` itself requires `fhir2`.

ibacher:
Alright, last commit here is somewhat important because it's somewhat changing behaviour, but I think in correct ways.

1. I implemented Dennis's suggestions above. This only changes things in "correct" ways.
2. The task lists was calling `useVisit()` to get the active visit, but we've deprecated the behaviour in the chart where possible so that the visit is only passed once and the "active" visit it available via props in various places. Now the task list gets the "current" visit via these props, consistent with the rest of the chart.
3. (This is the larger change) Previously a visit tagged `NEXT_VISIT` would use a new query to get the patient's most recent visit (including stopped visits). For these purposes, I re-interpreted `NEXT_VISIT` as the visit subsequent to the current visit. For non-retrospective use-cases, this should basically be equivalent, but for retrospective use-cases, this is a behaviour change. FWIW, if we fully build this out to support RDE, I'm not sure that `NEXT_VISIT` makes much sense in that context. RDE should probably use actual due dates.

vasharma05:
Thank you @ibacher and @denniskigen for looking into the PR and making the changes too. I'm really grateful for your efforts, and will push the final changes as described above such that this PR is merged in.
Thanks!

ibacher (approved):

ibacher:
I that case, I'm going to push the button to merge this in. We can patch anything else up.

PR #3207, opened 2026-04-04
(fix) Prevent order action crashes when encounter visit data is missing

aditya123000:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a conventional commit label. See existing PR titles for inspiration.
- [x] My work is based on designs, which are linked or shown either in the Jira ticket or the description below. (See also: Styleguide)

## Summary

This PR prevents order action crashes when order payloads are missing embedded `encounter` or `visit` data.

It updates the medication and shared order basket builders to safely handle missing `encounter.visit`, and disables revise, renew, and discontinue-style actions when the required visit context is unavailable.

Without this change, several code paths guarded `order.encounter?.uuid`, but still directly dereferenced `order.encounter.visit`, which could cause runtime failures in order action flows.

## Screenshots

N/A - this change is in null-safety handling, action guarding, and test coverage. No visible UI change was introduced.

## Related Issue

N/A

## Other

Local test run:

```bash
node .yarn/releases/yarn-4.10.3.cjs jest packages/esm-patient-medications-app/src/api/api.test.ts --runInBand
node .yarn/releases/yarn-4.10.3.cjs jest packages/esm-patient-medications-app/src/components/medications-details-table.test.tsx --runInBand
node .yarn/releases/yarn-4.10.3.cjs jest packages/esm-patient-orders-app/src/utils/index.test.ts --runInBand
node .yarn/releases/yarn-4.10.3.cjs jest packages/esm-patient-orders-app/src/components/order-details-table.test.tsx --runInBand

```

### DIFF
--- a/packages/esm-patient-common-lib/src/orders/types/order.ts
+++ b/packages/esm-patient-common-lib/src/orders/types/order.ts
@@ -70,7 +70,7 @@ export interface OrderBasketItem {
   orderNumber?: string;
   scheduledDate?: Date;
   encounterUuid?: string;
-  visit: Visit;
+  visit: Visit | null;
 }
 
 export type OrderUrgency = 'ROUTINE' | 'STAT' | 'ON_SCHEDULED_DATE';
@@ -142,7 +142,7 @@ export interface Order {
   drug: Drug | null;
   duration: number | null;
   durationUnits: OpenmrsResource | null;
-  encounter: Encounter;
+  encounter: Encounter | null;
   frequency: OpenmrsResource | null;
   instructions?: string | null;
   numRefills: number | null;
@@ -317,19 +317,19 @@ export interface TestOrderBasketItem extends OrderBasketItem {
 }
 
 export interface OrderBasketWindowProps {
-  encounterUuid: string;
+  encounterUuid: string | null;
   onOrderBasketSubmitted?: (encounterUuid: string, postedOrders: Array<Order>) => void;
 }
 
 export interface ExportedOrderBasketWindowProps {
-  encounterUuid: string;
+  encounterUuid: string | null;
   drugOrderWorkspaceName: string;
   labOrderWorkspaceName: string;
   generalOrderWorkspaceName: string;
   patient: fhir.Patient;
   patientUuid: string;
-  visitContext: Visit;
-  mutateVisitContext: () => void;
+  visitContext: Visit | null;
+  mutateVisitContext: (() => void) | null;
   onOrderBasketSubmitted?: (encounterUuid: string, postedOrders: Array<Order>) => void;
   /**
    * An optional array of order type UUIDs to display. If not provided, all panels are shown.
--- a/packages/esm-patient-common-lib/src/orders/types/order.ts
+++ b/packages/esm-patient-common-lib/src/orders/types/order.ts
@@ -70,7 +70,7 @@ export interface OrderBasketItem {
   orderNumber?: string;
   scheduledDate?: Date;
   encounterUuid?: string;
-  visit: Visit;
+  visit: Visit | null;
 }
 
 export type OrderUrgency = 'ROUTINE' | 'STAT' | 'ON_SCHEDULED_DATE';
@@ -142,7 +142,7 @@ export interface Order {
   drug: Drug | null;
   duration: number | null;
   durationUnits: OpenmrsResource | null;
-  encounter: Encounter;
+  encounter: Encounter | null;
   frequency: OpenmrsResource | null;
   instructions?: string | null;
   numRefills: number | null;
@@ -322,12 +322,12 @@ export interface TestOrderBasketItem extends OrderBasketItem {
 }
 
 export interface OrderBasketWindowProps {
-  encounterUuid: string;
+  encounterUuid: string | null;
   onOrderBasketSubmitted?: (encounterUuid: string, postedOrders: Array<Order>) => void;
 }
 
 export interface ExportedOrderBasketWindowProps {
-  encounterUuid: string;
+  encounterUuid: string | null;
   drugOrderWorkspaceName: string;
   labOrderWorkspaceName: string;
   generalOrderWorkspaceName: string;
@@ -339,8 +339,8 @@ export interface ExportedOrderBasketWindowProps {
   allergyFormWorkspaceName?: string;
   patient: fhir.Patient;
   patientUuid: string;
-  visitContext: Visit;
-  mutateVisitContext: () => void;
+  visitContext: Visit | null;
+  mutateVisitContext: (() => void) | null;
   onOrderBasketSubmitted?: (encounterUuid: string, postedOrders: Array<Order>) => void;
   /**
    * An optional array of order type UUIDs to display. If not provided, all panels are shown.
--- a/packages/esm-patient-common-lib/src/orders/types/order.ts
+++ b/packages/esm-patient-common-lib/src/orders/types/order.ts
@@ -142,7 +142,7 @@ export interface Order {
   drug: Drug | null;
   duration: number | null;
   durationUnits: OpenmrsResource | null;
-  encounter: Encounter | null;
+  encounter: Encounter;
   frequency: OpenmrsResource | null;
   instructions?: string | null;
   numRefills: number | null;
--- a/packages/esm-patient-common-lib/src/orders/types/order.ts
+++ b/packages/esm-patient-common-lib/src/orders/types/order.ts
@@ -322,12 +322,12 @@ export interface TestOrderBasketItem extends OrderBasketItem {
 }
 
 export interface OrderBasketWindowProps {
-  encounterUuid: string | null;
+  encounterUuid: string;
   onOrderBasketSubmitted?: (encounterUuid: string, postedOrders: Array<Order>) => void;
 }
 
 export interface ExportedOrderBasketWindowProps {
-  encounterUuid: string | null;
+  encounterUuid: string;
   drugOrderWorkspaceName: string;
   labOrderWorkspaceName: string;
   generalOrderWorkspaceName: string;
--- a/packages/esm-patient-medications-app/src/api/api.ts
+++ b/packages/esm-patient-medications-app/src/api/api.ts
@@ -301,7 +301,7 @@ export function buildMedicationOrder(order: Order, action: OrderAction): DrugOrd
         }
       : null,
     encounterUuid: order.encounter?.uuid,
-    visit: order.encounter.visit,
+    visit: order.encounter?.visit ?? null,
   };
 }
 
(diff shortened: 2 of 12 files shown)


denniskigen (changes requested):
Thanks for the defensive coding approach, @aditya123000. I checked the backend — `Order.encounter` is a `NOT NULL` column in the database with a JPA `nullable = false` annotation, and the REST API always includes it in all representations. So `order.encounter` being null shouldn't happen in practice.

Given that, changing the shared `Order` type to `Encounter | null` would be inaccurate — it would spread null checks to every consumer for a field the backend guarantees is present. The existing code that dereferences `order.encounter.uuid` directly is correct.

The `visit` field on `Encounter` is a different story — that one *is* nullable (encounters can exist without visits). If you've seen crashes related to `order.encounter.visit`, could you share a reproduction case? That would help scope the fix to just the visit nullability rather than the encounter itself.

Good catch on the `mutateVisitContext` eager evaluation in `order-details-table.component.tsx` though — that's a real bug worth fixing separately.

aditya123000:
Thanks for the clarification. I'll update the PR to focus only on the nullable visit case and the mutateVisitContext issue. I'll also try to provide a reproduction case for the visit-related crash.

aditya123000:
@denniskigen Gentle ping on this PR
Would really appreciate a review when you get time.

Josephkagimu1 (changes requested):
Hi @aditya123000 thanks for this PR , are you still working on this ? , do you mind adding a link to the Jira ticket ?
also the encounter: Encounter | null — Dennis flagged this in review (April 7): Order.encounter is NOT NULL in the DB and always returned by the REST API. You replied (June 8) you'd revert it, but it's still Encounter | null in the diff. Please revert encounter to non-null and keep only visit: Visit | null.
Deleted tests — the PR removes all existing prepMedicationOrderPostData and bucketMedicationOrders tests (~300 lines). These cover real submission logic that still exists. Please restore them and add the new null-safety tests alongside, not as replacements.
Also: rebase against main (4 months stale) .

The visit null-safety changes and disabling order actions when visit is missing look good — just need these two fixes.

packages/esm-patient-medications-app/src/api/api.ts
```diff
@@ -296,8 +296,7 @@ export function buildMedicationOrder(order: Order, action: OrderAction): DrugOrd
         }
       : null,
     encounterUuid: order.encounter?.uuid,
-    previousOrderDateActivated: action === 'REVISE' ? order.dateActivated : undefined,
-    visit: order.encounter.visit,
+    visit: order.encounter?.visit ?? null,
```
denniskigen:
```suggestion
    previousOrderDateActivated: action === 'REVISE' ? order.dateActivated : undefined,
    visit: order.encounter?.visit ?? null,
```

Can we keep the revision start-date floor? Without this value, users can choose a revised start date before the order it replaces, and the backend will normally reject the revision. [#3198](https://github.com/openmrs/openmrs-esm-patient-chart/pull/3198) added this value so the form can enforce that minimum.

packages/esm-patient-orders-app/src/utils/index.ts
```diff
@@ -100,8 +100,7 @@ export function buildMedicationOrder(order: Order, action?: OrderAction): DrugOr
         }
       : null,
     encounterUuid: order.encounter?.uuid,
-    previousOrderDateActivated: action === 'REVISE' ? order.dateActivated : undefined,
-    visit: order.encounter.visit,
+    visit: order.encounter?.visit ?? null,
```
denniskigen:
```suggestion
    previousOrderDateActivated: action === 'REVISE' ? order.dateActivated : undefined,
    visit: order.encounter?.visit ?? null,
```

Ditto. The orders app needs to preserve the same revision minimum.

packages/esm-patient-medications-app/src/components/medications-details-table.component.tsx
```diff
@@ -401,50 +401,68 @@ function OrderBasketItemActions({
   const { t } = useTranslation();
   const isTablet = useLayoutType() === 'tablet';
   const alreadyInBasket = items.some((x) => x.uuid === medication.uuid);
+  const encounterUuid = medication.encounter?.uuid ?? null;
+  const visitContext = medication.encounter?.visit ?? null;
+  const visitUuid = visitContext?.uuid;
+  const hasVisitContext = Boolean(encounterUuid && visitUuid);
```
denniskigen: Can we keep these actions available when `visit` is null? On systems with `visits.enabled=false`, order encounters intentionally have no visit, so this disables modify, renew, and discontinue for every order.

Keep the null-safe `visitContext` handling, but let the basket's existing `visitRequired` check decide whether a visit is needed. The test should use `encounter: { ..., visit: null }`, not `encounter: null`, and assert that the actions remain enabled.

packages/esm-patient-orders-app/src/components/order-details-table.component.tsx
```diff
@@ -613,19 +613,27 @@ function OrderBasketItemActions({ orderItem, patient }: OrderBasketItemActionsPr
   const { orders, setOrders } = useOrderBasket<OrderBasketItem>(patient, grouping, postDataPrepFn);
   const alreadyInBasket = orders.some((x) => x.uuid === orderItem.uuid);
   const { mutate: globalMutate } = useSWRConfig();
+  const encounterUuid = orderItem.encounter?.uuid ?? null;
+  const visitContext = orderItem.encounter?.visit ?? null;
+  const visitUuid = visitContext?.uuid;
+  const hasVisitContext = Boolean(encounterUuid && visitUuid);
```
denniskigen: Ditto. This disables modify and cancel for every order on systems with visits disabled. Visit presence should not control whether these actions are available.

denniskigen (changes requested):
Thanks for fixing the nullable visit handling, @aditya123000. I found two workflow regressions that still need addressing.

packages/esm-patient-medications-app/src/components/medications-details-table.component.tsx
```diff
@@ -460,13 +479,13 @@ function OrderBasketItemActions({
           id="modify"
           itemText={t('modify', 'Modify')}
           onClick={handleModifyClick}
-          disabled={alreadyInBasket}
+          disabled={alreadyInBasket || (visitRequired && !visitContext)}
```
denniskigen:
```suggestion
          disabled={alreadyInBasket}
```

Can we keep Modify available here and remove the matching early return? A `REVISE` posts directly against `finalizedOrder.encounterUuid`, so it doesn't require `visitContext`. The test should keep Modify enabled while renew and discontinue remain gated.

packages/esm-patient-orders-app/src/components/order-details-table.component.tsx
```diff
@@ -701,7 +714,7 @@ function OrderBasketItemActions({ orderItem, patient }: OrderBasketItemActionsPr
       <OverflowMenu aria-label={t('actionsMenu', 'Actions menu')} align="left" flipped selectorPrimaryFocus="#modify">
         <OverflowMenuItem
           className={styles.menuItem}
-          disabled={alreadyInBasket}
+          disabled={alreadyInBasket || (visitRequired && !visitContext)}
```
denniskigen:
```suggestion
          disabled={alreadyInBasket}
```

Ditto. Drug, lab, and general revisions all post against the existing encounter. Can we keep Modify enabled and update the test to gate only Cancel?

packages/esm-patient-common-lib/src/orders/types/order.ts
```diff
@@ -322,12 +322,12 @@ export interface TestOrderBasketItem extends OrderBasketItem {
 }
 
 export interface OrderBasketWindowProps {
-  encounterUuid: string;
+  encounterUuid: string | null;
```
denniskigen:
```suggestion
  encounterUuid: string;
```

Can we keep this as `string` here and in `ExportedOrderBasketWindowProps`? The nullable version came from the now-reverted `Order.encounter` change. Every launcher still supplies a string, so widening these published interfaces makes downstream consumers handle a state this fix doesn't produce.

denniskigen (changes requested):
This is very close, @aditya123000. The previous feedback is addressed, but I found two remaining workflow and type issues.

tools/vitest.shared.ts
```diff
@@ -1,3 +1,4 @@
+import { fileURLToPath } from 'node:url';
```
denniskigen: Nit: This is a Windows path fix and looks unrelated to this change. It's probably better to land it in its own PR.

packages/esm-patient-medications-app/src/components/medications-details-table.test.tsx
```diff
@@ -0,0 +1,159 @@
...
+    </button>
+  ));
+  OverflowMenuItem.displayName = 'OverflowMenuItem';
+
+  return {
+    ...originalModule,
+    OverflowMenu,
+    OverflowMenuItem,
+  };
+});
```
denniskigen:
```suggestion
```

Can we remove this mock and exercise Carbon's real overflow menu? It renders every item while the menu is closed, changes `menuitem` elements into buttons, and bypasses the portal, focus, and close behavior. The focused suites pass without it.

packages/esm-patient-orders-app/src/components/order-details-table.test.tsx
```diff
@@ -33,6 +35,31 @@ const mockUseReactToPrint = vi.mocked(useReactToPrint);
...
+    </button>
+  ));
+  OverflowMenuItem.displayName = 'OverflowMenuItem';
+
+  return {
+    ...originalModule,
+    OverflowMenu,
+    OverflowMenuItem,
+  };
+});
```
denniskigen:
```suggestion
```

Ditto. The orders tests can use the real menu too.

denniskigen (changes requested):
Thanks, @aditya123000. This is very close now. The revision date floor is back, and the visit gating is correct for renew, discontinue, and cancel. Modify should remain available because revisions post against the existing encounter. My remaining feedback also covers the public `encounterUuid` type and the Carbon test mock, which bypasses the real overflow-menu behavior.

aditya123000:
Thanks for the thorough re-review — fixed all of these:

- Modify is now always enabled regardless of visit context (only Renew/Discontinue/Cancel are gated by visitRequired), since revisions post against the existing encounter as you noted.

- Reverted encounterUuid back to non-nullable string in both OrderBasketWindowProps and ExportedOrderBasketWindowProps — you're right that the nullable version was a leftover from the reverted Order.encounter change and every launcher already supplies a string.

- Removed the @carbon/react mock in both test files; tests now exercise the real overflow menu (open via the "Actions menu" trigger, then query real menuitem elements) rather than a simplified stand-in.

- Dropped the unrelated Windows-path fix from tools/vitest.shared.ts — will land that separately.

- Rebased onto latest main.

PR #3221, opened 2026-04-11
(feat) O3-5561: Add reference ranges toggletip to vitals header

vivektarun:
## Requirements

## Summary

Implements O3-5561. Adds a reference range panel to the vitals and biometrics header strip.

A single ⓘ icon button appears next to the "Vitals and biometrics" heading. Clicking it opens a Toggletip panel listing normal reference ranges for each configured vital sign concept. The panel is hidden when no reference ranges are available.

Key decisions:
- One tooltip for the whole strip rather than one per card — `NumericObservation` is reused in dense views (e.g. the vitals table) where per-cell icons would clutter the UI.
- `vitals-header-item.component.tsx` was refactored to use `NumericObservation` from `@openmrs/esm-framework`, removing a duplicate custom implementation and its associated stylesheet.

Added 3 unit tests in vitals-header.test.tsx:
- Toggletip button renders when reference ranges are configured
- Toggletip button is hidden when conceptRangeMap is empty
- Panel opens with correct content on click
### Before
image

### After
image
image

### Tests
image

## Related Issue

https://issues.openmrs.org/browse/O3-5561

## Other

- `vitals-header-item.scss` deleted — styling is now fully owned by `NumericObservation`.
- Two new i18n keys added: `referenceRanges` and `viewReferenceRanges`.


packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx
```diff
@@ -119,6 +119,71 @@ const VitalsHeader: React.FC<VitalsHeaderProps> = ({
...
+                </ToggletipButton>
+                <ToggletipContent>
+                  <div className={styles.referenceRangeContent}>
+                    <p className={styles.referenceRangeHeading}>{t('referenceRanges', 'Reference ranges')}</p>
+                    <table className={styles.referenceRangeTable}>
+                      <tbody>
+                        {[
+                          {
+                            label: t('bp', 'BP'),
+                            uuid: config.concepts.systolicBloodPressureUuid,
```
denniskigen: The reference-range panel fetches separate systolic and diastolic BP concepts, but this row only looks up `systolicBloodPressureUuid` and renders it under the combined "BP" label. That means the panel will show e.g. "90–120 mmHg" and silently omit the diastolic normal range (60–80 mmHg in the existing metadata), which makes the blood-pressure reference guidance incomplete/misleading.

Please either render separate systolic/diastolic rows or format both ranges together in the BP row.

packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.test.tsx
```diff
@@ -373,6 +381,56 @@ describe('VitalsHeader', () => {
...
+    } as ReturnType<typeof useVitalsAndBiometrics>);
+
+    renderWithSwr(<VitalsHeader {...testProps} />);
+
+    await waitForLoadingToFinish();
+
+    const toggletipButton = screen.getByRole('button', { name: /view reference ranges/i });
+    await user.click(toggletipButton);
+
+    expect(screen.getByText(/reference ranges/i)).toBeInTheDocument();
```
denniskigen: Nit: this assertion only checks for the heading text ("Reference ranges"), which is present in the DOM regardless of whether any range rows actually rendered. Consider asserting on an actual range value from the mock data (e.g. a specific "60–100" string) to verify the rows are populated correctly.

packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx
```diff
@@ -119,6 +119,71 @@ const VitalsHeader: React.FC<VitalsHeaderProps> = ({
...
+                            uuid: config.concepts.oxygenSaturationUuid,
+                            unit: conceptUnits.get(config.concepts.oxygenSaturationUuid),
+                          },
+                          {
+                            label: t('temperatureAbbreviated', 'Temp'),
+                            uuid: config.concepts.temperatureUuid,
+                            unit: conceptUnits.get(config.concepts.temperatureUuid),
+                          },
+                        ]
+                          .filter(({ uuid }) => conceptRangeMap?.get(uuid))
```
ibacher: These should also filter out any entries where the UUID is `null` or blank.

packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx
```diff
@@ -119,6 +119,71 @@ const VitalsHeader: React.FC<VitalsHeaderProps> = ({
...
+                <ToggletipButton
+                  className={styles.referenceRangeButton}
+                  label={t('viewReferenceRanges', 'View reference ranges')}
+                >
+                  <Information size={16} />
+                </ToggletipButton>
+                <ToggletipContent>
+                  <div className={styles.referenceRangeContent}>
+                    <p className={styles.referenceRangeHeading}>{t('referenceRanges', 'Reference ranges')}</p>
+                    <table className={styles.referenceRangeTable}>
```
ibacher, line 162: I'm not overly sold here on using a table. At the very least, the units should be just represented with the value, like we do everywhere. If we are using a table, the reference range should probably be right-aligned rather than left-aligned.

ibacher (commented):
Minor, but I think the `i` icon should come at the end of the line. It's less important that the freshness of the vitals for the ability to see the full history.

vivektarun:
@ibacher @denniskigen I am on it.

vivektarun:
Addressed all feedback — separated BP rows, improved tests, moved ⓘ to end, added UUID filtering, and updated panel layout.


Attached screenshot

packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx
```diff
@@ -135,6 +135,77 @@ const VitalsHeader: React.FC<VitalsHeaderProps> = ({
                 {t('vitalsHistory', 'Vitals history')}
               </ConfigurableLink>
             )}
+            {conceptRangeMap?.size > 0 && (
+              <Toggletip align="bottom-left">
```
denniskigen: The Toggletip renders a `<span>` wrapper that becomes the flex child of `.headerItems`. Currently `align-self: center` is on the button (`.referenceRangeButton`), but the button is nested inside that `<span>`, so the property has no effect on flex layout.

To align the icon with the other header items, move `align-self: center` to the Toggletip wrapper:

```tsx
<Toggletip className={styles.referenceRangeToggletip}>
```

```scss
.referenceRangeToggletip {
  align-self: center;
}
```

Then remove `align-self: center` from `.referenceRangeButton`.

packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx
```diff
@@ -135,6 +135,77 @@ const VitalsHeader: React.FC<VitalsHeaderProps> = ({
...
+                >
+                  <Information size={16} />
+                </ToggletipButton>
+                <ToggletipContent>
+                  <div className={styles.referenceRangeContent}>
+                    <p className={styles.referenceRangeHeading}>{t('referenceRanges', 'Reference ranges')}</p>
+                    <table className={styles.referenceRangeTable}>
+                      <tbody>
+                        {[
+                          {
```
denniskigen: Nit: "systolic" and "diastolic" render lowercase while the other labels use sentence case ("Heart rate", "R. rate", "SpO2", "Temp"). These should be "Systolic" and "Diastolic" to be consistent.

CleanShot 2026-04-13 at 22 03 00@2x

packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.test.tsx
```diff
@@ -61,6 +63,12 @@ mockUseConfig.mockReturnValue({
 } as ConfigObject);
 
 describe('VitalsHeader', () => {
+  beforeEach(() => {
+    mockUseVitalsConceptMetadata.mockReturnValue(
+      mockVitalsConceptMetadata as ReturnType<typeof useVitalsConceptMetadata>,
```
denniskigen, line 67: The module-level `jest.mock` at line 53 already stubs `useVitalsConceptMetadata` with `mockVitalsConceptMetadata`. This `beforeEach` sets the same return value on the `jest.mocked()` reference. Both paths end up at the same result, but it's confusing to have two sources of truth.

Consider removing the `mockImplementation` for `useVitalsConceptMetadata` from the module-level mock (line 53) and letting this `beforeEach` be the single source of truth. That way the "empty map" test override is clearly the only place that changes behavior.

denniskigen (dismissed):

packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx
```diff
@@ -135,6 +135,77 @@ const VitalsHeader: React.FC<VitalsHeaderProps> = ({
...
+                  className={styles.referenceRangeButton}
+                  label={t('viewReferenceRanges', 'View reference ranges')}
+                >
+                  <Information size={16} />
+                </ToggletipButton>
+                <ToggletipContent>
+                  <div className={styles.referenceRangeContent}>
+                    <p className={styles.referenceRangeHeading}>{t('referenceRanges', 'Reference ranges')}</p>
+                    <table className={styles.referenceRangeTable}>
+                      <tbody>
```
denniskigen, line 163: The panel shows `lowNormal`/`hiNormal` values from concept metadata. OpenMRS distinguishes between normal, critical, and absolute ranges, so "Reference ranges" is broader than what's actually displayed here. "Normal ranges" would be more precise. Same applies to the toggletip button label and the i18n keys (`referenceRanges` / `viewReferenceRanges`).

denniskigen:
Thanks for iterating on the feedback, @vivektarun. Mind posting an updated screenshot?

vivektarun:
I have addressed all the changes and 

> Thanks for iterating on the feedback, @vivektarun. Mind posting an updated screenshot?

sure

vivektarun:
@denniskigen please have a look
image

omsrivastava512:
Or maybe this?

https://github.com/user-attachments/assets/b6d5f963-4dc8-4397-97fe-bf1486a81445

---

Also for BP, this should work:

image

denniskigen:
@omsrivastava512 Thanks for the demos! I still feel that the singular [toggletip](https://react.carbondesignsystem.com/?path=/docs/components-toggletip--overview) (click-triggered on desktop / tap-triggered on tablet) approach in this PR is the right choice here.

Carbon's guidance treats a toggletip as a click/Enter-triggered disclosure for supplemental content, especially around an info icon, while a [definition tooltip](https://react.carbondesignsystem.com/?path=/docs/components-definitiontooltip--overview) is meant for defining a term or giving brief help within text. In this PR, the trigger is an info icon in a header, and the content is a small panel with multiple rows of normal ranges, which is closer to a disclosure/popover than an inline definition. Carbon also explicitly notes that toggletips commonly use an information icon button as the trigger.

I'd only switch to a definition tooltip if we changed the UI to attach brief, read-only explanatory text directly to a term like "Normal ranges" and kept the content very short. For the current "info icon opens a small reference panel" pattern, toggletip is the more semantically correct choice.

@bmamlin, @paulsonder, any thoughts?

packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx
ibacher: On this, it may be slightly nicer if we could render the BP in a unified way (**100-120 / 60-80 mmHg**) since these aren't interpreted separately.

ibacher (approved):

ibacher:
Thanks @vivektarun! I've implemented a combined BP range and slightly tweaked the position of the tool tip to work better with narrower displays.

vivektarun:
Thanks @ibacher  for the review, the feedback, and merging the PR!

vivektarun:
Thanks, @denniskigen, for your feedback

omsrivastava512:
hey guys...our new reference table apparantly **‘no habla español!**’

image

---

image

---

image

omsrivastava512:
Same with French.

image

I've created an issue over this. I can take a quick pass at fixing this. If it turns out to be more involved, we can revisit or @vivektarun can take it up.

PR #3222, opened 2026-04-11
Null fix of session user

Abhisy2506:
### Issue 

https://openmrs.atlassian.net/browse/O3-5605

## Summary

The PrintComponent was attempting to access session.user.display without a null check. Because useSession() is an asynchronous hook, the user object is null during the initial loading state or in cases of network latency/session timeout.

## Fix

Applied optional chaining to the session user object: {session?.user?.display}.
This ensures the component renders safely even if the session data hasn't arrived yet, maintaining consistency with how other components in the repository handle the session state.


ibacher:
Thank you for your contribution. Unfortunately, due to the volume of contributions, we cannot currently accept contributions that do not minimally utilize our PR template or the conventions noted therein.

PR #3225, opened 2026-04-13
(feat) O3-5610: Add result date column to the lab order results table

Omoshlawi:
## Requirements

## Summary

Currently, users can only see the order date for a lab order, not when the result itself was recorded. This PR surfaces the result date alongside each observation.

- Add a **Result date** column to the lab order results table on both desktop (`IndividualResultsTable`) and tablet (`LabSetPanel`) layouts, formatted with `formatDate(..., { mode: 'standard', time: true })` via a new shared `formatResultDate` helper.
- Redistribute column widths to make room for the new column (all three layouts still sum to 100%):
  - Desktop and tablet *with* reference range: `testName` 50% → 36%, `value` 25% → 17%, new `resultDate` 22%, `referenceRange` 25%.
  - Tablet *without* reference range: `testName` 60% → 50%, `value` 40% → 28%, new `resultDate` 22%.
  - `resultDate` is sized at 22% so the rendered `DD-MMM-YYYY, HH:MM AM` format stays on a single line at typical table widths.
- Fix a latent bug on the tablet panel subtitle where the rendered time was always `00:00`. `panel.date` is a date-only string (`YYYY-MM-DD`) produced by `filter-context`, so deriving the time from it never worked. The subtitle now parses the first entry's `obsDatetime` instead.
- Update the affected unit and e2e tests to account for the new column (row accessible-name strings and `td:nth-child` selectors).
- Type `patientUuid` as `string` on `IndividualResultsTable` and return `[]` instead of `0` when there are no entries, so `DataTable` never receives a non-array `rows` prop.

## Screenshots

Screenshot 2026-04-13 at 13 18 28

## Related Issue

https://openmrs.atlassian.net/browse/O3-5610

https://thepalladiumgroup.atlassian.net/browse/KHP3-9086


Omoshlawi:
cc @makombe @donaldkibet @denniskigen

omeriinnocent (commented):
@Omoshlawi ,wish you can explain your description more clearer for your jira ticket description and pr description, though your work is open.

donaldkibet:
@Omoshlawi address the failing build and text

Omoshlawi:
> Currently end users can only see the order date for lab order hence the need for also viewing the result date

Done ✅

packages/esm-patient-tests-app/src/test-results/individual-results-table-tablet/lab-set-panel.component.tsx
```diff
@@ -42,6 +42,7 @@ const LabSetPanel: React.FC<LabSetPanelProps> = ({ panel, activePanel, setActive
               key: 'testName',
               header: t('testName', 'Test name'),
             },
+            { id: 'resultDate', key: 'resultDate', header: t('resultDate', 'Result Date') },
```
denniskigen: "Result Date" is title case but the adjacent header "Test name" is sentence case. This should be "Result date" to match the convention in this file. The translation key default value in `en.json` should also be updated.

packages/esm-patient-tests-app/src/test-results/individual-results-table-tablet/lab-set-panel.component.tsx
```diff
@@ -74,9 +75,13 @@ const LabSetPanel: React.FC<LabSetPanelProps> = ({ panel, activePanel, setActive
         ? panel.entries.map((test) => {
             const units = test.units ?? '';
             const range = formatRangeWithUnits(test.range, units);
+            const resultDate = test.obsDatetime
```
denniskigen: This formatting block appears three times across two files (twice here, once in `individual-results-table.component.tsx`). Consider extracting a small helper to avoid the repetition:

```ts
function formatResultDate(obsDatetime: string) {
  return obsDatetime ? formatDate(parseDate(obsDatetime), { mode: 'standard', time: true }) : '--';
}
```

denniskigen (commented):
Thanks for working on this, @Omoshlawi. Two things below.

denniskigen:
The "Result Date" column wraps the date across three lines because both tables use `table-layout: fixed` with column widths that don't account for the new column.

**`individual-results-table.scss`** — update the column widths:

```scss
.col-testName {
  width: 35%;
}

.col-resultDate {
  width: 20%;
}

.col-value {
  width: 20%;
}

.col-referenceRange {
  width: 25%;
}
```

**`lab-set-panel.scss`** — same for both variants:

```scss
.tableWithRange {
  .col-testName {
    width: 35%;
  }

  .col-resultDate {
    width: 20%;
  }

  .col-value {
    width: 20%;
  }

  .col-referenceRange {
    width: 25%;
  }
}

.tableWithoutRange {
  .col-testName {
    width: 40%;
  }

  .col-resultDate {
    width: 20%;
  }

  .col-value {
    width: 40%;
  }
}
```

The `getColumnClass` function in both components already generates the class name from the header key (`col-resultDate`), so these styles will be picked up automatically.

This would give you this UI:

CleanShot 2026-04-15 at 10 43 24@2x

Omoshlawi:
Thanks @denniskigen 

All addressed ✅

denniskigen:
Nice work on the Result date column, @Omoshlawi! One small bug I noticed in the tablet view (`lab-set-panel.component.tsx`):

`panel.date` is a date-only string like `"2025-10-24"`, so `new Date(panel.date)` on [line 30](https://github.com/openmrs/openmrs-esm-patient-chart/blob/42f71012cfd60ec74bb65b113288af1f4fc29376/packages/esm-patient-tests-app/src/test-results/individual-results-table-tablet/lab-set-panel.component.tsx#L30) gives midnight UTC. The header then calls `getHours()`/`getMinutes()` on [lines 116–120](https://github.com/openmrs/openmrs-esm-patient-chart/blob/42f71012cfd60ec74bb65b113288af1f4fc29376/packages/esm-patient-tests-app/src/test-results/individual-results-table-tablet/lab-set-panel.component.tsx#L116-L121), which converts that midnight UTC to local time, showing a wrong time in the panel header.

Since the obstree endpoint doesn't provide a panel-level datetime (only individual obs have `obsDatetime`), we can grab the first entry's datetime instead. Two changes:

**Line 30** — use `obsDatetime` from the first entry when available:
```suggestion
const date = panel.entries[0]?.obsDatetime ? parseDate(panel.entries[0].obsDatetime) : new Date(panel.date);
```
(`parseDate` is already imported.)

**Lines 116–121** — replace the manual `getHours()`/`getMinutes()` with `formatDate`:
```tsx
{formatDate(date, {
  mode: 'wide',
  time: true,
})}
```

This only affects the tablet lab-set-panel view. The desktop `individual-results-table` view uses `time: false` in the header and is unaffected.

Omoshlawi:
> Nice work on the Result date column, @Omoshlawi! One small bug I noticed in the tablet view (`lab-set-panel.component.tsx`):
> 
> `panel.date` is a date-only string like `"2025-10-24"`, so `new Date(panel.date)` on [line 30](https://github.com/openmrs/openmrs-esm-patient-chart/blob/42f71012cfd60ec74bb65b113288af1f4fc29376/packages/esm-patient-tests-app/src/test-results/individual-results-table-tablet/lab-set-panel.component.tsx#L30) gives midnight UTC. The header then calls `getHours()`/`getMinutes()` on [lines 116–120](https://github.com/openmrs/openmrs-esm-patient-chart/blob/42f71012cfd60ec74bb65b113288af1f4fc29376/packages/esm-patient-tests-app/src/test-results/individual-results-table-tablet/lab-set-panel.component.tsx#L116-L121), which converts that midnight UTC to local time, showing a wrong time in the panel header.
> 
> Since the obstree endpoint doesn't provide a panel-level datetime (only individual obs have `obsDatetime`), we can grab the first entry's datetime instead. Two changes:
> 
> **Line 30** — use `obsDatetime` from the first entry when available:
> 
> ```
> const date = panel.entries[0]?.obsDatetime ? parseDate(panel.entries[0].obsDatetime) : new Date(panel.date);
> ```
> 
> (`parseDate` is already imported.)
> 
> **Lines 116–121** — replace the manual `getHours()`/`getMinutes()` with `formatDate`:
> 
> ```tsx
> {formatDate(date, {
>   mode: 'wide',
>   time: true,
> })}
> ```
> 
> This only affects the tablet lab-set-panel view. The desktop `individual-results-table` view uses `time: false` in the header and is unaffected.

Addressed ✅

denniskigen (approved):
LGTM. Thanks, @Omoshlawi!

PR #3227, opened 2026-04-14
(fix) O3-5444: Show inline warning when adding a duplicate condition

praneeth622:
## Requirements

- [x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [Contributing Guidelines](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines).

## Summary

Addresses [O3-5444](https://openmrs.atlassian.net/browse/O3-5444): The Conditions module currently allows the same condition to be added to a patient's problem list multiple times with no warning. This creates duplicate clinical records.

This PR adds an **inline warning** (using Carbon's `InlineNotification`) that appears when a clinician selects a condition that already exists on the patient's problem list. The warning:

- **Matches on condition concept alone** (not filtered by clinical status), per the design discussion on the JIRA ticket
- **Shows different messages** depending on the existing condition's status:
  - **Active match**: _"Diabetes mellitus is already on this patient's active problem list. Saving will create a duplicate record."_
  - **Inactive match**: _"Diabetes mellitus was previously recorded but is now inactive. You may want to edit the existing record to reactivate it instead."_
- **Is non-blocking** — the clinician can still save if they have a valid clinical reason (e.g., recurrent conditions with different onset dates)
- **Disappears** when the search input is cleared

This approach follows the inline warning pattern already used in the allergies module (`allergy-form.workspace.tsx`) and aligns with the design feedback from @ibacher and @burkemamlin on the JIRA ticket.

### Why this approach

A previous attempt ([PR #3055](https://github.com/openmrs/openmrs-esm-patient-chart/pull/3055)) was declined because it used a blocking confirmation modal and lacked test coverage. This PR addresses both concerns:
- Uses an **inline warning** instead of a modal, as recommended by @ibacher
- Matches on **concept alone** (not concept + status), as recommended by @burkemamlin
- Includes **5 new test cases** covering all scenarios

### Before (no warning — duplicate saved silently)

image

### After (with fix — warning shown)

image

## Related Issue

https://openmrs.atlassian.net/browse/O3-5444

## Other

N/A

[O3-5444]: https://openmrs.atlassian.net/browse/O3-5444?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


praneeth622:
Hi @denniskigen @ibacher, this PR addresses the duplicate condition warning discussed in O3-5444.

I reviewed the prior attempt in #3055 and the JIRA discussion before implementing. Key decisions based on your feedback:

Inline warning instead of a blocking modal (per @ibacher’s recommendation)
Concept-only matching, warns regardless of clinical status (per @burkemamlin’s recommendation)
Different messages for active vs inactive matches, with the inactive case suggesting reactivation
Non-blocking, save remains enabled for legitimate clinical use cases (e.g., recurrent conditions)
The pattern follows the existing InlineNotification usage in the allergies form. All existing tests pass and 5 new test cases are included.

@VeronicaMuthee  would appreciate your verification as well since you reported the issue. Let me know if any adjustments are needed.

packages/esm-patient-conditions-app/src/conditions/conditions-widget.component.tsx
```diff
@@ -265,6 +276,32 @@ const ConditionsWidget: React.FC<ConditionsWidgetProps> = ({
...
+                  title={
+                    isActiveDuplicate
+                      ? t('duplicateActiveConditionTitle', 'This condition is already active')
+                      : t('duplicateInactiveConditionTitle', 'This condition was previously recorded')
+                  }
+                  subtitle={
+                    isActiveDuplicate
+                      ? t(
+                          'duplicateActiveConditionSubtitle',
+                          "{{conditionName}} is already on this patient's active problem list. Saving will create a duplicate record.",
```
ibacher:
```suggestion
                          "{{conditionName}} is already on this patient's active problem list. Saving will create a two records for the same condition.",
```

packages/esm-patient-conditions-app/src/conditions/conditions-widget.component.tsx
```diff
@@ -265,6 +276,32 @@ const ConditionsWidget: React.FC<ConditionsWidgetProps> = ({
...
+                          { conditionName: selectedCondition.display },
+                        )
+                      : t(
+                          'duplicateInactiveConditionSubtitle',
+                          '{{conditionName}} was previously recorded but is now inactive. You may want to edit the existing record to reactivate it instead.',
```
ibacher:
```suggestion
                          '{{conditionName}} was previously recorded but is now inactive. You may want to edit the existing record instead.',
```

ibacher (commented):
Implementation seems fine, although I want to be careful with the wording here. I'm not completely sure about my suggestions either.

One improvement I think we should do is removing the title. We want this to be non-interrupting and informative and the bold text both adds little value in this context over the body text and may make this unnecessarily prominent.

praneeth622:
@ibacher , applied your wording suggestions verbatim for both the active and inactive cases. Also addressed Copilot's real concern: conditions.find() returned whichever duplicate was first in the array, so an existing active record could get masked by an earlier inactive entry. Now uses filter + priority pick (active wins). Added a test for the active+inactive coexistence case.

Ready for another look when you have a moment.

denniskigen (commented):
Thanks for iterating on this, @praneeth622. The approach is solid: a non-blocking inline notification that matches the allergies pattern, matching on concept alone (per Burke's feedback), and genuinely thorough test coverage across active/inactive/no-match/clear/both-duplicates/save-still-works. Nice to see the previous round of feedback fully addressed.

A few small inline comments below. The typo on the active subtitle default string is the one worth catching before merge.

packages/esm-patient-conditions-app/src/conditions/conditions-widget.component.tsx
```diff
@@ -265,6 +276,27 @@ const ConditionsWidget: React.FC<ConditionsWidgetProps> = ({
...
+                <InlineNotification
+                  hideCloseButton
+                  kind="warning"
+                  lowContrast
+                  className={styles.duplicateWarning}
+                  subtitle={
+                    isActiveDuplicate
+                      ? t(
+                          'duplicateActiveConditionSubtitle',
+                          "{{conditionName}} is already on this patient's active problem list. Saving will create a two records for the same condition.",
```
denniskigen: Two small things on this line:

1. **Typo**: `create a two records` should be `create two records`. 
2. **Drift from `en.json`**: the default here (`a two records`) and the value in `translations/en.json` line 24 (`two records`) don't match.

```suggestion
                          "{{conditionName}} is already on this patient's active problem list. Saving will create a duplicate record.",
```

packages/esm-patient-conditions-app/src/conditions/conditions-widget.component.tsx
```diff
@@ -265,6 +276,27 @@ const ConditionsWidget: React.FC<ConditionsWidgetProps> = ({
...
+                          { conditionName: selectedCondition.display },
+                        )
+                      : t(
+                          'duplicateInactiveConditionSubtitle',
+                          '{{conditionName}} was previously recorded but is now inactive. You may want to edit the existing record instead.',
```
denniskigen: The PR description says this message should read `...edit the existing record to reactivate it instead.` The `to reactivate it` phrase makes the suggested action more concrete for clinicians (who may otherwise wonder what editing achieves). Consider adding it back (and updating `en.json` to match).

packages/esm-patient-conditions-app/src/conditions/conditions-widget.component.tsx
```diff
@@ -87,6 +88,16 @@ const ConditionsWidget: React.FC<ConditionsWidgetProps> = ({
   const debouncedSearchTerm = useDebounce(searchTerm);
   const { searchResults, isSearching } = useConditionsSearch(debouncedSearchTerm);
 
+  const duplicateCondition = useMemo(() => {
+    if (!selectedCondition || !conditions?.length) {
+      return null;
+    }
+    const matches = conditions.filter((c) => c.conceptId === selectedCondition.uuid);
+    return matches.find((c) => c.clinicalStatus.toLowerCase() === 'active') ?? matches[0] ?? null;
+  }, [selectedCondition, conditions]);
```
denniskigen, line 102: Small suggestion: you can drop the intermediate `matches` array and express the same intent as two `find` calls:

```tsx
const duplicateCondition = useMemo(() => {
  if (!selectedCondition || !conditions?.length) {
    return null;
  }
  const uuid = selectedCondition.uuid;
  return (
    conditions.find((c) => c.conceptId === uuid && c.clinicalStatus?.toLowerCase() === 'active') ??
    conditions.find((c) => c.conceptId === uuid) ??
    null
  );
}, [selectedCondition, conditions]);
```

Not a blocker, just a touch cleaner.

packages/esm-patient-conditions-app/src/conditions/conditions-widget.component.tsx
```diff
@@ -87,6 +88,16 @@ const ConditionsWidget: React.FC<ConditionsWidgetProps> = ({
...
+
+  const isActiveDuplicate = duplicateCondition?.clinicalStatus?.toLowerCase() === 'active';
```
denniskigen: Minor consistency nit: line 96 uses `c.clinicalStatus.toLowerCase()` (no `?.`) while this line uses `duplicateCondition?.clinicalStatus?.toLowerCase()`. The `Condition` type declares `clinicalStatus: string` (non-optional), so either works, but picking one style avoids implying the field is sometimes nullable when it isn't.

packages/esm-patient-conditions-app/src/conditions/conditions-form.test.tsx
```diff
@@ -272,3 +281,255 @@ describe('Conditions form', () => {
...
+    });
+
+    renderConditionsForm();
+
+    const conditionSearchInput = screen.getByRole('searchbox', { name: /enter condition/i });
+    await user.type(conditionSearchInput, 'Hypertension');
+    await user.click(screen.getByRole('menuitem', { name: /hypertension/i }));
+
+    expect(screen.getByText(/already on this patient.*active problem list/i)).toBeInTheDocument();
+    expect(screen.getByText(/hypertension is already on this patient's active problem list/i)).toBeInTheDocument();
```
denniskigen: Nit: these two assertions check effectively the same text, with the second being strictly more specific than the first. You can drop the first one.

praneeth622:
Thanks @denniskigen @ibacher, pushed 18b9a33:

1. Typo fixed (`create a two records` → `create two records`); component now matches `en.json`.
2. Restored `to reactivate it` on the inactive warning + updated the matching test regex.
3. Memo simplified to two chained `find` calls per @denniskigen's suggestion, much cleaner.
4. Dropped the stray `?.` so both sites use the same style (`clinicalStatus` is non-optional).
5. Removed the redundant `.*active problem list` assertion, kept the stricter one.
6. Re the title @ibacher  already dropped last round, notification is body-only now.

All 31 tests pass locally. Ready for another look.

denniskigen (approved):
Thanks, @praneeth622! Could you give it a final pass, @ibacher?

denniskigen:
I took the liberty of pushing one small follow-up commit (6ac8db8a) for copy polish:

- Added `title="Possible duplicate"` to the `InlineNotification` so the warning has a short title above the body, instead of being subtitle-only. 
- Tightened the active-duplicate subtitle. "Saving will create two records for the same condition" was effectively the definition of "duplicate", so it's now just "Saving will create a duplicate."
- Tightened the inactive-duplicate subtitle: "but" became "and" (the two facts are sequential, not contrastive), and "You may want to edit the existing record to reactivate it instead" became "Consider reactivating the existing record instead of creating a new one."
- Updated the matching test assertions.

Screenshot showing the latest changes:

CleanShot 2026-04-30 at 12 51 28@2x

denniskigen (approved):
Thanks, @praneeth622!

PR #3230, opened 2026-04-14
(fix) O3-3230: Null-safe patient identifier handling

GouravPartap7:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a conventional commit

## Summary
- Adds safer handling for incomplete identifier data.
- Prevents runtime errors when identifier metadata is missing.

## Screenshots
- N/A (no UI changes)

## Related Issue
- Fixes #3230

## Other
- No breaking changes


GouravPartap7:
Hi @openmrs maintainers 👋  
All checks are passing now and branch is updated.  
Kindly review this PR when possible. Thanks!

praneeth622 (changes requested):
hi, thanks for the fix. i checked out the branch, read through the diff, and ran the tests for the four packages. the new tests you added all pass.

i have one concern before this can go in.

there is another place in the code that has the same identifier filter and map. it is in packages/esm-patient-tests-app/src/test-results/print-modal/print-modal.extension.tsx around lines 88 to 98. it uses patient.identifier.filter on the same exclude list and then maps the values. the file already uses optional chaining so it is not crashing today, but the whole point of moving this into a helper was to stop having the same logic in many places. if we leave this one behind then the next time someone changes the logic they will have to remember to update both spots. please update this file to call getIncludedPatientIdentifierValues too.

small nit: in order-details-table.component.tsx you placed getIncludedPatientIdentifierValues before getDrugOrderByUuid in the import block. the rest of the block is sorted alphabetically so getDrug should come first. lint will likely flag this on the next ci run.

praneeth622:
one more thing i want to flag. the new helper in patient-details.ts calls identifier.value.trim() and then pushes the trimmed value into the returned array. the old code in the three files you changed did not do this. it passed the raw value straight through.

this has two effects that were not in the original code.

first, whitespace only values are now silently dropped. the old code would include them in the list. the test 'filters out identifiers with empty values' locks this new behavior in but the pr description does not mention it.

second, the value that gets returned to the ui is the trimmed value, not the stored value. if a backend sends an identifier with padding on purpose, or if there is any downstream code that compares these values to what is stored in the server, things can silently go out of sync.

the ticket title says null safe handling so i was expecting the change to be only about not crashing on missing metadata. if you want to also filter empty values and trim, that is fine but please call it out in the pr body so reviewers can decide. otherwise the safer move is to drop the trim and the empty filter, and keep this pr focused on null safety only.

also a small question on the tests. in vitals-overview.test.tsx you added a jest.mock for the whole esm-framework module at the top of the file just to replace age. this runs for every test in the file, not just the new one. none of the existing tests assert on age so it works today, but a mock with that broad a scope makes it easy for someone later to miss that age is mocked. was there a specific failure you were trying to avoid? if not, please move it inside the single test that needs it using jest.spyOn, or drop it since age with the real mockPatient.birthDate works fine.

denniskigen:
Hi @GouravPartap7, thanks for sending this in. Before going deeper on the code, a quick check on scope.

The PR title references O3-3230, but that Jira ticket is "Develop Rule Builder Component", a cancelled 2024 form-builder ticket unrelated to patient identifiers. The `Fixes #3230` in the body also self-references this PR rather than an actual GitHub issue.

Could you share the ticket or issue that motivated this change? Is there:
- A Jira ticket on openmrs.atlassian.net describing the bug?
- A GitHub issue?

The crash is real in principle (per FHIR spec, `identifier.type.coding[0].code` can throw when type metadata is missing), but standard OpenMRS-backed deployments usually fill those fields, so it helps to know where this was actually observed. It also helps us scope the fix appropriately.

If there's no existing ticket, the OpenMRS flow is to file one on Jira first and claim it before opening the PR. Happy to help you walk through that if it's your first time contributing :)

PR #3240, opened 2026-04-17
(fix) O3-5618: Validate condition selection before submit triggers

omsrivastava512:
## Requirements

## Summary

The conditions form allowed users to click **Save & close** after typing arbitrary text in the condition search box, without actually selecting a concept from the results. This would trigger form submission, briefly show a loading spinner, and then silently bail, leaving the form stuck.

### Root Cause

The `conditionName` field served a dual purpose: it stored the raw search text *and* was the field Zod validated for "did the user select a condition?". Since Zod only checked whether the string was non-empty, any typed text, even garbage passed validation. The actual concept selection was tracked in a separate local `selectedCondition` state that Zod had no visibility into.

### Fix

Separated the two concerns into two distinct fields:

- **`conditionName`**: continues to hold the raw search input text (drives the search box display). No validation rule.
- **`conditionUuid`**: a new hidden field that holds the UUID of the selected concept. This is what Zod validates. It is only set when the user explicitly clicks a result from the dropdown.

### Changes

**`conditions-form.workspace.tsx`**
- Replaced `conditionName` with  `conditionUuid` in the Zod schema
- Relaxed `conditionName` to a plain `z.string()` with no validation rule
- Added `conditionUuid: ''` to `defaultValues`
- Changed `mode` from `'all'` to `'onSubmit'`: validation now runs on submit, not on every keystroke, which is consistent with the form's intent

**`conditions-widget.component.tsx`**
- Added `setValue` to the `useFormContext` destructure
- `handleConditionChange`: on concept click, calls `setValue('conditionUuid', selectedCondition.uuid, { shouldValidate: true })` to write the UUID and immediately clear the error if one was showing
- Search `onChange`: calls `setValue('conditionUuid', '')` on every keystroke to deselect any prior concept pick
- Search `onClear`: similarly clears `conditionUuid`
- All error UI references (`className`, `aria-labelledby`, `renderIcon`, error message paragraph) updated from `errors?.conditionName` → `errors?.conditionUuid`
- Added `setIsSubmittingForm(false)` in the `handleCreate` early-return guard as a safety net for any edge case where the function is reached without a valid selection

## Screenshots

Earlier:

https://github.com/user-attachments/assets/33413c0d-7c6f-497c-9b42-c2b447e64596

Now:

https://github.com/user-attachments/assets/dd5b354d-2a84-4578-8e7e-1fdf58a3cc55

## Related Issue

https://openmrs.atlassian.net/browse/O3-5618


denniskigen (approved):
Thanks, @omsrivastava512. This LGTM. I've taken the liberty of adding a regression test for the specific bug scenario this PR fixes.

PR #3242, opened 2026-04-17
(fix) O3-5619: Fix reference range tooltip content overflow

omsrivastava512:
## Requirements

## Summary

The reference ranges toggletip was overflowing its container in locales with longer label translations (e.g. French "Fréquence cardiaque").

image

### Root Cause

The limitation is with the Carbon `ToggletipContent` shell capping the width from the outside as it had a hardcoded max-width that when the content is wider than, it bleeds out.

### Changes

- Replaced `<table>` with a CSS Grid layout for consistent column alignment
- Adjusted CSS padding and alignment to prevent overflow

### Shortcoming Acknowledged

The tooltip content may still overflow and appear slightly misaligned in longer labels.

image

### Alternate Hack

The above shortcoming can be fixed by overriding the Carbon's shell CSS and unsetting the max-width.

## Screenshots
image
|
image
|
image

## Related Issue

https://issues.openmrs.org/browse/O3-5619

## Other

Issue introduced by #3221 

You can find other approaches discussed in the [Slack Discussion](https://openmrs.slack.com/archives/CKS32D55G/p1776446217327599?thread_ts=1774608750.600939&cid=CKS32D55G).


denniskigen (commented):
Thanks for the thorough writeup, the before/after screenshots and the Slack link made the context easy to follow. The grid refactor is a reasonable direction, and the acknowledgment of the remaining shortcoming is appreciated.

That said, after digging into Carbon's `Toggletip` internals, I think the patch doesn't reach the element that's actually causing the overflow, which would explain why long translations still bleed out. Details in comments 1 and 2 below, which I think are the core items to address before this lands.

packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx
```diff
@@ -156,55 +156,51 @@ const VitalsHeader: React.FC<VitalsHeaderProps> = ({
...
-                                <td className={styles.referenceRangeLabel}>{label}</td>
-                                <td className={styles.referenceRangeValue}>
-                                  {unit ? `${rangeValue} ${unit}` : rangeValue}
-                                </td>
-                              </tr>
-                            );
-                          })}
-                      </tbody>
-                    </table>
+                <ToggletipContent className={styles.referenceRangeContent}>
```
denniskigen, line 159: **Class is applied outside the Carbon width cap**

`ToggletipContent` forwards `className` to the outer `PopoverContent`, but Carbon then wraps children in an inner `<div class="cds--toggletip-content">` that carries `max-inline-size: 18rem` and `padding: $spacing-05` (see `@carbon/styles/scss/components/toggletip/_toggletip.scss`). The styles added here land on the outer element, so the inner cap that's actually constraining the content is untouched, and the new `padding-inline` stacks on top of Carbon's existing padding.

One option is to target the nested Carbon element from `referenceRangeToggletip`, something like:

```scss
.referenceRangeToggletip :global(.cds--toggletip-content) {
  max-inline-size: unset; // or a larger value that fits your longest translated label
  padding: layout.$spacing-05; // keep or adjust, but don't double it
}
```

With that, you can drop the new `padding-inline` on `.referenceRangeContent` and the `min-width` becomes optional.

packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.scss
```diff
@@ -157,20 +158,25 @@
...
   color: colors.$gray-30;
   white-space: nowrap;
 }
 
 .referenceRangeValue {
   @include type.type-style('body-compact-01');
-  padding: layout.$spacing-02 0 layout.$spacing-02 layout.$spacing-05;
   font-weight: 600;
   white-space: nowrap;
   text-align: right;
```
denniskigen, line 190: **Grid still cannot shrink long translations**

The new row uses `grid-template-columns: auto 1fr` with `white-space: nowrap` on both columns. A translated label with wide `max-content` (like "Fréquence cardiaque") forces the row wider than the container instead of wrapping, which is the same class of overflow this PR aims to fix.

Consider making the label track shrinkable and wrappable while keeping the numeric value intact:

```scss
.referenceRangeRow {
  grid-template-columns: minmax(0, 1fr) max-content;
}

.referenceRangeLabel {
  white-space: normal;
}

// .referenceRangeValue keeps white-space: nowrap
```

packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx
```diff
@@ -156,55 +156,51 @@ const VitalsHeader: React.FC<VitalsHeaderProps> = ({
...
+                  <p className={styles.referenceRangeHeading}>{t('normalRanges', 'Normal ranges')}</p>
+                  <div className={styles.referenceRangeTable}>
```
denniskigen: Nit (non-blocking): replacing the `<table>` with `<div>/<span>` loses the label-value semantics the previous markup offered. A description list would be a natural fit and still works with the grid CSS:

```tsx
<dl className={styles.referenceRangeTable}>
  <div className={styles.referenceRangeRow}>
    <dt className={styles.referenceRangeLabel}>{t('heartRate', 'Heart rate')}</dt>
    <dd className={styles.referenceRangeValue}>{rangeValue}</dd>
  </div>
  ...
</dl>
```

Happy to take this up separately if you'd prefer to keep this PR scoped.

packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.scss
```diff
@@ -157,20 +158,25 @@
 }
 
 .referenceRangeTable {
```
denniskigen: Nit (non-blocking): `referenceRangeTable` is no longer a table. Something like `referenceRangeList` would read more accurately now.

packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx
```diff
@@ -121,7 +121,7 @@ const VitalsHeader: React.FC<VitalsHeaderProps> = ({
           ? `≥ ${range.lowNormal}`
           : range.hiNormal != null
             ? `≤ ${range.hiNormal}`
-            : '—';
+            : '-';
```
denniskigen: Was this intentional? It looks unrelated to the overflow fix, and it's now inconsistent with line 119 (`${range.lowNormal}–${range.hiNormal}`) which uses an en dash.

packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx
omsrivastava512: Yes, this is exactly what I meant in the "Alternate Hack". However it's somewhat fragile if the classname changes, isn't it? 

Although I was thinking about smth like this:

```css
 .referenceRangeToggletip {
 :global(.cds--toggletip-content) {
    max-width: unset;
    width: max-content;
  }
 }
 ```

Comment (denniskigen) [reply packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx:159]: Fair call on fragility. Two notes:

1. Targeting `cds--*` classnames via `:global()` is a well-established pattern in OpenMRS frontend modules (around 90 occurrences across the patient-chart packages alone), and the names are generally stable within a Carbon major version. A short comment above the override explaining what it targets helps if anything does change later.

2. On `width: max-content`: that works, but it's orthogonal to comment 2. Without `white-space: normal` on the label, a very long translation grows the toggletip horizontally rather than wrapping. If you're OK with that tradeoff, your snippet is a legitimate fix. If you'd rather cap the width and have the label wrap, raising (rather than unsetting) `max-inline-size` combined with the shrinkable-label grid from comment 2 achieves that.

Either direction works, pick whichever tradeoff feels right here.

File: packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx (hunk 5 above)
Comment (omsrivastava512) [reply packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx]: Yes, I intended to add two hyphens instead of one, I missed that. Unrelated but a tiny change so I thought we could push.

File: packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx (hunk 3 above)
Comment (omsrivastava512) [reply packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx]: Sure, this seems right. I think we can do this in this PR itself. Let me make changes.

File: packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.scss (hunk 4 above)
Comment (omsrivastava512) [reply packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.scss]: Acknowledged.

File: packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx (hunk 1 above)
Comment (omsrivastava512) [reply packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx:159]: I'll have to look into this, I think with wraps, the label and the values could look misaligned. Let me implement both and come back here.

Comment (omsrivastava512) [reply packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx:159]: So, I reduced some padding of the `.referenceRangeContent` and added wrap to the label class, I think that alone works works well for the longest translated label. 

image

-----

This is how it looks on a even longer hardcoded string.

image

-----

Do you still think we should touch the Carbon shell, change the max-width? or the wrap looks fine?

Comment (omsrivastava512) [reply packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx:159]:
```css
:global(.cds--toggletip-content) {
    max-inline-size: 24rem;
  }
```

image

packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.extension.tsx
```diff
@@ -121,7 +121,7 @@ const VitalsHeader: React.FC<VitalsHeaderProps> = ({
           ? `≥ ${range.lowNormal}`
           : range.hiNormal != null
             ? `≤ ${range.hiNormal}`
-            : '—';
+            : '––';
```
denniskigen: `'––'` is two en dashes (U+2013 U+2013). The original was `'—'` (U+2014, em dash), and line 119 uses a single en dash for the low–high range. Two en dashes is not a standard typographic form. Would you mind reverting this to `'—'`?

denniskigen (approved):
LGTM. Thanks, @omsrivastava512!

omsrivastava512:
> LGTM. Thanks, @omsrivastava512!

um...wait, the dashes still don't look the same.

omsrivastava512:
Yes, now it's consistent. Apologies for the numerous commits. I'm assuming they get squashed. 😓

denniskigen:
> Yes, now it's consistent. Apologies for the numerous commits. I'm assuming they get squashed. 😓

They do, so it's all good.

PR #3254, opened 2026-04-22
(fix) O3-5614: Fix delete-condition retry and conditions search input

omeriinnocent:
## Requirements

## Summary
**Delete Condition Modal improvements:**

* Added an early return and error logging if `patientUuid` is missing, preventing unintended API calls and making debugging easier.
* Ensured the `mutate` function is always called with a string (defaulting to `''` if `patientUuid` is undefined) to avoid type errors.
* Improved error handling by properly typing the error object for the snackbar subtitle and ensuring the `isDeleting` state is reset in a `finally` block, preventing the delete button from staying disabled after an error.

**Conditions Widget improvements:**

* Added a type parameter (`<HTMLInputElement>`) to `useRef` for `searchInputRef` to improve type safety.
* Simplified the logic for the search input's `value` prop, ensuring it always receives a string value and reducing potential UI inconsistencies.

**Test updates:**

* Updated the delete condition modal test to expect the delete button to be enabled after an error, matching the new behavior.
## Screenshots
delete condition modal screenshort


packages/esm-patient-conditions-app/src/conditions/conditions-widget.component.tsx
```diff
@@ -247,6 +247,7 @@ const ConditionsWidget: React.FC<ConditionsWidgetProps> = ({
                         if (debouncedSearchTerm) {
                           return value;
                         }
+                        return '';
```
denniskigen: This regresses the search input. The IIFE is still gated on `debouncedSearchTerm`, but with `return ''` the Search is now controlled to `''` during the debounce window. On each keystroke, `searchTerm` and the RHF `value` update synchronously while `debouncedSearchTerm` stays empty for ~300ms (and the timer resets on every new keystroke), so this branch returns `''` and React clears the input. Fast typing can leave only the last character visible.

The input should be bound to the live RHF value; only the API lookup via `useConditionsSearch(debouncedSearchTerm)` should use the debounced term:

```tsx
value={selectedCondition?.display ?? value ?? ''}
```

That also avoids the uncontrolled→controlled flip the original `undefined` fall-through caused, which I'm guessing prompted this change.

denniskigen (changes requested):
Thanks for the fix, @omeriinnocent! One concern on the Search `value` change that I think needs addressing before merge.

omeriinnocent:
Hi @denniskigen , have Fixed the search input regression by binding it to the live RHF value instead of the debounced term. This keeps typing responsive during the debounce window, while useConditionsSearch(debouncedSearchTerm) remains the only debounced part. Also verified with the targeted conditions form tests.

packages/esm-patient-conditions-app/src/conditions/conditions-widget.component.tsx
```diff
@@ -248,15 +248,8 @@ const ConditionsWidget: React.FC<ConditionsWidgetProps> = ({
...
-                      renderIcon={errors?.conditionUuid && ((props) => <WarningFilled fill="red" {...props} />)}
-                      value={(() => {
-                        if (selectedCondition) {
-                          return selectedCondition.display;
-                        }
-                        if (debouncedSearchTerm) {
-                          return value;
-                        }
-                      })()}
+                      renderIcon={errors?.conditionName && ((props) => <WarningFilled fill="red" {...props} />)}
```
denniskigen: This looks like a stray revert. Everywhere else in this file uses `errors?.conditionUuid`. The zod schema only validates `conditionUuid` (`conditionName` has no required rule), so `errors.conditionName` never populates and the red warning icon won't render on an invalid submit. #3240 switched these over for that reason.

```suggestion
                      renderIcon={errors?.conditionUuid && ((props) => <WarningFilled fill="red" {...props} />)}
```

denniskigen (commented):
One more finding to address:

omeriinnocent:
Hi @denniskigen , i have addressed the changes

denniskigen (approved):
Looks good to me. Thanks for the iteration, @omeriinnocent!

omeriinnocent:
humbled, thanks much @denniskigen .

PR #3258, opened 2026-04-22
fix(attachments): show Add button as icon-only on tablet and mobile

moconnell9:
## Requirements
- [x] My code follows the guidelines for [contributing to this project](https://o3-docs.openmrs.org/docs/frontend-modules/contributing)
- [x] I have performed a self-review of my own code

## Summary

Two bugs fixed in the Attachments widget:

**1. Upload modal buttons missing on mobile/tablet**  
The Cancel and "Add attachment" submit buttons in the upload modal 
(`file-review.component.tsx`) were wrapped in `<UserHasAccess privilege="Create Attachments">`. 
Because `UserHasAccess` renders null on its first render (before the session 
observable emits), the buttons were invisible on mobile/tablet devices where 
slower networks mean the session hasn't loaded by the time the modal opens. 
A user who has already passed the entry-point privilege check to open the 
modal should never have their in-modal controls hidden.

Fix: removed `UserHasAccess` from inside the modal footer. The privilege 
check is now correctly placed at the two entry points that open the modal 
(the Add button in the CardHeader and the EmptyState launch button).

**2. Add button not visible on tablet/phone viewports**  
The CardHeader Add button was always rendered as text+icon regardless of 
screen size, causing it to be clipped or missing on smaller viewports.

Fix: use `isDesktop(useLayoutType())` to render the button as icon-only 
(`hasIconOnly`) on non-desktop viewports, matching the Carbon Design System 
pattern used elsewhere in the app.

## Related Issue

N/A

## Other

The root cause of the modal button bug (first-render null in `UserHasAccess`) 
is fixed upstream in a companion PR against `openmrs-esm-core`. This PR 
fixes the symptom by moving the privilege gate to the correct location 
regardless.

### DIFF
--- a/packages/esm-patient-attachments-app/src/attachments/attachments-overview.component.tsx
+++ b/packages/esm-patient-attachments-app/src/attachments/attachments-overview.component.tsx
@@ -134,11 +134,13 @@ const AttachmentsOverview: React.FC<AttachmentsOverviewProps> = ({ patientUuid }
 
   if (!attachments.length) {
     return (
-      <EmptyState
-        displayText={t('attachmentsInLowerCase', 'attachments')}
-        headerTitle={t('attachmentsInProperFormat', 'Attachments')}
-        launchForm={showAddAttachmentModal}
-      />
+      <UserHasAccess privilege="Create Attachments">
+        <EmptyState
+          displayText={t('attachmentsInLowerCase', 'attachments')}
+          headerTitle={t('attachmentsInProperFormat', 'Attachments')}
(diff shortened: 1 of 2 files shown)


Josephkagimu1 (commented):
Hei @moconnell9 , thanks for your effort at this PR , but the 'Title' format isn't that needed , 
Also include the ticket link under 'Related issue' ,
Read through this guide , may be helpful : https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477199/Pull+Request+Tips

PR #3261, opened 2026-04-23
(feat) O3-5659: Add ability to group order encounter type by order type

donaldkibet:
## Requirements

## Summary
This PR adds ability to create encounters for every test order if configured. Defaults back to initial workflow if not configure.

See https://openmrs.atlassian.net/browse/O3-5659
## Screenshots
https://github.com/user-attachments/assets/01cb6499-46a2-48e2-be0e-aeb32ef6732b


makombe:
@donaldkibet add a jira ticket for this. Thanks

donaldkibet:
ping @ibacher @denniskigen

packages/esm-patient-orders-app/src/config-schema.ts
```diff
@@ -68,6 +68,24 @@ export const configSchema = {
...
       'Controls whether users can add extra tests while entering lab results in the test-results workspace.',
   },
+  orderTypeEncounterTypeMap: {
+    _type: Type.Array,
+    _elements: {
+      _type: Type.Object,
+      orderBasketGrouping: {
+        _type: Type.String,
+        _description:
+          'The order basket store grouping key. For general order types this is the orderTypeUuid. For drug orders this is "medications".',
```
chibongho, line 78: I'm not that familiar with this code. Why should drug orders have a special key when it has its own orderType?

This also makes it if we want to place both a drug order and a test order (or some other order type) at the same time, they will be recorded as separate encounters. Is that what we want?  @ibacher @denniskigen can you chime in?

donaldkibet: Per the requirement: each order type maps to its own corresponding encounter type (e.g., drug orders → Drug encounter type). This is an opt-in feature; the default behavior — a single shared encounter type across all order types — remains unchanged.

packages/esm-patient-common-lib/src/orders/postOrders.ts
```diff
@@ -62,6 +101,77 @@ export async function postOrdersOnNewEncounter(
...
+        encounterType: encounterTypeUuid,
+        ...(encounterDate ? { encounterDatetime: encounterDate } : {}),
+        visit: currentVisit?.uuid,
+        obs: [],
+        orders,
+      };
+      return postEncounter(encounterPostData, abortController);
+    });
+
+  return Promise.all(encounterPromises);
```
denniskigen, line 172: When `orderTypeEncounterTypeMap` is configured and a basket spans multiple groupings, this fires one `postEncounter` per type concurrently and rejects on the first failure. The `catch` in `OrderBasket.handleSave` only shows a generic error and leaves the basket intact, and the basket items carry no "already posted" marker, so the user's natural retry will resubmit every grouping from scratch, including the one(s) that already succeeded. That produces duplicate encounters and duplicate orders, which is hard to recover from after the fact.

Could we switch to `Promise.allSettled` and have the caller clear/error only the failed groupings, so successful encounters aren't resubmitted on retry?

packages/esm-patient-common-lib/src/orders/postOrders.ts
```diff
@@ -24,6 +24,45 @@ function getOrdersPayloadFromOrderBasket(patientUuid: string, ordererUuid: strin
...
+    encounterTypeMap.map(({ orderBasketGrouping, encounterTypeUuid }) => [orderBasketGrouping, encounterTypeUuid]),
+  );
+
+  const result = new Map<string, Array<OrderPost>>();
+
+  Object.entries(patientItems).forEach(([grouping, groupOrders]) => {
+    const encounterType = groupingToEncounterType.get(grouping) ?? defaultEncounterType;
+
+    if (!result.has(encounterType)) {
+      result.set(encounterType, []);
```
denniskigen, line 49: Minor cleanup: the empty array is added to `result` before we validate `dataPrepFn`, so a missing dataPrep fn leaves an empty Map entry that the downstream `.filter(orders.length > 0)` then has to discard. Not a bug, just slightly tidier to move the `dataPrepFn` check above `result.set` so the Map only ever holds entries we'll actually populate. Non-blocking.

packages/esm-patient-orders-app/src/order-basket/order-basket.component.tsx
```diff
@@ -112,22 +113,24 @@ const OrderBasket: React.FC<OrderBasketProps> = ({
...
         await closeWorkspace({ discardUnsavedChanges: true });
         mutateEncounterUuid();
         // Only revalidate current visit since orders create new encounters
         mutateVisitContext?.();
         invalidateVisitAndEncounterData(mutate, patientUuid);
         clearOrders();
         await mutateOrders();
-        onOrderBasketSubmitted?.(postedEncounter.uuid, postedEncounter.orders);
+        const allPostedOrders = postedEncounters.flatMap((e) => e.orders ?? []);
+        onOrderBasketSubmitted?.(postedEncounters[0]?.uuid, allPostedOrders);
```
denniskigen, line 133: Worth thinking through: `onOrderBasketSubmitted` is typed as `(encounterUuid: string, postedOrders: Array<Order>) => void`, but when the new map produces multiple encounters we now pass only the first UUID along with a flattened orders list from all of them. External consumers like `openmrs-esm-dispensing-app`'s `fill-prescription-button.component.tsx` use that UUID to open a modal scoped to the encounter, so they would silently miss the others.

Today both known external callers pass `visibleOrderPanels: [singleOrderTypeUuid]`, so in practice their baskets only ever produce one encounter and the bug is latent. Still, the in-tree order basket has no such restriction, so any consumer that mounts it unrestricted is exposed.

Two options that would resolve this:

1. Extend the callback to `(encounterUuids: Array<string>, postedOrders: Array<Order>) => void` and update the two consumers, or
2. Keep the signature and document the truncation in the type's JSDoc, only invoking the callback when exactly one encounter is produced.

Curious which feels right to you?

packages/esm-patient-common-lib/src/orders/postOrders.ts
```diff
@@ -24,6 +24,45 @@ function getOrdersPayloadFromOrderBasket(patientUuid: string, ordererUuid: strin
...
+
+    groupOrders.forEach((order) => {
+      result.get(encounterType).push(dataPrepFn(order, patientUuid, null, ordererUuid));
+    });
+  });
+
+  return result;
+}
+
 export async function postOrdersOnNewEncounter(
```
denniskigen, line 66: Now that the only in-tree caller has been migrated, would you consider marking `postOrdersOnNewEncounter` `@deprecated` with a pointer to the new function? External modules may still import it, so deletion isn't safe, but the JSDoc would steer new code in the right direction.

One small behavioural difference to confirm is intentional: the old function still POSTs an encounter when the basket is empty, whereas the new one returns `[]` (no POST). The Save button is disabled when the basket is empty so this should be theoretical, but worth noting in case anything else relied on the old behaviour.

packages/esm-patient-orders-app/src/config-schema.ts
denniskigen, line 78: Picking up on @chibongho's question: it would help future config authors if the description said *why* drug orders use the literal `"medications"` while other types use the `orderTypeUuid`. Something like "drug orders use `'medications'` because the order basket store groups them under that fixed key rather than by orderType UUID." That's the main ask.

Optional: we could also export `'medications'` as a named constant from `esm-patient-common-lib` to give the codebase a single source of truth (it's currently a literal in `medications-details-table.component.tsx` and a few tests). Config authors would still type the string into JSON either way, so this is purely an internal cleanup, not a blocker.

packages/esm-patient-common-lib/src/orders/postOrders.ts
```diff
@@ -62,6 +101,77 @@ export async function postOrdersOnNewEncounter(
...
+ * When `orderTypeEncounterTypeMap` is non-empty, basket orders are grouped by their store
+ * grouping key and each unique encounter type gets its own encounter POST.  Orders whose
+ * grouping is not listed in the map fall back to `defaultOrderEncounterType`.
+ *
+ * When `orderTypeEncounterTypeMap` is empty (the default), all orders are placed in a single
+ * encounter using `defaultOrderEncounterType`, preserving the original behaviour.
+ *
+ * @returns All created encounters (one per unique encounter type that received at least one order).
+ */
+export async function postOrdersOnNewEncountersByType({
```
denniskigen, line 132: The PR template has the tests box unchecked, and the new `getOrdersGroupedByEncounterType` / `postOrdersOnNewEncountersByType` carry the heart of the feature. Could we add unit tests in `postOrders.test.ts` (or a sibling) covering at least:

- empty map produces a single encounter (parity with the old behaviour)
- groupings present in the map split into per-type encounters
- groupings absent from the map fall back to `defaultOrderEncounterType`
- multiple groupings mapping to the same encounter type are merged into one
- missing `postDataPrepFunctions[grouping]` warns and skips

denniskigen (commented):
Thanks for picking this up, @donaldkibet.

In `order-basket.component.tsx`, the `else` branch around line 155 (the `postOrders(patientUuid, orderEncounterUuid, ...)` call) posts every basket grouping straight onto the single `orderEncounterUuid` and never consults `orderTypeEncounterTypeMap`. Per the comment at line 111, `orderEncounterUuid` is preset when visits are disabled and an order encounter already exists for the day, so in that mode the mapping works for the first basket of the day (which goes through the new encounter path) but silently degrades on every basket after that, with mapped orders landing in the default encounter instead of their configured type.

Was leaving the existing-encounter path out of the mapping intentional (e.g. "we only ever create one encounter per day in no-visits mode")? If so, a short comment in that branch would help future readers. If not, we'd probably want it to split as well, for example by routing groupings whose mapping matches the existing encounter's type into `postOrders(orderEncounterUuid, ...)` and groupings with a different mapping into a fresh `postEncounter` call.

PR #3263, opened 2026-04-24
(feat) Polish custom order-type panels to match drug/lab styling

denniskigen:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number.
- [x] My work is based on designs, which are linked or shown either in the Jira ticket or the description below.
- [ ] My work includes tests or is validated by existing tests.

## Summary

Custom order-type panels in the order basket rendered a bare glyph on a white card while Drug and Lab panels rendered a themed chip with a tinted icon backdrop, colored left border, and colored glyph. This PR brings the custom panels to visual parity across desktop and tablet.

Adds an optional `accentColor` field on each `orderTypes` config entry, e.g., `"accentColor": "teal"`. Allowed values: `purple`, `cyan`, `teal`, `magenta`, `orange`, `red`, `green`, `gray` (default `purple`). The accent drives four properties per tile: left border, top border, icon backdrop, and glyph color.

Also lines up tile height, container background, chevron size, icon sizing, h4 typography, and margin between tiles so the four panel types look like peers. Fixes the tablet bottom border that was previously dead because of a broken descendant selector (`.tabletTile .collapsedTile` became `.tabletTile.collapsedTile`).

Zeroheight page for the O3 order basket: https://zeroheight.com/23a080e38/p/71ad91-order-basket

## Screenshots

Tested using this sample config from AMRS O3:

```json
{
  "@openmrs/esm-patient-orders-app": {
    "orderEncounterType": "d8e0f981-62e0-4c2a-b3ca-890c51afc350",
    "orderTypes": [
      {
        "orderTypeUuid": "ff4485a4-f071-4423-aeb2-db6efce52b83",
        "orderableConceptSets": [
          "e83f4f97-91eb-4858-89d8-371535c5c131",
          "759a4bd6-79ab-40a9-9836-1e30783f7ae5"
        ],
        "label": "Radiology Orders",
        "icon": "omrs-icon-image-medical",
        "accentColor": "teal"
      },
      {
        "orderTypeUuid": "2315ab24-9a4e-4b36-b189-8e74d2c77394",
        "orderableConceptSets": [
          "eef060d3-d503-4695-a9ba-47f5af44135b"
        ],
        "label": "Procedure Orders",
        "icon": "omrs-icon-syringe",
        "accentColor": "orange"
      }
    ]
  }
}
```

#### Desktop
CleanShot 2026-04-24 at 12 11 39@2x

#### Tablet
CleanShot 2026-04-24 at 12 15 20@2x

[...]

PR #3264, opened 2026-04-24
(fix) Restore drug tile tablet bottom border and remove dead selectors

denniskigen:
## Requirements

- [x] This PR has a title that briefly describes the work done.

## Summary

Follow-up to #3263. Two cleanups in `esm-patient-medications-app` and `esm-patient-tests-app`:

1. **Drug's tablet bottom border was dead.** `drug-order-basket-panel.scss` used `.tabletTile .collapsedTile` (descendant), but both classes sit on the same element, so the rule never matched. Changed to the compound `.tabletTile.collapsedTile` so the 1px grey bottom border now renders, matching lab and the general order panels.
2. **Removed dead h4 rules** in both drug and lab that target `.desktopTile .container h4` / `.tabletTile .container h4`. Those components render `.desktopContainer` / `.tabletContainer`, so the selectors never matched. h4 already falls through to Carbon's `heading-03` default, so removal is a no-op visually. The `@use '@carbon/type'` import is dropped in both files since nothing references it anymore.

## Screenshots

No visible change on desktop. Drug tile gains a 1px grey bottom border on tablet (previously absent).

## Related Issue

None. Follow-up to #3263.

## Other

Thanks to @NethmiRodrigo for spotting the dead h4 rules during review of #3263.

### DIFF
--- a/packages/esm-patient-medications-app/src/drug-order-basket-panel/drug-order-basket-panel.scss
+++ b/packages/esm-patient-medications-app/src/drug-order-basket-panel/drug-order-basket-panel.scss
@@ -1,6 +1,5 @@
 @use '@carbon/colors';
 @use '@carbon/layout';
-@use '@carbon/type';
 @use '@openmrs/esm-styleguide/src/vars' as *;
 
 .desktopContainer {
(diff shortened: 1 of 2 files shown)


NethmiRodrigo (approved):
Thanks @denniskigen!

PR #3265, opened 2026-04-24
(feat) O3-4578: Prevent saving duplicate immunization doses

Josephkagimu1:
## Requirements

## Problem
Previously, users could submit the same vaccine and dose number multiple times for a patient, creating duplicate records. There was no validation to warn or block this.

## Summary
Adds a UX-level guard that prevents duplicate dose entries for the same vaccine in the immunization form.

**`immunizations-form.workspace.tsx`**
- Fetches existing immunizations via the `useImmunizations` hook
- Detects duplicates in `onSubmit` — compares the incoming dose number against already-recorded doses for the same vaccine, excluding the record currently being edited
- Surfaces an inline validation error via `setError('doseNumber', ...)` when a duplicate is detected, instead of using a snackbar
- Clears the stale duplicate-dose error when the user switches to a different vaccine
- Disables the Save button while existing immunizations are loading (`isLoadingImmunizations`)

**`dose-input.component.tsx`**
- Wires `fieldState.error` from `useController` into both the sequence `<Dropdown>` and the free-form `<NumberInput>` via `invalid` / `invalidText` so the inline error renders on either dose-input variant

**`immunizations-form.test.tsx`**
- Adds `should prevent saving a duplicate dose for the same vaccine`
- Adds `should allow re-saving when editing an existing immunization record`

**`translations/en.json`**
- Adds `duplicateDoseError`: "Dose {{dose}} has already been recorded for this vaccine"

## Note
This is a UX-level guard, not a server-side integrity guarantee. Two clinicians on different sessions could still race to create duplicates. Server-side enforcement would require a separate backend change.

### Initial implementation

Initial duplicate-blocked UI

### After review (inline error on the dose field)

Inline error after review

## Related Issue
[O3-4578](https://openmrs.atlassian.net/browse/O3-4578)

### Other

[O3-4578]: https://openmrs.atlassian.net/browse/O3-4578?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


omsrivastava512 (changes requested):
Hey @Josephkagimu1 thank you for this PR and especially for including the tests. However, I see some unintended files were included in your commits, could you delete them?

Also, before diving into the code review, I'd appreciate if you bring this to the coffee break today where we should discuss the approach first. Personally, I'm not very certain about blocking the user, a simple warning/confirmation is what feels right to me.

Lastly, your PR body formatting is a bit off (just bringing to your notice).

Josephkagimu1:
> Hey @Josephkagimu1 thank you for this PR and especially for including the tests. However, I see some unintended files were included in your commits, could you delete them?
> 
> Also, before diving into the code review, I'd appreciate if you bring this to the coffee break today where we should discuss the approach first. Personally, I'm not very certain about blocking the user, a simple warning/confirmation is what feels right to me.
> 
> Lastly, your PR body formatting is a bit off (just bringing to your notice).

@omsrivastava512 , thanks for that information , where are those unintended files included , kindly help me identify it such that I delete that .

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.workspace.tsx
```diff
@@ -189,10 +189,42 @@ const ImmunizationsForm: React.FC<PatientWorkspace2DefinitionProps<{}, {}>> = ({
       mutate,
     ],
   );
+
+  const wrappedSubmit = useCallback(
```
denniskigen: Could the duplicate check live with the rest of the form validation rather than as a snackbar? Every other rule on this form is surfaced inline via zod + `invalidText`, and a duplicate dose is conceptually a `doseNumber` field error, so a toast feels inconsistent and is easier to miss.

Two options that fit the existing pattern:

- Add a `superRefine` to the zod schema that checks `existingImmunizations` / `immunizationToEditMeta` and reports the issue on `doseNumber`.
- Or run the duplicate check after `handleSubmit` validation and call `setError('doseNumber', { message: t('duplicateDoseError', …) })`.

Either approach also avoids a smaller issue with the current shape: the check runs before `handleSubmit(onSubmit)`, so if a duplicate is found while other fields are invalid, the duplicate toast fires and zod's inline errors never surface.

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.workspace.tsx
```diff
@@ -189,10 +189,42 @@ const ImmunizationsForm: React.FC<PatientWorkspace2DefinitionProps<{}, {}>> = ({
...
+    (e: React.FormEvent) => {
+      e.preventDefault();
+
+      const data = formProps.getValues();
+
+      const isDuplicate = existingImmunizations?.some((group) => {
```
denniskigen: Two edge cases worth handling here:

1. `useImmunizations` returns `[]` while the request is loading, so if a user submits before the history loads, `isDuplicate` is `false` and the check silently passes. Easiest fix is to pull `isLoading` from the hook and add it to the Save button's `disabled` prop.
2. `Number(null) === Number(null)` is `0 === 0`, so two null-ish dose numbers will false-match. The schema allows `doseNumber: null`, so an existing dose with no recorded number could match a new submission with no number set. An explicit guard would be safer:
   ```tsx
   dose.doseNumber != null &&
   data.doseNumber != null &&
   Number(dose.doseNumber) === Number(data.doseNumber)
   ```

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.test.tsx
```diff
@@ -458,6 +472,60 @@ describe('Immunizations Form', () => {
       expect.any(AbortController),
     );
   });
+
+  it('should prevent saving a duplicate dose for the same vaccine', async () => {
```
denniskigen, line 473: Nice test. Two small additions would round it out:

- A case for edit mode — re-saving the record currently being edited (so `dose.immunizationObsUuid === immunizationToEditMeta?.immunizationObsUuid`) should *not* trigger the warning. That branch is the non-obvious half of the logic.
- The current assertion only checks `kind: 'warning'`. Asserting on the title (or its translation key) would catch regressions where the wrong snackbar fires.

denniskigen (commented):
Thanks for tackling this — the duplicate-check logic is on the right track and the test demonstrates the intended behavior. A few things to address before merge, mostly around where the check lives and a couple of edge cases.

One thing not inline-able: there are three empty files at the repo root (`#`, `npx`, `yarn`) that look like accidental shell artifacts. Could you delete them?

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.test.tsx
Josephkagimu1: Hei @denniskigen  thanks for the guidance , I made some changes , as i was directed , let me hope its now fine , let me know if there is still something to fix , thanks

Josephkagimu1: > Nice test. Two small additions would round it out:
> 
> * A case for edit mode — re-saving the record currently being edited (so `dose.immunizationObsUuid === immunizationToEditMeta?.immunizationObsUuid`) should _not_ trigger the warning. That branch is the non-obvious half of the logic.
> * The current assertion only checks `kind: 'warning'`. Asserting on the title (or its translation key) would catch regressions where the wrong snackbar fires.

Changes have been made , hope the PR is now fine to be merged , is there any additional change needed ? @denniskigen , @omsrivastava512

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.workspace.tsx
```diff
@@ -62,7 +62,6 @@ const ImmunizationsForm: React.FC<PatientWorkspace2DefinitionProps<{}, {}>> = ({
             message: t('vaccinationDateCannotBeInTheFuture', 'Vaccination date cannot be in the future'),
           },
         ),
-      // null means unset; when provided, must be an integer ≥ 1
       doseNumber: z.union([z.number({ coerce: true }).int().min(1), z.null()]).optional(),
```
denniskigen, line 67: Small thing — looks like a couple of intentional comments got dropped in this area during the `wrappedSubmit` cleanup. Could we restore them? They explain non-obvious behavior:

```ts
// Normalize both dates to start of day in local timezone
```

(above the `inputDate`/`today` lines in the `.refine`) and

```ts
// null means unset; when provided, must be an integer ≥ 1
```

(above this `doseNumber` line).

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.workspace.tsx
```diff
@@ -124,6 +124,27 @@ const ImmunizationsForm: React.FC<PatientWorkspace2DefinitionProps<{}, {}>> = ({
 
   const onSubmit = useCallback(
     async (data: ImmunizationFormInputData) => {
+      const isDuplicate = existingImmunizations?.some((group) => {
```
denniskigen, line 137: Worth a brief comment here noting that `existingImmunizations` is `undefined` during loading, so this check silently no-ops in that window — and that the disabled Save button (`isLoadingImmunizations`) is what keeps that safe. Just so a future change doesn't accidentally re-enable submit-while-loading and let duplicates slip through.

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.workspace.tsx
```diff
@@ -124,6 +124,27 @@ const ImmunizationsForm: React.FC<PatientWorkspace2DefinitionProps<{}, {}>> = ({
...
+      });
+
+      if (isDuplicate) {
+        setError('doseNumber', {
+          type: 'manual',
+          message: t('duplicateDoseError', 'Dose {{dose}} has already been recorded for this vaccine', {
+            dose: data.doseNumber,
+          }),
+        });
+        return;
```
denniskigen, line 155: Minor UX issue worth handling: once `setError('doseNumber', ...)` fires, the error clears naturally when the user edits the **dose** field (RHF re-validates), but it stays stuck if they change the **vaccine** to one without a conflict. Could we clear it when the vaccine changes? Something like:

```ts
useEffect(() => {
  clearErrors('doseNumber');
}, [vaccineUuid, clearErrors]);
```

### Duplicate validation error showing correctly
CleanShot 2026-05-04 at 14 35 53@2x

### Validation error persists even after changing the vaccine:
CleanShot 2026-05-04 at 14 42 27@2x

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.test.tsx
```diff
@@ -458,6 +472,113 @@ describe('Immunizations Form', () => {
...
+
+    render(<ImmunizationsForm {...testProps} />);
+
+    await user.click(screen.getByRole('button', { name: /Save/i }));
+
+    // Should NOT show duplicate error — we're editing the same record
+    expect(screen.queryByText(/has already been recorded/i)).not.toBeInTheDocument();
+
+    // Should allow save
+    expect(mockSavePatientImmunization).toHaveBeenCalledTimes(1);
```
denniskigen: Nice that you covered the edit case. One suggestion to make this test a stronger guard against regressions: as written, it would also pass if the duplicate check were broken to always return `false`, since both paths end up calling `savePatientImmunization`. Could we tighten this assertion to prove the call actually carries the edit UUID?

```ts
expect(mockSavePatientImmunization).toHaveBeenCalledWith(
  expect.anything(),
  'existing-uuid',
  expect.any(AbortController),
);
```

That way the test specifically pins the "edit-mode bypass" behavior rather than just "save was called."

denniskigen (commented):
Nice work on this — the inline-error UX is the right call, and the edit-mode exclusion is handled cleanly. One thing worth calling out in the PR description: this is a UX guard, not a server-side integrity guarantee. Two clinicians on different sessions could still race to create duplicates. Not something to address in this PR, but good to note so it doesn't get mistaken for full prevention.

Left a few inline suggestions below.

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.test.tsx
Josephkagimu1: Hi @denniskigen I made all the required changes , kindly check to confirm that i fixed everything needed , thanks

denniskigen (commented):
Thanks for the quick turnaround on these. LGTM, @Josephkagimu1!

denniskigen (approved):

Josephkagimu1:
> Thanks for the quick turnaround on these. LGTM, @Josephkagimu1!

Your welcome @denniskigen , I am also happy and appreciate your guidance at every step , thanks

PR #3277, opened 2026-05-04
Add Claude PR review caller workflow (pilot)

dkayiwa:
## Summary

Adds a thin caller workflow that invokes the reusable Claude PR review workflow in `openmrs/openmrs-contrib-gha-workflows`. Triggers on `[opened, ready_for_review, reopened]` (no `synchronize`) to avoid re-reviewing on every push during the pilot.

## Why

Part of a 10-repo pilot of AI-assisted PR review (see openmrs/openmrs-contrib-gha-workflows#36). We're evaluating Claude Code Action as a potential replacement for CodeRabbit, which has been hitting free-tier limits.

## Dependencies

⚠️  **Do not merge before openmrs/openmrs-contrib-gha-workflows#36 is merged.** This PR is opened as a draft until then; the caller pins the reusable workflow to `@main`.

## Cost guardrails (pilot)

- Model: `claude-sonnet-4-6` (cheaper of the two main options).
- Trigger: `[opened, ready_for_review, reopened]` only — explicitly excludes `synchronize`, so each PR gets one review per state-change, not one per push.
- Skips draft PRs.

## Test plan

- [ ] Merge openmrs/openmrs-contrib-gha-workflows#36.
- [ ] Mark this PR ready for review.
- [ ] Verify the next non-draft PR opened against this repo gets a Claude review comment.
- [ ] Compare review quality / signal-to-noise vs. CodeRabbit over the next 1–2 weeks.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.github/workflows/claude-review.yml
```diff
@@ -0,0 +1,25 @@
...
+on:
+  pull_request_target:
+    types: [opened, ready_for_review, reopened]
+
+jobs:
+  review:
+    permissions:
+      contents: read
+      pull-requests: write
+      issues: write
```
denniskigen, line 12: Small least-privilege thought: the Claude Code Action typically only needs `pull-requests: write` to post review comments. If `claude-pr-review.yml` doesn't actually create or edit issues, `issues: write` could likely be dropped here. Worth a quick check against the upstream workflow's declared needs.

.github/workflows/claude-review.yml
```diff
@@ -0,0 +1,25 @@
...
+    with:
+      extra_instructions: |
+        This is a TypeScript/React OpenMRS frontend repository (part of OpenMRS 3.x microfrontend architecture). Pay particular attention to:
+
+        - i18n: user-facing strings must use the t() function from useTranslation, not hard-coded text. Flag hard-coded English strings in JSX, alerts, and toasts.
+        - Type safety: avoid `any` in new code; prefer specific types or `unknown` with type guards. Flag new uses of `any`.
+        - Data fetching: use SWR/React Query patterns consistent with the rest of the codebase; don't introduce ad-hoc fetch wrappers.
+        - Public exports: flag breaking changes to anything exported from index.ts or package entry points.
+        - Testing: use @testing-library/react idioms; avoid testing implementation details. Flag tests that mock OpenMRS framework primitives instead of using the official testing utilities.
+        - Performance: flag obvious re-render hazards (new object/array literals as hook deps, missing useMemo/useCallback on hot paths) but skip nits.
```
denniskigen, line 25: Solid starting point for the pilot. After looking at the conventions in [o3-docs](https://o3-docs.openmrs.org) and the most-used `@openmrs/esm-framework` symbols across this repo, a few of these bullets restate generic React/TS advice that Claude already applies without prompting, while some of the highest-leverage O3-specific rules are missing. A few suggestions worth considering:

- **Data fetching:** the line mentions React Query, but this codebase is SWR-only via `openmrsFetch`. The convention from o3-docs is to colocate fetching in `*.resource.ts` files with hooks named `use<Resource>` that memoize their return value (especially for hooks consumed in table rows).
- **Testing:** the current bullet says to flag tests that "mock OpenMRS framework primitives," but per o3-docs the framework mocks at `@openmrs/esm-framework/mock` are the official path. The rule there is "don't mock components, render real implementations."
- **i18n:** worth adding the `en.json` / Transifex rule, so the reviewer doesn't ask contributors to update other locale files. Also `<Trans>` for strings containing HTML.
- **Framework primitives over ad-hoc:** judging by import counts across this repo, the most common violations are around `showSnackbar` (not custom toasts), `showModal` (not bespoke modals), `launchWorkspace` / `launchWorkspace2` for workspaces, `formatDate` / `parseDate` for dates, and `useConfig` + `defineConfigSchema` for configurable values. A bullet flagging custom toasts, bespoke modals, raw `dayjs.format()`, and hardcoded tunables would catch a lot.
- **Imports:** two small but frequent ones worth adding are `lodash-es` over `lodash` (tree-shaking) and `@carbon/react` as the sole UI library (flag MUI/Chakra/Mantine/etc.).

A denser revised draft, dropping the generic React/TS bits to make room:

```yaml
      extra_instructions: |
        This is an OpenMRS 3.x frontend repo (TS/React, microfrontend
        architecture). Conventions: https://o3-docs.openmrs.org. Flag
        deviations from the following project-specific rules; skip generic
        React nits.

        - Prefer `@openmrs/esm-framework` primitives over ad-hoc equivalents:
          `showSnackbar` for toasts, `showModal` for modals, `launchWorkspace`
          (or `launchWorkspace2` where the surrounding code uses v2) for
          workspaces, `formatDate`/`parseDate`/`formatDatetime` for dates,
          `useConfig` + `defineConfigSchema` for configurable values. Flag
          custom toast/modal components, raw `dayjs.format()`, and hardcoded
          tunables that should be config.
        - Data fetching: colocate in `*.resource.ts`. Use `openmrsFetch` with
          `useSWR` / `useSWRImmutable` / `useOpenmrsSWR`; no ad-hoc fetch
          wrappers. Name hooks `use<Resource>`. Wrap hook return objects in
          `useMemo` (especially for hooks consumed in table rows). Return
          `isLoading`, `error`, `data` distinctly. Pass `null` URL for
          conditional fetching. No infinite retries.
        - i18n: user-facing strings via `t()` from `useTranslation`; use
          `<Trans>` for strings containing HTML. `en.json` is the source of
          truth and is regenerated by `extract-translations`; don't hand-edit,
          and don't ask contributors to update other locale files (Transifex
          syncs them).
        - UI library: `@carbon/react` only. Flag imports from other component
          libraries (MUI, Chakra, Mantine, Radix, etc.).
        - Imports: use `lodash-es`, not `lodash` (tree-shaking). Use
          `import type { … }` for type-only imports.
        - Testing: render real components rather than mocking them; use the
          mocks from `@openmrs/esm-framework/mock`. Don't test implementation
          details.
        - Public exports: flag breaking changes to anything exported from
          `index.ts` or package entry points.
        - Performance: flag missing `useMemo` on hook return values
          (especially in row-rendered components).
```

PR #3280, opened 2026-05-05
fix(ts): remove broken index signature on DataTable render prop in encounters-table

vinisha231:
## Problem

The `EncountersTable` component had an inline type annotation on the Carbon `DataTable` render prop with an index signature `[key: string]: unknown`:

```tsx
{({
  rows, headers, getHeaderProps, getTableProps, ...
}: {
  headers: Array<{ header: React.ReactNode; key: string }>;
  rows: Array<...>;
  [key: string]: unknown;  // ← this is the problem
}) => {
```

The index signature causes TypeScript to widen all destructured values to `unknown`, breaking every helper function call (`getTableProps()`, `getHeaderProps()`, `getToolbarProps()`, etc.) with errors like:

```
TS2349: This expression is not callable. Type '{}' has no call signatures.
TS2339: Property 'length' does not exist on type 'unknown'.
```

## Fix

Remove the manual inline type annotation entirely. Carbon's `DataTable` is fully generic and TypeScript correctly infers `DataTableRenderProps` for the render prop — no annotation needed.

## Testing
- `yarn tsc --noEmit` passes on `esm-patient-chart-app`
- No runtime behavior changes

Fixes build failure in `esm-patient-chart-app`.


vinisha231:
Closing to reopen with correct branch base and title format.

PR #3282, opened 2026-05-06
(fix) Use POST when fetching concept references

denniskigen:
## Requirements

## Summary

Mirrors openmrs/openmrs-esm-form-engine-lib#669 in patient-chart. Migrates the two remaining `/conceptreferences` callers from a chunked GET pattern (`chunk(uuids, 10)` plus `Promise.all` over N requests) to a single POST that takes the references array in the body.

Affected hooks:

- `esm-generic-patient-widgets-app`: `useConcepts`
- `esm-patient-tests-app`: `useOrderReasons`

Benefits:

- One round trip per render instead of `ceil(N / 10)`. A widget configured with 50 concepts goes from 5 sequential SWR fetches to 1.
- Removes the implicit cap from chunking against an unstated URL-length limit. POST sidesteps URL-length concerns entirely.
- Cleaner failure semantics: one request, one error path, no risk of silently merging partial results from multiple chunks.

Behavior preserved end-to-end: both hooks return the same `{ concepts, isLoading }` / `{ orderReasons, isLoading }` shape, the empty-input short-circuit (`null` SWR key) still applies, and consumers (`useObs`, `test-order-form`) are untouched.

## Other

Requires REST API >= 4.0.0, when `POST /conceptreferences` landed. No formal floor declaration mechanism exists in patient-chart (none was added for the form-engine-lib precedent either), so this is documented here rather than enforced.

### DIFF
--- a/packages/esm-generic-patient-widgets-app/src/resources/useConcepts.ts
+++ b/packages/esm-generic-patient-widgets-app/src/resources/useConcepts.ts
@@ -1,6 +1,7 @@
 import { openmrsFetch, type FetchResponse, restBaseUrl, showSnackbar } from '@openmrs/esm-framework';
-import chunk from 'lodash/chunk';
 import useSWRImmutable from 'swr/immutable';
+
+const conceptRepresentation = 'custom:(uuid,display,datatype,answers)';
 
 export interface ConceptReferenceResponse {
   [key: string]: {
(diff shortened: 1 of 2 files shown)


NethmiRodrigo (approved):
LGTM, thanks!

ibacher (approved):

PR #3284, opened 2026-05-06
(fix) O3-4578: Use warning for duplicate immunization doses

ibacher:
## Requirements

## Summary

This is a fix for O3-4578 / #3265 which makes it a validation error to enter the same dose number for the same vaccine. This switches things so that a warning is shown to the user, but the warning is non-blocking.

The "dose number" for vaccinations is a sequence number in a schedule and not an incrementing counter of the number of times a patient has been given a dose. As Andy Kanter has repeatedly pointed out as we've developed immunization-related functionality, ["just because a person gets a shot doesn't mean the shot was valid"](https://talk.openmrs.org/t/immunization-vs-vaccination/30144/9).  Consequently, blocking users from submitting a record of a second application of the same dose forces users to record clinically incorrect data. Presumably, we should have some way of marking known-bad or known-erroneous doses as such, but that's beyond the scope of this ticket.

## Screenshots

Immunization

## Related Issue

https://issues.openmrs.org/browse/O3-4578


denniskigen (approved):
LGTM. Thanks, @ibacher!

PR #3297, opened 2026-05-11
(fix) O3-5657: Various fixes to the procedures form

dilankavishka:
## Requirements

## Summary

- Replace the year and month `Select`s with `ComboBox`es and fix horizontal alignment of the duration fields
- Fix validation bypass when toggling "Is start date known?" from Yes to No while a start date was already entered (stale `startDateTime` was left in form state and silently passed validation)
- Fix the unreachable saving state on the submit button so the spinner now appears while the form is submitting
- Fix a crash when selecting a procedure type by guarding against a null `selectedItem`
- Add `estimatedStartDate` to the `RawProcedure` type so the payload contract matches what is actually sent (and the backend persists)
- Build the estimated-date comparison in local time so end-date validation does not reject the first valid day of the estimated period in non-UTC zones
- Re-trigger `startDateTime` validation after submit so the "Start date is required" error clears as soon as the user picks a year
- Add month-name translations and rename the duration legend to "Procedure duration"
- Add tests for the toggle bug and the estimated-date save path

## Screenshots

1. Start date time error
Screenshot 2026-05-11 at 17 53 42

2. When selecting "No" for "Is start date known?"

Before:
Screenshot 2026-05-11 at 17 57 09

After:
Screenshot 2026-05-11 at 17 58 19

## Related Issue

https://issues.openmrs.org/browse/O3-5657


dilankavishka:
@denniskigen @NethmiRodrigo Could you please review this PR?

packages/esm-patient-procedures-app/src/procedures/procedures-form.component.tsx
```diff
@@ -86,21 +86,36 @@ const ProceduresFormComponent: React.FC<ProceduresFormComponentProps> = ({
   const [estimatedYear, setEstimatedYear] = useState(initialEstimatedDate?.split('-')[0] ?? '');
   const [estimatedMonth, setEstimatedMonth] = useState(initialEstimatedDate?.split('-')[1] ?? '');
 
+  React.useEffect(() => {
```
NethmiRodrigo: Lets add `useEffect` as an import directly instead of doing this

packages/esm-patient-procedures-app/src/procedures/procedures-form.component.tsx
```diff
@@ -86,21 +86,36 @@ const ProceduresFormComponent: React.FC<ProceduresFormComponentProps> = ({
...
-    { value: '08', label: 'August' },
-    { value: '09', label: 'September' },
-    { value: '10', label: 'October' },
-    { value: '11', label: 'November' },
-    { value: '12', label: 'December' },
+  const yearOptions = Array.from({ length: currentYear - 1900 + 1 }, (_, i) => ({
+    id: String(currentYear - i),
+    label: String(currentYear - i),
+  }));
+  const monthOptions = [
```
NethmiRodrigo: A couple of things:
1. Since these are constants, we could either move them up above the component, or wrap them in a `useMemo`
2. The month strings should be wrapped in a `t` function
3. A future question for @jayasanka-sack, why did we not use a datepicker here?

packages/esm-patient-procedures-app/src/procedures/procedures-form.component.tsx
NethmiRodrigo, line 246: So, there’s a bug here. The form default opens with `isStartDateKnown = true`. User picks a real date, so the `startDateTime` is now a Date object in form state. The user then toggles, `isStartDateKnown` to false. The `onChange` handler for the toggle does:
```
onChange={({ index }) => {
  setIsStartDateKnown(index === 0);   // → false
  setEstimatedYear('');
  setEstimatedMonth('');
}}
```
which sets the `estimatedYear` and `estimatedMonth` to an empty string, but it doesn’t update the `startDateTime` in the form state. The `useEffect` here runs, and goes into the `else` branch and sets the `estimatedStartDate` also to an empty string. Say the user does not pick a year. The validation that you added in the `procedures-form.workspace`: `Boolean(data.startDateTime) || Boolean(data.estimatedStartDate)` doesn’t give an error because the stale `startDateTime` was never cleared.
The fix would be to clear `startDateTime` on the `No` branch of the toggle.

Lets add a test to cover this scenario.

packages/esm-patient-procedures-app/src/procedures/procedures-form.workspace.test.tsx
NethmiRodrigo, line 1: Can we add a test here to cover the estimated date scenario, for example:

```
 it('saves an estimated start date when "Is start date known?" is "No"', async () => {
   const user = userEvent.setup();
   mockUseConceptSearch.mockReturnValue({ searchResults: searchedProcedure, isSearching: false });
   mockSaveProcedure.mockResolvedValue({ status: 201 } as unknown as FetchResponse);
   renderProceduresForm();

   await user.type(screen.getByRole('searchbox', { name: /enter procedure/i }), 'App');
   await user.click(screen.getByRole('menuitem', { name: /appendectomy/i }));
   await user.type(screen.getByRole('searchbox', { name: /enter body site/i }), 'Site');
   await user.click(screen.getByRole('menuitem', { name: /appendectomy/i }));
   const statusGroup = screen.getByRole('group', { name: /^status/i });
   await user.click(within(statusGroup).getByRole('combobox'));
   await user.click(screen.getByRole('option', { name: /appendectomy/i }));
   const procedureTypeGroup = screen.getByRole('group', { name: /procedure type/i });
   await user.click(within(procedureTypeGroup).getByRole('combobox'));
   await user.click(screen.getByRole('option', { name: /surgery/i }));

   const known = screen.getByRole('group', { name: /is start date known/i });
   await user.click(within(known).getByText(/^no$/i));

   await user.click(screen.getByRole('combobox', { name: /year/i }));
   await user.click(screen.getByRole('option', { name: '2023' }));
   await user.click(screen.getByRole('combobox', { name: /month/i }));
   await user.click(screen.getByRole('option', { name: /may/i }));

   await user.click(screen.getByRole('button', { name: /save & close/i }));
   await waitFor(() =>
     expect(mockSaveProcedure).toHaveBeenCalledWith(
       expect.objectContaining({ startDateTime: null, estimatedStartDate: '2023-05' }),
     ),
   );
 });
```

packages/esm-patient-procedures-app/src/procedures/procedures-form.component.tsx
dilankavishka: Done

packages/esm-patient-procedures-app/src/procedures/procedures-form.component.tsx
dilankavishka: Done

packages/esm-patient-procedures-app/src/procedures/procedures-form.component.tsx
dilankavishka: Updated

packages/esm-patient-procedures-app/src/procedures/procedures-form.workspace.test.tsx
dilankavishka: Updated

packages/esm-patient-procedures-app/src/procedures/procedures-form.component.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
```
denniskigen, line 1: `Select` and `SelectItem` are no longer used from the `carbon/react` import. Please remove them.

denniskigen:
On `packages/esm-patient-procedures-app/src/types.ts` (`RawProcedure`):

`RawProcedure` is missing `estimatedStartDate`, but `handleSave` sends it at `procedures-form.component.tsx:171` and the backend persists it (`Procedure.estimatedStartDate`, exposed via `ProcedureResource.getCreatableProperties()`).

Could we add it here?

```ts
estimatedStartDate?: string | null;
```

denniskigen:
On `packages/esm-patient-procedures-app/src/procedures/procedures-form.workspace.tsx:39-40`:

`new Date(data.estimatedStartDate)` parses `YYYY` / `YYYY-MM` as UTC midnight, while the date picker value is normalized to local midnight in `DateTimeField` via `setHours(0, 0, 0, 0)`. In UTC+ time zones, an end date on the first valid day of the estimated period gets rejected.

Example in `Africa/Nairobi`: estimated start `2023-05` parses to `2023-05-01T00:00:00.000Z`, but selecting May 1 produces local midnight, `2023-04-30T21:00:00.000Z`, so `endDateTime >= start` is false. The tests run under `TZ=UTC`, so they miss this.

Could we build the comparison date in local time?

```ts
const [year, month] = data.estimatedStartDate.split('-').map(Number);
const start = new Date(year, Number.isFinite(month) ? month - 1 : 0, 1);
return data.endDateTime >= start;
```

packages/esm-patient-procedures-app/src/procedures/procedures-form.component.tsx
dilankavishka: done

packages/esm-patient-procedures-app/src/procedures/procedures-form.component.tsx
```diff
@@ -86,22 +84,43 @@ const ProceduresFormComponent: React.FC<ProceduresFormComponentProps> = ({
...
+      setValue('estimatedStartDate', '');
+    } else if (estimatedYear) {
+      setValue('estimatedStartDate', estimatedMonth ? `${estimatedYear}-${estimatedMonth}` : estimatedYear, {
+        shouldValidate: true,
+      });
+    } else {
+      setValue('estimatedStartDate', '');
+    }
+  }, [isStartDateKnown, estimatedYear, estimatedMonth, setValue]);
+
```
denniskigen, line 96: Nice work on the toggle clearing logic, @dilankavishka. While testing locally I hit a small validation issue that I think originates here. Wanted to share what I found and a fix that worked end to end.

Repro:
1. Toggle "Is start date known?" to **No**. "Start date is required" appears immediately, before the user has touched the year combo.
2. Pick **2023** for Year. The error stays visible.
3. Click **Save & close**. Only then does the error clear.

There are two things going on. First, the error doesn't clear after picking a year because `setValue('estimatedStartDate', '2023', { shouldValidate: true })` runs validation, but `shouldValidate` only updates errors on the path it was called with. The refine's error lives at `path: ['startDateTime']`, so the stale error there is left untouched. We need an explicit `trigger('startDateTime')` to refresh that path. Second, the error fires immediately on toggle because `setValue('startDateTime', null, { shouldValidate: true })` in the toggle handler revalidates on the spot, before the user has had a chance to fill the year.

Just those two changes still leave the error appearing on toggle, because this `useEffect` itself runs whenever `isStartDateKnown` changes and re-triggers validation. The RHF idiom for progressive validation is to gate the manual `trigger` on `isSubmitted`, so we only push back on the user after they've actually tried to submit:

```ts
const {
  formState: { errors, isSubmitted },
  setValue,
  trigger,
  ...
} = useFormContext<ProceduresFormSchema>();

useEffect(() => {
  const value =
    !isStartDateKnown && estimatedYear
      ? estimatedMonth ? `${estimatedYear}-${estimatedMonth}` : estimatedYear
      : '';
  setValue('estimatedStartDate', value);
  if (isSubmitted) {
    trigger('startDateTime');
  }
}, [isStartDateKnown, estimatedYear, estimatedMonth, isSubmitted, setValue, trigger]);
```

And in the toggle handler at line 260:

```ts
if (!isKnown) {
  setValue('startDateTime', null);
}
```

I verified locally that this gives:

- Toggle pre-submit: no premature error.
- Submit with year empty: error appears.
- Pick year post-submit: error clears.
- Clear year post-submit: error reappears.

Could we also add a test for the "clears after picking a year" path? That's the actual regression the fix addresses, and the current test at `procedures-form.workspace.test.tsx:464` only checks the error after submit:

```ts
// After the existing submit assertion:
await user.click(screen.getByRole('combobox', { name: /year/i }));
await user.click(screen.getByRole('option', { name: '2023' }));
expect(screen.queryByText(/start date is required/i)).not.toBeInTheDocument();
```

Side note while we're here: the original three-branch form collapses to two since the `if (isStartDateKnown)` and final `else` both write `''`. The snippet above already incorporates that.

packages/esm-patient-procedures-app/src/procedures/procedures-form.component.tsx
```diff
@@ -123,6 +142,7 @@ const ProceduresFormComponent: React.FC<ProceduresFormComponentProps> = ({
   const [errorSaving, setErrorSaving] = useState(null);
 
   const handleSave = useCallback(async () => {
+    setIsSubmittingForm(true);
     const procedureType = getValues('procedureType');
```
denniskigen, line 144: Tiny nit: since the `useEffect` above keeps `estimatedStartDate` in form state in sync, `handleSave` can read it from there:

```ts
const estimatedStartDate = getValues('estimatedStartDate') || null;
```

Then `estimatedYear`, `estimatedMonth`, and `isStartDateKnown` come out of the `useCallback` deps. Single source of truth.

denniskigen:
Tangential, but can we do something about the duplicate `Duration` label text?

CleanShot 2026-05-11 at 23 10 18@2x

dilankavishka:
> Tiny nit: since the `useEffect` above keeps `estimatedStartDate` in form state in sync, `handleSave` can read it from there:
> 
> ```ts
> const estimatedStartDate = getValues('estimatedStartDate') || null;
> ```
> 
> Then `estimatedYear`, `estimatedMonth`, and `isStartDateKnown` come out of the `useCallback` deps. Single source of truth.

@denniskigen To keep the consistency and the better understanding shall we rename the title to `Length of procedure` ? 
Screenshot 2026-05-12 at 10 13 28

so as like this, the length of procedure is 2 hours.

packages/esm-patient-procedures-app/src/procedures/procedures-form.component.tsx
dilankavishka: Done. 

https://github.com/user-attachments/assets/f7353858-59a0-4172-84e5-25f0c73f19a1

packages/esm-patient-procedures-app/src/procedures/procedures-form.component.tsx
dilankavishka: Updated

denniskigen:
> > Tiny nit: since the `useEffect` above keeps `estimatedStartDate` in form state in sync, `handleSave` can read it from there:
> > ```ts
> > const estimatedStartDate = getValues('estimatedStartDate') || null;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then `estimatedYear`, `estimatedMonth`, and `isStartDateKnown` come out of the `useCallback` deps. Single source of truth.
> 
> @denniskigen To keep the consistency and the better understanding shall we rename the title to `Length of procedure` ? Screenshot 2026-05-12 at 10 13 28
> 
> so as like this, the length of procedure is 2 hours.

Would "Procedure duration" be a more accurate legend label here? @VeronicaMuthee

denniskigen (approved):
Thanks for the careful iteration on this one, @dilankavishka. Tracking down the stale `startDateTime` after the toggle and wiring up the validation re-trigger so the error clears as soon as the user picks a year both took some thought, and the new tests lock both behaviours in nicely. Good catch on the procedure-type null guard too, and tightening the `RawProcedure` type closes a contract gap that had been hiding in plain sight.

Pushed one small follow-up to switch the duration translation key to camelCase for consistency with the rest of `en.json`. Approving.

dilankavishka:
> Thanks for the careful iteration on this one, @dilankavishka. Tracking down the stale `startDateTime` after the toggle and wiring up the validation re-trigger so the error clears as soon as the user picks a year both took some thought, and the new tests lock both behaviours in nicely. Good catch on the procedure-type null guard too, and tightening the `RawProcedure` type closes a contract gap that had been hiding in plain sight.
> 
> Pushed one small follow-up to switch the duration translation key to camelCase for consistency with the rest of `en.json`. Approving.

Thank you Dennis!

PR #3301, opened 2026-05-12
(chore) Migrate from jest to vitest

denniskigen:
## Requirements

## Summary

Migrates patient-chart's test runner from jest to vitest across all 21 workspace packages that use the JS test stack (`esm-form-entry-app` uses Angular's `ng test` / Karma and is left alone), aligning patient-chart with the patterns adopted across the rest of the O3 apps.

Changes:
- **Root deps:** Replace `jest`, `@swc/jest`, `jest-cli`, `jest-environment-jsdom`, and `@types/jest` with `vitest`, `@vitest/coverage-v8`, `jsdom`, and bump `cross-env` (7.x to 10.x). Bump `@testing-library/dom` (10.1 to 10.4), `@testing-library/jest-dom` (6.4 to 6.8), and `@testing-library/react` (15 to 16) so the `/vitest` subpath is available. Delete the root `jest.config.js` and every per-package `jest.config.js` shim that re-exported it.
- **vitest config:** A single root `vitest.config.ts` shared by all packages. Uses `jsdom`; aliases `.scss`/`.css` to `identity-obj-proxy` via the anchored `^.*\.s?css$` regex; uses an exact-match regex (`/^@openmrs\/esm-framework$/`) for the framework so it does not capture the `/mock` subpath; aliases `@carbon/charts-react`, `react-i18next`, `tools`, `__mocks__`, and `uuid` the way the jest config did via `moduleDirectories` + `moduleNameMapper`; pins `fakeTimers.toFake` so it does not break React's scheduler; excludes the `packages/esm-form-entry-app` subtree from discovery so vitest ignores the Angular tests. Sets `process.env.TZ = 'UTC'` so the test scripts can drop `cross-env`. Ports the coverage configuration (80% global threshold and the `**/src/**/*.component.tsx` scope) from the deleted jest config.
- **Setup:** Update `tools/setup-tests.ts` to load `@testing-library/jest-dom/vitest`, swap every `jest.*` mock for `vi.*`, and rewrite the `ResizeObserver` and `IntersectionObserver` mocks as ES classes (vitest's `vi.fn().mockImplementation(...)` is not constructable, so React's effect-time `new ResizeObserver(...)` call fails when the mock is a plain function).
- **Per-package scripts:** Update every package's test scripts to `yarn run -T vitest run --passWithNoTests` etc. (`esm-form-entry-app` is excluded.) The `-T` flag tells yarn 4 to resolve the binary from the top-level workspace.
- **Test files (135 of them):** Convert to the vitest API: `jest.*` swapped for `vi.*`, `jest.Mock`/`jest.SpyInstance` type aliases replaced with `Mock`/`MockInstance` from vitest. Explicit `import { vi, describe, it, expect, ... } from 'vitest'` added to every test file. `vi.mock` factories that used `jest.requireActual` rewritten as the async `vi.mock(path, async () => ({ ...((await vi.importActual(path)) as object), override }))` form. Several `vi.mock(name, () => () => <jsx />)` factories wrapped in `{ default: ... }` so vitest accepts the module object shape. `vi.spyOn(...).mockImplementation()` calls passed an explicit `() => {}` since vitest's `mockImplementation` requires an argument. `xit` and `xdescribe` rewritten as `it.skip` / `describe.skip`. `mockResolvedValueOnce` casts changed from `ReturnType<typeof X>` to `Awaited<ReturnType<typeof X>>` because vitest's mock typing expects the unwrapped Promise value.
- **`encounters-table` editability tests:** The "Encounter editability" describe block previously relied on the implicit jest behaviour where `jest.restoreAllMocks` only undid spies; vitest's `vi.restoreAllMocks` resets all mocks including `vi.fn` implementations. Replace the blanket `vi.restoreAllMocks()` with a scoped `dateNowSpy.mockRestore()`, and add an explicit `mockUseConfig.mockImplementation` in the first editability test.
- **Accessible-name selectors:** Tolerate the newer testing-library accessible-name computation in the `visit-tag` tooltip and the Carbon danger-button "danger Delete" selectors by switching to whitespace-tolerant regexes.
- **Task workspace timers:** In `esm-patient-task-list-app` `task-list-view` and `task-details-view` tests, switch `vi.useFakeTimers()` to `vi.useFakeTimers({ shouldAdvanceTime: true })` so `userEvent.click` resolves while system time is mocked.
- **vitals-overview:** Replace the inline `ResizeObserver` mock with the class form for the same constructable-mock reason.
- **Two skips:** `past-medications` and `active-medications` "renders a tabular overview" cases hit a Carbon DataTable column rendering difference under jsdom + the newer testing-library. Flagged with `.skip` and a `TODO`.
- **Delete** `jest.config.js` and every per-package `jest.config.js` shim.

## Screenshots

N/A — tooling change only.

## Related Issue

N/A

## Other

911 tests pass across 20 packages (esm-form-entry-app left on `ng test`; small handful of skips documented above). All packages green under tsc.


NethmiRodrigo (approved):
Thanks @denniskigen!

ibacher (changes requested):
Hold off on this. Revising some things. Same for management.

ibacher (approved):

denniskigen:
Great catch! Thanks, @ibacher!

PR #3313, opened 2026-05-15
(fix) O3-5653: Wire body site and status concepts in config schema

dilankavishka:
## Requirements

## Summary

This PR is for wire the `bodySiteConceptUuid` and the `statusConceptUuid` in the config schema.

## Screenshots

1. Body site
Screenshot 2026-05-15 at 10 14 25

2. Status
Screenshot 2026-05-15 at 10 14 34

## Related Issue

https://issues.openmrs.org/browse/O3-5653

## Other

N/A

### DIFF
--- a/packages/esm-patient-procedures-app/src/config-schema.ts
+++ b/packages/esm-patient-procedures-app/src/config-schema.ts
@@ -32,12 +32,12 @@ export const configSchema = {
     _type: Type.UUID,
     _description:
       'Value used to constrain the body-site concept search. Its meaning is determined by `bodySiteConceptSourceType`. For `Concept classs`, provide the concept class name.',
-    _default: '8d491c7a-c2cc-11de-8d13-0010c6dffd0f',
+    _default: '8d491c7a-c2cc-11de-8d13-0010c6dffd0f', // Anatomy concept class
   },
   bodySiteConceptSourceType: {
     _type: Type.String,
     _description: sourceTypeDescription,
-    _default: 'Concept set',
+    _default: 'Concept class',
     _validators: [validators.oneOf(conceptSourceTypes)],
   },
   // TODO: Update the following value once the concept set is available for procedure status, currently using
--- a/packages/esm-patient-procedures-app/src/config-schema.ts
+++ b/packages/esm-patient-procedures-app/src/config-schema.ts
@@ -40,16 +40,17 @@ export const configSchema = {
     _default: 'conceptClass',
     _validators: [validators.oneOf(conceptSourceTypes)],
   },
+  // TODO: Update the following value once the procedure-status concept set is available.
   statusConceptUuid: {
     _type: Type.UUID,
     _description:
       'UUID used to constrain the procedure-status concept search. Its meaning is determined by `statusConceptSourceType`.',
-    _default: '167157AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
+    _default: '365b8d02-2786-4ac4-a8b6-2bb4f22e4bc2',
   },
   statusConceptSourceType: {
     _type: Type.String,
     _description: sourceTypeDescription,
-    _default: 'conceptSet',
+    _default: 'answerTo',
     _validators: [validators.oneOf(conceptSourceTypes)],
   },
   durationUnitConceptUuid: {
--- a/packages/esm-patient-procedures-app/src/config-schema.ts
+++ b/packages/esm-patient-procedures-app/src/config-schema.ts
@@ -28,7 +28,7 @@ export const configSchema = {
     _default: 'Concept class',
     _validators: [validators.oneOf(conceptSourceTypes)],
   },
-  bodySiteConceptSource: {
+  bodySiteConceptUuid: {
     _type: Type.UUID,
     _description:
       'Value used to constrain the body-site concept search. Its meaning is determined by `bodySiteConceptSourceType`. For `Concept classs`, provide the concept class name.',
--- a/packages/esm-patient-procedures-app/src/config-schema.ts
+++ b/packages/esm-patient-procedures-app/src/config-schema.ts
@@ -29,29 +29,27 @@ export const configSchema = {
     _validators: [validators.oneOf(conceptSourceTypes)],
   },
   bodySiteConceptUuid: {
-    _type: Type.UUID,
+    _type: Type.String,
     _description:
-      'UUID used to constrain the body-site concept search. Its meaning is determined by `bodySiteConceptSourceType`.',
-    _default: '',
+      'UUID or concept class name used to constrain the body-site concept search. Its meaning is determined by `bodySiteConceptSourceType`.',
+    _default: 'Anatomy',
   },
   bodySiteConceptSourceType: {
     _type: Type.String,
     _description: sourceTypeDescription,
-    _default: 'any',
+    _default: 'conceptClass',
     _validators: [validators.oneOf(conceptSourceTypes)],
   },
-  // TODO: Update the following value once the concept set is available, currently using
-  //  https://app.openconceptlab.org/#/orgs/CIEL/sources/CIEL/concepts/168857/ for demo
   statusConceptUuid: {
     _type: Type.UUID,
     _description:
       'UUID used to constrain the procedure-status concept search. Its meaning is determined by `statusConceptSourceType`.',
-    _default: '365b8d02-2786-4ac4-a8b6-2bb4f22e4bc2',
+    _default: '167157AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
   },
   statusConceptSourceType: {
     _type: Type.String,
     _description: sourceTypeDescription,
-    _default: 'answerTo',
+    _default: 'conceptSet',
     _validators: [validators.oneOf(conceptSourceTypes)],
   },
   durationUnitConceptUuid: {
--- a/packages/esm-patient-procedures-app/src/config-schema.ts
+++ b/packages/esm-patient-procedures-app/src/config-schema.ts
@@ -1,13 +1,13 @@
 import { Type, validators } from '@openmrs/esm-framework';
 
-export const conceptSourceTypes = ['conceptClass', 'conceptSet', 'answerTo', 'any'] as const;
+export const conceptSourceTypes = ['Concept class', 'Concept set', 'Answer to', 'any'] as const;
 export type ConceptSourceType = (typeof conceptSourceTypes)[number];
 
 const sourceTypeDescription =
   'How the paired UUID filters concept search results: ' +
-  '"conceptClass" filters by concept class (REST `class`); ' +
-  '"conceptSet" returns set members (REST `memberOf`); ' +
-  '"answerTo" returns answers to the given coded question (REST `answerTo`); ' +
+  '"Concept class" filters by concept class (REST `class`); ' +
+  '"Concept set" returns set members (REST `memberOf`); ' +
+  '"Answer to" returns answers to the given coded question (REST `answerTo`); ' +
   '"any" ignores the UUID and searches all concepts.';
 
 export const configSchema = {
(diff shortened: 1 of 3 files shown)


dilankavishka:
@NethmiRodrigo @denniskigen Could you review this please?

packages/esm-patient-procedures-app/src/config-schema.ts
```diff
@@ -29,29 +29,27 @@ export const configSchema = {
     _validators: [validators.oneOf(conceptSourceTypes)],
   },
   bodySiteConceptUuid: {
-    _type: Type.UUID,
+    _type: Type.String,
     _description:
-      'UUID used to constrain the body-site concept search. Its meaning is determined by `bodySiteConceptSourceType`.',
-    _default: '',
+      'UUID or concept class name used to constrain the body-site concept search. Its meaning is determined by `bodySiteConceptSourceType`.',
+    _default: 'Anatomy',
```
NethmiRodrigo: Why do we use the concept class name here instead of the UUID?

NethmiRodrigo (approved):

packages/esm-patient-procedures-app/src/config-schema.ts
dilankavishka: If I got correctly, the concept class assigned as 'Anatomy' to all body site concepts. Because of this, we have to filter the dropdown dynamically using the class name ('Anatomy') rather than a Concept Set UUID.

Here : https://openmrs.atlassian.net/browse/O3-5652?focusedCommentId=194703

packages/esm-patient-procedures-app/src/config-schema.ts
```diff
@@ -29,29 +29,27 @@ export const configSchema = {
...
     _validators: [validators.oneOf(conceptSourceTypes)],
   },
-  // TODO: Update the following value once the concept set is available, currently using
-  //  https://app.openconceptlab.org/#/orgs/CIEL/sources/CIEL/concepts/168857/ for demo
   statusConceptUuid: {
     _type: Type.UUID,
     _description:
       'UUID used to constrain the procedure-status concept search. Its meaning is determined by `statusConceptSourceType`.',
-    _default: '365b8d02-2786-4ac4-a8b6-2bb4f22e4bc2',
+    _default: '167157AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
```
denniskigen, line 49: This UUID resolves to "Medication dispense status" ([OCL link](https://app.openconceptlab.org/#/search/?q=167157&isTable=true&isList=false&page=1&type=concepts)) — it's a ConvSet whose FHIR mapping is http://hl7.org/fhir/CodeSystem/medicationdispense-status and whose members are: Preparation, In progress, Cancelled, On hold, Completed, Entered in error report status, Discontinued, Refused, Unknown.

Screenshot 2026-05-15 at 12 26 02

Several of those are medication-specific (notably "Refused", "Preparation", "Discontinued") and don't make sense to me as procedure statuses. The HL7 set we actually want is FHIR procedure-status / event-status: preparation, in-progress, not-done, on-hold, stopped, completed, entered-in-error, unknown.

Could you find the CIEL ConvSet whose members align with the FHIR procedure-status value set and use that UUID instead?

dilankavishka: @VeronicaMuthee Could you help me with this please?

cc: @jayasanka-sack

packages/esm-patient-procedures-app/src/procedures/procedures-form.component.tsx
```diff
@@ -128,14 +128,25 @@ const ProceduresFormComponent: React.FC<ProceduresFormComponentProps> = ({
     { uuid: bodySiteConceptUuid, sourceType: bodySiteConceptSourceType },
     getValues('bodySite'),
   );
-  const { searchResults: statusOptions } = useConceptSearch('', {
+  const { searchResults: rawStatusOptions } = useConceptSearch('', {
     uuid: statusConceptUuid,
     sourceType: statusConceptSourceType,
   });
-  const { searchResults: durationUnitOptions } = useConceptSearch('', {
+  const statusOptions = useMemo(() => {
```
denniskigen: Question, not a blocker: is this deduplication fixing an observed duplicate-results bug from the REST response, or is it defensive?

If duplicates were observed, a short comment naming the source would help future readers understand why we normalize here. If this is only defensive, we may not need the extra mapping step in the form component.

packages/esm-patient-procedures-app/src/config-schema.ts
denniskigen: I traced this through the REST implementation, and this needs to be a UUID for the current code path.

When the user types in the body-site search, `/concept?name=...&searchType=fuzzy&class=Anatomy` is handled by `ConceptSearchHandler1_8`, which resolves the class with `conceptService.getConceptClassByUuid(conceptClass)`. Passing `"Anatomy"` returns `null`, so the class filter is silently dropped and the search is effectively unconstrained.

When no name is typed, `/concept?class=Anatomy` falls through to `ConceptResource1_8#doSearch`, which only handles `answerTo` and `memberOf`; it does not read `class`.

Please use the Anatomy concept-class UUID instead: `8d491c7a-c2cc-11de-8d13-0010c6dffd0f`. This matches the existing `procedureConceptUuid` pattern, where `8d490bf4-...` is the Procedure class UUID.

denniskigen (changes requested):
Thanks for picking this up, @dilankavishka. I think there are still two correctness issues to resolve before merge:

1. `bodySiteConceptUuid = "Anatomy"` does not constrain the body-site search with the current REST implementation. The concept search path expects `class` to be a concept-class UUID, and the no-name path does not handle `class` at all. I left details on the existing line 35 thread.

2. The status concept-set issue from the existing line 47 thread is still unresolved: `167157...` is the CIEL Medication dispense status set, mapped to `http://hl7.org/fhir/CodeSystem/medicationdispense-status`. We need the CIEL set aligned to FHIR Procedure/event status instead.

packages/esm-patient-procedures-app/src/config-schema.ts
denniskigen: Suggested concrete fix:

```suggestion
  bodySiteConceptUuid: {
    _type: Type.UUID,
    _description:
      'UUID used to constrain the body-site concept search. Its meaning is determined by `bodySiteConceptSourceType`.',
    _default: '8d491c7a-c2cc-11de-8d13-0010c6dffd0f',
```

This keeps `bodySiteConceptSourceType: "conceptClass"`, but gives the REST `class` parameter the UUID it expects.

packages/esm-patient-procedures-app/src/config-schema.ts
```diff
@@ -29,29 +29,27 @@ export const configSchema = {
...
   },
   statusConceptSourceType: {
     _type: Type.String,
     _description: sourceTypeDescription,
-    _default: 'answerTo',
+    _default: 'conceptSet',
```
denniskigen: Until we have the correct CIEL ConvSet for FHIR Procedure/event status, I would keep the existing TODO/demo wiring rather than switching to the medication-dispense set:

```suggestion
  // TODO: Update the following value once the procedure-status concept set is available.
  statusConceptUuid: {
    _type: Type.UUID,
    _description:
      'UUID used to constrain the procedure-status concept search. Its meaning is determined by `statusConceptSourceType`.',
    _default: '365b8d02-2786-4ac4-a8b6-2bb4f22e4bc2',
  },
  statusConceptSourceType: {
    _type: Type.String,
    _description: sourceTypeDescription,
    _default: 'answerTo',
```

This avoids presenting the medication-dispense status set as the procedure status set while the correct terminology is still being identified.

packages/esm-patient-procedures-app/src/procedures/procedures-form.component.tsx
dilankavishka: Yes! this fix is because getting the duplication and sure I'll put a comment in there

packages/esm-patient-procedures-app/src/config-schema.ts
dilankavishka: Updated.

packages/esm-patient-procedures-app/src/config-schema.ts
dilankavishka: Reverted.

VeronicaMuthee:
> Thanks for picking this up, @dilankavishka. I think there are still two correctness issues to resolve before merge:
> 
> 1. `bodySiteConceptUuid = "Anatomy"` does not constrain the body-site search with the current REST implementation. The concept search path expects `class` to be a concept-class UUID, and the no-name path does not handle `class` at all. I left details on the existing line 35 thread.
> 2. The status concept-set issue from the existing line 47 thread is still unresolved: `167157...` is the CIEL Medication dispense status set, mapped to `http://hl7.org/fhir/CodeSystem/medicationdispense-status`. We need the CIEL set aligned to FHIR Procedure/event status instead.

@denniskigen, the codes are similar, though with the event status codes ~  https://fhir.hl7.org/fhir/valueset-event-status.html. @jayasanka-sack, is it Andy who recommended the use of medicationdispense-status concept? Meanwhile i have asked Andy if we can have event status added in OCL.

packages/esm-patient-procedures-app/src/config-schema.ts
```diff
@@ -32,16 +32,15 @@ export const configSchema = {
...
   bodySiteConceptSourceType: {
     _type: Type.String,
     _description: sourceTypeDescription,
-    _default: 'any',
+    _default: 'conceptClass',
     _validators: [validators.oneOf(conceptSourceTypes)],
   },
-  // TODO: Update the following value once the concept set is available, currently using
-  //  https://app.openconceptlab.org/#/orgs/CIEL/sources/CIEL/concepts/168857/ for demo
+  // TODO: Update the following value once the procedure-status concept set is available.
```
ibacher: Doesn't seem like this comment should be touched here.

packages/esm-patient-procedures-app/src/config-schema.ts
```diff
@@ -32,16 +32,15 @@ export const configSchema = {
...
     _description:
       'UUID used to constrain the body-site concept search. Its meaning is determined by `bodySiteConceptSourceType`.',
-    _default: '',
+    _default: '8d491c7a-c2cc-11de-8d13-0010c6dffd0f',
   },
   bodySiteConceptSourceType: {
     _type: Type.String,
     _description: sourceTypeDescription,
-    _default: 'any',
+    _default: 'conceptClass',
```
ibacher: 1. It's not clear why we're changing this
2. The point of the configuration file is to be human readable and editable. We should be using "Concept class" (or any capitalization of that) as a value. Requiring specific strings seems like a bad idea.
3. I think it's a category error to try to construct concept sets from concept classes. This should just default to a concept set.

packages/esm-patient-procedures-app/src/config-schema.ts
```diff
@@ -32,16 +32,15 @@ export const configSchema = {
     _type: Type.UUID,
     _description:
       'UUID used to constrain the body-site concept search. Its meaning is determined by `bodySiteConceptSourceType`.',
-    _default: '',
+    _default: '8d491c7a-c2cc-11de-8d13-0010c6dffd0f',
```
ibacher: I think this configuration option is poorly named. Concept classes should generally be identified by their name (which is unique) and not their UUID.

dilankavishka: Updated like this: `bodySiteConceptSource`

packages/esm-patient-procedures-app/src/config-schema.ts
dilankavishka: Updated

packages/esm-patient-procedures-app/src/config-schema.ts
dilankavishka: Updated

dilankavishka:
@denniskigen @jayasanka-sack @ibacher Could you please take another look at this?

packages/esm-patient-procedures-app/src/config-schema.ts
```diff
@@ -25,33 +25,33 @@ export const configSchema = {
   procedureConceptSourceType: {
     _type: Type.String,
     _description: sourceTypeDescription,
-    _default: 'conceptClass',
+    _default: 'Concept class',
     _validators: [validators.oneOf(conceptSourceTypes)],
   },
-  bodySiteConceptUuid: {
+  bodySiteConceptSource: {
```
jayasanka-sack:
```suggestion
  bodySiteConceptUuid: {
```

packages/esm-patient-procedures-app/src/procedures/procedures-form.component.tsx
```diff
@@ -128,15 +128,30 @@ const ProceduresFormComponent: React.FC<ProceduresFormComponentProps> = ({
     { uuid: bodySiteConceptUuid, sourceType: bodySiteConceptSourceType },
     getValues('bodySite'),
   );
-  const { searchResults: statusOptions } = useConceptSearch('', {
+  const { searchResults: rawStatusOptions } = useConceptSearch('', {
     uuid: statusConceptUuid,
     sourceType: statusConceptSourceType,
   });
-  const { searchResults: durationUnitOptions } = useConceptSearch('', {
+  // Deduplicate results by UUID. The OpenMRS REST API can sometimes return duplicate concepts in searches.
```
jayasanka-sack: this can be handled in useConceptSearch hook. Fixed at b8453bab

packages/esm-patient-procedures-app/src/config-schema.ts
jayasanka-sack: @ibacher I'm not sure about this. The concept endpoint expect the concept class's UUID. 
https://github.com/openmrs/openmrs-module-webservices.rest/blob/5da3a8603977eb23368ac9c2a905ac7f7a376c8f/omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/search/openmrs1_8/ConceptSearchHandler1_8.java#L125

jayasanka-sack (approved):
Thanks for getting this done, @dilankavishka!

PR #3316, opened 2026-05-20
(fix) Lower task list backend dependency

denniskigen:
## Requirements

## Summary

Lowers the patient task list app backend dependency from the unreleased `tasks >=2.0.0` requirement to `tasks >=1.1.0-SNAPSHOT`. Reviewing `openmrs-module-tasks` showed that `1.1.0-SNAPSHOT` is the published backend version that supports the current rationale payload shape in `activity.detail.reasonCode`.

This keeps the frontend dependency metadata aligned with the backend module that is available to the reference application distro. It also unblocks adding `@openmrs/esm-patient-task-list-app` to the core E2E environment in openmrs/openmrs-esm-core#1779, where implementer tools currently reports the unsatisfied `tasks >=2.0.0` requirement as an unresolved backend dependency toast.

## Other

### DIFF
--- a/packages/esm-patient-task-list-app/src/routes.json
+++ b/packages/esm-patient-task-list-app/src/routes.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.openmrs.org/routes.schema.json",
   "backendDependencies": {
-    "tasks": ">=2.0.0",
+    "tasks": ">=1.1.0-SNAPSHOT",
     "webservices.rest": ">=3.1.0"
   },
   "modals": [


ibacher:
This will be the 2.0.0 release, which is why I put that in though. It's a breaking change to the API, hence, major version bump.

denniskigen:
> This will be the 2.0.0 release, which is why I put that in though. It's a breaking change to the API, hence, major version bump.

Thanks, Ian. That makes a lot of sense.

PR #3318, opened 2026-05-21
(chore) Use reusable build and release GHA workflows

NethmiRodrigo:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspection.

## Summary

Replaces inline CI job definitions in `.github/workflows/ci.yml` with calls to reusable workflows from [openmrs/openmrs-contrib-gha-workflows](https://github.com/openmrs/openmrs-contrib-gha-workflows).

Also fixes a build failure caused by a duplicate `--concurrency` flag. The reusable `build-frontend-module` workflow infers the verify command for monorepos as `yarn verify --concurrency=5`, but the `verify` script in `package.json` already includes `--concurrency=5`, causing Turbo to receive the flag twice and fail with:

```
the argument '--concurrency <CONCURRENCY>' cannot be used multiple times
```

The fix passes an explicit `verify_command: yarn verify` input to the workflow, bypassing the inference step so the flag is only applied once (via the `package.json` script).

## Related

Follows the pattern established in https://github.com/openmrs/openmrs-esm-dispensing-app/pull/812.


denniskigen:
This drops the existing `patch/*` prerelease publishing behavior.

The caller still triggers on `patch/*`, and the reusable workflow's prerelease job will run for those refs. But because `support_patch_branches` defaults to `false`, patch branches are versioned as `-pre.${run_number}` instead of `-patch.${run_number}`, and the `Publish pre-release (patch)` step is skipped. The old workflow explicitly ran `yarn run ci:prepublish-patch`, which publishes with the `patch` npm tag.

Can we pass `support_patch_branches: true` here?

```yaml
release:
  needs: build
  uses: openmrs/openmrs-contrib-gha-workflows/.github/workflows/release-frontend-module.yml@main
  with:
    support_patch_branches: true
  secrets:
    NPM_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
    BOT_PAT: ${{ secrets.OMRS_BOT_GH_TOKEN }}
```

NethmiRodrigo:
> This drops the existing `patch/*` prerelease publishing behavior.
> 
> The caller still triggers on `patch/*`, and the reusable workflow's prerelease job will run for those refs. But because `support_patch_branches` defaults to `false`, patch branches are versioned as `-pre.${run_number}` instead of `-patch.${run_number}`, and the `Publish pre-release (patch)` step is skipped. The old workflow explicitly ran `yarn run ci:prepublish-patch`, which publishes with the `patch` npm tag.
> 
> Can we pass `support_patch_branches: true` here?
> 
> ```yaml
> release:
>   needs: build
>   uses: openmrs/openmrs-contrib-gha-workflows/.github/workflows/release-frontend-module.yml@main
>   with:
>     support_patch_branches: true
>   secrets:
>     NPM_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
>     BOT_PAT: ${{ secrets.OMRS_BOT_GH_TOKEN }}
> ```

Thanks for catching this! Fixed

NethmiRodrigo:
It's the tasks list e2e tests that are failing

denniskigen:
I discussed this potential explanation for the tasks module-related e2e test failures with @ibacher yesterday:

The issue is in `tasks-omod:2.0.0`, not the CI backend build job. In 2.0.0, the CarePlan controller was moved to: `org.openmrs.module.tasks.web.controller` and mapped as `/ws/rest/v1/tasks/careplan`.

That is wrong for OpenMRS Webservices REST. The `/ws` prefix is added by the Webservices REST servlet, and that dispatcher scans: `org.openmrs.module.webservices.rest.web`. So the controller is not registered where `/openmrs/ws/rest/v1/tasks/careplan` is actually served. That matches the CI 404s.

Backend fix:

1. Restore/move the controller to: `omod/src/main/java/org/openmrs/module/webservices/rest/web/controller/tasks/TasksCarePlanController.java`

2. Use this package: `org.openmrs.module.webservices.rest.web.controller.tasks`

3. Map it as: @RequestMapping("/rest/v1/tasks/careplan"), not: @RequestMapping("/ws/rest/v1/tasks/careplan")

4. Keep the improved 2.0.0 controller logic, but expose it through the Webservices REST package.

5. Update tests so they do not only use standalone MockMvc against /ws/rest/v1/... They should cover the Webservices REST controller path, or at minimum instantiate TasksCarePlanController and call /rest/v1/tasks/careplan.

Then release a patched tasks-omod version, for example 2.0.1, and bump tasks.version in the distro to that fixed version.

So the fix is not to run the backend build job in CI. The backend module itself needs a patch because the 2.0.0 release stopped registering the task endpoint in the real Webservices REST dispatcher.

denniskigen (approved):
Thanks, Nethmi!

PR #3320, opened 2026-05-21
(feat) O3-5077: Support recording clinical procedures

jayasanka-sack:
## Requirements

## Summary

A microfrontend for [OpenMRS 3](https://o3-docs.openmrs.org) that records, displays, and manages clinical procedures in the patient chart. 
Jira: [O3-5077](https://openmrs.atlassian.net/browse/O3-5077)

## Screenshots

https://github.com/user-attachments/assets/a1d0e9a9-9676-4ab5-b74a-1b82dd3e2abf

## Related Issue

https://openmrs.atlassian.net/browse/O3-5077

## Features

- **Procedure overview widget** — compact table in the patient summary that lists each procedure with its date, paginated according to the configured page size.
- **Procedures detailed summary** — full-page table that adds procedure type, body site, start/end timestamps, status, and inline notes. Rows are expandable to show additional detail without leaving the page.
- **Record & edit form** — a side-panel workspace for creating and updating procedures. Supports:
  - Fuzzy concept search for procedure name, body site, status, and duration unit — all driven by configurable concept UUIDs.
  - Exact start date/time or estimated start date (year + optional month) for procedures where the precise date is not known.
  - End date/time and duration with unit.
  - Procedure type selection from the OpenMRS procedure type registry.
  - Free-text notes.
- **Delete confirmation** — modal that voids a procedure record.

---

#### Backend

This app requires the [openmrs-module-emrapi](https://github.com/openmrs/openmrs-module-emrapi) module, which exposes the `/ws/rest/v1/procedure` REST endpoints used for all CRUD operations.

---

## Configuration

The app is configured through the [OpenMRS config system](https://o3-docs.openmrs.org/docs/frontend-modules/loading-modules#module-configuration). All keys live under `@openmrs/esm-patient-procedures-app`.

| Key | Type | Default | Description |
|---|---|---|---|
| `overviewPageSize` | Number | `5` | Rows per page in the procedures overview widget. |
| `detailedViewPageSize` | Number | `10` | Rows per page in the procedures detailed summary view. |
| `procedureConceptUuid` | UUID | `8d490bf4-…` | Scopes the procedure concept search. |
| `procedureConceptSourceType` | String | `Concept class` | How `procedureConceptUuid` filters results: `Concept class`, `Concept set`, `Answer to`, or `any`. |
| `bodySiteConceptUuid` | UUID | `8d491c7a-…` (Anatomy) | Scopes the body-site concept search. |
| `bodySiteConceptSourceType` | String | `Concept class` | Same options as above. |
| `statusConceptUuid` | UUID | `167157AAAA…` | Scopes the status concept search. |
| `statusConceptSourceType` | String | `Concept set` | Same options as above. |
| `durationUnitConceptUuid` | UUID | `1732AAAAA…` | Scopes the duration-unit concept search. |
| `durationUnitConceptSourceType` | String | `Answer to` | Same options as above. |

---

## Form fields

| Field | Required | Notes |
|---|---|---|
| Procedure | Yes | Fuzzy concept search |
| Procedure type | Yes | Dropdown from `/ws/rest/v1/proceduretype` |
| Body site | Yes | Fuzzy concept search |
| Start date known? | — | Toggle between exact datetime and estimated date |
| Start date/time | Yes (if date is known) | Full date + time picker |
| Estimated start date | Yes (if date is unknown) | Year required, month optional |
| End date/time | No | Must be on or after start date |
| Duration | No | Positive integer; requires a unit when set |
| Duration unit | Conditional | Required when duration is provided |
| Status | Yes | Concept search |
| Notes | No | Free text |

---

## Estimated start date

When the exact start date of a procedure is not known — common for historical procedures — clinicians can switch the start date toggle to **No** and enter a year (required) and month (optional). The stored value is an ISO partial date string (`YYYY` or `YYYY-MM`). 

---
### Related

- Ticket: [O3-5077](https://openmrs.atlassian.net/browse/O3-5077)
- Requirements: [Procedure History — Confluence](https://openmrs.atlassian.net/wiki/spaces/projects/pages/589267006/Procedure+History)
- Architecture discussion: [OpenMRS Talk](https://talk.openmrs.org/t/procedures-app-architecture-questions/47948/19)
- Backend module: [openmrs-module-emrapi](https://github.com/openmrs/openmrs-module-emrapi/)

[O3-5077]: https://openmrs.atlassian.net/browse/O3-5077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[O3-5077]: https://openmrs.atlassian.net/browse/O3-5077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/esm-patient-procedures-app/src/workspaces/procedures-form/procedures.form.component.tsx
```diff
@@ -0,0 +1,458 @@
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
```
jayasanka-sack, line 1: form.component.tsx

jayasanka-sack:
@NethmiRodrigo @denniskigen I know this is a pretty large PR, but could you please take a look but whenever you get a chance? A few parts have already been reviewed in separate PRs so hopefully that helps make it a bit easier to go through. Thanks!

packages/esm-patient-procedures-app/src/procedures.resource.ts
```diff
@@ -0,0 +1,130 @@
+import { useEffect, useState } from 'react';
+import useSWR, { useSWRConfig } from 'swr';
+import { openmrsFetch, restBaseUrl, useDebounce } from '@openmrs/esm-framework';
+import { type ConceptSourceType } from './config-schema';
+import {
+  type ConceptReference,
+  type ProcedureApiResponse,
+  type ProcedureTypeApiResponse,
+  type RawProcedure,
+} from './types';
```
NethmiRodrigo: nit:

```suggestion
import type {
  ConceptReference,
  ProcedureApiResponse,
  ProcedureTypeApiResponse,
  RawProcedure,
} from './types';
```

packages/esm-patient-procedures-app/src/procedures.resource.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+  if (source.uuid && source.sourceType !== 'any') {
+    params.set(sourceTypeToRestParam[source.sourceType], source.uuid);
+  }
+  return `${restBaseUrl}/concept?${params.toString()}`;
+};
+
+export const useProcedureTypes = () => {
+  const url = `${restBaseUrl}/proceduretype?v=full`;
+  const { data, isLoading } = useSWR<{ data: ProcedureTypeApiResponse }, Error>(url, openmrsFetch);
```
NethmiRodrigo, line 29: Shouldn’t we return the error here and not render the dropdown if there is an error and display an error message instead? Same for `useConceptSearch` and `useConceptById`?

packages/esm-patient-procedures-app/src/procedures.resource.ts
```diff
@@ -0,0 +1,130 @@
...
+    totalCount: data?.data?.totalCount ?? 0,
+    error,
+    isLoading,
+    isValidating,
+  };
+};
+
+export const deleteProcedure = async (procedureId: string) => {
+  return await openmrsFetch(`${restBaseUrl}/procedure/${procedureId}`, {
+    method: 'DELETE',
```
denniskigen, line 85: The emrapi `ProcedureResource` requires a non-empty void reason when deleting a procedure. A bare `DELETE` will fail before voiding the record.

Could we pass a reason to the endpoint, probably collected or supplied by `delete-procedure.modal.tsx`, and add a test covering the request shape?

packages/esm-patient-procedures-app/src/components/concept-search-field/concept-search-field.component.tsx
```diff
@@ -0,0 +1,109 @@
...
+    <>
+      <ResponsiveWrapper>
+        <Search
+          labelText={label}
+          placeholder={placeholder}
+          onChange={(e) => {
+            field.setSearchTerm(e.target.value);
+            field.setSelectedConcept(null);
+          }}
+          onClear={field.clear}
```
denniskigen: Typing into the field clears `selectedConcept` locally, and `onClear` calls `field.clear()`, but neither path clears the RHF-tracked UUID in the parent.

That means a select, clear, submit flow can still pass Zod validation with the stale UUID, then `procedures.form.component.tsx` dereferences `selectedConcept!.uuid` while it is null. Could we propagate `onChange(null)` from the typing and clear paths, or make `selectedConcept` the single source of truth? A test for select, clear, then submit would help lock this down.

packages/esm-patient-procedures-app/src/routes.json
```diff
@@ -0,0 +1,46 @@
+{
+  "$schema": "https://json.openmrs.org/routes.schema.json",
+  "workspaces2": [
```
denniskigen, line 7: Could we add `backendDependencies` for this app? The procedures UI depends on the emrapi `/ws/rest/v1/procedure` and `/proceduretype` endpoints. I verified those resources landed in emrapi `3.4.0`, so something like this would make missing backend support fail early and clearly:

```json
"backendDependencies": {
  "webservices.rest": ">=2.2.0",
  "emrapi": ">=3.4.0"
}
```

packages/esm-patient-procedures-app/src/config-schema.ts
```diff
@@ -0,0 +1,87 @@
...
+    _default: 'Concept class',
+    _validators: [validators.oneOf(conceptSourceTypes)],
+  },
+  // TODO: Update the following value once the concept set is available for procedure status, currently using
+  // https://app.openconceptlab.org/#/orgs/CIEL/sources/CIEL/concepts/167157/ (Medication dispense status)
+  statusConceptUuid: {
+    _type: Type.UUID,
+    _description:
+      'UUID used to constrain the procedure-status concept search. Its meaning is determined by `statusConceptSourceType`.',
+    _default: '167157AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
```
denniskigen, line 54: Could we avoid shipping the medication-dispense status concept as the default for procedures, or link a follow-up ticket here with clear distro configuration guidance? Otherwise an out-of-the-box install may show dispense-status values in the procedure Status dropdown, which could be confusing for clinicians.

packages/esm-patient-procedures-app/src/routes.json
```diff
@@ -0,0 +1,46 @@
...
+      "order": 8,
+      "meta": {
+        "slot": "patient-chart-procedures-dashboard-slot",
+        "path": "procedures"
+      }
+    },
+    {
+      "name": "procedure-delete-confirmation-dialog",
+      "component": "procedureDeleteConfirmationDialog"
+    }
```
denniskigen, line 50: Most sibling chart apps register dialogs under the top-level `modals` array rather than `extensions`. Could we move `procedure-delete-confirmation-dialog` there to match the surrounding route conventions?

packages/esm-patient-procedures-app/src/components/overview/procedures-overview.component.tsx
```diff
@@ -0,0 +1,152 @@
...
+import styles from './procedures-overview.scss';
+
+type ProceduresOverviewProps = {
+  patientUuid: string;
+};
+
+const ProceduresOverview: React.FC<ProceduresOverviewProps> = ({ patientUuid }) => {
+  const { overviewPageSize } = useConfig<ConfigObject>();
+  const { t } = useTranslation();
+  const launchProceduresForm = useCallback(() => launchWorkspace2('procedures-form-workspace'), []);
```
denniskigen: The detailed summary launches this workspace with `{ formContext: 'creating' }`, while the overview launches it without props. It works because the form treats missing `formContext` like create mode, but passing the same prop here would make the two entry points easier to reason about and keep the tests aligned.

packages/esm-patient-procedures-app/src/components/action-menu/procedures-action-menu.component.tsx
```diff
@@ -0,0 +1,54 @@
...
+    const dispose = showModal('procedure-delete-confirmation-dialog', {
+      closeDeleteModal: () => dispose(),
+      procedureUuid,
+      patientUuid,
+    });
+  };
+
+  return (
+    <Layer className={styles.layer}>
+      <OverflowMenu aria-label="Edit or delete procedure" align="left" size={isTablet ? 'lg' : 'sm'} flipped>
```
denniskigen: Could we wrap the overflow menu `aria-label` in `t(...)`? The visible menu item labels are translated already, so this would keep the assistive text consistent for non-English locales.

packages/esm-patient-procedures-app/src/components/overview/procedures-overview.component.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const headerTitle = t('procedures', 'Procedures');
+  const displayText = t('procedures_lower', 'procedures');
+  const pageUrl = `\${openmrsSpaBase}/patient/${patientUuid}/chart/Procedures`;
```
denniskigen: Could we update this link to use `/chart/procedures` to match the dashboard path in `routes.json`? The route metadata now registers the dashboard as lowercase `procedures`, and the chart legacy alias map does not include `Procedures`. As written, clicking See all from the overview can resolve to an unknown dashboard and fall back to the default dashboard instead of opening Procedures.

denniskigen:
Thanks for the work on this. Two inline comments are merge blockers in my view: the missing void reason on delete, and the concept clear path that can submit stale form state and crash. The rest are smaller suggestions.

packages/esm-patient-procedures-app/src/components/detailed-summary/procedures-detailed-summary.component.tsx
```diff
@@ -0,0 +1,218 @@
...
+  );
+
+  const tableRows: ProcedureTableRow[] = useMemo(
+    () =>
+      procedures?.map((p) => ({
+        id: p.uuid,
+        display: p.display,
+        procedureType: p.procedureType.name,
+        bodySite: p.bodySite.display ?? '--',
+        startDateTimeRender: p.estimatedStartDate ?? p.startDateTime,
```
NethmiRodrigo, line 95: Are both of these ISO timestamps or a mixture of YYYY/YYYY-MM partial date strings? If they are a mixture, would the sorting work as intended?

NethmiRodrigo (commented):
Thanks @jayasanka-sack!

Josephkagimu1 (commented):
Hei @jayasanka-sack  , I appreciate the effort for this PR , I kindly request you check again the PR description order to get rid of description errors .

packages/esm-patient-procedures-app/src/components/detailed-summary/procedures-detailed-summary.component.tsx
jayasanka-sack: The startDate is always calculated and persisted as a full timestamp on save in the backend, even when the user enters an estimated date like 2024 or 2024-03. The partial string lives in a separate estimatedStartDate field. The sorting comparator always uses the persisted startDateTime.

packages/esm-patient-procedures-app/src/config-schema.ts
jayasanka-sack: Valid point. However, as the required concept set is not yet available, we decided to go with the one closest to the requirement. A TODO link with the ticket is already linked a few lines above.

packages/esm-patient-procedures-app/src/components/concept-search-field/concept-search-field.component.tsx
jayasanka-sack: Good catch, I updated this to use RHF as the single source of truth. c755dd6

packages/esm-patient-procedures-app/src/routes.json
jayasanka-sack: Fixed with 31278d91248c5992ed3507dc970ab35b47b78575

packages/esm-patient-procedures-app/src/routes.json
jayasanka-sack: fixed with 7d8d3dade59a23262239e91256c728d1786097ba

packages/esm-patient-procedures-app/src/components/overview/procedures-overview.component.tsx
jayasanka-sack: fixed with 7d8d3dade59a23262239e91256c728d1786097ba

packages/esm-patient-procedures-app/src/procedures.resource.ts
jayasanka-sack: fixed with eba13bd4177f2d00a8684e9919a6acb55497b18f

packages/esm-patient-procedures-app/src/procedures.resource.ts
jayasanka-sack: Very valid suggestion. Fixed with 5bc10df49276c847bc2e45bdb93ae1ad347f7281

packages/esm-patient-procedures-app/src/procedures.resource.ts
jayasanka-sack: The reason is an optional field, therefore currently it let users deleting records, but this it is better to let the user to add a reason. I'll open this as a followup PR as soon as this get's merged.

jayasanka-sack (commented):
Thanks @denniskigen @NethmiRodrigo for your reviews. I've updated the PR. Please have another round when you get a chance.

denniskigen (approved):
LGTM

packages/esm-patient-procedures-app/src/config-schema.ts
```diff
@@ -0,0 +1,87 @@
...
+  procedureConceptSourceType: {
+    _type: Type.String,
+    _description: sourceTypeDescription,
+    _default: 'Concept class',
+    _validators: [validators.oneOf(conceptSourceTypes)],
+  },
+  bodySiteConceptUuid: {
+    _type: Type.UUID,
+    _description:
+      'Value used to constrain the body-site concept search. Its meaning is determined by `bodySiteConceptSourceType`. For `Concept class`, provide the concept class name.',
```
denniskigen: Tiny wording fix: since this config value is typed as `Type.UUID` and the default is the Anatomy concept class UUID, this should say UUID rather than name.

```suggestion
      'UUID used to constrain the body-site concept search. Its meaning is determined by `bodySiteConceptSourceType`. For `Concept class`, provide the concept class UUID.',
```

packages/esm-patient-procedures-app/.eslintrc
```diff
@@ -0,0 +1,6 @@
+{
```
denniskigen, line 1: Do we need a package-level ESLint config here? The repo already has a root `.eslintrc`, and almost no sibling packages carry their own config. Since this only makes `@typescript-eslint/no-explicit-any` stricter for this one package, I’d suggest removing it unless there’s a deliberate reason procedures should have a different lint policy from the rest of the monorepo.

packages/esm-patient-procedures-app/src/config-schema.ts
jayasanka-sack: fixed, thanks!

packages/esm-patient-procedures-app/.eslintrc
jayasanka-sack: The root config allows any types, which we eventually need to disable, but that will take some effort. This override prevents lazy contributors from introducing any types in the future for this package.

.github/workflows/e2e.yml
```diff
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - main
+      - feat/procedure-history
```
ibacher, line 7: This should not have been committed.

packages/esm-patient-procedures-app/src/config-schema.ts
```diff
@@ -0,0 +1,87 @@
...
+      'UUID used to constrain the procedure-status concept search. Its meaning is determined by `statusConceptSourceType`.',
+    _default: '167157AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
+  },
+  statusConceptSourceType: {
+    _type: Type.String,
+    _description: sourceTypeDescription,
+    _default: 'Concept set',
+    _validators: [validators.oneOf(conceptSourceTypes)],
+  },
+  durationUnitConceptUuid: {
```
ibacher, line 62: The configuration options for concepts MUST support other reference types than just UUIDs. It safer to mark these as strings, which allows us to support concept maps as well as UUIDs. Concept maps should, in every case, be preferred as defaults (e.g., CIEL:1732 rather than the UUID here).

PR #3332, opened 2026-05-25
(fix) O3-5707: Relaunch patient chart workspace group when visit context changes

donaldkibet:
## Requirements

## Summary

This PR fixes an issue where the `patient-chart workspace` group kept stale `groupProps` after the visit context changed.

Previously, the workspace group was launched only once on mount using a boolean ref. If the visit context later changed — for example when a visit ended or a new one started — the patient chart store was updated but `launchWorkspaceGroup2` was never called again, causing workspaces opened from the patient chart to receive stale visit props.

This change tracks the UUID of the last successfully launched visit context and relaunches the workspace group whenever that UUID changes. The launch state is only recorded after `launchWorkspaceGroup2` returns true, so a user declining to close unsaved workspaces does not suppress future re-launch attempts.

### DIFF
--- a/packages/esm-patient-chart-app/src/patient-chart/patient-chart.resources.ts
+++ b/packages/esm-patient-chart-app/src/patient-chart/patient-chart.resources.ts
@@ -1,15 +1,16 @@
 import { useEffect, useMemo, useRef } from 'react';
 import useSWR from 'swr';
 import {
-  closeWorkspaceGroup2,
   launchWorkspaceGroup2,
   openmrsFetch,
   restBaseUrl,
+  showSnackbar,
   usePatient,
   useVisit,
   type Visit,
 } from '@openmrs/esm-framework';
 import { type PatientWorkspaceGroupProps, usePatientChartStore } from '@openmrs/esm-patient-common-lib';
+import { useTranslation } from 'react-i18next';
 
 const defaultVisitCustomRepresentation =
   'custom:(uuid,display,voided,indication,startDatetime,stopDatetime,' +
@@ -49,6 +50,7 @@ export function useVisitByUuid(visitUuid: string | null, representation: string 
  * @returns
  */
 export function usePatientChartPatientAndVisit(patientUuid: string) {
+  const { t } = useTranslation();
   const { isLoading: isLoadingPatient, patient } = usePatient(patientUuid);
   const {
     patientUuid: storePatientUuid,
@@ -58,7 +60,7 @@ export function usePatientChartPatientAndVisit(patientUuid: string) {
     setVisitContext,
   } = usePatientChartStore(patientUuid);
 
-  const isVisitContextValid = visitContext && visitContext.patient.uuid === patientUuid;
+  const isVisitContextValid = visitContext?.patient.uuid === patientUuid;
   const {
     visit: newVisitContext,
     mutate: newMutateVisitContext,
@@ -70,10 +72,11 @@ export function usePatientChartPatientAndVisit(patientUuid: string) {
     mutate: mutateActiveVisit,
   } = useVisit(isVisitContextValid ? null : patientUuid);
 
-  const isWorkspaceGroupLaunched = useRef(false);
   const launchedVisitContextUuid = useRef<string | null | undefined>(undefined);
+  const isWorkspaceGroupLaunchPending = useRef(false);
 
   useEffect(() => {
+    let cancelled = false;
     const initializeWorkspaceGroup = async () => {
       if (!isValidatingVisitContext && !isValidatingActiveVisit && patient) {
         let groupProps: PatientWorkspaceGroupProps = null;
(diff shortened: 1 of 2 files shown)


jwnasambu (commented):
Thanks @donaldkibet for the work done. Kindly attach the related ticket to this PR please.

packages/esm-patient-chart-app/src/patient-chart/patient-chart.resources.ts
jwnasambu: Kindly aren't we supposed to remove `isWorkspaceGroupLaunched` from the useEffect dependency array entirely?

packages/esm-patient-chart-app/src/patient-chart/patient-chart.resources.ts
```diff
@@ -71,40 +71,51 @@ export function usePatientChartPatientAndVisit(patientUuid: string) {
...
+        setVisitContext(groupProps.visitContext, groupProps.mutateVisitContext);
 
-      setVisitContext(groupProps.visitContext, groupProps.mutateVisitContext);
+        const visitContextUuid = groupProps.visitContext?.uuid ?? null;
+        const visitContextChanged = visitContextUuid !== launchedVisitContextUuid.current;
 
-      if (!isWorkspaceGroupLaunched.current) {
-        launchWorkspaceGroup2('patient-chart', groupProps);
-        isWorkspaceGroupLaunched.current = true;
+        if (!isWorkspaceGroupLaunched.current || visitContextChanged) {
```
denniskigen: Now that we have `launchedVisitContextUuid`, does `isWorkspaceGroupLaunched` still carry independent information? Both refs are only written together inside `if (launched)`, and `launchedVisitContextUuid` starts as `undefined` while `visitContextUuid` is always `string | null` (because of the `?? null`), so `visitContextChanged` is always true on the first run. That makes `!isWorkspaceGroupLaunched.current ||` redundant with the UUID comparison, as far as I can see. Happy to be wrong if there's a case I'm missing.

packages/esm-patient-chart-app/src/patient-chart/patient-chart.resources.ts
```diff
@@ -71,40 +71,51 @@ export function usePatientChartPatientAndVisit(patientUuid: string) {
...
+          const launched = await launchWorkspaceGroup2('patient-chart', groupProps);
+          if (launched) {
```
denniskigen: Race condition: if the effect's dependencies change while the `await launchWorkspaceGroup2` is suspended (which happens when the "close workspaces?" prompt is shown to the user), a second invocation of `initializeWorkspaceGroup` will start, read the still-stale `launchedVisitContextUuid.current`, and fire a second `launchWorkspaceGroup2` call — potentially showing two overlapping confirmation modals. A cancellation flag in the effect cleanup would prevent this:

```ts
let cancelled = false;
const initializeWorkspaceGroup = async () => {
  // ...
  const launched = await launchWorkspaceGroup2('patient-chart', groupProps);
  if (launched && !cancelled) {
    isWorkspaceGroupLaunched.current = true;
    launchedVisitContextUuid.current = visitContextUuid;
  }
};
initializeWorkspaceGroup();
return () => { cancelled = true; };
```

packages/esm-patient-chart-app/src/patient-chart/patient-chart.resources.ts
```diff
@@ -71,40 +71,51 @@ export function usePatientChartPatientAndVisit(patientUuid: string) {
...
+            isWorkspaceGroupLaunched.current = true;
+            launchedVisitContextUuid.current = visitContextUuid;
+          }
+        }
       }
-    }
+    };
+
+    initializeWorkspaceGroup();
   }, [
```
denniskigen, line 142: The promise returned by `initializeWorkspaceGroup()` is fire-and-forget. If `launchWorkspaceGroup2` ever rejects (e.g. `promptForClosingWorkspaces` throws), it becomes an unhandled rejection. At minimum add `.catch(console.error)`, or wrap the `await` in a try/catch inside the function.

donaldkibet: Resolved this

packages/esm-patient-chart-app/src/patient-chart/patient-chart.resources.ts
```diff
@@ -70,41 +72,71 @@ export function usePatientChartPatientAndVisit(patientUuid: string) {
...
 
-      setVisitContext(groupProps.visitContext, groupProps.mutateVisitContext);
+        const visitContextUuid = groupProps.visitContext?.uuid ?? null;
+        const visitContextChanged = visitContextUuid !== launchedVisitContextUuid.current;
 
-      if (!isWorkspaceGroupLaunched.current) {
-        launchWorkspaceGroup2('patient-chart', groupProps);
-        isWorkspaceGroupLaunched.current = true;
+        if (visitContextChanged && !isWorkspaceGroupLaunchPending.current) {
+          isWorkspaceGroupLaunchPending.current = true;
```
denniskigen, line 113: This still has a race: while the first `launchWorkspaceGroup2` is suspended on the close-workspaces prompt, a later effect run for a newer visit context will hit `isWorkspaceGroupLaunchPending.current` and skip launching the updated props. When the first call resolves, that effect is already cancelled, so it also skips updating `launchedVisitContextUuid`; since changing the ref does not trigger another render, the newer visit context is never pushed into the workspace group until some unrelated dependency changes. Could we queue/retry the latest `groupProps` after the pending launch settles, or otherwise compare again after clearing the pending flag?

denniskigen (approved):
Thanks!

PR #3339, opened 2026-05-26
(fix) Relaunch workspace group with latest visit when context changes mid-flight

denniskigen:
## Requirements

## Summary

Builds on #3332. After that fix landed, a secondary issue remained: if the visit context changed *while* a workspace group launch was already in progress, the in-flight launch would complete with stale props and no relaunch would follow.

This PR replaces the inline async logic in `usePatientChartPatientAndVisit` with a `launchLatestWorkspaceGroup` callback that uses a while loop to re-read `latestWorkspaceGroupProps` after each resolved launch. If props changed while a launch was pending, the loop picks up the latest visit context and relaunches immediately — no separate effect run needed.

Also removes the redundant `launchWorkspaceGroup2` call from `VisitForm`. That call was a holdover from before the hook handled relaunching reactively; keeping it caused a double-launch on visit save.

## Related Issue

https://openmrs.atlassian.net/browse/O3-5707

## Other

Adds a regression test that directly exercises the mid-flight race: holds the first launch via a controlled promise, changes the active visit while it's pending, resolves the first launch, and asserts the while loop picks up the latest visit and relaunches.

### DIFF
--- a/packages/esm-patient-chart-app/src/patient-chart/patient-chart.resources.ts
+++ b/packages/esm-patient-chart-app/src/patient-chart/patient-chart.resources.ts
@@ -24,6 +24,23 @@ const defaultVisitCustomRepresentation =
   'attributes:(uuid,display,attributeType:(name,datatypeClassname,uuid),value),' +
   'location:(uuid,name,display))';
 
+type WorkspaceGroupLaunchKey = {
+  patientUuid: string | null;
+  visitContextUuid: string | null;
+};
+
+// The workspace group is current only when it was launched for this patient and visit context.
+function getWorkspaceGroupLaunchKey(groupProps: PatientWorkspaceGroupProps | null): WorkspaceGroupLaunchKey {
+  return {
+    patientUuid: groupProps?.patientUuid ?? null,
+    visitContextUuid: groupProps?.visitContext?.uuid ?? null,
+  };
+}
+
+function workspaceGroupLaunchKeysEqual(a: WorkspaceGroupLaunchKey | null, b: WorkspaceGroupLaunchKey | null) {
+  return a?.patientUuid === b?.patientUuid && a?.visitContextUuid === b?.visitContextUuid;
+}
+
 export function useVisitByUuid(visitUuid: string | null, representation: string = defaultVisitCustomRepresentation) {
   const url = `${restBaseUrl}/visit/${visitUuid}?v=${representation}`;
   const { data, ...rest } = useSWR<{ data: Visit }>(visitUuid ? url : null, openmrsFetch);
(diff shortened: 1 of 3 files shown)


packages/esm-patient-chart-app/src/patient-chart/patient-chart.resources.ts
```diff
@@ -72,12 +72,61 @@ export function usePatientChartPatientAndVisit(patientUuid: string) {
...
+        const launched = await launchWorkspaceGroup2('patient-chart', groupProps);
+
+        if (!isMounted.current) {
+          return;
+        }
 
+        if (!launched) {
+          needsLaunch = false;
+          continue;
+        }
```
ibacher, line 136: Could we add a comment here? I feel like the naming of the variables makes this a little weird...

packages/esm-patient-chart-app/src/patient-chart/patient-chart.resources.ts
```diff
@@ -72,12 +72,61 @@ export function usePatientChartPatientAndVisit(patientUuid: string) {
...
+        const visitContextUuid = groupProps?.visitContext?.uuid ?? null;
+
+        if (visitContextUuid === launchedVisitContextUuid.current) {
+          needsLaunch = false;
+          continue;
+        }
+
+        const launched = await launchWorkspaceGroup2('patient-chart', groupProps);
+
+        if (!isMounted.current) {
```
ibacher, line 128: Shouldn't the `isMounted` check come before launching the workspace group?

ibacher (commented):
Couple of small questions. Do we need similar guards for if the patient changes in-flight?

packages/esm-patient-chart-app/src/patient-chart/patient-chart.resources.ts
denniskigen: Yep, good point. I moved the mounted guard before `launchWorkspaceGroup2` and kept the post-await guard as well so we do not launch after unmount and also avoid updating local refs after an in-flight launch resolves.

packages/esm-patient-chart-app/src/patient-chart/patient-chart.resources.ts
denniskigen: Thanks, I added a short comment above the launch-key helper explaining that the workspace group is current only when it was launched for the same patient and visit context.

denniskigen:
Thanks for the patient-change catch, Ian. I now track the workspace launch identity by patient UUID plus visit context UUID, so a patient switch relaunches the group even when both patients have no active visit. I also added regression coverage for the settled patient switch and the in-flight patient switch case.

ibacher (approved):

PR #3345, opened 2026-05-29
(fix) De-duplicate repeated lab results in the test results viewer

cerveaubleu64:
## Requirements

## Summary

This PR fixes a bug where lab results were displayed multiple times in the patient-chart lab results viewer. The root cause was that the de-duplication key used `flatName` instead of `conceptUuid`, causing the same observation to render once per order branch. The fix keys de-duplication on `conceptUuid` so identical observations (same concept + datetime + value) collapse into a single row.

### Problem

On the patient-chart lab results viewer, a single lab result is rendered multiple times — once per order placed for that test. A NEW order plus its auto-generated DISCONTINUE order made the same result show twice; two orders, four times, and so on.

### Root cause

The /obstree backend returns the same orderable-test concept under several branches of the orderable-tests tree (effectively one branch per order), each branch carrying the same single observation. In filter-context's tableData builder the de-duplication key is based on test.flatName, which differs per branch, so the identical observation is not collapsed and renders once per branch.

### Fix

Key the de-duplication on the observation's conceptUuid instead of flatName. The same observation (same concept + datetime + value) reached via different branches now collapses to a single row, while genuinely different tests stay separate. One-line change, no new dependencies, no new i18n keys.


packages/esm-patient-tests-app/src/test-results/filter/filter-context.tsx
```diff
@@ -135,8 +135,14 @@ const FilterProvider = ({ roots, isLoading, children }: FilterProviderProps) =>
...
-          // Use a more specific key that includes the test name to avoid over-deduplication
-          const testKey = `${test.flatName}_${obs.obsDatetime}_${obs.value}`;
+          // Dedupe by the observation's concept identity, not flatName. The obstree
+          // backend can return the same concept under several branches of the
+          // orderable-tests tree (e.g. one branch per order), each carrying the
+          // same single obs. Keying on flatName (which differs per branch) left
+          // those as separate rows, so a single result showed up N times. Concept
+          // uuid + datetime + value identifies the actual observation, while still
+          // keeping genuinely different tests apart.
+          const testKey = `${test.conceptUuid ?? test.flatName}_${obs.obsDatetime}_${obs.value}`;
```
denniskigen: Thanks for tackling this. On dev3, `/obstree` can return the same observation through multiple branches, so deduping is the right direction. I think this key is a bit too broad because it dedupes before we group rows by the rendered panel. The current CI failure shows this with the existing `mockResults` fixture: the Complete blood count panel, Platelets included, is mounted under two top-level roots, Hematology and Bloodwork, so the same observations appear under both. Keying only on concept identity keeps the first and drops the second, which reduces the tree view from 3 tables to 2 and the Platelets rows from 2 to 1. 

Could we scope the dedupe to the rendered group as well, for example include the group key derived from `flatName`, or dedupe after grouping? A focused regression test for this fixture shape would also help lock the intended behavior in.

NethmiRodrigo:
@denniskigen could you re-review this? I missed this before and I think it would be good if we can get this in, if it looks good

jayasanka-sack (approved):
Looks good to me, thanks @cerveaubleu64 and @NethmiRodrigo!

PR #3348, opened 2026-06-02
(fix) Recalculate revised medication quantities

denniskigen:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a conventional commit label. See existing PR titles for inspiration.

## Summary

When the outpatient-quantity setting is on, the drug order form auto-calculates "Quantity to dispense" from `dose × frequencyPerDay × duration`. This works for new orders, but silently broke for **revise** and **renew**: editing the dose or duration on an existing order recomputed nothing, and no recalculate link ever appeared.

The cause is how an order's frequency is reconstructed. `buildMedicationOrder` (`api/api.ts`) hardcodes `frequencyPerDay: null` because the value isn't on the order payload. OpenMRS core models `frequencyPerDay` on the `OrderFrequency` definition, and the REST layer serializes an order's frequency as a `REF` (uuid + display only), so the per-day multiplier is dropped. With it null, `calculatedQuantity` is always null on a revised order, so the form can't offer a recalculation until the clinician re-picks the frequency from the dropdown.

### What changed

- Resolve `frequencyPerDay` from the order config (`orderFrequencies`) by frequency UUID when the watched frequency doesn't carry it. That config is the same definition the backend treats as the source of truth. Freshly selected frequencies keep using their own value.
- Only show the "Apply calculated quantity" link when the calculated value differs from the current quantity, so an inherited (or deliberately set) quantity stays put on open and the link surfaces only when there's a real discrepancy to act on.

Manual overrides stay sticky and quantities are never auto-overwritten; recalculation remains an explicit, opt-in click.

### Recalculation available when dosing inputs change for revise and renew flows

https://github.com/user-attachments/assets/50e6be9f-d9ad-4f79-90a7-9e2d651c9417

## Other

 Follow-up to the auto-calculate dispense quantity feature (#3052).

### DIFF
--- a/packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
+++ b/packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
@@ -225,6 +225,17 @@ export function DrugOrderForm({
     initialOrderBasketItem?.isQuantityManual ?? (isExistingOrder && initialOrderBasketItem?.pillsDispensed != null),
   );
 
+  const watchedFrequencyPerDay = useMemo(() => {
+    if (watchedFrequency?.frequencyPerDay != null) {
+      return watchedFrequency.frequencyPerDay;
+    }
+
+    return (
+      orderConfigObject?.orderFrequencies?.find(({ valueCoded }) => valueCoded === watchedFrequency?.valueCoded)
+        ?.frequencyPerDay ?? null
+    );
+  }, [orderConfigObject?.orderFrequencies, watchedFrequency?.frequencyPerDay, watchedFrequency?.valueCoded]);
+
   const calculatedQuantity = useMemo(() => {
     if (watchedIsFreeText || watchedAsNeeded) {
       return null;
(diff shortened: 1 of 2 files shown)


VeronicaMuthee (approved):
Thank you very much, @denniskigen, for these changes. LGTM

NethmiRodrigo (approved):
Thanks @denniskigen! LGTM, the only thing - `esm-patient-orders-app` has an identical `buildMedicationOrder` that also hardcodes `frequencyPerDay: null`, so REVISE/RENEW orders entered via the orders-app path are not fixed by this PR, is that something we want to fix in this PR itself?

denniskigen:
> Thanks @denniskigen! LGTM, the only thing - `esm-patient-orders-app` has an identical `buildMedicationOrder` that also hardcodes `frequencyPerDay: null`, so REVISE/RENEW orders entered via the orders-app path are not fixed by this PR, is that something we want to fix in this PR itself?

Good catch, @NethmiRodrigo. `esm-patient-orders-app` does have its own `buildMedicationOrder`, but I verified that this path is covered because it launches the same `add-drug-order` workspace. This PR hydrates `frequencyPerDay` in that workspace from order config when the incoming basket item has `frequencyPerDay: null`, so both the Medications page and Orders page modify flows get the recalculation link.

CleanShot 2026-06-03 at 15 45 03@2x

PR #3353, opened 2026-06-04
(fix) Allow recording allergies from the order basket outside the patient chart

denniskigen:
## Requirements

## Summary

The "Add allergy" (+) control that appears in the order basket's drug search launches the allergy form. That form is registered to the `patient-chart` workspace group and reads the patient from its group props, so it only works inside the patient chart. When the order basket runs outside the chart (for example, in the ward app), the `patient-chart` group isn't active, so launching the form opened that group with null group props and crashed every action button registered to it (orders, notes, forms, vitals) with `Cannot read properties of null (reading 'patientUuid')`.

This decouples the allergy form from the chart group, following the same pattern the order basket itself already uses to run outside the chart:

- Extract a presentational `AllergyForm`, and add an `exportedAllergyFormWorkspace` that sources the patient from window props. Hosts register it into their own workspace window so it opens in their group rather than the chart's.
- The "add allergy" control now uses an optional, host-provided `launchAllergyForm`. Inside the chart it still launches the chart workspace (behavior unchanged); outside the chart the host launches the exported workspace as a child of its own window. The launcher is threaded through the exported add-drug-order workspace and the `allergy-list-pills-slot`.

The ward registration that consumes this lives in the companion PR openmrs/openmrs-esm-patient-management#2561.

#### **Before:** clicking the allergy **+** raised a stack of `Cannot read properties of null (reading 'patientUuid')` errors and the form never opened.

https://github.com/user-attachments/assets/33e5371b-6b7e-431c-b65d-7d35a475d81e

#### **After:** the allergy form opens cleanly within the ward, with no console errors, and closing it returns to the order basket.

https://github.com/user-attachments/assets/a610dae8-ebdf-4592-9fd6-89a8d5967dc9

## Other

Companion PR: openmrs/openmrs-esm-patient-management#2561.


chibongho (approved):
LGTM. Thanks for the fix!

PR #3359, opened 2026-06-05
(fix) O3-5709: Change the conditions left-nav icon to distinguish it from the task list

solomonfortune:
## Requirements

## Summary

The Conditions item in the patient chart's left navigation sidebar was using `omrs-icon-list-checked`, the same icon as the Task list in the right rail. This created a false visual suggestion of a semantic relationship between the two. Conditions represent a patient's longitudinal clinical problem list, while the Task list contains operational action items for the clinician.

Changed the Conditions nav link icon from `omrs-icon-list-checked` to `omrs-icon-table-of-contents` in `packages/esm-patient-conditions-app/src/dashboard.meta.ts`. This is an interim fix: `table-of-contents` is an already-registered, unclaimed glyph that reads as a list rather than a checklist, so it breaks the shared-icon collision with the Task list without implying another domain. The durable fix is a dedicated Conditions glyph in esm-styleguide (so `ConditionsIcon` stops aliasing `ListCheckedIcon`).

### Before

CleanShot 2026-06-09 at 13 41 52@2x

### After

The Conditions item now uses `omrs-icon-table-of-contents`, distinct from the Task list's checklist.

CleanShot 2026-06-09 at 13 41 04@2x

## Related Issue

https://openmrs.atlassian.net/browse/O3-5709

## Other

N/A

### DIFF
--- a/__mocks__/vitals-and-biometrics.mock.ts
+++ b/__mocks__/vitals-and-biometrics.mock.ts
@@ -5105,6 +5105,7 @@ export const formattedVitals = [
     bmi: null,
     muac: 23,
     bloodPressureRenderInterpretation: 'normal',
+    note: 'Pt reports severe L chest pain',
   },
   {
     id: '1',
@@ -5119,6 +5120,7 @@ export const formattedVitals = [
     spo2: 90,
     bmi: 23,
     bloodPressureRenderInterpretation: 'normal',
+    note: 'Follow up in 2 weeks. Patient also mentioned having some trouble sleeping lately and asked for some advice on sleep hygiene. Advised to avoid caffeine in the afternoon.',
   },
   {
     id: '2',
--- a/e2e/specs/vitals.spec.ts
+++ b/e2e/specs/vitals.spec.ts
@@ -5,7 +5,7 @@ import { BiometricsAndVitalsPage } from '../pages';
 test('Add, edit and delete patient vitals', async ({ page, patient }) => {
   const vitalsPage = new BiometricsAndVitalsPage(page);
   const headerRow = vitalsPage.vitalsTable().locator('thead > tr');
-  const dataRow = vitalsPage.vitalsTable().locator('tbody > tr');
+  const dataRow = vitalsPage.vitalsTable().locator('tbody > tr').first();
 
   await test.step('When I visit the vitals and biometrics page', async () => {
     await vitalsPage.goTo(patient.uuid);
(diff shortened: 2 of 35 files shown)


denniskigen (commented):
Thanks for tackling this, @solomonfortune — you're right that Conditions and the Task list sharing `omrs-icon-list-checked` creates a misleading visual link, so the underlying motivation is sound. A few things need addressing before this can go in, though.

**1. The direction is right, but `omrs-icon-user-xray` is the wrong target icon**

Moving *Conditions* (rather than the Task list) is the correct call: `list-checked` is literally a checklist glyph, which fits the Task list — and the Task list's launch button uses `ListCheckedIcon` directly (`esm-patient-task-list-app/src/launch-button/task-list-launch-button.extension.tsx`). Conditions only collides with it because the framework's `ConditionsIcon` is *aliased* to `ListCheckedIcon`; it never got its own dedicated glyph.

The problem is the replacement: `omrs-icon-user-xray` is registered (so it'll render), but it's the **imaging-orders icon** — the styleguide flags this explicitly:

```ts
// userXray is also used for imaging orders
addIconSvg('omrs-icon-user-xray', userXray);
```

So this just swaps one false association (Task list) for another (imaging/radiology), with a glyph that doesn't read as "clinical problem list." Could you give Conditions a dedicated, semantically appropriate icon and confirm the choice with the design team on O3-5709, rather than reusing imaging's icon?

**2. The PR bundles unrelated changes that aren't in the description**

The summary describes a one-line icon change, but the diff also includes:

- `feat: default clinical status to active when recording a new condition` (`conditions-form.workspace.tsx`) — a separate behavioral change with its own commit, unrelated to O3-5709.
- A `tsconfig.json` change adding `ignoreDeprecations: "6.0"` to this package only (no other package sets it) — looks unrelated/accidental.

Could you split these out? The icon fix should stand alone under O3-5709, and the clinical-status default deserves its own PR where it can be reviewed on its merits.

**3. On the clinical-status change specifically** (for whichever PR it lands in)

Defaulting to `'active'` makes the existing `clinicalStatusValidation` (the "A clinical status is required" refinement, which only applies when creating) unreachable from the UI — so that validation becomes dead code, and the test was updated to drop its assertions. If we keep the default, we should also remove the now-redundant validation; if we keep the validation, we shouldn't pre-fill. Right now it's inconsistent.

Happy to re-review once the icon target is confirmed and the changes are split. Thanks again!

denniskigen:
Hi @solomonfortune, thanks again for spotting this and pushing it forward. We're mid release cycle and wanted the icon fix in quickly, so I pushed directly to your branch (that's why the commits changed here).

The icon now uses `omrs-icon-table-of-contents` rather than `omrs-icon-user-xray`. Your read on the problem was right: Conditions and the Task list sharing a glyph is misleading. The catch with `user-xray` is that it's the imaging-orders icon, so it would have swapped the Task-list association for a radiology one. `table-of-contents` is already registered and reads as a plain list, so it separates Conditions from the Task list's checklist without picking up a new meaning.

`table-of-contents` is an interim choice. The durable fix is to give Conditions its own glyph in esm-styleguide so `ConditionsIcon` stops aliasing `ListCheckedIcon`. The candidate we're leaning toward is Carbon's `Catalog` icon: it reads as a bound registry of entries, which is essentially what a problem list is, and it stays clearly distinct from both the Task list's checklist and the imaging glyph. It's also a clean drop-in, since it comes straight from Carbon, the same source as the rest of the styleguide's icons (it registers exactly like the existing `procedure-order`, which is also a Carbon 32-grid SVG).

conditions-icon-preview

NethmiRodrigo (approved):
Thanks @denniskigen! LGTM

PR #3374, opened 2026-06-10
(fix) Deduplicate overlapping test result branches in the results viewer

denniskigen:
## Requirements

## Summary

Follow-up to #3345. The results viewer receives the same physical observation through several obstree branches: the default `resultsViewerConcepts` roots overlap (Bloodwork contains Hematology), and each root's tree carries a full copy of a concept's obs list. The filter reducer merged same-flatName duplicates by concatenating those copies, so one result was held N times. The visible symptoms and related edge cases:

- The "Results (N)" header overcounts: one Platelets result shows "Results (2)", two results show "Results (4)", while the table renders a single row.
- Two genuinely distinct results with the same value and datetime (one result per order, entered against the same encounter) collapse into one row, because de-duplication keyed on `datetime + value` cannot tell a copy from a real duplicate.
- Filtering by a checkbox whose branch was collapsed into another branch's row hides that row even though the count still includes it.

### Fix

Treat obs lists, not individual obs, as the unit of de-duplication. Branch copies are whole-list copies, so equal obs *within* one list are distinct real results while equal lists *across* branches are duplicates:

- `filter-reducer`: same-flatName duplicates now merge with a multiset union (for each `obsDatetime + value` identity, keep the highest count seen in any single branch) instead of concatenating. The merge also copies nodes instead of mutating them, so the SWR-cached trees are no longer modified and `INITIALIZE` is idempotent across re-dispatches.
- `filter-context`: a normalized branch list (keyed by rendered panel + `conceptUuid`) collapses duplicate branches that differ in flatName, using the same multiset merge. `tableData`, `timelineData`, `totalResultsCount`, and `filteredResultsCount` all derive from this one list, so the header count is structurally aligned with the rendered rows. Entries carry the flatNames that collapsed into them.
- `tree-view` / `grouped-timeline`: checkbox filtering matches entries against all contributing flatNames, so filtering by either alias of a collapsed branch keeps the row visible.

This preserves the behavior #3345 established: the same concept legitimately rendered under two different panels stays as two rows. One intentional change in the Over time view: a branch collapsed in the table is now also collapsed in the timeline instead of appearing once per source branch.

### Before

Results count is 4 yet the list only has one row

CleanShot 2026-06-10 at 12 24 54@2x

Results count is 5 yet the list only has four rows

CleanShot 2026-06-10 at 10 59 55@2x

### After

Results counts match row counts

CleanShot 2026-06-10 at 10 54 36@2x

CleanShot 2026-06-10 at 11 56 40@2x

## Other

The `obsDatetime + value` identity is the best available until the backend exposes obs uuids. A follow-up to add `uuid` to the obs map in the REST module's `ObsTreeResource1_9` would make the de-duplication exact and let the multiset merge degenerate to a simple set union.

### DIFF
--- a/packages/esm-patient-tests-app/src/test-results/filter/filter-context.test.tsx
+++ b/packages/esm-patient-tests-app/src/test-results/filter/filter-context.test.tsx
@@ -262,6 +262,91 @@ describe('FilterContext', () => {
       });
     });
 
+    it('should collapse duplicate branches in timeline data and keep alias filters working', async () => {
+      const user = userEvent.setup();
+      const sharedConceptUuid = '729AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA';
+      const primaryFlatName = 'Hematology-Complete blood count-Platelets';
+      const selectedFlatName = 'Chemistry-Complete blood count-Platelets';
+      const sharedObs = [{ obsDatetime: '2024-11-04T05:48:00.000Z', value: '56.0', interpretation: 'LOW' as const }];
+
+      const roots: Array<TreeNode> = [
+        {
+          display: 'Hematology',
+          flatName: 'Hematology',
+          hasData: true,
+          subSets: [
+            {
+              display: 'Complete blood count',
+              flatName: 'Hematology-Complete blood count',
+              hasData: true,
+              subSets: [
+                {
+                  display: 'Platelets',
+                  flatName: primaryFlatName,
+                  conceptUuid: sharedConceptUuid,
+                  hasData: true,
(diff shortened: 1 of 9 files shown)


ibacher (approved):

ibacher:
I'm beginning to wonder if the ObsTree shape isn't what we need and we could be sending this data in a more useful way. But definitely that "Other" bit should become a ticket.

denniskigen:
@ibacher I've filed https://openmrs.atlassian.net/browse/RESTWS-1038. Please have a look and validate that it has the correct context.

PR #3379, opened 2026-06-12
(test) Add form validation coverage for vitals, drug order, and visit forms

denniskigen:
## Requirements

## Summary

Adds test coverage for form validation paths that previously had none, splitting assertions by level: exact per-message checks live in unit tests, while e2e tests assert the user journey.

- **Unit** (`drug-order-form.test.tsx`): submitting an empty drug order surfaces every required-field error (dosage, route, frequency, indication, quantity to dispense, refills) and does not save. Verified by mutation: temporarily relaxing the route schema makes the test fail on exactly "Route is required".
- **E2E, drug orders**: saving a drug order with missing required fields is blocked, the basket item stays Incomplete, and filling the fields produces a New order.
- **E2E, vitals (empty form)**: the save button is disabled while the form is pristine, and submitting a dirtied-then-cleared form shows the "Please fill at least one field" error.
- **E2E, vitals (out of range)**: saving SpO2 101 shows the field-level range error and the invalid-values notification. The test first verifies via API that the SpO2 concept defines `hiAbsolute` below 101, so a misconfigured environment fails fast with a clear message instead of a mystery failure. It also waits for the input's `max` attribute to be populated before saving, because the client-side range check silently no-ops until the `conceptreferencerange` request resolves (the test raced this under parallel load and the form saved 101 successfully).
- **E2E, visits**: starting a visit without selecting a visit type shows the "Missing visit type" error and no visit starts. This is folded into the existing start-and-end-visit journey rather than a standalone test, since the error strings themselves are already covered by `visit-form.test.tsx`; the e2e adds the recovery path (fix the error, then a real save succeeds).

### DIFF
--- a/e2e/specs/drug-orders.spec.ts
+++ b/e2e/specs/drug-orders.spec.ts
@@ -366,4 +366,94 @@ test.describe('Drug Order Tests', () => {
       ).toBeVisible();
     });
   });
+
+  test('Block saving a drug order with missing required fields', async ({ page, patient }) => {
+    const orderBasket = page.locator('[data-extension-slot-name="order-basket-slot"]');
+    const medicationsPage = new MedicationsPage(page);
+
+    await test.step('When I visit the medications page', async () => {
+      await medicationsPage.goTo(patient.uuid);
+    });
+
+    await test.step('And I click the Add button on the medications details table to launch the order basket', async () => {
+      const launchOrderBasketButton = page
+        .getByRole('button', { name: /record active medications/i })
+        .or(page.getByRole('button', { name: 'Add', exact: true }))
+        .first();
+      await expect(launchOrderBasketButton).toBeVisible();
+      await launchOrderBasketButton.click();
+    });
+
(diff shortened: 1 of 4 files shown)


chibongho (approved):
LGTM.

One minor thing: Claude says `expect(...).not.toBeVisible()` and  `expect(...).toBeHidden()` are almost functionally equivalent, but I think the former is more semantically correct  when we don't expect an element to be there (vs it being there but hidden). That said, we use both in our code base right now.

denniskigen:
Good call on the semantics. I went to make the swap, but our eslint config (`playwright/no-useless-not`) flags `not.toBeVisible()` and recommends `toBeHidden()`, so I've kept `toBeHidden()` here to stay lint-clean and consistent with the rest of the suite (it's the dominant form today). If we'd rather standardize on `not.toBeVisible()` for the absence checks, relaxing that rule and sweeping the suite would be the cleaner way to do it. Thanks for reviewing!

PR #3383, opened 2026-06-14
(fix) O3-5739: Decode HTML entities in sticky note content to render special characters correctly

sourav-jyoti:
## Requirements

## Summary
When a sticky note containing angle brackets (< or >) is saved, the saved view renders the literal HTML entity codes (&lt;, &gt;) instead of the original characters
### reason : 
https://dev3.openmrs.org/openmrs/ws/rest/v1/obs?patient=[]   

=> response  "value": "&lt;div/&gt; fixed &amp; &lt;/div&gt;",

### Before
image

### After

image

## Related Issue
https://openmrs.atlassian.net/browse/O3-5739


NethmiRodrigo:
Thanks @sourav-jyoti! I made a change to go for a more lightweight approach - 
1. Replaced the DOMParser-based decodeHtmlEntities with a detached <textarea> because it parses the input as a full HTML document, which does more than we want
2. Moved the helper to a feature-scoped `utils.ts`. `decodeHtmlEntities` is a generic string helper, not data-fetching logic, so it no longer lives in `sticky-note.resource.ts`. Moved it (and its unit tests) to `sticky-notes/utils.ts` / `sticky-notes/utils.test.ts` and updated the imports in the panel and modal.
3. Decode the existing note value once in the modal. The edit modal was re-decoding `existingNote.value` inside the `isUnchanged` check on every render (i.e. every keystroke), even though that value never changes. It's now memoized once and reused for both the initial textarea value and the comparison.

sourav-jyoti:
thanks @NethmiRodrigo , i came up with the same approach with textarea  before but later on i thought may its not react friendly manipulating dom .

packages/esm-patient-notes-app/src/sticky-notes/utils.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Decodes HTML entities (e.g. &lt; &gt; &amp;) that the REST API may return
+ * in obs values back into their original characters.
+ */
+export function decodeHtmlEntities(text: string): string {
+  const textarea = document.createElement('textarea');
+  textarea.innerHTML = text;
+  return textarea.value;
+}
```
ibacher, line 9: This is a case where I strongly feel we should use a library. While this works around the problem, it does so in the worst possible way by effectively introducing a XSS mechanism. Please use something like `html-entities` for this.

NethmiRodrigo (approved):
Thanks @sourav-jyoti! Tested it locally and works for me

PR #3390, opened 2026-06-16
(fix) Settle previous-encounter Promise so form loads can't hang

denniskigen:
## Requirements

## Summary

`PatientPreviousEncounterService.getPreviousEncounter` wraps an RxJS subscription in a `Promise` but left two paths unsettled:

- the error callback only logged via `console.error` and never called `reject()`, so a failed encounters request left the Promise pending forever; and
- the `if (encounters)` block had no `else`, so a `null`/`undefined` response also never settled.

`fe-wrapper.component.ts` `mergeMap`s the form-load pipeline into this Promise, so an unsettled Promise leaves the form workspace stuck in a loading state with no error surfaced, since the swallowed error never reaches the pipeline's `catchError`.

This change rejects on the error path and resolves to an empty result on a falsy response so both paths settle. A new spec covers the match, no-match, null-response, and both failure paths.

### DIFF
--- a/packages/esm-form-entry-app/src/app/openmrs-api/patient-previous-encounter.service.spec.ts
+++ b/packages/esm-form-entry-app/src/app/openmrs-api/patient-previous-encounter.service.spec.ts
@@ -1,0 +1,100 @@
+import { TestBed } from '@angular/core/testing';
+import { of, throwError } from 'rxjs';
+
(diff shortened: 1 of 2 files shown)


ibacher (approved):
Thanks @denniskigen!

PR #3395, opened 2026-06-17
(feat) O3-5746: Support conditional order reasons via hide expression

donaldkibet:
## Requirements

## Summary
This PR extends `orderReasons` config to accept either a plain concept UUID string or an object `{ conceptUuid, hideWhenExpression }`, allowing individual order reason options to be conditionally hidden per patient context.
## Screenshots
https://github.com/user-attachments/assets/9bb78b0d-472f-46ee-b61b-ed8d0c41e8a3

### DIFF
--- a/packages/esm-patient-tests-app/src/config-schema.ts
+++ b/packages/esm-patient-tests-app/src/config-schema.ts
@@ -35,7 +35,8 @@ export const configSchema = {
   },
   orderTypeUuid: {
     _type: Type.UUID,
-    _description: "UUID identifying this extension's order type for order basket panel filtering. Must match orders.labOrderTypeUuid if that value is overridden.",
+    _description:
+      "UUID identifying this extension's order type for order basket panel filtering. Must match orders.labOrderTypeUuid if that value is overridden.",
     _default: '52a447d3-a64a-11e3-9aeb-50e549534c5e',
   },
   orders: {
@@ -107,9 +108,9 @@ export const configSchema = {
         _type: Type.Array,
         _elements: {
           _type: Type.ConceptUuid,
-          _description: 'Array of coded concepts that represent reasons for ordering a lab test',
         },
-        _description: 'Coded Lab test order reason options',
+        _description:
+          'Coded Lab test order reason options. Each entry may be a plain concept UUID string, or an object with `conceptUuid` (string) and an optional `hideWhenExpression`.',
         _default: [],
       },
     },
@@ -128,9 +129,16 @@ export interface LabTestReason {
   label?: string;
 }
 
+export interface OrderReasonReferenceConfig {
+  conceptUuid: string;
+  hideWhenExpression?: string;
+}
+
+export type OrderReasonReference = string | OrderReasonReferenceConfig;
+
 export interface OrderReason {
   labTestUuid: string;
-  orderReasons: Array<string>;
+  orderReasons: Array<OrderReasonReference>;
   required: boolean;
 }
 
--- a/packages/esm-patient-tests-app/src/config-schema.ts
+++ b/packages/esm-patient-tests-app/src/config-schema.ts
@@ -1,4 +1,4 @@
-import { Type } from '@openmrs/esm-framework';
+import { Type, validator } from '@openmrs/esm-framework';
 
 export const configSchema = {
   resultsViewerConcepts: {
@@ -107,7 +107,13 @@ export const configSchema = {
       orderReasons: {
         _type: Type.Array,
         _elements: {
-          _type: Type.ConceptUuid,
+          _validators: [
+            validator(
+              (v) =>
+                typeof v === 'string' || (typeof v === 'object' && v !== null && typeof v.conceptUuid === 'string'),
+              'must be a concept UUID string or an object with a `conceptUuid` string',
+            ),
+          ],
         },
         _description:
           'Coded Lab test order reason options. Each entry may be a plain concept UUID string, or an object with `conceptUuid` (string) and an optional `hideWhenExpression`.',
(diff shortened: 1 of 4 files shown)


CynthiaKamau (approved):

donaldkibet:
ping @denniskigen @NethmiRodrigo

denniskigen (commented):
Thanks for working on this, @donaldkibet. I've left a few review comments inline.

packages/esm-patient-tests-app/src/test-orders/add-test-order/test-order-form.component.tsx
```diff
@@ -137,7 +137,7 @@ export function LabOrderForm({
     (config.labTestsWithOrderReasons?.find((c) => c.labTestUuid === defaultValues?.testType?.conceptUuid) || {})
       .orderReasons || [];
 
-  const { orderReasons } = useOrderReasons(orderReasonUuids);
+  const { orderReasons } = useOrderReasons(orderReasonUuids, patient);
```
denniskigen: One edge case I think is worth handling before merge: when a test is `required: true` but every configured reason has a `hideWhenExpression` that hides it for the current patient, `orderReasons` comes back empty, so the ComboBox stops rendering (the `orderReasons.length > 0` guard below). But `orderReasonRequired` is still computed from the unfiltered config up at lines 80-85, so Zod keeps blocking submit.

I rendered this to be sure: clicking Save fires the form's `onError` with "Order reason is required", the order never saves, and there's no field and no error message on screen, so the clinician has no way to resolve it.

Could we gate the requirement on the visible list, e.g. require `orderReason` only when `orderReasons.length > 0` in the schema at lines 103-109? Worth a line in the PR on the trade-off: a required reason effectively becomes optional when all options are hidden, which seems right since there's nothing left to pick, but good to make it a conscious call. A test with `required: true` and all reasons hidden would lock this in.

packages/esm-patient-tests-app/src/config-schema.ts
```diff
@@ -107,9 +108,9 @@ export const configSchema = {
         _type: Type.Array,
         _elements: {
           _type: Type.ConceptUuid,
```
denniskigen: One thing to update here: `_elements` is still typed as `Type.ConceptUuid`, so the config framework validates each entry as a UUID string and rejects the `{ conceptUuid, hideWhenExpression }` object form this PR adds. I ran the actual validator against a sample config and each object entry logs:

`Invalid configuration for ...orderReasons[1]: must be a valid UUID or a 36-character hexadecimal string`

The value still passes through so the feature works, but anyone who adopts the object form gets an error on every config load (and the implementer tools will likely render each entry as a plain concept field too). Since the schema has no union type, one option that validates both shapes cleanly is to drop `_type` and add a custom element validator:

```ts
orderReasons: {
  _type: Type.Array,
  _elements: {
    _validators: [
      validator(
        (v) =>
          typeof v === 'string' ||
          (typeof v === 'object' && v !== null && typeof v.conceptUuid === 'string'),
        'must be a concept UUID string or an object with a `conceptUuid` string',
      ),
    ],
  },
  _description: '...',
  _default: [],
},
```

I tested this variant: strings and well-formed objects pass with no errors, and a bad entry (object missing `conceptUuid`, or a number) still gets flagged.

packages/esm-patient-tests-app/src/test-orders/api.ts
```diff
@@ -64,17 +65,45 @@ export function usePatientLabOrders(patientUuid: string, status: 'ACTIVE' | 'any
...
+        return evaluatedExpression ? null : concept.conceptUuid;
+      }
+
+      return concept.conceptUuid;
+    }),
+  );
+
+  return visibleConceptUuids
+    .filter(
+      (result): result is PromiseFulfilledResult<string> => result.status === 'fulfilled' && result.value !== null,
```
denniskigen, line 97: Small one, not blocking: when `evaluateAsBooleanAsync` rejects (invalid syntax, a non-boolean result, or referencing something undefined), `Promise.allSettled` plus the fulfilled-only filter quietly drops that reason from the request, and `useOrderReasons` never sees an error to surface in the existing snackbar. So a typo in a `hideWhenExpression` makes an option silently vanish, which is hard to debug. Could we catch per item and fail open (keep the reason visible) and log the failure? I'd steer away from switching to `Promise.all` here, since one bad expression would then drop every reason and reject the whole fetch.

packages/esm-patient-tests-app/src/test-orders/api.ts
```diff
@@ -64,17 +65,53 @@ export function usePatientLabOrders(patientUuid: string, status: 'ACTIVE' | 'any
...
+      if (typeof concept === 'string') {
+        return concept;
+      }
+
+      if (concept.hideWhenExpression) {
+        try {
+          const evaluatedExpression = await evaluateAsBooleanAsync(concept.hideWhenExpression, {
+            patient: patient,
+            openmrsFetch,
+          });
```
ibacher, line 80: I'm somewhat suspicious of including this in the implementation. It would be better if we can build out a notion _what data are available to expressions_ rather than trying to allow expressions to load external data. The design of the expression evaluation framework, while it doesn't prevent this, is meant to discourage this pattern as it makes it much harder to reason about what's happening. It would be better to work out _what kinds of things_ we might need to work with and see if we can make those available.

ibacher: Basically, `openmrsFetch()` should never be in the context of `evaluate()` without a really strong justification.

ibacher (changes requested):
I think the use of `openmrsFetch()` in the evaluate context at least needs a justification before this can be merged.

PR #3401, opened 2026-06-18
(feat) O3-5753: Defer drug search API calls until minimum character threshold is reached

donaldkibet:
## Requirements

## Summary
 See https://openmrs.atlassian.net/browse/O3-5753

## Screenshots
https://github.com/user-attachments/assets/ebb8eb39-be1c-4250-b322-239c99b3648d

### DIFF
--- a/packages/esm-patient-medications-app/src/config-schema.ts
+++ b/packages/esm-patient-medications-app/src/config-schema.ts
@@ -64,9 +64,12 @@ export const configSchema = {
   },
   minimumCharacterLengthForDrugSearch: {
     _type: Type.Number,
+    _validators: [
+      validator((v: unknown) => typeof v === 'number' && Number.isInteger(v) && v >= 1, 'Must be a positive integer'),
+    ],
     _description:
       'The minimum number of characters that must be entered in the drug search input before the search is performed.',
-    _default: 3,
+    _default: 2,
   },
 };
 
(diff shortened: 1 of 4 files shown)


packages/esm-patient-medications-app/src/config-schema.ts
```diff
@@ -62,6 +62,12 @@ export const configSchema = {
       _type: Type.String,
     },
   },
+  minimumCharacterLengthForDrugSearch: {
```
denniskigen, line 65: Nice addition, and the description is clear. Since `debounceDelayInMs` just above validates its bounds, could we add a validator here too so a stray `0`, negative value, or decimal gets caught during config validation instead of silently disabling the threshold?

```ts
_validators: [validator((v: unknown) => Number.isInteger(v) && v >= 1, "Must be a positive integer")],
```

packages/esm-patient-medications-app/src/config-schema.ts
```diff
@@ -62,6 +62,12 @@ export const configSchema = {
...
+    _type: Type.Number,
+    _description:
+      'The minimum number of characters that must be entered in the drug search input before the search is performed.',
+    _default: 3,
```
ibacher: I think the default here should be 2 mirroring the default for user search, unless there's a strong reason we want this to be 3.

donaldkibet: Not really. updated config to 2

makombe (approved):
Thanks @donaldkibet now looks fine to me  cc @ibacher @denniskigen

denniskigen (approved):
Nice work iterating on this, @donaldkibet! Thanks.

PR #3403, opened 2026-06-18
(fix) Filter conditions query to problem-list-item category

NethmiRodrigo:
## Requirements

## Summary

The conditions widget fetched all FHIR `Condition` resources for a patient regardless of category, which surfaced non–problem-list entries (e.g. encounter-diagnosis conditions) in the conditons list.

This PR adds a `category` filter to the `useConditions` FHIR query so it only returns problem-list-item conditions, so that encounter-diagnoses (diagnoses added from the visit note form) do not appear in the conditions table:

```
${fhirBaseUrl}/Condition?patient=${patientUuid}&category=http://terminology.hl7.org/CodeSystem/condition-category|problem-list-item&_count=100
```

These changes do not affect the search for conditions when trying to record a condition for a patient.

## Screenshots

Before - 

https://github.com/user-attachments/assets/0cc1ebb2-f88e-47b0-8011-7e4f558cf23f

After - 

https://github.com/user-attachments/assets/5bbba769-9a6f-4636-9945-76d26017841d

## Other

https://talk.openmrs.org/t/why-do-we-display-diagnoses-in-the-conditions-list/49608/4


packages/esm-patient-conditions-app/src/conditions/conditions.resource.ts
```diff
@@ -127,7 +127,8 @@ export type FormFields = {
 };
 
 export function useConditions(patientUuid: string) {
-  const conditionsUrl = `${fhirBaseUrl}/Condition?patient=${patientUuid}&_count=100`;
+  const conditionsCategory = 'http://terminology.hl7.org/CodeSystem/condition-category|problem-list-item';
+  const conditionsUrl = `${fhirBaseUrl}/Condition?patient=${patientUuid}&category=${conditionsCategory}&_count=100`;
```
denniskigen, line 131: Thanks for narrowing this query to problem-list conditions.

One compatibility question: this package still declares `fhir2 >=1.2` in `src/routes.json`, but `Condition.category` search support only starts in FHIR2 2.8.0 (added in [FM2-667](https://openmrs.atlassian.net/browse/FM2-667)). On older FHIR2 versions the resource provider doesn't declare the `category` search param at all, so deployments that satisfy the currently declared dependency but run FHIR2 <2.8.0 would fail this request instead of showing conditions.

Could we either bump the conditions app's backend dependency to `fhir2 >=2.8.0`, or add a fallback for older FHIR2 versions?

NethmiRodrigo: Good catch — I've bumped the conditions app's backend dependency to `fhir2 >=2.8.0` to match where `Condition.category` search support was introduced.

denniskigen:
One small suggestion for the PR description: could we clarify that this only affects the existing Conditions list, not the Add Condition search/autocomplete?

For example, concepts like Pneumonia may still appear in the search because users can add them to the problem list. What should no longer happen is Pneumonia appearing in the Conditions table only because it was recorded as an encounter diagnosis during a visit.

NethmiRodrigo:
> One small suggestion for the PR description: could we clarify that this only affects the existing Conditions list, not the Add Condition search/autocomplete?
> 
> For example, concepts like Pneumonia may still appear in the search because users can add them to the problem list. What should no longer happen is Pneumonia appearing in the Conditions table only because it was recorded as an encounter diagnosis during a visit.

Done

packages/esm-patient-conditions-app/src/conditions/conditions.resource.ts
ibacher: Although I think this is mostly harmless. The FHIR2 module tries hard to just ignore parameters it doesn't understand or can't process (IIRC, the search bundle's `self` link tells you which items were actually read).

ibacher (approved):

denniskigen (approved):
Thanks, @NethmiRodrigo!

PR #3406, opened 2026-06-18
(fix) Attach visit to visit note submissions

NethmiRodrigo:
## Requirements

## Summary

This attaches the active visit uuid to the saved visit note payload in `VisitNotesForm` (`visit: activeVisit?.uuid ?? null`), guarding against a null active visit. This fixes the issue of visit notes not being attached to the active visit.

## Screenshots

Before - 

https://github.com/user-attachments/assets/a166bcd7-e70d-450b-8f8e-f8d81a7034f9

After -

https://github.com/user-attachments/assets/cacd12df-e777-414b-a1b2-8472d7ad81cc

## Other
Removes the dead `testsFilter` `useMemo` (and the now-orphaned `ExternalOverviewProps` import) from `VisitSummary`. The encounter-based filtering it represented now lives inside the `TestsSummary` component, which builds the filter itself from its `encounters` prop — so the parent's copy was unused.


ibacher (approved):

packages/esm-patient-notes-app/src/notes/visit-notes-form.workspace.tsx
```diff
@@ -394,6 +396,7 @@ const VisitNotesForm: React.FC<PatientWorkspace2DefinitionProps<VisitNotesFormPr
               },
             ]
           : [],
+        visit: activeVisit?.uuid ?? null,
```
denniskigen: Nice work tracking this down. Attaching the active visit so notes show up against it is exactly right, and it's a genuinely useful fix.

One thing on the edit path: this same payload gets reused by `updateVisitNote`, which POSTs to `/encounter/{uuid}`, and `visit` is an updatable property on the encounter resource (the base `getUpdatableProperties()` pulls it in from `getCreatableProperties()`). So when someone edits a note that belongs to an older visit, this would move its encounter onto today's active visit, and when there's no active visit it'd send `visit: null` and detach the encounter from its visit altogether.

Could we only attach it when creating the note?

```suggestion
        // Only attach the visit when creating a note. On edit, omitting `visit` leaves the encounter's
        // existing visit untouched rather than reassigning it to (or detaching it from) the active visit.
        ...(!isEditing && activeVisit?.uuid && { visit: activeVisit.uuid }),
```

Bonus: this also avoids sending `visit: null` on create when there's no active visit, since the key just gets omitted.

denniskigen (commented):
Thanks for the fix! One follow-up on the edit path inline.

packages/esm-patient-notes-app/src/notes/visit-notes-form.workspace.tsx
NethmiRodrigo: Done!

ibacher:
It might be worth switching this from activeVisit to currentVisit... I can't remember where we landed on that exactly, but currently currentVisit it meant to represent "the visit the user is operating in" whereas activeVisit is more informational.

It also might be good to ensure that we have a visit before allowing submission when not editing a note and blocking submission otherwise. This would prevent the potential case where, e.g., the submit fires before the `useVisit` hook returns.

NethmiRodrigo:
> It might be worth switching this from activeVisit to currentVisit... I can't remember where we landed on that exactly, but currently currentVisit it meant to represent "the visit the user is operating in" whereas activeVisit is more informational.

@ibacher looks like currentVisit is now deprecated - 
https://github.com/openmrs/openmrs-esm-core/blob/c0e53dfeb4c369ca9709805d722d29889f77be61/packages/framework/esm-react-utils/src/useVisit.ts#L19-L21
So I updated it to use the patient chart store like the doc suggested.

> It also might be good to ensure that we have a visit before allowing submission when not editing a note and blocking submission otherwise. This would prevent the potential case where, e.g., the submit fires before the `useVisit` hook returns.

This, I will look into, I’m not sure if we can get a loading state out of the context store. Since it is [made to return null if the patient UUID doesn’t match](https://github.com/openmrs/openmrs-esm-patient-chart/blob/f4a4466d6444e96b3a8e68c84eb8de39275da4e6/packages/esm-patient-common-lib/src/store/patient-chart-store.ts#L37), blocking on null values doesn’t seem right to me, if I’m understanding correctly...

packages/esm-patient-notes-app/src/notes/visit-notes-form.workspace.tsx
```diff
@@ -38,6 +38,7 @@ import {
   useFeatureFlag,
   useLayoutType,
   useSession,
+  useVisit,
```
denniskigen: `useVisit` is no longer used now that the visit comes from `usePatientChartStore`, so this import can be dropped.

```suggestion
```

denniskigen (approved):
Thanks, @NethmiRodrigo!

PR #3408, opened 2026-06-20
(feat) O3-5757: Add patient specific audit history tab

RajPrakash681:
## Requirements

- [x] This PR has a title that briefly describes the work done, including a conventional commit label.
- [x] My work is based on designs (timeline-style audit view; mirrors the admin Audit Log dashboard's diff UI).

## Summary

This PR adds a **patient-specific Audit History** tab to the patient chart, implemented as a new microfrontend package, **`@openmrs/esm-patient-audit-history-app`**.

When a clinician opens a patient's chart, the **Audit History** tab shows a chronological, paginated record of every change made to that patient's record — **who** changed it, **when**, the **type** of change (Created / Updated / Deleted), and the **field-level before → after diff**. It complements the system-wide admin Audit Log dashboard by scoping the same audit data to a single patient, where it's most useful in the clinical workflow.

Data comes from the [`auditlogweb`](https://github.com/openmrs/openmrs-module-auditlogweb) module's `GET /ws/rest/v1/auditlogs/patients?uuid={patientUuid}` endpoint (the patient's Hibernate Envers revisions, newest-first, server-side paginated). The widget degrades gracefully when the module isn't installed.

**What's included**
- New widget package registered into the patient chart via the `patient-chart-dashboard-slot` extension + a dedicated content slot, scoped to the open patient's UUID.
- Paginated Carbon `DataTable`; each row expands to a field-level diff (CSS-grid aligned) plus the related records changed in the same revision.
- Full states: loading skeleton, empty, error, module-not-installed, and no-permission (gated on the `View Audit Log` privilege).
- Unit tests for the component and the data hook.

**Scope notes / backend follow-ups (auditlogweb)**
- The endpoint currently surfaces changes to the Patient/Person record itself (e.g. gender, birthdate). Edits to related entities (name, address, phone, identifiers) and their field-level values are not yet returned — a backend enhancement.
- The patient endpoint does not yet accept date / event-type filters, so no filter bar is shown for now.

## Screenshots
[this](https://drive.google.com/file/d/1tnkmKltjfXKe1FwXCxMIfZSdUHoJ9fvF/view?usp=sharing) and[ this](https://drive.google.com/file/d/1n5DRiir01ZeRZNQPW271wWxSpPW8zrGC/view?usp=sharing)

## Related Issue

[AUDIT_LOG](https://openmrs.atlassian.net/browse/O3-5757)

### Automated (unit tests, types, lint)

```sh
yarn install
yarn turbo run test typescript lint --filter=@openmrs/esm-patient-audit-history-app
```

Expected: tests pass, no type errors, no lint errors.

### Manual (in the patient chart)

**Prerequisites**

- An OpenMRS backend running with the [`auditlogweb`](https://github.com/openmrs/openmrs-module-auditlogweb) module installed and **Hibernate Envers enabled** (`hibernate.integration.envers.enabled=true` and `hibernate.hbm2ddl.auto=update` in the runtime properties). Sanity-check the endpoint:
  `GET /openmrs/ws/rest/v1/auditlogs/patients?uuid={patientUuid}` → returns `{ totalLogs, logs: [...] }`.
- A user with the **View Audit Log** privilege.

**Steps**

1. Start this package against your backend:

   ```sh
   yarn start --sources 'packages/esm-patient-audit-history-app' --backend <your-backend-url>
   ```

2. Log in and open any patient's chart.
3. Confirm **Audit History** appears in the chart's left navigation; click it.
4. If the patient has no history yet, generate some — edit the patient and change a **Person/Patient-level field** such as **gender** or **birthdate**, then reopen the tab.
5. Verify the table:
   - Revisions are listed **newest-first**, paginated, with columns **Date & time · Event (Created / Updated / Deleted) · Changed by**.
   - Expanding a row shows the **field-level diff** (Field / Previous value → New value), with old values struck through in red and new values in green.
   - When a revision touched other records in the same save, an **"Also changed in this revision"** list appears (e.g. PersonName, PatientIdentifier).
   - Pagination controls move between pages.

**Edge states to verify**

- **Empty:** open a patient with no recorded changes → empty state.
- **No permission:** log in as a user **without** the `View Audit Log` privilege → permission message instead of the table.
- **Module not installed:** point the dev server at a backend **without** `auditlogweb` → "auditlogweb module is not installed" message (graceful, no crash).

### Notes

- The endpoint currently surfaces changes to the **Patient/Person record itself** (e.g. gender, birthdate). Edits to related entities (name, address, phone, identifiers) are listed as labels only, and standalone edits to them are not yet returned — these are backend follow-ups in `auditlogweb`.


RajPrakash681:
@olewandowski1 @ManojLL Please have a look!

packages/esm-patient-audit-history-app/src/constants.ts
```diff
@@ -0,0 +1,26 @@
+export const moduleName = '@openmrs/esm-patient-audit-history-app';
+
+export const PRIVILEGE_VIEW_AUDIT_LOG = 'View Audit Log';
```
denniskigen: Thanks for wiring the privilege gate through config, that's the right pattern here. One catch: the `auditlogweb` module registers this privilege as "View Audit Logs" (plural) in its config.xml, and its `AuditService` methods enforce the same plural string via `@Authorized(AuditLogConstants.VIEW_AUDIT_LOGS)`. Since `userHasAccess` does an exact string match (with a superuser bypass), a non superuser who has been granted the module's actual privilege will still see the "no permission" message with this default. Testing as admin wouldn't surface this because superusers skip the check. Could you change the default to "View Audit Logs" and update the mock in audit-history.test.tsx to match?

packages/esm-patient-audit-history-app/src/audit-history/audit-history.component.tsx
```diff
@@ -0,0 +1,197 @@
...
+  const { tableRows, originalById } = useMemo(() => {
+    const occurrences = new Map<string, number>();
+    const rows = logs.map((log) => {
+      const baseKey = `${log.revisionID}-${log.changedOn}`;
+      const occurrence = occurrences.get(baseKey) ?? 0;
+      occurrences.set(baseKey, occurrence + 1);
+      const id = occurrence === 0 ? baseKey : `${baseKey}-${occurrence}`;
+      return {
+        id,
+        changedOn: log.changedOn,
```
denniskigen, line 83: A question on the timestamps: the backend serializes `changedOn` as `dd/MM/yyyy HH:mm:ss` in GMT, and this column renders the string as is, so viewers outside UTC see shifted times in a fixed format. Was GMT display intended here? If the goal is the chart convention (viewer-local via `formatDatetime`), you could parse the value as UTC with the dayjs `utc` and `customParseFormat` plugins using the `BACKEND_DATETIME_FORMAT` constant already defined in constants.ts, then format with `formatDatetime`. If UTC is intended, labeling the column as UTC would avoid misreading.

denniskigen (commented):
Thanks for adding this patient-chart audit history entry point, @RajPrakash681! I left a few backend-contract comments around the auditlogweb privilege name, timestamp handling for the endpoint's GMT-serialized dates, and release/version coordination for the new patient endpoint.

packages/esm-patient-audit-history-app/src/audit-history/audit-history.component.tsx
```diff
@@ -0,0 +1,197 @@
...
+  const layout = useLayoutType();
+  const responsiveSize = isDesktop(layout) ? 'sm' : 'lg';
+  const headerTitle = t('auditHistory', 'Audit history');
+
+  const resolvedPatientUuid = patientUuid ?? patient?.id ?? '';
+  const pageSize = auditHistoryPageSize ?? DEFAULT_PAGE_SIZE;
+
+  const [page, setPage] = useState(1);
+
+  const { logs, totalLogs, isLoading, isValidating, error, mutate } = usePatientAuditHistory(
```
denniskigen, line 56: Small one: this hook runs before the privilege early return below, so the request still fires for users who fail the gate and the backend then rejects it. You could compute the access check first and pass an empty uuid when it fails, since the hook already maps an empty uuid to a null SWR key.

packages/esm-patient-audit-history-app/src/audit-history/audit-log-diff.component.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <p className={styles.noChanges}>
+        {t('noFieldChangesRecorded', 'No field-level changes recorded for this revision.')}
+      </p>
+    );
+  }
+
+  return (
+    <div className={styles.diffWrapper}>
+      <p className={styles.diffTitle}>{t('changeDetails', 'Change details')}</p>
+      <div className={styles.diffGrid} role="table" aria-label={t('changeDetails', 'Change details')}>
```
denniskigen, line 63: Minor accessibility note: `role="table"` expects `role="row"` children wrapping the columnheader and cell elements. As written, screen readers may not associate the cells with their headers. Wrapping the header trio and each change trio in a `role="row"` div would make the grid announce correctly.

packages/esm-patient-audit-history-app/src/routes.json
```diff
@@ -0,0 +1,26 @@
+{
+  "$schema": "https://json.openmrs.org/routes.schema.json",
+  "backendDependencies": {
+    "webservices.rest": ">=2.2.0"
+  },
+  "extensions": [
```
denniskigen, line 16: Right now the Audit History link renders on deployments without a compatible auditlogweb module, where it leads to a permanent "module not installed" message. Could we gate these extensions so the tab only appears when it can work? The framework supports extension privileges, and optionalBackendDependencies can enable a feature flag when a version-satisfying backend module is installed. That would let this package hide the dashboard link unless auditlogweb with the patient audit endpoint is present.

packages/esm-patient-audit-history-app/src/audit-history/audit-history.component.tsx
```diff
@@ -0,0 +1,197 @@
...
+  if (isLoading) {
+    return <DataTableSkeleton role="progressbar" compact={isDesktop(layout)} zebra />;
+  }
+
+  if (error) {
+    if ((error as unknown as FetchError)?.response?.status === 404) {
+      return (
+        <Layer>
+          <Tile className={styles.messageTile}>
+            <p>{t('moduleNotInstalled', 'The auditlogweb module is not installed on this server.')}</p>
```
denniskigen: Suggestion on this message: a 404 here does not necessarily mean the module is absent. The latest tagged `auditlogweb` version, 1.0.0, predates `/auditlogs/patients`, so a deployment can have `auditlogweb` installed and still hit this branch. Something like "The auditlogweb module is not installed or does not support patient audit history" would point implementers at the actual issue.

packages/esm-patient-audit-history-app/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "@openmrs/esm-patient-audit-history-app",
+  "version": "12.1.0",
```
denniskigen: Could you update this package metadata to match current main? The new package starts at 12.1.0 and peers on `@openmrs/esm-framework` 9.x, while the existing patient-chart packages are at 12.3.4 and peer on framework 10.x. Leaving it as-is would publish this package out of step with its siblings.

packages/esm-patient-audit-history-app/src/constants.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const DEFAULT_PAGE_SIZE = 10;
+export const DEFAULT_PAGE_SIZES = [10, 20, 50];
```
denniskigen: Looks like `DEFAULT_PAGE_SIZES` isn't used anywhere. Fine to drop unless you're planning to wire page-size options into the pagination.

denniskigen (commented):
Duplicate review removed.

denniskigen:
Overall this is using the right O3/Carbon patterns, but I do not think the audit details are maximally clear for users yet. The biggest issue is that the expanded rows expose backend/internal representations: raw Java field names like `deathdateEstimated`, raw values like `Concept#88`, Java date strings inside diffs, and related entity names like `PersonName`. A label/value formatting layer, plus backend follow-up to return display values and ISO dates, would make the audit trail much more usable for privileged staff.

A few related observations:

- **Voiding**: in OpenMRS the common form of deletion is a void, which Envers records as MOD, so a voided patient shows up as a blue "Updated" row with the voided flag change buried in the expansion. The red "Deleted" tag only fires on a purge, which is rare for patient records. Promoting a voided transition to its own "Voided" tag would match what users scanning for deletions expect.

- **Created rows**: from reading the backend path (fetchPreviousRevision returns null for a first revision, and computeFieldDiffs then marks every populated field as changed against an empty old value), I would expect expanding a "Created" row to show the whole record, including rows like patientId and dead: false. The unit test fixture models the ADD case with changes: [], so I was not sure which behavior your manual testing showed. If it is the full dump, a "Record created" summary for ADD rows, or extending TECHNICAL_FIELDS to cover id and default-boolean noise, might keep the expansion focused. (The same null-previous path would presumably apply to the first audited edit of a patient that predates enabling Envers, though that is an inference from the code, not something I have reproduced.)

- **Row affordance**: rows give no hint whether expanding is worthwhile, and some expand to just the "no field-level changes" message. A small "Changes" count column would let users know what is inside before expanding. Relatedly, wording like "Also changed in this revision" could be more useful, for example "Changed in the same save".

None of these need to block the PR; the label/value formatting layer is the piece I would prioritize.

RajPrakash681:
@denniskigen please have a look again.

ibacher:
I have some dislike for including this in the patient-chart-repo unless there's a strong reason we can't include it with the rest of the audit frontend. For the most part, the organization of microfrontends is a little convoluted, but the idea is that related functionality should live together not just because of where it is in the app, but because it likely shared components, etc. The patient-chart repo is not just for "things in the patient-chart", but "things that primarily live in the patient chart".

jayasanka-sack:
Good work, @RajPrakash681!

So to echo what Ian said: from what I've seen you've been opening PRs to add new apps across relevant repos, e.g. esm-auditlog-app in admin-tools (https://github.com/openmrs/openmrs-esm-admin-tools/pull/284) and esm-patient-audit-history-app in patient chart. The assumption seems to be that each piece should live as its own app in its own repo.

Only issue is that kinda creates a problem down the line. Say someone doesn't need audit functionality at all, they'd have to go rip out a bunch of audit apps scattered with different features. Not great for consumers, and not great for us either since now any new feature means PRs across multiple repos just to ship one thing.

Good news is our microfrontend setup already handles this with extension slots, so a single app can push content into whatever slots it needs. So probably the cleaner move here is just spinning up one new repo and bundling all the audit frontend stuff into a single app.

Happy to set that repo up for you. Lmk if you wanna hop on a huddle or 1:1 to talk through it more.

RajPrakash681:
> Good work, @RajPrakash681!
> 
> So to echo what Ian said: from what I've seen you've been opening PRs to add new apps across relevant repos, e.g. esm-auditlog-app in admin-tools ([openmrs/openmrs-esm-admin-tools#284](https://github.com/openmrs/openmrs-esm-admin-tools/pull/284)) and esm-patient-audit-history-app in patient chart. The assumption seems to be that each piece should live as its own app in its own repo.
> 
> Only issue is that kinda creates a problem down the line. Say someone doesn't need audit functionality at all, they'd have to go rip out a bunch of audit apps scattered with different features. Not great for consumers, and not great for us either since now any new feature means PRs across multiple repos just to ship one thing.
> 
> Good news is our microfrontend setup already handles this with extension slots, so a single app can push content into whatever slots it needs. So probably the cleaner move here is just spinning up one new repo and bundling all the audit frontend stuff into a single app.
> 
> Happy to set that repo up for you. Lmk if you wanna hop on a huddle or 1:1 to talk through it more.

Thanks jaye and yes I would love your time please let me know whenever possible😇

denniskigen:
Closing this in favor of openmrs/openmrs-esm-audit-log-app#1, which moves the patient audit history tab into a dedicated audit log app alongside the admin dashboard, as Ian and Jayasanka suggested above. Thanks, @RajPrakash681. The integration now happens purely through routes.json, so no patient-chart changes are needed.

PR #3427, opened 2026-06-27
(fix) O3-5459: Patient persists in the queue after visit deletion

Josephkagimu1:
## Requirements

## Summary
When a patient's active visit is deleted, the associated queue entry was not being removed. This caused the patient to persist in the service queue despite having no active visit.

## Problem
A patient is added to a service queue when they check in with an active visit. If a clinician navigates to the patient summary → Actions → Delete Active Visit, the visit is successfully voided but the queue entry remains. The patient continues to appear in "Patients currently in queue" even though their visit no longer exists.

## Solution
After successfully deleting a visit, the system now fetches any associated queue entry for that visit and voids it via the REST API. This ensures the queue stays in sync with visit lifecycle events.

## Changes
Modified useDeleteVisit.tsx to void the associated queue entry when a visit is deleted
Added a voidQueueEntryForVisit helper that fetches the queue entry for the visit via /visit-queue-entry?visit={visitUuid} and DELETEs it via /queue-entry/{uuid}
Queue entry removal is wrapped in a try/catch so visit deletion still succeeds even if the queue API call fails
Added 3 new test cases to useDeleteVisit.test.tsx:
Verifies queue entry is voided on visit deletion
Verifies no DELETE call when no queue entry exists
Verifies visit deletion completes even if queue entry voiding fails

## Related Issue
https://openmrs.atlassian.net/browse/O3-5459


Josephkagimu1:
CC : @denniskigen , @VeronicaMuthee

VeronicaMuthee:
Thanks for digging into this, @Josephkagimu1 — the approach works in your screenshots, and the diagnosis (a deleted visit leaving its queue entry behind) is spot on. After digging into the backend I want to share what's actually causing the bug, because it changes where the best fix lives, and to walk back part of my earlier comment.

**Root cause (the important part).** The queue module already closes queue entries when their visit ends — see [`AutoCloseVisitQueueEntryTask`](https://github.com/openmrs/openmrs-module-queue/blob/main/api/src/main/java/org/openmrs/module/queue/tasks/AutoCloseVisitQueueEntryTask.java):

```java
criteria.setIsEnded(false);
criteria.setHasVisit(true);
...
if (visitStopDatetime != null) {
    queueEntry.setEndedAt(visitStopDatetime);
}
```

It keys entirely on the visit's `stopDatetime`. But **deleting a visit voids it (`voided = true`) and never sets `stopDatetime`**, so this guard is never satisfied and the entry is left open. That's exactly why *ending* a visit removes the patient from the queue with no queue call at all, while *deleting* one doesn't. The gap is in the backend's void handling, not really in the frontend.

So the cleanest fix is in `openmrs-module-queue`: have visit-void close the associated open queue entry the same way ending a visit does (setting `endedAt`). That fixes every client at once and lives right next to the existing logic. Might be worth a queue-module ticket — happy to help draft it.

If we keep a client-side fix here as an interim mitigation, a few things are worth tightening regardless of endpoint:

1. **Match the module's semantics — end the entry, don't void it.** The backend *ends* entries (`setEndedAt`), preserving the record; `DELETE /queue-entry/{uuid}` voids it and loses the queue history.
2. **Don't report success when the queue step fails.** The `catch` currently only `console.warn`s, so the user sees "Visit deleted successfully" even if the entry wasn't removed — which silently reproduces the original bug. Surfacing a secondary warning would be clearer.
3. **Look up the queue entry before voiding the visit.** The visit is voided first and the entry is fetched after; if the lookup filters on active visits, it can come back empty and skip the DELETE.
4. **Undo symmetry.** `restoreDeletedVisit` restores the visit but doesn't bring the queue entry back, so after Delete → Undo the patient won't reappear in the queue.
5. **Tests.** The new cases mock `openmrsFetch` end-to-end, so they pass for any endpoint/shape — they confirm internal wiring but can't catch a wrong resource or response shape. Worth noting so we don't lean on them for that.

None of this is a knock on the work — the diagnosis is right and the interim fix is close. Mostly it's about pushing the real fix down to where the lifecycle already lives.

---
_Generated by [Claude Code](https://claude.ai/code)_

denniskigen:
Thanks @Josephkagimu1 for the investigation here. Your diagnosis was right, and your screenshots turned out to be load-bearing evidence: they demonstrated that `visit-queue-entry` still works (it is served by [`Legacy1xRestController`](https://github.com/openmrs/openmrs-module-queue/blob/main/omod/src/main/java/org/openmrs/module/queue/web/Legacy1xRestController.java) in queue 2.x/3.x, contrary to earlier concern on this thread) and that the whole flow, including the lookup after the visit is voided, behaves as you described.

That said, after tracing this to the root cause, I'd recommend we close this PR in favor of a backend fix that already exists. Summary of what we found, verified against a running RefApp backend (core 2.8.7, queue 3.0.0) with REST calls only:

- The bug reproduces with no frontend involved: `DELETE /visit/{uuid}` leaves the visit's queue entries active. So this was never a frontend defect, and a frontend fix would patch only one client while REST integrations and other apps stay broken.
- The queue module already contains code intended to fix exactly this: the [O3-4416](https://openmrs.atlassian.net/browse/O3-4416) void branch in [`VisitWithQueueEntriesSaveHandler`](https://github.com/openmrs/openmrs-module-queue/blob/main/api/src/main/java/org/openmrs/module/queue/api/VisitWithQueueEntriesSaveHandler.java). It never fires on the RefApp: its `visit.getVoided()` guard depends on core's `BaseVoidHandler` running first, and with both handlers at the default `@Handler` order, core resolves the tie by `HashMap` iteration order. That is also why *ending* a visit removes the patient from the queue while *deleting* one doesn't: the end-visit path reads `stopDatetime`, which is already set, and depends on nothing else.
- The fix is [openmrs-module-queue#96](https://github.com/openmrs/openmrs-module-queue/pull/96), which hooks `voidVisit` and `purgeVisit` directly via AOP advice, immune to the ordering issue. I built it and verified it live: visit delete voids the queue entries with the reason propagated, and visit purge no longer fails on the FK constraint. It also runs inside the visit-void transaction, and it fixes every client at once. It's now approved.

Once that PR merges and ships in a queue release, this behavior arrives with the distro with no frontend change needed, and this PR's added call would become a no-op (the entries are already voided by the time the lookup runs). So the most useful places for your energy, if you're interested: testing openmrs/openmrs-module-queue#96 against your reproduction scenario here, and two follow-ups that came out of the review, retiring the superseded handler branch and handling `unvoidVisit` so delete-then-undo restores the queue entry. Either would carry your investigation through to the root cause, and this kind of tracing is exactly what makes those contributions go well.


[O3-4416]: https://openmrs.atlassian.net/browse/O3-4416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Josephkagimu1:
Thanks @denniskigen and @VeronicaMuthee for the thorough review and for tracing this to the root cause. I agree the backend fix in openmrs/openmrs-module-queue#96 is the right approach 
Am Closing this PR in favor of that backend fix.

PR #3444, opened 2026-07-01
(chore) O3-5783: Direct dev-dep bumps in patient chart

Josephkagimu1:
## Requirements

## Summary
Bumps `@playwright/test`, `webpack-dev-server`, and `vitest` to their latest versions to resolve known security vulnerabilities. Updating `vitest` also transitively pulls in a patched version of `vite`. No application code changes — dev dependencies only. All 65 tasks pass via `yarn verify`.

## Related Issue
https://openmrs.atlassian.net/browse/O3-5783


solomonfortune:
@Josephkagimu1 I noticed that e2e tests failed and i think that happened because the file upload locator is no longer unique. getByRole('button', { name: ... }) now resolves to both the upload button and the underlying file input, causing Playwright's strict mode to fail before the filechooser event is triggered. I think the most robust solution is to interact with the file input directly using locator('input[type="file"]').setInputFiles(...), as this avoids locator ambiguity and the native file chooser altogether, making the tests more reliable and less susceptible to UI or accessibility changes. Otherwise thanks for taking this up.

PR #3458, opened 2026-07-05
(feat) O3-5798: Add past visit indicator to patient banner

Janani-Balasooriya:
## Requirements

## Summary
When a user switches into a past visit context, the workspace panel correctly shows "Adding to: [past visit]", but the patient banner at the top still just says "Active Visit", nothing tells the user they're actually charting into an old, closed visit. This can cause real confusion during retrospective data entry, since it's easy to assume you're working in the current visit when you're not.

This PR adds a small tag to the patient banner that reads "Past visit: {date}" whenever this context is active. It uses the same visit-context data the workspace indicator already relies on, so the two stay in sync automatically.
The tag disappears once the user switches back to the active visit.

## Screenshots

BEFORE - the banner still says "Active Visit" even while charting into a past visit, with no way to tell you're not in the current visit:

before

https://github.com/user-attachments/assets/f35c77c1-a100-4c65-814f-7ad4adae5cc7

AFTER - the banner now shows a "Past visit" tag alongside "Active Visit", so it's always clear which visit you're actually entering data into:

after

Short recording showing the tag appearing when switching into a past visit and disappearing when switching back:

https://github.com/user-attachments/assets/ea27902e-288b-4668-8e40-d82facbdbec3

## Related Issue
https://openmrs.atlassian.net/browse/O3-5798

## Other
N/A

### DIFF
--- a/packages/esm-patient-banner-app/src/banner-tags/past-visit-tag.extension.tsx
+++ b/packages/esm-patient-banner-app/src/banner-tags/past-visit-tag.extension.tsx
@@ -1,0 +1,49 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { Tag, Toggletip, ToggletipButton, ToggletipContent } from '@carbon/react';
+import { type Visit, formatDate, formatDatetime, parseDate, useFeatureFlag } from '@openmrs/esm-framework';
+import { usePatientChartStore, useSystemVisitSetting } from '@openmrs/esm-patient-common-lib';
+import styles from './past-visit-tag.scss';
+
+interface PastVisitTagProps {
+  patientUuid: string;
+}
+
+function PastVisitTag({ patientUuid }: PastVisitTagProps) {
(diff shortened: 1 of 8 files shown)


slubwama:
@Janani-Balasooriya  Thanks for this. Given that the visit context would be in the past, Its important to remove the active visit while in a past visit context.

Janani-Balasooriya:
@slubwama  Thanks for the feedback! That makes sense, showing both together could read as contradictory. I'll update it so "Active Visit" is hidden while the past-visit context is active, and only the "Past visit" tag shows.

Will push a follow-up commit shortly.

Janani-Balasooriya:
@slubwama  Pushed the fix. "Active Visit" now hides itself while a past-visit context is active. Let me know if this looks right. Thank you!

after 2

slubwama:
@denniskigen help review the  rest to ensure It matches to the standards. We need this expedited. Since its needed really soon

slubwama (approved):
LGTM

PR #3460, opened 2026-07-07
(chore) O3-5788: Upgrade webpack-bundle-analyzer to 5.3.0 to fix ws DoS vulnerability

solomonfortune:
## Requirements

## Summary
`webpack-bundle-analyzer@4.10.2` pulls in `ws@7.5.10`, affected by [GHSA-96hv-2xvq-fx4p](https://github.com/advisories/GHSA-96hv-2xvq-fx4p) (memory-exhaustion DoS, fixed in `ws@7.5.11`). No `4.x` patch fixes this — `4.10.2` is the last `4.x` release ever published. The fix only lands in `5.2.0+`, which bumps `ws` to `^8.19.0`.

`webpack-bundle-analyzer` isn't declared directly anywhere in this repo — it's pulled in transitively via `@openmrs/rspack-config`, `@openmrs/webpack-config`, `@rspack/cli`, and `openmrs`. Added a `resolutions` override in root `package.json` to force `5.3.0` repo-wide.

No breaking API changes between `4.x` and `5.x` per the [official changelog](https://github.com/webpack/webpack-bundle-analyzer/blob/main/CHANGELOG.md) — only requirement is Node >= 20.9.0, already satisfied here.

## Screenshots
N/A (build tooling change, no UI impact)

## Related Issue
https://issues.openmrs.org/browse/O3-5788

## Other
- Verified `ws` now resolves to `8.21.0` under `webpack-bundle-analyzer` (`yarn why ws`)
- `yarn build` and `yarn analyze` both pass in `esm-patient-chart-app`, confirming the analyzer still works
- Unrelated: a full `turbo run build` shows `esm-form-entry-app` failing (Angular/Sass loader parsing issue, pre-existing, unaffected by this change — that package doesn't depend on `webpack-bundle-analyzer`). Not touched by this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,8 @@
     "*.{css,scss,ts,tsx}": "prettier --write --list-different"
   },
   "resolutions": {
-    "sass": "^1.54.3"
+    "sass": "^1.54.3",
+    "webpack-bundle-analyzer": "5.3.0"
   },
   "packageManager": "yarn@4.10.3"
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -2339,17 +2339,17 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@discoveryjs/json-ext@npm:0.5.7, @discoveryjs/json-ext@npm:^0.5.7":
-  version: 0.5.7
-  resolution: "@discoveryjs/json-ext@npm:0.5.7"
-  checksum: 10/b95682a852448e8ef50d6f8e3b7ba288aab3fd98a2bafbe46881a3db0c6e7248a2debe9e1ee0d4137c521e4743ca5bbcb1c0765c9d7b3e0ef53231506fec42b4
-  languageName: node
-  linkType: hard
-
-"@discoveryjs/json-ext@npm:0.6.3, @discoveryjs/json-ext@npm:^0.6.1":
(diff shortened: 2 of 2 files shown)


solomonfortune:
@jayasanka-sack This also seems to be done. Its up for review

jayasanka-sack (changes requested):
Hi @solomonfortune, thanks for the PR.  

As stated in the epic, you should not patch transitive packages directly. Instead fix the direct dependency that pulls them in. If that comes from the @openmrs/rspack-config, that means you have to fix that module first.

solomonfortune:
> Hi @solomonfortune, thanks for the PR.
> 
> As stated in the epic, you should not patch transitive packages directly. Instead fix the direct dependency that pulls them in. If that comes from the @openmrs/rspack-config, that means you have to fix that module first.

Thanks for catching this. re-reading the epic (O3-5784) confirms it explicitlyrules out resolutions entries except for the dompurify exception. My mistake.
I think I traced it: webpack-bundle-analyzer@4.10.2 is pulled in via @openmrs/rspack-config, @openmrs/webpack-config, @rspack/cli, and openmrs. The first two are OpenMRS-owned (openmrs-esm-core). I'll open a PR there bumping webpack-bundle-analyzer to 5.3.0 directly, then come back here for a normal version bump once that's released. 
@rspack/cli is third-party and out of OpenMRS's control will flag separately since we can't fix it upstream ourselves. Reverting the resolutions entry here in the meantime.

I stand to be corrected in case this is still not true

PR #3462, opened 2026-07-08
(feat) O3-5797: Add full day visit checkbox to past visit form

Janani-Balasooriya:
## Requirements

## Summary
Currently, creating a past visit requires users to specify the visit time. This often results in visit windows that do not encompass all encounters intended to belong to the visit, which can cause validation errors when creating or importing encounters whose timestamps fall outside the visit duration. In many retrospective documentation workflows, users only know the date on which the patient visited the facility, not the exact arrival or departure time, so requiring time selection adds unnecessary complexity.

This PR adds a "Full day visit" checkbox to the past visit form, checked by default. When checked, the form collapses down to a single date field and the visit is set to span from 00:00 to 23:59 on that date. Unchecking it restores the existing behavior of specifying exact start and end times.

The checkbox also correctly reflects the actual state of a visit being edited: it defaults to checked only when the existing visit's stored times genuinely span a full day, and stays unchecked if the visit was created with specific manual times.

### Before

Creating a past visit requires manually filling in both a start and end date/time:

before

### After

Default - the checkbox is checked by default, collapsing to a single date field:

after default

Unchecked - unchecking restores the full manual start/end date and time fields:
after uncheck

Video demonstration
https://github.com/user-attachments/assets/46545af7-6aef-4333-a9c9-cfe9d5f41da0

## Related Issue
https://openmrs.atlassian.net/browse/O3-5797

## Other
N/A


slubwama:
@denniskigen  help review this to ensure It matches to the standards. We need this expedited. Since its needed really soon

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.resource.ts
```diff
@@ -232,6 +233,19 @@ export function useVisitFormSchemaAndDefaultValues(visitToEdit: Visit, earliestA
...
+    // precision) because the backend doesn't necessarily preserve the exact 23:59:59.999 the
+    // frontend computes via dayjs(...).endOf('day') on round-trip through the REST API. Visits
+    // with specific manual times must not be misrepresented as full-day. New visits are handled
+    // separately by a useEffect in exported-visit-form.workspace.tsx that defaults this to true
+    // when the user switches a brand new visit's status to "past" via the content switcher.
+    const isVisitToEditFullDay =
+      Boolean(visitToEdit?.startDatetime) &&
+      Boolean(visitToEdit?.stopDatetime) &&
+      dayjs(visitToEdit.startDatetime).isSame(dayjs(visitToEdit.startDatetime).startOf('day'), 'minute') &&
+      dayjs(visitToEdit.stopDatetime).isSame(dayjs(visitToEdit.stopDatetime).endOf('day'), 'minute');
```
denniskigen, line 248: This checks that the visit starts at midnight and ends at 23:59, but not that both happen on the same day. So a visit from Jan 1 00:00 to Jan 3 23:59 shows up with the checkbox already checked and only Jan 1 in the date field. If the user then saves any change, the visit quietly shrinks from three days to one. I confirmed this in a test by checking what gets sent to `updateVisit`.

```suggestion
    const isVisitToEditFullDay =
      Boolean(visitToEdit?.startDatetime) &&
      Boolean(visitToEdit?.stopDatetime) &&
      dayjs(visitToEdit.startDatetime).isSame(dayjs(visitToEdit.stopDatetime), 'day') &&
      dayjs(visitToEdit.startDatetime).isSame(dayjs(visitToEdit.startDatetime).startOf('day'), 'minute') &&
      dayjs(visitToEdit.stopDatetime).isSame(dayjs(visitToEdit.stopDatetime).endOf('day'), 'minute');
```

A `mockFullDayPastVisit` variant whose stop is on a later day would cover this.

packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx
```diff
@@ -74,14 +77,31 @@ const VisitDateTimeSection: React.FC<VisitDateTimeSectionProps> = ({
...
+          )}
+        />
+      )}
       <VisitDateTimeField
-        dateField={{ name: 'visitStartDate', label: t('startDate', 'Start date') }}
+        dateField={{ name: 'visitStartDate', label: isFullDay ? t('date', 'Date') : t('startDate', 'Start date') }}
         timeField={{ name: 'visitStartTime', label: t('startTime', 'Start time') }}
         timeFormatField={{ name: 'visitStartTimeFormat', label: t('startTimeFormat', 'Start time format') }}
         minDate={earliestStartDate}
         maxDate={minOf(Date.now(), firstEncounterDateTime, selectedVisitStopDateTime?.getTime())}
```
denniskigen: When the checkbox is checked, the end date fields are hidden, but their old values still feed into this `maxDate`. That stops the user from ever moving a full day visit to a later date. I confirmed this by checking the `maxDate` prop that gets passed to `OpenmrsDatePicker`.

```suggestion
        maxDate={minOf(Date.now(), firstEncounterDateTime, isFullDay ? undefined : selectedVisitStopDateTime?.getTime())}
```

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.resource.ts
```diff
@@ -289,6 +305,7 @@ export function useVisitFormSchemaAndDefaultValues(visitToEdit: Visit, earliestA
           visitStopDate,
           visitStopTime,
           visitStopTimeFormat,
+          isFullDayVisit,
         } = data;
 
         const visitStartDateTime = convertToDate(visitStartDate, visitStartTime, visitStartTimeFormat);
```
denniskigen: When the checkbox is checked, the form saves midnight and 23:59 of `visitStartDate`, but these two lines still validate the hidden time fields. So the schema can reject a save that would actually be fine, and the error message lands on `visitStartTime`, a field that isn't on screen. To the user it looks like the Save button just doesn't work. Building the values here the same way the submit handler does fixes that, and also lets the end-time checks below run in full day mode:

```ts
        const isFullDay = visitStatus === 'past' && isFullDayVisit;
        const visitStartDateTime = isFullDay
          ? visitStartDate
            ? dayjs(visitStartDate).startOf('day').toDate()
            : null
          : convertToDate(visitStartDate, visitStartTime, visitStartTimeFormat);
        const visitStopDateTime = isFullDay
          ? visitStartDate
            ? dayjs(visitStartDate).endOf('day').toDate()
            : null
          : convertToDate(visitStopDate, visitStopTime, visitStopTimeFormat);
```

The error paths in the checks below should then point at `visitStartDate` when `isFullDay` is true, since it's the only field the user can see.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.resource.ts
```diff
@@ -331,7 +348,7 @@ export function useVisitFormSchemaAndDefaultValues(visitToEdit: Visit, earliestA
           }
         }
 
-        if (visitStatus === 'past') {
+        if (visitStatus === 'past' && !isFullDayVisit) {
```
denniskigen: Skipping these checks when the checkbox is checked loses two protections the manual flow has. Picking today saves an end time of tonight at 23:59, which is still in the future; the manual flow blocks that, and the backend won't, since [VisitValidator](https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/validator/VisitValidator.java) has no future check. And if the visit has encounters on more than one day, the save fails on the backend with a snackbar error instead of the inline message the manual flow shows. With the values built as in the previous suggestion, restoring the original condition brings both checks back:

```suggestion
        if (visitStatus === 'past') {
```

packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx
```diff
@@ -74,14 +77,31 @@ const VisitDateTimeSection: React.FC<VisitDateTimeSectionProps> = ({
           ? t('visitStartDate', 'Visit start date')
           : t('visitStartAndEndDate', 'Visit start and end date')}
```
denniskigen: With the checkbox checked there's only one date field, but the header still reads "Visit start and end date". Could read "Visit date" in full-day mode.

denniskigen (changes requested):
Really nice work, @Janani-Balasooriya.

I've left a few suggestions inline, most of which you can commit directly. The two most important ones are the full-day detection and the date picker's max date. The others come from the same root cause: when the checkbox is checked, the form saves times built from `visitStartDate`, but the validation still looks at the hidden time fields.

@ibacher should a full day visit be allowed for today? Its end time would be tonight at 23:59, which is still in the future. The backend accepts that, so it's our call on the client side.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.resource.ts
denniskigen: One more piece for this fix: the `visitStartTime` and `visitStopTime` format checks in the base schema still run when the checkbox is checked. So if the user types a bad time and then checks "Full day visit", the form still refuses to save, and the error sits on a field that's no longer on screen. I reproduced this by typing an invalid time and then checking the box.

Could we skip the time format checks in full day mode too, for example by moving them into `superRefine` so they only run for manual times, or by splitting the schema into full day and manual variants? A test that types an invalid time, checks the box, and saves successfully would cover this.

ibacher:
> should a full day visit be allowed for today?

I think for the purposes it's meant to serve, it should not apply to visits today. I also think, stylistically, that this should use a toggle and not a checkbox.

Janani-Balasooriya:
> Really nice work, @Janani-Balasooriya.
> 
> I've left a few suggestions inline, most of which you can commit directly. The two most important ones are the full-day detection and the date picker's max date. The others come from the same root cause: when the checkbox is checked, the form saves times built from `visitStartDate`, but the validation still looks at the hidden time fields.
> 
> @ibacher should a full day visit be allowed for today? Its end time would be tonight at 23:59, which is still in the future. The backend accepts that, so it's our call on the client side.

Thanks so much for the detailed review, really appreciate you taking the time to go through this. I've gone through all the inline comments and fixed each one, along with the max date issue you flagged. Just pushed a commit with all the fixes. Replying to each comment below with details.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.resource.ts
Janani-Balasooriya: Fixed, thanks for catching this. Added a check that confirms start and stop are on the same calendar day, not just midnight and 23:59. Added a test with a 3 day visit span confirming the checkbox correctly stays unchecked.

packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx
Janani-Balasooriya: Fixed. maxDate now ignores the stale end date when full day mode is checked, so it's only bounded by today and the first encounter date.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.resource.ts
Janani-Balasooriya: Fixed both of these. Added an isFullDay flag and used it to build visitStartDateTime and visitStopDateTime the same way the submit handler does, so validation matches what actually gets saved. Error paths now point to visitStartDate when isFullDay is true.

For the time format checks, added a useEffect that clears the time fields when full day mode is on, so the optional regex check passes cleanly instead of failing on a leftover invalid value. Added tests covering both, one confirming a valid full day save, and one confirming an invalid time followed by checking the box still saves successfully.

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.resource.ts
Janani-Balasooriya: Fixed, restored the original condition so both checks run in full day mode too, using the correctly built values. Added tests confirming each one fires.

packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx
Janani-Balasooriya: Fixed, the header now reads "Visit date" when full day mode is on and "Visit start and end date" otherwise. Added a test covering this.

Janani-Balasooriya:
> > should a full day visit be allowed for today?
> 
> I think for the purposes it's meant to serve, it should not apply to visits today. I also think, stylistically, that this should use a toggle and not a checkbox.

Thanks for the input on both of these. For the today restriction, I've already fixed it, the date picker now excludes today as a selectable date when full day visit is checked, capped at yesterday instead. If someone has today selected and then checks the box, it auto corrects the date back to yesterday rather than letting them hit a validation error on save.

For the toggle, agreed that makes more sense stylistically. I'll be doing that as a separate commit right after this one and will comment here once it's up.

packages/esm-patient-chart-app/src/visit/visit-form/exported-visit-form.workspace.tsx
```diff
@@ -180,6 +182,16 @@ const ExportedVisitForm: React.FC<Workspace2DefinitionProps<ExportedVisitFormPro
...
 
+  // When creating a new visit (not editing an existing one), the user can switch the visit status
+  // to "past" via the content switcher. defaultValues only defaults isFullDayVisit to true when
+  // visitToEdit is itself already a past visit, so this keeps the checkbox checked-by-default
+  // for that case too, without affecting the edit flow (which is excluded via !visitToEdit).
+  useEffect(() => {
+    if (!visitToEdit && visitStatus === 'past') {
+      setValue('isFullDayVisit', true);
+    }
+  }, [visitStatus, visitToEdit, setValue]);
```
ibacher: This feels like it needs an "unset" pair if I move off of the "past" tab. I'd also suggest that we make this a configurable option (i.e., what the default value is) rather than hard-code it.

packages/esm-patient-chart-app/src/visit/visit-form/exported-visit-form.workspace.tsx
```diff
@@ -283,13 +295,19 @@ const ExportedVisitForm: React.FC<Workspace2DefinitionProps<ExportedVisitFormPro
...
         visitStopTimeFormat,
+        isFullDayVisit,
       } = data;
 
       const { handleCreateExtraVisitInfo, attributes: extraAttributes } = extraVisitInfo ?? {};
       const hasStartTime = ['ongoing', 'past'].includes(visitStatus);
       const hasStopTime = 'past' === visitStatus;
-      const startDatetime = convertToDate(visitStartDate, visitStartTime, visitStartTimeFormat);
-      const stopDatetime = convertToDate(visitStopDate, visitStopTime, visitStopTimeFormat);
+      const isFullDay = isFullDayVisit && visitStatus === 'past';
```
ibacher, line 314: While redundant, I think it might read better to put `!isVisitToEdit` here.

slubwama:
@Janani-Balasooriya still working on this?? you do have comments

packages/esm-patient-chart-app/src/visit/visit-form/exported-visit-form.workspace.tsx
Janani-Balasooriya: Fixed both. Merged this into a single effect that always sets isFullDayVisit to match the current status, so it resets correctly whether entering or leaving "past", rather than needing two separate effects. Also moved the default into config, added a new defaultToFullDayVisit boolean in config-schema.ts, defaulting to true so existing behavior doesn't change unless an implementation configures it otherwise. Added tests covering the round trip (uncheck, leave past, come back, defaults correctly) and the config override case.

packages/esm-patient-chart-app/src/visit/visit-form/exported-visit-form.workspace.tsx
Janani-Balasooriya: Thanks, I looked into this before changing it and wanted to check my understanding. This line decides how startDatetime/stopDatetime get built for the save payload, including when editing an existing full day visit. If I swap in !isVisitToEdit here, isFullDay would become false whenever isVisitToEdit is true, so editing an existing full day visit would stop building midnight to 23:59 datetimes. Am I missing something that makes them equivalent in this spot, or would that change behavior? Happy to make the change if I'm wrong about this.

Janani-Balasooriya:
> @Janani-Balasooriya still working on this?? you do have comments

@slubwama  Yes, still actively working on it. I've addressed all of denniskigen's review comments and one of ibacher's two comments so far, just waiting on his confirmation on the second one. Next up is switching the checkbox to a toggle per his suggestion, and fixing a failing E2E test, both coming in the next couple of commits. Thank you!

Janani-Balasooriya:
@slubwama  @denniskigen  @ibacher Thanks everyone for your patience and the thorough review so far.

All of denniskigen's review comments are resolved and pushed. For ibacher's feedback, the checkbox has been replaced with a toggle, screenshots below (on and off states).

Still waiting on ibacher's thoughts on one open question about the !isVisitToEdit suggestion above, once that's resolved this should be ready for another look. Thanks again!

image
image

packages/esm-patient-chart-app/src/visit/visit-form/visit-form.resource.ts
denniskigen: Thanks, the full-day validation now correctly checks the computed end-of-day value. One issue remains: the end-time errors are still assigned to `visitStopTime`, which is hidden in full-day mode. I reproduced this locally by changing a multi-day visit to a full-day visit. Clicking Update blocked the save, but displayed no validation message or invalid field. In that scenario it's the `endTimeMustBeAfterMostRecentEncounter` check that fires, and its error lands on the hidden field.

Could we attach these errors to the visible date field in full-day mode? There are two `path: ['visitStopTime']` occurrences in the past-visit block (the `futureEndTime` check and the `endTimeMustBeAfterMostRecentEncounter` check), and both would become:

```ts
path: [isFullDay ? 'visitStartDate' : 'visitStopTime'],
```

I'm writing this as a plain snippet rather than a suggestion because these lines aren't part of the diff, so GitHub can't apply a suggestion to them.

Janani-Balasooriya: Fixed, thanks for the clear steps to reproduce it. Updated both occurrences (futureEndTime and endTimeMustBeAfterMostRecentEncounter checks) to use isFullDay ? 'visitStartDate' : 'visitStopTime', same pattern as the earlier start-time fix. Also strengthened the existing tests to assert the error path directly, and added a UI-level test that reproduces your exact scenario, editing a multi-day visit into full-day mode and confirming the error now shows on the visible date field.

slubwama:
@denniskigen @ibacher  any other issues that need to be resolved

PR #3471, opened 2026-07-11
(chore) O3-5792: Bump vitest to 4.1.10 to pull vite >= 8.0.16

sudeshkar:
## Requirements

## Summary

Resolves **GHSA-fx2h-pf6j-xcff** (HIGH) — a `server.fs.deny` bypass via alternate paths that lets a request read outside the served directory. It affects `vite@8.0.12` and is fixed in `8.0.16`. The bug is Windows-only, so it cannot be reproduced locally on Linux or macOS.

The affected `vite` is pulled in transitively by `vitest` / `@vitest/mocker`, so **there is no runtime exposure** — the blast radius is the test runner on Windows dev machines and Windows CI runners.

### What changed

- Bumped `vitest` from `^4.1.2` to `^4.1.10` in the root manifest and in each of the 21 workspaces that declare it.
- Bumped `@vitest/coverage-v8` to `^4.1.10` in lockstep — `vitest` peer-pins it to its own exact version.
- Re-resolved the transitive `vite` descriptor, which now lands on `8.2.0`.

### Why the version bump alone isn't sufficient

`vitest@4.1.10` declares the *same* permissive `vite` range as `4.1.6` (`^6.0.0 || ^7.0.0 || ^8.0.0`). Because the descriptor string is unchanged, Yarn reuses its existing lockfile entry and happily keeps the vulnerable `vite@8.0.12` pinned. The descriptor therefore has to be re-resolved explicitly for the fix to actually take effect — a bump plus a plain `yarn install` silently leaves the advisory in place.

### Why there is no `resolutions` entry

`vite` appears twice in this tree, and only one of the two is affected:

| Dependent | vite | Status |
| --- | --- | --- |
| `@angular/build@20.3.32` | `7.3.6` (exact pin) | Unaffected — untouched by this PR |
| `vitest@4.1.10` | `8.0.12` → **`8.2.0`** | Patched |

A root-level `"vite"` resolution is not scoped to a parent, so Yarn applies it to **every** `vite` descriptor in the tree. That would also override `@angular/build`'s exact `vite@7.3.6` pin and force the Angular build toolchain used by `esm-form-entry-app` onto a major version it does not support — a real regression risk in exchange for no security benefit, since vite 7 is not affected by this advisory.

This also keeps the change consistent with the parent epic's guidance to fix the direct dependency rather than patching transitives via `resolutions`.

## Screenshots

*N/A — dependency-only change, no UI impact.*

## Related Issue

https://issues.openmrs.org/browse/O3-5792

Parent epic: https://issues.openmrs.org/browse/O3-5784

## Other

Verification performed locally:

- `yarn install --immutable` — passes, lockfile is consistent and CI-stable.
- `yarn verify` (lint + typescript + test) — **65/65 tasks pass**.
- `yarn test` — all **22 workspace test suites pass**.
- `yarn npm audit --recursive --severity moderate` — **zero** advisories remain against `vite`.
- `yarn why vite` confirms the split resolution:

```
├─ @angular/build@npm:20.3.32
│  └─ vite@npm:7.3.6 (via npm:7.3.6)
│
└─ vitest@npm:4.1.10
   └─ vite@npm:8.2.0 (via npm:^6.0.0 || ^7.0.0 || ^8.0.0)
```

The remaining audit findings belong to the other packages tracked under the parent epic and are out of scope here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "@types/webpack-env": "^1.18.8",
     "@typescript-eslint/eslint-plugin": "^8.0.0",
     "@typescript-eslint/parser": "^8.0.0",
-    "@vitest/coverage-v8": "^4.1.2",
+    "@vitest/coverage-v8": "^4.1.10",
     "babel-preset-minify": "^0.5.2",
     "concurrently": "^8.2.1",
     "cross-env": "^10.1.0",
@@ -74,7 +74,7 @@
     "swr": "2.2.5",
     "turbo": "^2.9.14",
     "typescript": "^5.0.0",
-    "vitest": "^4.1.2",
+    "vitest": "^4.1.10",
     "webpack-dev-server": "^5.2.4"
   },
   "lint-staged": {
--- a/packages/esm-form-engine-app/package.json
+++ b/packages/esm-form-engine-app/package.json
@@ -54,6 +54,6 @@
   "devDependencies": {
     "@openmrs/esm-patient-common-lib": "workspace:*",
     "openmrs": "next",
-    "vitest": "^4.1.2"
+    "vitest": "^4.1.10"
   }
 }
--- a/packages/esm-generic-patient-widgets-app/package.json
+++ b/packages/esm-generic-patient-widgets-app/package.json
@@ -54,6 +54,6 @@
   "devDependencies": {
     "@openmrs/esm-patient-common-lib": "workspace:*",
     "openmrs": "next",
-    "vitest": "^4.1.2"
+    "vitest": "^4.1.10"
   }
 }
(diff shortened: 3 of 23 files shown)


sudeshkar:
@jayasanka-sack kindly review this

Josephkagimu1 (commented):
LGTM ✅ Verified that vite resolves to 8.2.0 (>= 8.0.16), tests pass (65/65 tasks, 22 suites), and Angular's @angular/build vite 7.3.6 pin is preserved. The approach of re-resolving the transitive descriptor rather than using a blanket resolutions entry is the right call.
what is remaining , resolve the conflicts  for this PR .

PR #3472, opened 2026-07-13
(fix) O3-5789: Upgrade webpack-dev-server & engine.io to pull ws >= 8.21.0

solomonfortune:
## Requirements
- [ ] My work is based on designs, which are linked or shown either in the Jira ticket or the description below. (See also: [Styleguide](http://om.rs/o3ui)) — N/A, dependency security fix.

## Summary

Fixes both `ws` advisories in O3-5789 ([GHSA-96hv-2xvq-fx4p](https://github.com/advisories/GHSA-96hv-2xvq-fx4p), [GHSA-58qx-3vcg-4xpx](https://github.com/advisories/GHSA-58qx-3vcg-4xpx)) by re-resolving three transitive chains to patched versions, all within existing semver ranges — no major versions bumped.

- `ws@8.17.1` came via `karma` → `socket.io` → `engine.io@~6.5.2`. Bumped `karma` `~6.4.2` → `~6.4.4` (patch-level) in `packages/esm-form-entry-app/package.json`, which pulls `socket.io@4.8.3` → `engine.io@~6.6.0` → `ws@~8.21.0`.
- `ws@8.20.1` came via `webpack-dev-server`'s own `ws: ^8.18.0` range, which already allowed `8.21.0` — just needed `yarn up -R` to force re-resolution. No `package.json` change needed here.
- `ws@8.17.1` also came independently via `socket.io-adapter@2.5.5`, fixed the same way (`yarn up -R` → `socket.io-adapter@2.5.8`).
- Deliberately avoided upgrading `webpack-dev-server` to the newly-released `6.0.0` — out of scope for this fix.

## Screenshots

N/A — no UI changes.

## Related Issue

https://issues.openmrs.org/browse/O3-5789

## Other

Verified via `yarn why ws`: all resolutions now `8.21.0` (except unrelated `ws@7.5.11` via `webpack-bundle-analyzer`, a different major version outside this ticket's scope). `yarn start` confirmed dev server boots and HMR works.

### DIFF
--- a/packages/esm-form-entry-app/package.json
+++ b/packages/esm-form-entry-app/package.json
@@ -94,7 +94,7 @@
     "eslint": "^8.57.0",
     "jasmine-core": "~5.1.1",
     "jasmine-spec-reporter": "~7.0.0",
-    "karma": "~6.4.2",
+    "karma": "~6.4.4",
     "karma-chrome-launcher": "~3.2.0",
     "karma-coverage-istanbul-reporter": "~3.0.3",
     "karma-jasmine": "~5.1.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -16679,16 +16679,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"launch-editor@npm:^2.14.1":
-  version: 2.14.1
-  resolution: "launch-editor@npm:2.14.1"
-  dependencies:
-    picocolors: "npm:^1.1.1"
-    shell-quote: "npm:^1.8.4"
-  checksum: 10/335d12ca437280e77070657531c251b6c91c62bc653f70ab66ddd2a6e50131b1b043480411c5b93d54955a0a6eb8ec01e9a5b5cfe2d887341d878d19394a126b
-  languageName: node
-  linkType: hard
-
 "launch-editor@npm:^2.6.1":
   version: 2.10.0
   resolution: "launch-editor@npm:2.10.0"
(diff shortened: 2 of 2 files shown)


solomonfortune:
@jayasanka-sack please review this in your time of convenience.

Josephkagimu1 (commented):
Thanks for the fix! @solomonfortune  , The ws@8.x bumps look good and address the security advisories in [Ticket](https://openmrs.atlassian.net/browse/O3-5789)

One request: could we scope this down to only the changes required by the ticket? I noticed a few extra transitive bumps that came along for the ride:

ws@7.5.10 → 7.5.11 (via webpack-bundle-analyzer) — not covered by the ticket, which targets ws@8.x only
webpack-dev-server 5.2.4 → 5.2.6 — pulls in new launch-editor and shell-quote deps beyond what's needed
socket.io 4.7.5 → 4.8.3 — a minor version bump (not patch), which also swaps out cookie and @types/cookie
While these are likely low-risk, keeping the lockfile diff minimal makes it easier to review, bisect, and roll back if needed. Could you re-run the resolution targeting only engine.io and karma to avoid the unrelated changes? That way the PR stays 1:1 with the ticket scope. 
Thanks , continue the good work !

solomonfortune:
> Thanks for the fix! @solomonfortune , The [ws@8.x](mailto:ws@8.x) bumps look good and address the security advisories in [Ticket](https://openmrs.atlassian.net/browse/O3-5789)
> 
> One request: could we scope this down to only the changes required by the ticket? I noticed a few extra transitive bumps that came along for the ride:
> 
> ws@7.5.10 → 7.5.11 (via webpack-bundle-analyzer) — not covered by the ticket, which targets [ws@8.x](mailto:ws@8.x) only webpack-dev-server 5.2.4 → 5.2.6 — pulls in new launch-editor and shell-quote deps beyond what's needed socket.io 4.7.5 → 4.8.3 — a minor version bump (not patch), which also swaps out cookie and @types/cookie While these are likely low-risk, keeping the lockfile diff minimal makes it easier to review, bisect, and roll back if needed. Could you re-run the resolution targeting only engine.io and karma to avoid the unrelated changes? That way the PR stays 1:1 with the ticket scope. Thanks , continue the good work !

Thanks for the catch! i have re-scoped and pushed. webpack-dev-server now stays at 5.2.4 and unchanged. 
Two things I couldn't avoid though. one is that socket.io 4.7.5 → 4.8.3 is required, not incidental, since 4.7.5 hard-pins engine.io@~6.5.2. two is that there's no way to reach the patched engine.io/ws without it; and the webpack-bundle-analyzer → ws@7.5.11 bump is a side effect of yarn up -R matching every ws range project-wide, with no way to scope it per-chain using this tooling.

PR #3474, opened 2026-07-14
(fix) O3-5798: Only hide the active visit tag when the past visit tag renders

denniskigen:
## Requirements

## Summary

Follow-up to #3458, which added a "Past visit" tag to the patient banner and hid the "Active Visit" tag while a past visit is in context.

The two conditions could disagree: the active visit tag hid whenever the visit context held an ended visit, but the past visit tag that replaces it only renders when system visits are enabled and the `rde` feature flag is on. A past visit can land in the visit context without RDE, for example after editing a past visit's details from the visit history, which sets the edited visit as the visit context on save. In that state the banner showed no visit tag at all even though the patient has an active visit, and with RDE off the visit context switcher isn't available to recover with, so the tag stayed hidden until a full page reload.

This extracts the gating into a shared `useIsInPastVisitContext` hook used by both tags, so the active visit tag only hides when the past visit tag actually renders in its place. Also removes two unused style classes that were copied over from `visit-tag.scss`.

Verified live against the local RefApp distro: with RDE off, editing a past visit while an active visit is ongoing dropped the Active Visit tag from the banner on current `main`, and the tag stays in place with this change. Also covered by a new regression test.

### Before

With RDE off, editing a past visit while an active visit is ongoing drops the Active Visit tag from the patient header.

https://github.com/user-attachments/assets/9b8f357a-d1c7-48e5-a4b5-0cf664939d69

### After

Active visit context is retained

https://github.com/user-attachments/assets/88a52dad-7ed9-4216-aee6-343a30d9e935

## Related Issue

https://openmrs.atlassian.net/browse/O3-5798

### DIFF
--- a/packages/esm-patient-banner-app/src/banner-tags/past-visit-tag.extension.tsx
+++ b/packages/esm-patient-banner-app/src/banner-tags/past-visit-tag.extension.tsx
@@ -1,8 +1,9 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { Tag, Toggletip, ToggletipButton, ToggletipContent } from '@carbon/react';
-import { type Visit, formatDate, formatDatetime, parseDate, useFeatureFlag } from '@openmrs/esm-framework';
-import { usePatientChartStore, useSystemVisitSetting } from '@openmrs/esm-patient-common-lib';
+import { type Visit, formatDate, formatDatetime, parseDate } from '@openmrs/esm-framework';
+import { usePatientChartStore } from '@openmrs/esm-patient-common-lib';
+import { useIsInPastVisitContext } from '../hooks/useIsInPastVisitContext';
 import styles from './past-visit-tag.scss';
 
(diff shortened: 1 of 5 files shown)


ibacher (approved):
Although should we maybe unify the "Past" and "Active" visit tags into one extension?

denniskigen:
Set to draft pending design review from @chibongho @fiona855.

@chibongho's comment: The biggest issue we have is that the active visit tag in the patient banner is being interpreted  as "our current visit context is the active visit", and not "this patient currently has an active visit". I think there we should have better indications that we're currently in RDE mode, but I don't think that should be in the patient banner. I think we'll need more discussion around the design.

denniskigen:
@chibongho is this still viable in its current shape?

chibongho:
@denniskigen I think we should close this PR, as the RDE design has drifted away from this.

denniskigen:
Cool, closing the related issue too.

PR #3484, opened 2026-07-16
(feat) O3-5677: Add print visit summary button to visit history row actions

BlessedAmrita:
## Requirements

## Summary

Adds a "Print visit summary" button to each visit row in the patient chart's Visits tab. Clicking it opens a modal that previews the visit summary PDF inline, with a Download button always available in the footer as a fallback for browsers that cannot render the PDF in an iframe.

The button and modal live in `esm-patient-label-printing-app`, which already depends on the `patientdocuments` backend module (it hosts the existing patient identifier sticker printing). The button registers into the `visit-detail-overview-actions` slot at order 2, after the edit and delete visit buttons.

The PDF is fetched from the existing backend endpoint `GET /ws/rest/v1/patientdocuments/visitSummary?visitUuid=<uuid>` as a blob via `openmrsFetch`, rather than pointing an iframe at the URL directly, so that HTTP failures can be surfaced distinctly. Each failure maps to its own translated message in the modal: 403 (not authorized), 404 (visit not found), any other status (generation failed), and no response (network error).

The button is gated on the `App: Can generate a Visit Summary` privilege and is hidden by default behind the `showPrintVisitSummaryButton` config flag (default `false`), matching the pattern of the existing sticker button.

## Dependencies

This is the frontend half of the feature. For it to be usable by non-admin users on a distribution:

- The `patientdocuments` module version shipped must contain the visit summary endpoint (present on `main`, `1.2.0-SNAPSHOT`; not in the released `1.1.0`). `backendDependencies` is left at the current bound and should be tightened to `>=1.2.0` once that version is released.
- The `App: Can generate a Visit Summary` privilege must be registered on the server. It is currently enforced by the backend but not created at startup, so on a fresh install only superusers pass the check. Tracked separately as a backend fix.

## Screenshots 

https://github.com/user-attachments/assets/11b30607-260c-47b3-9957-fd0c731bc784

## Related Issue

https://openmrs.atlassian.net/browse/O3-5677
Parent epic: https://openmrs.atlassian.net/browse/O3-5667


BlessedAmrita:
cc: @NethmiRodrigo @wikumChamith

packages/esm-patient-label-printing-app/src/print-visit-summary/print-visit-summary-action-button.component.tsx
```diff
@@ -0,0 +1,56 @@
...
+import type { ConfigObject } from '../config-schema';
+
+interface PrintVisitSummaryActionButtonProps {
+  visit: Visit;
+  patient: fhir.Patient;
+
+  /**
+   * If true, renders as IconButton instead
+   */
+  compact?: boolean;
```
ibacher, line 16: How would someone ever change this value?

packages/esm-patient-label-printing-app/src/print-visit-summary/print-visit-summary.modal.tsx
```diff
@@ -0,0 +1,120 @@
...
+      subtitle: t(
+        'visitSummaryGenerationFailed',
+        'The server could not generate the visit summary. Try again — if the problem persists, contact your administrator.',
+      ),
+    },
+    network: {
+      title: t('visitSummaryNetworkErrorTitle', 'Network error'),
+      subtitle: t(
+        'visitSummaryNetworkError',
+        'The visit summary could not be retrieved. Check your internet connection and try again.',
```
ibacher:
```suggestion
        'The visit summary could not be retrieved. Check your network connection and try again.',
```

packages/esm-patient-label-printing-app/src/print-visit-summary/print-visit-summary.modal.tsx
```diff
@@ -0,0 +1,120 @@
...
+      })
+      .catch((error) => {
+        if (abortController.signal.aborted) {
+          return;
+        }
+        setErrorType(getVisitSummaryPdfErrorType(error));
+      });
+
+    return () => {
+      abortController.abort();
```
ibacher, line 42: Yay!! Someone correctly used an abortController!

packages/esm-patient-label-printing-app/src/print-visit-summary/print-visit-summary.modal.tsx
```diff
@@ -0,0 +1,120 @@
...
+      if (createdObjectUrl) {
+        window.URL.revokeObjectURL(createdObjectUrl);
```
ibacher, line 44: Shouldn't this also unset the objectUrl?

packages/esm-patient-label-printing-app/src/print-visit-summary/print-visit-summary-action-button.component.tsx
BlessedAmrita: `compact` arrives via the extension slot state. In this repo only `visit-actions-cell.component.tsx` renders `visit-detail-overview-actions`, and it always passes `compact: true`, so nothing can currently make this false. The other two extensions in the same slot keep the identical prop and branch, which is why I matched them and just clarified the JSDoc.

That said, if the intent is that this slot is always compact, the prop and the branch are both dead weight and I'm happy to take them out. What would you prefer?

denniskigen:
Solid work, @BlessedAmrita. One update on the Dependencies section: the privilege registration gap you flagged has since been fixed. https://github.com/openmrs/openmrs-module-patientdocuments#31 declares `App: Can generate a Visit Summary` in `config.xml` on the module's `main`, so both backend prerequisites are now just waiting on the 1.2.0 release.

wikumChamith:
> Solid work, @BlessedAmrita. One update on the Dependencies section: the privilege registration gap you flagged has since been fixed. [openmrs/openmrs-module-patientdocuments#31](https://github.com/openmrs/openmrs-module-patientdocuments?rgh-link-date=2026-07-23T20%3A21%3A48Z#31) declares `App: Can generate a Visit Summary` in `config.xml` on the module's `main`, so both backend prerequisites are now just waiting on the 1.2.0 release.

@denniskigen because we are using 1.2.0-SNAPSHOT in distro, can't we move forward with this? 

https://github.com/openmrs/openmrs-distro-referenceapplication/blob/6a40b247c26c860afa1436024042001e56a2a27e/distro/pom.xml#L31

packages/esm-patient-label-printing-app/src/print-visit-summary/print-visit-summary.modal.tsx
```diff
@@ -0,0 +1,124 @@
...
+      title: t('visitSummaryNetworkErrorTitle', 'Network error'),
+      subtitle: t(
+        'visitSummaryNetworkError',
+        'The visit summary could not be retrieved. Check your network connection and try again.',
+      ),
+    },
+  };
+
+  return (
+    <div>
```
denniskigen:
```suggestion
    <>
```

Change the closing tag to `</>` too. `showModal` [expects the modal sections at the component top level](https://github.com/openmrs/openmrs-esm-core/blob/3757bfe5/packages/framework/esm-styleguide/src/modals/index.tsx#L188-L194). This wrapper makes the whole modal a single grid item, so the footer can render outside the panel and become inaccessible on shorter viewports.

packages/esm-patient-label-printing-app/src/print-visit-summary/print-visit-summary.scss
```diff
@@ -0,0 +1,14 @@
...
+
+.modalBody {
+  display: flex;
+  flex-direction: column;
+  min-height: layout.$spacing-13;
+}
+
+.previewFrame {
+  width: 100%;
+  min-height: 70vh;
```
denniskigen:
```suggestion
  min-height: 50vh;
```

Even with the modal sections at the top level, `70vh` is taller than Carbon's content row on shorter viewports. This adds an outer modal scrollbar around the PDF viewer, which already has its own scrolling.

denniskigen (commented):
Good work, @BlessedAmrita. The fetch and error handling look solid. I've left one layout issue that can make the footer inaccessible on shorter viewports.

packages/esm-patient-label-printing-app/src/print-visit-summary/print-visit-summary-action-button.component.tsx
denniskigen: You read it right, @BlessedAmrita, I checked all three. `edit-visit-details.component.tsx` and `delete-visit-action-item.component.tsx` carry the same `compact?: boolean` with the same branch, and `visit-actions-cell.component.tsx` is the only host, always passing `compact: true`. Let's keep it here so the three stay identical. If we want the prop gone it should come out of all three at once in its own PR, not something to hold this up over.

denniskigen:
@wikumChamith yes, nothing blocks on the module version, the distro is on 1.2.0-SNAPSHOT. Worth tightening `backendDependencies` once 1.2.0 is released, and `^1.2.0-SNAPSHOT` rather than `>=1.2.0` so we don't flag the distro while it's still on the snapshot. Nothing for this PR to do there. The only thing outstanding from me is the layout issue I left above.

packages/esm-patient-label-printing-app/src/print-visit-summary/print-visit-summary.modal.tsx
BlessedAmrita: Fixed, sections are at the top level now, so the footer stays in the panel.

packages/esm-patient-label-printing-app/src/print-visit-summary/print-visit-summary.scss
BlessedAmrita: Done, drops the outer scrollbar.

denniskigen (approved):
LGTM. Thanks, @BlessedAmrita!

PR #3490, opened 2026-07-18
(fix) O3-5817: Re-register the built-in endpoint data source during form creation

denniskigen:
## Requirements

## Summary

The form engine registers its built-in `endpoint` data source (openmrs/openmrs-ngx-formentry#178) once, when `FormEntryModule` is constructed. This app clears every registered data source at the start of each form creation and never restores the built-in one. Questions using `"datasource": { "name": "endpoint" }` therefore rendered without options, and the engine logged `No data source registered under 'endpoint'`.

The fix re-registers a fresh `EndpointDataSource` right after the clearing step, before the configurable data sources, so a custom data source registered under the same name still overrides it.

Includes the first spec for `FormCreationService`, covering both behaviors.

## Related Issue

https://openmrs.atlassian.net/browse/O3-5817

## Other

Companion engine hardening PR: openmrs/openmrs-ngx-formentry#180


maikofelix47 (approved):
LGTM

PR #3496, opened 2026-07-21
(fix) O3-5822: Dedupe diagnoses in visit-level rollups

denniskigen:
## Requirements

## Summary

When the same diagnosis is recorded in multiple encounters of one visit (for example, two visit notes in the same visit both asserting Diabetes mellitus), the visit-level rollups repeat the tag once per encounter. Both the Diagnoses column in the visit history table and the expanded visit summary aggregate every encounter's diagnoses without deduplication before passing them to `DiagnosisTags`. Multi-encounter and multi-day visits are the worst case, since each progress note adds another copy of the same tag.

This adds a shared `dedupeDiagnoses` helper and applies it in both views. Coded diagnoses dedupe by concept UUID and non-coded diagnoses by exact text. The list is sorted by rank before deduplication, so a diagnosis that is primary in any encounter of the visit shows as primary.

This is a display-level change only. Encounter-level diagnosis records are unchanged, and per-encounter data still reflects every assertion as recorded.

### Before

Duplicate "Diabetes mellitus" tags

CleanShot 2026-07-21 at 17 05 44@2x

### After

Only one top-level "Diabetes mellitus" tag

CleanShot 2026-07-21 at 21 47 01@2x

## Related Issue

https://openmrs.atlassian.net/browse/O3-5822

## Other

The ticket flags deduplicating non-coded diagnoses by exact text as a proposed display rule rather than settled behavior. Happy to adjust if reviewers would rather keep non-coded entries separate.

### DIFF
--- a/packages/esm-patient-chart-app/src/visit/dedupe-diagnoses.test.ts
+++ b/packages/esm-patient-chart-app/src/visit/dedupe-diagnoses.test.ts
@@ -1,0 +1,63 @@
+import { describe, expect, it } from 'vitest';
+import { type Diagnosis } from '@openmrs/esm-framework';
+import { dedupeDiagnoses } from './dedupe-diagnoses';
+
+const codedDiagnosis = (uuid: string, display: string, rank: number): Diagnosis =>
+  ({
+    uuid: `diagnosis-${uuid}-rank-${rank}`,
+    display,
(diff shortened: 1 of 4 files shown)


ibacher (approved):

VeronicaMuthee:
@denniskigen thanks for getting this fixed.

PR #3499, opened 2026-07-22
(fix) O3-5824: Keep dashboard grid widgets contained at narrow viewports

denniskigen:
## Requirements

## Summary

Dashboard widgets that are wider than the space available to them — like the vitals table — spill past the edge of the page and get cut off. There's no scrollbar, so the cut-off content (the SpO2 column, the table pagination, the Add button) is effectively hidden. This happens on tablets, on smaller desktop windows, and at any desktop size when an open workspace narrows the chart.

Two CSS Grid defaults cause it:

1. The single-column dashboard layout uses a plain `1fr` column, which can't shrink below its content's natural width. The two-column layout already avoids this with `minmax(0, 1fr)`.
2. The widgets are grid items, which default to `min-width: auto` and also refuse to shrink below their content.

This PR fixes both: the single-column layout now uses `minmax(0, 1fr)`, and widgets get `min-width: 0`. Fixing only the column is not enough — the widget itself still overflows (verified: the column shrank to 678px while the widget stayed at 831px). With both in place, widgets stay contained and Carbon's built-in table scrolling takes over, so wide tables scroll inside their own cards.

Tested at every width from 672px to 1920px, on both sides of each layout breakpoint, and with a workspace open: nothing is clipped, and the two-column layout on large screens is unchanged. Full measurements are in the ticket.

### Before

before-1024

### After

after-1024

## Related Issue

https://openmrs.atlassian.net/browse/O3-5824

### DIFF
--- a/packages/esm-patient-chart-app/src/patient-chart/chart-review/dashboard-view.scss
+++ b/packages/esm-patient-chart-app/src/patient-chart/chart-review/dashboard-view.scss
@@ -36,7 +36,7 @@
 // proper rendering when the workspace area is opened.
 @container dashboard (width <= 68.25rem) {
   .dashboard {
-    grid-template-columns: 1fr;
+    grid-template-columns: minmax(0, 1fr);
   }
 }
 
@@ -48,6 +48,13 @@
   }
 }
 
+.extension {
+  // Grid items default to min-width: auto, which lets wide content (e.g. data
+  // tables) expand the item beyond its track and get clipped. min-width: 0
+  // keeps widgets contained so tables scroll within their own cards instead.
+  min-width: 0;
+}
+
 .extension:only-child {
   grid-column: 1 / -1;
 }
@@ -57,7 +64,7 @@
 }
 
 :global(.omrs-breakpoint-lt-desktop) .dashboard {
-  grid-template-columns: 1fr;
+  grid-template-columns: minmax(0, 1fr);
 }
 
 :global(.omrs-breakpoint-lt-tablet) .container {


packages/esm-patient-chart-app/src/patient-chart/chart-review/dashboard-view.scss
```diff
@@ -36,7 +36,7 @@
 // proper rendering when the workspace area is opened.
 @container dashboard (width <= 68.25rem) {
   .dashboard {
-    grid-template-columns: 1fr;
+    grid-template-columns: minmax(0, 1fr);
   }
 }
```
ibacher, line 41: Do we just want:

```scss
.dashboard {
 grid-template-columns: minmax(0, 1fr);
}
```

Unifying this rule and the global one below and removing the need to do any container (or implicit container) queries?

ibacher (approved):

packages/esm-patient-chart-app/src/patient-chart/chart-review/dashboard-view.scss
denniskigen: Not with that unconditional rule as written. It would override the base two-column layout used on large desktops. At 1920px, I verified that the dashboard changes from two 608px columns to one 1238px column.

**Current two-column layout**

Current two-column dashboard layout at 1920px

**With the proposed unconditional rule**

Single-column dashboard layout at 1920px with the proposed unconditional rule

We still need both scoped overrides:

- The container query reacts when a workspace narrows the dashboard without changing the viewport. At 1440px with Visit Note open, the dashboard is only about 674px wide.
- The `lt-desktop` rule covers the Attachments dashboard, which is intentionally excluded from container sizing at line 29. Without it, Attachments would not collapse on tablet.

We could revisit the Attachments exclusion separately, but that would change its existing layout behavior and needs dedicated testing.

PR #3501, opened 2026-07-22
(chore) O3-5803: Bump @angular/* to 20.3.26 to resolve Dependabot security alerts

Josephkagimu1:
## Requirements

## Summary
Bumps `@angular/core`, `@angular/common`, and `@angular/compiler` to 20.3.26  (from 20.3.21) to resolve 9 Dependabot security alerts (5 high, 4 moderate) related to XSS sanitization bypasses.

Added `@angular-devkit/core` as an explicit devDependency for `esm-form-entry-app` to fix a missing module error caused by `hoistingLimits`.

## Related Issue
https://openmrs.atlassian.net/browse/O3-5803


Josephkagimu1:
CC : @pirupius , @denniskigen , @jabahum

denniskigen (commented):
Thanks, @Josephkagimu1! You got a tricky part right: Angular uses two separate version lines, so the framework packages go to 20.3.26 while the CLI and devkit go to 20.3.32. Good catch on @angular-devkit/core too.

I pushed two small commits to tie things together:

- The first runs `yarn dedupe`. `yarn up -R "@angular/*"` does not match `@schematics/angular`, so it stayed at 20.3.26 while `@angular/cli` moved to 20.3.32. That left two copies of it, plus two each of `@angular-devkit/core` and `@angular-devkit/schematics`. `yarn dedupe --check` found 48 duplicates here and none on main. There are none now. Worth running after any bump.

- The second changes `package.json` from ^20.3.21 to ^20.3.26. The ticket called this optional, but it is what makes the fix reach other people. The lockfile only fixes this repo. Anyone whose lockfile already has 20.3.21 keeps it, because ^20.3.21 still allows it and nothing tells yarn to look again. All eleven had to change together, since each Angular package requires the others at exactly the same version. `@angular/cdk` and `@angular/material` stay at ^20.2.14 because they are on the other version line.

denniskigen (approved):
Thanks, @Josephkagimu1! You got a tricky part right: Angular uses two separate version lines, so the framework packages go to 20.3.26 while the CLI and devkit go to 20.3.32. Good catch on `@angular-devkit/core` too.

I pushed two small commits to tie things together:

- The first runs `yarn dedupe`. `yarn up -R "@angular/*"` does not match `@schematics/angular`, so it stayed at 20.3.26 while `@angular/cli` moved to 20.3.32. That left two copies of it, plus two each of `@angular-devkit/core` and `@angular-devkit/schematics`. `yarn dedupe --check` found 48 duplicates here and none on main. There are none now. Worth running after any bump.

- The second changes `package.json` from ^20.3.21 to ^20.3.26. The ticket called this optional, but it is what makes the fix reach other people. The lockfile only fixes this repo. Anyone whose lockfile already has 20.3.21 keeps it, because ^20.3.21 still allows it and nothing tells yarn to look again. All eleven had to change together, since each Angular package requires the others at exactly the same version. `@angular/cdk` and `@angular/material` stay at ^20.2.14 because they are on the other version line.

Josephkagimu1:
Thanks @denniskigen for that .

PR #3507, opened 2026-07-24
(feat) Order basket should try to rely on server-side dates

ibacher:
## Requirements

## Summary

This PR attempts to address an issue where it still possible for the order basket to attempt to create an order encounter dated to a time which is in the future from the server's perspective. Future-dated encounters are rejected by the backend.

When an order is enqueued, we effectively use the client's date / time as the `startDate` field for the order. Then, when submitting the order we calculate the Encounter date time as the earliest of: the `scheduledDate` of any order, the `startDate` of any order, or the current date. All of these datetimes are client-local. In most circumstances, network delays and processing times means that as long as the client and server are synced to the same NTP server, this is largely unproblematic (NTP drift is normally well < 100ms), but if they are not synced, it possible for the client to be in the future by a small lead from server (a few seconds), which isn't made up form during request submission and processing, resulting in the server rejecting the request.

This tries to work around this problem by inferring, at the time of entry, whether an order is being back-dated or forward-dated ("user-selected date") or just relying on the default (i.e., the order date is the today). If the order date is today, we submit an order (and encounter) with no datetime, allowing the server side to fill in these datetime fields. Because they use the server timestamps, these timestamps should always be earlier or equal to the moment the encounter validation fires.


denniskigen (approved):
LGTM. Thanks, @ibacher!

PR #3508, opened 2026-07-24
(test) Assert real pluralized wait time in vitals header overdue tag

denniskigen:
## Requirements

## Summary

The `displays correct overdue tag for vitals 5 days old` test asserted `These vitals are 5 day old` (singular "day"). It only passed because the shared `react-i18next` test mock does not implement pluralization and renders the singular default children of the `<Trans>`. The running app renders `These vitals are 5 days old` via the `daysOldVitals_other` key.

This switches the test file to the real `react-i18next` so the overdue tag resolves the plural keys from `en.json`, and corrects the assertion to `These vitals are 5 days old`. The component render now verifies the right plural form end to end, complementing the key-resolution test added in #3045.

Two setup notes, both commented inline: the shared vitest config aliases `react-i18next` to a passthrough mock, so the real module is resolved by absolute path to bypass the alias; and `window.i18next` is pointed at the real instance (as the app shell does) so the `@openmrs/esm-translations` init hook does not crash.


packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.test.tsx
```diff
@@ -58,6 +61,26 @@ vi.mock('../common/data.resource', async () => {
...
+});
+
+// The @openmrs/esm-framework mock stubs `window.i18next`, but @openmrs/esm-translations registers
+// an "initialized" hook that calls `window.i18next.loadNamespaces`. Point the global at the real
+// instance (as the app shell does) so init does not crash.
+window.i18next = i18n;
+
+i18n.use(initReactI18next).init({
+  lng: 'en',
+  initImmediate: false,
```
ibacher: Two minor things:

1. `initImmediate` has been renamed `initAsync` (found this out when looking into below)
2. Do we need this here? From the docs it seems like it wouldn't actually be doing anything ("This option only works for sync (blocking) loading backend", but if I've got this correct, when supplying `resources` directly, there's no "backend" involved).

ibacher (approved):

packages/esm-patient-vitals-app/src/vitals-and-biometrics-header/vitals-header.test.tsx
denniskigen: Good catch, thanks. Confirmed against the library source: `initImmediate` only survives as a back-compat shim onto `initAsync`, and since we pass inline `resources`, `init()` calls `load()` synchronously either way, so the option does nothing on this path.

denniskigen: For the record, the relevant branch in i18next 25.6.0's `init()`:

```js
if (this.options.resources || !this.options.initAsync) {
  load();
} else {
  setTimeout(load, 0);
}
```

`resources` short-circuits the check, so the flag can never influence behavior here.

PR #3509, opened 2026-07-25
(fix) O3-5763: Expose the visit notes form as a reusable workspace

UjjawalPrabhat:
## Requirements

## Summary

`VisitNotesForm` gets its visit from `usePatientChartStore`, which only the patient chart populates. Launched from anywhere else, `visitContext` is null, `visit` is dropped from the payload, and the backend's visit assignment handler picks a visit instead. The note saves and the success snackbar shows, but it can land on a different visit than the one the caller is displaying. Service Queues hits this today on its Current visit tab.

Splits the form the way the vitals form already is: the implementation takes `visitContext` as a workspace prop and is exported as `exportedVisitNotesFormWorkspace`; `visit-notes-form.workspace.tsx` stays as the chart-only wrapper. Chart behaviour is unchanged - `PatientWorkspaceGroupProps` already carries `visitContext`, so the wrapper passes through the same value the store was returning.

## Related Issue

[O3-5763](https://openmrs.atlassian.net/browse/O3-5763)

## Other

Related PR [here](https://github.com/openmrs/openmrs-esm-patient-management/pull/2600)

[O3-5763]: https://openmrs.atlassian.net/browse/O3-5763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/esm-patient-notes-app/src/notes/visit-notes-form.workspace.tsx
```diff
@@ -1,859 +1,32 @@
...
-                handleSearch(name);
-              }}
-              value={value instanceof Date ? value.toISOString() : value}
-              onBlur={onBlur}
-            />
-          </ResponsiveWrapper>
-          {fieldState?.error?.message && <p className={styles.errorMessage}>{fieldState?.error?.message}</p>}
-        </>
-      )}
+    <ExportedVisitNotesForm
```
NethmiRodrigo: I understand that this is the pattern that we’ve followed in this repo for some places (i.e., the [vitals and biometrics form](https://github.com/openmrs/openmrs-esm-patient-chart/blob/78729b9fea5c5de4cdc2c6a4756abe8fbe22cfb7/packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.workspace.tsx#L22) and the [lab results form](https://github.com/openmrs/openmrs-esm-patient-chart/blob/78729b9fea5c5de4cdc2c6a4756abe8fbe22cfb7/packages/esm-patient-orders-app/src/lab-results/lab-results-form.workspace.tsx#L36)), but the fact that we are creating the form itself inside the exported version and making the internal one use the one that’s meant to be used outside the patient chart feels ...unnatural to me. Do we have a reason for doing this @ibacher @denniskigen?

NethmiRodrigo: @chibongho tagging you as well in case you might know :D

denniskigen: Good question. The dependency direction in this PR is functionally correct: the patient chart workspace adapts its group props into the explicit props required outside the chart. However, I agree that making the chart workspace depend on a component named `ExportedVisitNotesForm` obscures the actual architecture.

I don't think there is a framework requirement for this structure. Perhaps the clearest approach would be to move the shared implementation into `visit-notes-form.component.tsx`, then make both `visit-notes-form.workspace.tsx` and `exported-visit-notes-form.workspace.tsx` thin adapters around it. That keeps the shared form independent of either launch context. It is also what most of our exported workspaces already do: [form entry](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-forms-app/src/forms/form-entry.workspace.tsx), [add drug order](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-medications-app/src/add-drug-order/add-drug-order.workspace.tsx), the order basket, add general order and add test order all keep the implementation in a `.component.tsx` and render `Workspace2` from there. The vitals and lab results forms you linked are the exceptions rather than the rule, so your instinct here is right.

I see this as a maintainability improvement rather than a correctness blocker.

NethmiRodrigo (approved):
Thanks @UjjawalPrabhat!

PR #3511, opened 2026-07-27
(test) O3-4825: Assert all vitals for every age group in age-based vitals e2e tests

VeronicaMuthee:
## Requirements

## Summary

Refactors the age-based vitals e2e suite (`e2e/specs/age-based-vitals.spec.ts`) to a single table-driven set of scenarios and adds the preschooler (3–<4 years) age group for [O3-4825](https://openmrs.atlassian.net/browse/O3-4825).

Previously each age group only asserted the temperature cell. Now every group (newborn, infant, child, preschooler, adult) runs two scenarios:

- All values sit comfortably inside the normal band, so nothing is flagged. This guards against false positives.
- All values sit comfortably past the critically-high cutoff, so temperature, blood pressure, pulse, and respiratory rate are each flagged critically high.

Pulse, respiratory rate, and blood pressure use age-specific reference ranges, so each group runs against its own generated patient to confirm the correct ranges resolve. Temperature is not age-specific (its range is fixed, splitting only at 3 months), so the same temperature values are used for every group.

Also extracts the `expectCellInterpretation` helper into `e2e/commands/test-helpers.ts` (typed via `VitalInterpretation`) so the assertion is shared instead of hand-rolled per test.

## Related Issue

https://openmrs.atlassian.net/browse/O3-4825

## Other

Expected reference ranges: https://docs.google.com/spreadsheets/d/1-afKi0PFI_ySQxuXKN0OTwYgU29WiMAZcy-k7SSZYN4/edit?gid=958690794#gid=958690794

[O3-4825]: https://openmrs.atlassian.net/browse/O3-4825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/e2e/specs/age-based-vitals.spec.ts
+++ b/e2e/specs/age-based-vitals.spec.ts
@@ -1,9 +1,38 @@
-import { expect } from '@playwright/test';
+import { expect, type Page } from '@playwright/test';
 import { test } from '../core';
 import { BiometricsAndVitalsPage } from '../pages';
 import { type Patient, deletePatient, endVisit, generateRandomPatient, startVisit } from '../commands';
 import { type Visit } from '@openmrs/esm-framework';
 import { calculateBirthdate, getAfterContent, getBackgroundColor } from '../commands/test-helpers';
+
+// Background color + trailing indicator that the vitals table renders for each interpretation
+// level (see the NumericObservation `cell` variant in @openmrs/esm-styleguide). The indicator is
+// emitted as CSS `::after` content, which the browser also folds into the cell's accessible name —
+// so the same string both locates the cell by role/name and is asserted as the `::after` content.
+const interpretationCellStyles: Record<string, { backgroundColor: string; indicator: string }> = {
+  normal: { backgroundColor: 'rgba(0, 0, 0, 0)', indicator: '' },
+  high: { backgroundColor: 'rgb(255, 242, 232)', indicator: ' ↑' },
+  low: { backgroundColor: 'rgb(255, 242, 232)', indicator: ' ↓' },
+  critically_high: { backgroundColor: 'rgb(255, 215, 217)', indicator: ' ↑↑' },
+  critically_low: { backgroundColor: 'rgb(255, 215, 217)', indicator: ' ↓↓' },
+};
+
+// Asserts that the table cell displaying `displayValue` (e.g. `38.5` or `145 / 100`) is styled for
+// the given interpretation: the colored background and the ↑/↓ indicator live on the inner `div`
+// rendered by NumericObservation, so we read the styling from there.
+async function expectCellInterpretation(page: Page, displayValue: string, interpretation: string) {
+  const { backgroundColor, indicator } = interpretationCellStyles[interpretation];
+  // Match the accessible name exactly: an unflagged cell's name is just its value, which would
+  // otherwise substring-match the blood-pressure cell (e.g. `100` in `100 / 60`) or the timestamp
+  // cell (e.g. `25` in a `10:25` time). The ↑/↓ indicator is folded into the accessible name.
+  const styledContent = page
+    .getByRole('cell', { name: `${displayValue}${indicator}`, exact: true })
+    .locator('div')
+    .first();
+
+  expect(await getBackgroundColor(styledContent)).toBe(backgroundColor);
+  expect(await getAfterContent(styledContent)).toBe(interpretation === 'normal' ? 'none' : `"${indicator}"`);
+}
 
 const ageGroups = [
   {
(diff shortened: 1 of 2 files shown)


VeronicaMuthee:
@jayasanka-sack @gitcliff @denniskigen, kindly review my PR.

dkayiwa:
@VeronicaMuthee this is so sweet of you, lady!!! 👍

gitcliff:
thanks @VeronicaMuthee LGTM

e2e/specs/age-based-vitals.spec.ts
```diff
@@ -1,10 +1,39 @@
...
+  // otherwise substring-match the blood-pressure cell (e.g. `100` in `100 / 60`) or the timestamp
+  // cell (e.g. `25` in a `10:25` time). The ↑/↓ indicator is folded into the accessible name.
+  const styledContent = page
+    .getByRole('cell', { name: `${displayValue}${indicator}`, exact: true })
+    .locator('div')
+    .first();
+
+  expect(await getBackgroundColor(styledContent)).toBe(backgroundColor);
+  expect(await getAfterContent(styledContent)).toBe(interpretation === 'normal' ? 'none' : `"${indicator}"`);
+}
```
dkayiwa: This is worth hoisting into `e2e/commands/test-helpers.ts`, next to `getBackgroundColor` and `getAfterContent`. The same assertion is hand-rolled four more times in the repo: `vitals.spec.ts:207-229` and `vitals.spec.ts:297-321`, plus lines 141-148 and 212-224 right below here. Moving it and switching those over would delete a fair bit of copy-paste, and it's the reason `high`, `low` and `critically_low` sit here as dead entries; nothing in the new tests reaches them.

Small thing while you're in here: typing the parameter as `keyof typeof interpretationCellStyles` rather than `string` turns a typo like `'critical_high'` into a compile error instead of a `Cannot destructure property 'backgroundColor' of undefined` at runtime.

e2e/specs/age-based-vitals.spec.ts
```diff
@@ -196,3 +225,192 @@ test.describe('Vitals validation for different age groups', () => {
     });
   });
 });
+
+// Preschooler (Preschool 3–<4 years) scenario for O3-4825. Records a critically unwell 3-year-old —
+// tachycardia, tachypnea, hypertension and fever — and asserts that every age-sensitive vital (heart
+// rate/pulse, breathing rate, blood pressure and temperature) is flagged critically high. The values
+// sit clear of the 3–<4 critical cut-offs (HR ≥136, R.rate ≥33, BP ≥140/95, temp ≥38), so the child's
+// age-specific reference ranges must resolve correctly for the flags to render.
```
dkayiwa: Temperature doesn't belong in this list. In the demo dictionary the temperature concept splits only at 3 months, so a preschooler and an adult both get 36 to 37.7 normal with a critical high of 38. I checked a 3-year-old and an adult against the running demo's reference ranges: pulse (86-123, crit 136 vs 60-100, crit 110), respiratory rate (21-29, crit 33 vs 12-18, crit 24), systolic (90-110, crit 140 vs 100-120, crit 180) and diastolic (47-75, crit 95 vs 60-80, crit 110) all move with age. Temperature is byte-for-byte the same.

So 38.5 flagging critically high tells you nothing about whether the 3 to <4 band resolved, and "every age-sensitive vital (... and temperature)" plus "the 3–<4 critical cut-offs (... temp >=38)" will mislead whoever picks values for the next age band. Dropping temperature from both sentences fixes it. The assertion itself is fine to keep.

e2e/specs/age-based-vitals.spec.ts
```diff
@@ -196,3 +225,192 @@ test.describe('Vitals validation for different age groups', () => {
...
+test.describe('Preschooler vital signs validation (3-<4 years)', () => {
```
dkayiwa: Two screens up, this file already generates these exact two scenarios (normal and critical) from the `ageGroups` table, so a preschooler could be a fifth row instead of a hand-written pair of tests. As it stands we end up with a third and fourth copy of the record-vitals flow in one file, and the next person adding an age group has two patterns to pick from. The new block also sits outside `Vitals validation for different age groups`, even though a preschooler is one.

The table doesn't fit as-is, granted: no pulse field, and it only ever checks the temperature cell. The change that makes it fit is fairly contained though. Give each row a `pulse` plus the expected interpretation per vital, and have the generated tests assert whichever vitals the row declares. That also gets the existing four groups' BP and respiration cells asserted, so it's more coverage than this PR adds for fewer lines.

I'd rather see it that way, but I wouldn't hold the PR up over it if you'd sooner land the coverage and follow up.

VeronicaMuthee: Thanks @dkayiwa for your reviews. I have moved expectCellInterpretation into test-helpers.ts (typed with keyof), refactored to the table-driven ageGroups pattern so every age group asserts pulse, BP, and respiratory rate (plus a no-false-positives scenario), and corrected the temperature framing. Left the wider dedup of the other hand-rolled assertions as a follow-up.

denniskigen (approved):
LGTM. Thanks for iterating on this, @VeronicaMuthee! The table-driven refactor @dkayiwa suggested came together cleanly, and pulling `expectCellInterpretation` into `test-helpers.ts` means every age group now asserts pulse, BP, and respiratory rate, not just the preschooler. That's more coverage than the PR started with.

The change grew past its original description along the way, so I tidied up the title and body to match what it does now.

VeronicaMuthee:
> LGTM. Thanks for iterating on this, @VeronicaMuthee! The table-driven refactor @dkayiwa suggested came together cleanly, and pulling `expectCellInterpretation` into `test-helpers.ts` means every age group now asserts pulse, BP, and respiratory rate, not just the preschooler. That's more coverage than the PR started with.
> 
> The change grew past its original description along the way, so I tidied up the title and body to match what it does now.

Thanks @denniskigen @dkayiwa for your reviews & suggestions.

dkayiwa:
@VeronicaMuthee thank you too for cooking! 😊

PR #3521, opened 2026-07-28
(feat) O3-5839: Differentiate the patient banner for deceased patients

VeronicaMuthee:
## Requirements

- [x] My work is based on designs, which are linked or shown in the Jira ticket or below.

## Summary

Implements the deceased-differentiation portion of O3-5839: differentiates the **patient header** (not the navigation bar) for deceased patients.

- Deceased patients get a somber dark-grey (`gray-70`) header band. Text/icons on the band are lightened, while Carbon components that carry their own light surfaces — identifier/status tags, the Actions overflow-menu, and toggletip/popover content (including the contact-details values) — are excluded so they stay legible.
- Ghost-button labels/icons (Actions trigger, three-dots, Show more/less toggle) are lightened, hover darkens instead of flashing white, and the keyboard focus ring is recoloured to meet WCAG 1.4.11 (box-shadow ring for the toggle; a dedicated `outline-color` for the Actions trigger).
- The Deceased tag is lifted to a light pill so it reads as a distinct chip on the dark band, and the InlineLoading spinner shown while contact details fetch is kept hollow.

Validated by the `mark-patient-deceased` E2E test. It asserts the band color, AA text contrast for contact details and tag labels, tag-to-band visual contrast, and the focus indicators for both banner controls. The spec uses shared computed-style contrast helpers, visible Playwright locators, and a stable hook for the contact-details panel.

## Screenshots

image
image
image
image

## Related Issue

https://openmrs.atlassian.net/browse/O3-5839

## Other

The always-visible / sticky-header portion of O3-5839 is deliberately **not** in this PR: `position: sticky` has no effect where the banner is mounted (its containing block is the banner's own height, and `#omrs-apps-container`'s overflow — not the viewport — is the scrollport), so genuinely pinning the header needs changes in `patient-chart.component.tsx` and esm-core. Thanks @dkayiwa for the detailed review.


VeronicaMuthee:
@denniskigen @dkayiwa @ibacher please review my PR

packages/esm-patient-banner-app/src/banner/patient-banner.scss
```diff
@@ -6,13 +7,62 @@
   border-top: 0.063rem solid $ui-03;
   border-bottom: 0.063rem solid $ui-03;
   background-color: $ui-01;
+  // Keep the patient header visible while the chart scrolls. The whole window is the scroll
+  // container, so we stick just below the top nav (mirrors #omrs-workspaces-container in the
+  // styleguide). The z-index sits above chart content but below the workspace panel (1000).
+  position: sticky;
+  top: var(--omrs-navbar-height);
+  z-index: 100;
```
dkayiwa: This block needs to come out before this merges, or move to its own follow-up, because it has no effect where the banner is actually mounted. I checked it on dev3 by injecting exactly these three declarations onto the live patient banner: before scrolling the header sits at 48px, and after scrolling the window to the bottom it sits at -769px, so it scrolls away exactly as it does today.

Two separate things block it. First, the extension system renders the banner into a `<div data-extension-id="patient-banner">` whose height is exactly the banner's own (98px on dev3), and a sticky box cannot be shifted outside its containing block, so there is zero travel available. Second, `#omrs-apps-container` sets `overflow-x: auto`, which makes the computed `overflow-y` `auto` as well, so that container rather than the viewport is the scrollport for everything inside it, and since it is `height: fit-content` its scrollHeight equals its clientHeight and it never scrolls vertically. That is also why the `#omrs-workspaces-container` precedent in the comment above doesn't carry over: that element is a direct grid child of `body`, not nested inside the apps container. I tried patching both, and pinning then lasts 220px, which is the spare height of the `<aside>` in `patient-chart.component.tsx` that wraps the header and info slots.

If this merges as is, the always-visible-header half of O3-5839 ships as dead CSS: the header keeps scrolling out of view while the ticket reads as delivered. The QA note on the ticket saying sticky-on-scroll was confirmed doesn't hold up either, so that's worth correcting there. Getting the header to genuinely stick means giving it a containing block that spans the scrolling chart content, which is a change in `patient-chart.component.tsx`, plus most likely the `overflow-x` on `#omrs-apps-container` in esm-core. That's a bigger piece of work than this PR, so I'd split it.

packages/esm-patient-banner-app/src/banner/patient-banner.scss
```diff
@@ -6,13 +7,62 @@
...
+// name, gender, age/birthdate and the "Actions" trigger render as plain text/icons directly on
+// the band, so we lighten those to keep them legible. We deliberately EXCLUDE two things that
+// carry their own light backgrounds and would be masked by light text: the contents of Carbon
+// tags (the identifier pill, visit/status tags) and the actions overflow-menu options popover
+// (rendered inline in the banner, not portaled). The exclusions use attribute selectors so CSS
+// Modules leaves the global Carbon class names untouched.
+.deceasedPatientContainer {
+  background-color: colors.$gray-70;
+
+  :is(span, p, a, h1, h2, h3, h4, h5, h6):not([class*='cds--tag'] *, [class*='cds--overflow-menu-options'] *) {
```
dkayiwa: The element list here misses the contact details panel, and the result is unreadable rather than merely low contrast. `PatientBannerContactDetails` renders its values in `<li>` and `<div>`, neither of which is in the `:is()` list, so they inherit the `color: $text-02` set on the contact-details wrapper, and `$text-02` is `#525252`, the same `gray-70` as the new band. Measured contrast is 1.00. Expand "Show more" on a deceased patient and you get the four headings and the relationship link in light text with blank space where the address lines, phone number, relationship type and age should be. Nothing throws, so CI stays green and this lands on users.

The same gap catches the past-visit toggletip from a different direction. Carbon's `PopoverContent` renders a `<span class="cds--popover-content">`, which the list does match, and its background is Carbon's `$layer` (`#f4f4f4`), so that content ends up `#f4f4f4` on `#f4f4f4`, also 1.00.

Adding `li` and `div` on its own isn't enough: the exclusions match only descendants of `cds--tag`, not the tag element itself, and Carbon's `Tag` renders a `<div>`. It would pick up the light colour and its excluded children would then inherit it, so the ID pill and the Deceased tag would go light-on-light. Excluding the roots as well as their descendants, and adding `cds--popover-content`, covers all of it. I measured the amended selector against the real banner markup: every contact-details value comes out at 7.1:1, and the ID pill, the Deceased tag and the overflow-menu items keep their dark text at 13.7:1 and above.

```suggestion
  :is(span, p, a, li, div, h1, h2, h3, h4, h5, h6):not(
      [class*='cds--tag'],
      [class*='cds--tag'] *,
      [class*='cds--overflow-menu-options'],
      [class*='cds--overflow-menu-options'] *,
      [class*='cds--popover-content'],
      [class*='cds--popover-content'] *
    ) {
```

The `svg` rule just below wants the same `cds--popover-content` exclusion, for icons rendered inside those popovers.

packages/esm-patient-banner-app/src/banner/patient-banner.scss
```diff
@@ -6,13 +7,62 @@
...
+  // the styleguide, so it needs !important to override on the dark band. Scoped to the trigger so
+  // the menu popover's own icons are left alone.
+  [class*='cds--overflow-menu__trigger'] svg {
+    fill: colors.$gray-10 !important;
+  }
+
+  // Carbon ghost buttons (the Actions trigger and the Show more/less toggle) use a *light* hover
+  // background, which hides our light label text on the dark band. Darken the hover instead so
+  // the label stays readable. Exclude the menu popover, whose items sit on their own light surface.
+  [class*='cds--btn--ghost']:hover:not([class*='cds--overflow-menu-options'] *) {
```
dkayiwa, line 84: Since you're already correcting hover on these buttons, focus needs the same treatment, and as it stands it's an accessibility regression. Carbon draws the focus ring for ghost buttons in `$button-focus-color`, which defaults to `$focus` (`#0f62fe`). Against the old `#f4f4f4` banner that measures 4.55:1; against the new `gray-70` band it measures 1.56:1. So a keyboard user tabbing to Actions or Show more/less on a deceased patient's chart gets a focus indicator they effectively cannot see, well under the 3:1 that WCAG 1.4.11 requires for non-text contrast, and unlike the hover case there's no pointer position to fall back on.

Carbon exposes that colour as a custom property, so one declaration on the container covers both buttons and takes it to 7.81:1. The styleguide already uses this pattern in `toggletip/_toggletip.scss`:

```scss
.deceasedPatientContainer {
  --cds-button-focus-color: #{colors.$white};
}
```

packages/esm-patient-banner-app/src/banner/patient-banner.test.tsx
```diff
@@ -59,4 +60,21 @@ describe('PatientBanner', () => {
 
     expect(within(getButtonCol()).getByText(toggleButtonText)).toBeInTheDocument();
   });
+
+  it('applies the active-patient styling for a living patient', () => {
```
dkayiwa: Both of these pass unchanged on `main`. The `isDeceased ? deceasedPatientContainer : activePatientContainer` toggle on line 53 of the component is identical on the base branch, and with `identity-obj-proxy` standing in for the stylesheet the class names resolve the same way either side, so nothing this PR actually changes (the band colour, the text recolouring, the sticky positioning) is exercised. The `mark-patient-deceased` spec you cite in the description only asserts that the Deceased tag is visible, so it doesn't cover it either.

Unit tests can't see the CSS here, so the place that can is Playwright. On a deceased patient, assert a computed colour or contrast ratio on one of the contact-details values, and assert the banner's `getBoundingClientRect().top` after scrolling the page. Either one would have caught the two problems above.

VeronicaMuthee: @dkayiwa many thanks for your reviews - please review the changes. 
- Sticky (comment 1): Thanks for scroll-testing this, and it's dead CSS for the reasons you gave (extension wrapper height + #omrs-apps-container overflow). Removed it from this PR and splitting the always-visible header into a follow-up; I'll correct the QA note on O3-5839. Thanks for the thorough diagnosis.
- Contact details (comment 2): "Confirmed — the values render in li/div and inherited $text-02, the same grey as the band. Adopted your selector (li/div + excluding tag/menu/popover roots and descendants) and added cds--popover-content to the svg rule."
- Focus (comment 3): "Added --cds-button-focus-color: #{colors.$white} on the container — good catch, hover alone wasn't enough."
- Tests (comment 4): "Agreed the unit tests were no-ops. Reverted them and extended the mark-patient-deceased e2e to assert the band colour and AA contrast on a contact-details value (which would have caught the readability bug)."

packages/esm-patient-banner-app/src/banner/patient-banner.scss
```diff
@@ -8,11 +9,69 @@
...
+// text and icons that sit directly on it. Carbon components that carry their own light surfaces
+// (identifier/status tags, the actions overflow-menu popover, and toggletip/popover content) are
+// excluded from the recolour so their dark-on-light text is never flipped to light-on-light. The
+// exclusions use attribute selectors so CSS Modules leaves the global Carbon class names alone.
+.deceasedPatientContainer {
+  background-color: colors.$gray-70;
+  // Carbon's ghost-button focus ring defaults to blue ($focus, #0f62fe), which is only 1.56:1
+  // against this band — below the 3:1 WCAG 1.4.11 needs for non-text contrast. Recolor it light
+  // so keyboard focus is visible on the Actions trigger and the Show more/less toggle.
+  --cds-button-focus-color: #{colors.$white};
```
dkayiwa, line 25: This only reaches the Show more/less toggle. The Actions trigger's focus ring is unchanged by it, so the part of my earlier note that said one declaration covered both buttons was wrong, and the trigger still needs a fix before this merges. Carbon gives `.cds--overflow-menu__trigger:focus` its own `outline: 2px solid var(--cds-focus)` in `overflow-menu/_overflow-menu.scss`, and an outline paints over the inset box-shadow that `--cds-button-focus-color` recolours. The trigger carries both class names (`cds--btn cds--btn--ghost cds--overflow-menu__trigger`, set in `patient-banner-actions-menu.component.tsx`), so both rules apply and the blue outline is what you see.

I rendered the trigger markup against Carbon 1.107.0's compiled CSS on the gray-70 band, once with this declaration and once without, then diffed the outer 2px ring: identical, 35328 of 38400 ring pixels at `rgb(15, 98, 254)` either way. The toggle does flip from blue to white, so the declaration is earning its keep, just not here.

As it stands, a keyboard user tabbing to Actions on a deceased patient's chart gets a 1.56:1 focus indicator (`#0f62fe` on `#525252`), against 4.55:1 on today's banner and the 3:1 that WCAG 1.4.11 wants. Nothing throws and no test looks at focus, so it lands on users.

Recolouring the trigger's own outline covers it, and it leaves the menu popover alone: those items draw their focus from `$focus` directly and sit on their own light surface, so they should stay blue. Measured on the same harness this takes the ring to `rgb(255, 255, 255)`, 7.81:1.

```suggestion
  // Carbon's ghost-button focus ring defaults to blue ($focus, #0f62fe), which is only 1.56:1
  // against this band, below the 3:1 WCAG 1.4.11 needs for non-text contrast. Recolor it light so
  // keyboard focus is visible on the Show more/less toggle.
  --cds-button-focus-color: #{colors.$white};

  // The Actions trigger needs its own rule: Carbon draws `.cds--overflow-menu__trigger:focus` with
  // an `outline`, which paints over the inset box-shadow the property above recolours. Scoped to
  // the trigger so the menu popover's items keep their blue ring on their own light surface.
  [class*='cds--overflow-menu__trigger']:focus {
    outline-color: colors.$white;
  }
```

e2e/specs/mark-patient-deceased.spec.ts
```diff
@@ -49,4 +49,35 @@ test('Mark a patient as deceased', async ({ page, patient }) => {
...
+      const relativeLuminance = ([r, g, b]: number[]) =>
+        [r, g, b]
+          .map((v) => v / 255)
+          .map((v) => (v <= 0.03928 ? v / 12.92 : Math.pow((v + 0.055) / 1.055, 2.4)))
+          .reduce((sum, v, i) => sum + v * [0.2126, 0.7152, 0.0722][i], 0);
+
+      const value = el.querySelector('ul li');
+      if (!value) {
+        throw new Error('No contact-details value found in the expanded banner');
+      }
```
dkayiwa: `ul li` doesn't resolve to a contact-details value. `PatientBannerActionsMenu` renders its popover into the banner unconditionally (only `display: none` hides it), Carbon's `OverflowMenuItem` is an `<li>` inside `ul.cds--overflow-menu-options__content`, and that whole subtree sits in `buttonCol`, ahead of the contact-details panel in document order. So `querySelector` hands back the first Actions menu item.

I rebuilt the banner DOM against the compiled Carbon 1.107.0 CSS plus this PR's rules and ran this exact code against it: it picks `li.cds--overflow-menu-options__option` ("Mark patient alive") with an empty `getClientRects()`, and reports 7.10. The address value it was meant to measure also comes out at 7.10, which is why the assertion passes and the E2E job is green.

That 7.10 is inherited rather than measured. Menu items are excluded from the recolour, so the `<li>` picks up `gray-10` from the ancestor `div.cds--overflow-menu`. The assertion therefore tracks whether some ancestor `div` is light, not whether the contact-details values are readable, which is the one thing this step exists to guard. Merged as is, the readability fix ships with its guard pointed at the wrong element, and a later change that leaves the wrapper divs light while breaking the values goes in green.

Skipping the menu subtree fixes the targeting:

```suggestion
      const value = Array.from(el.querySelectorAll('ul li')).find(
        (li) => !li.closest('[class*="cds--overflow-menu-options"]'),
      );
      if (!value) {
        throw new Error('No contact-details value found in the expanded banner');
      }
```

(`banner.locator('ul li:visible').first()` with a `toBeVisible()` ahead of the `evaluate` would work too and buys an auto-wait for the panel's fetches, but it restructures more of the step. I'd take the version above.)

The other half of the recolour has no guard either. I dropped just the two `cds--tag` root exclusions from the selector and the identifier pill fell to 1.20:1 on its own `#e0e0e0` tag, while this step still returned 7.10 and passed. A companion assertion that the pill keeps its dark text (13.71:1 today) would close that off, and it's the half my earlier note showed is easy to get wrong.

dkayiwa (commented):
The description still describes the sticky header and the unit tests. It says the header is kept always visible, has a Sticky header bullet stating the banner is now `position: sticky` below the top nav, and lists unit tests for the active vs deceased container class under both the summary and Other. None of that is in the code now, which is the right call, but the description is what the ticket and the merge commit carry, and someone reading O3-5839 later would take the always-visible half as delivered. That was the reason the dead CSS was worth pulling in the first place, so it's worth trimming this down to the recolour plus the e2e step before it goes in.

e2e/specs/mark-patient-deceased.spec.ts
VeronicaMuthee: @dkayiwa i have updated the PR with the suggested changes. Please let me know

dkayiwa (commented):
The bullet list at the top of the description still promises the header is kept always visible, says it "is now sticky, pinned just below the top nav while the chart scrolls", and lists unit tests asserting the container class. None of that is in the code, and the Other section further down now says the sticky half is deliberately out, so the description contradicts itself. The merge commit carries this text and it is what someone reading O3-5839 later goes by, which was the reason the dead CSS was worth pulling in the first place. Trimming those two bullets and the intro sentence would leave it matching what actually ships.

packages/esm-patient-banner-app/src/banner/patient-banner.scss
```diff
@@ -8,11 +9,69 @@
...
   background-color: $ui-01;
 }
 
+// Differentiate the header for deceased patients with a somber, dark grey band, lightening the
+// text and icons that sit directly on it. Carbon components that carry their own light surfaces
+// (identifier/status tags, the actions overflow-menu popover, and toggletip/popover content) are
+// excluded from the recolour so their dark-on-light text is never flipped to light-on-light. The
+// exclusions use attribute selectors so CSS Modules leaves the global Carbon class names alone.
+.deceasedPatientContainer {
+  background-color: colors.$gray-70;
```
dkayiwa, line 21: Did the design account for what this does to the Deceased tag itself? `deceased-patient-tag.scss` gives that pill `background-color: colors.$gray-80`, which read as a firm dark chip on the old `#f4f4f4` banner at 10.51:1. Against this band it is 1.48:1, so the pill edge all but disappears and the tag reads as loose text; in your first screenshot it is the faintest thing in the header, sitting next to an identifier pill that keeps 5.92:1 against the band. The label itself is fine (gray-20 on gray-80 is 8.75:1), so this is about the tag no longer reading as a tag rather than about legibility.

If the band is meant to carry the signal by itself then leaving it alone is a reasonable call. If not, `.tagOverride` in the same package is where to lift it.

packages/esm-patient-banner-app/src/banner/patient-banner.scss
```diff
@@ -8,11 +9,69 @@
...
+      [class*='cds--tag'] *,
+      [class*='cds--overflow-menu-options'],
+      [class*='cds--overflow-menu-options'] *,
+      [class*='cds--popover-content'],
+      [class*='cds--popover-content'] *
+    ) {
+    color: colors.$gray-10;
+  }
+
+  svg:not([class*='cds--tag'] *, [class*='cds--overflow-menu-options'] *, [class*='cds--popover-content'] *) {
```
dkayiwa: This rule also catches Carbon's loading spinner, which renders inside this banner. `PatientBannerContactDetails` puts an `InlineLoading` in each of its four sections while they fetch, and Carbon keeps that spinner hollow with `.cds--loading__svg { fill: transparent }`, which is a single class of specificity. `.deceasedPatientContainer svg:not(...)` outranks it, so both `<circle>` elements inherit `gray-10` and the spinner draws as a solid light disc with a blue arc round the edge instead of a ring. Measured against the compiled @carbon/styles 1.91.0 CSS plus this file's rules: `getComputedStyle` on `.cds--loading__stroke` gives `fill: rgb(244, 244, 244)` inside the container and `rgba(0, 0, 0, 0)` outside it.

Excluding the loading svg fixes it. I checked afterwards that the three icons this rule exists to recolour (gender, the three-dots trigger, the chevron) all still come out `rgb(244, 244, 244)`, and the "Loading ..." label stays light:

```suggestion
  svg:not(
    [class*='cds--tag'] *,
    [class*='cds--overflow-menu-options'] *,
    [class*='cds--popover-content'] *,
    [class*='cds--loading']
  ) {
```

Cosmetic, and only visible for as long as those fetches take, so take it or leave it.

packages/esm-patient-banner-app/src/banner/patient-banner.scss
dkayiwa: Still open in 772146ca: the SCSS is byte-identical to aa84afbc, so the trigger has no rule of its own yet. I re-measured against the compiled @carbon/styles 1.91.0 CSS that esm-styleguide resolves, with transitions disabled so the read isn't caught mid-animation. Focusing the trigger computes `outline: solid 2px rgb(15, 98, 254)` at `outline-offset: -2px`, and screenshotting the focused button gives 2200 of the 2384 pixels in that outer 2px band at `rgb(15, 98, 254)`. The white inset box-shadow from `--cds-button-focus-color` is there underneath, it just gets painted over. On this band that ring is 1.56:1 against the 3:1 WCAG 1.4.11 asks for, so a keyboard user tabbing to Actions on a deceased patient's chart gets an indicator they effectively cannot see, and since nothing throws and no test looks at focus it ships quietly. This is the piece I'd want in before it merges.

For contrast, the same measurement on Show more/less comes back 2656 of 2656 pixels at `rgb(255, 255, 255)`, which is the declaration doing its job on the button it can reach.

The comment on lines 23 to 25 also claims the declaration covers "the Actions trigger and the Show more/less toggle", so it wants correcting alongside the rule:

```scss
  // Carbon's ghost-button focus ring defaults to blue ($focus, #0f62fe), which is only 1.56:1
  // against this band, below the 3:1 WCAG 1.4.11 needs for non-text contrast. Recolor it light so
  // keyboard focus is visible on the Show more/less toggle.
  --cds-button-focus-color: #{colors.$white};

  // The Actions trigger needs its own rule: Carbon draws `.cds--overflow-menu__trigger:focus` with
  // an `outline`, which paints over the inset box-shadow the property above recolours. Scoped to
  // the trigger so the menu popover's items keep their blue ring on their own light surface.
  [class*='cds--overflow-menu__trigger']:focus {
    outline-color: colors.$white;
  }
```

With that in, the trigger's ring measures `rgb(255, 255, 255)`, 7.81:1, and the menu items are untouched: `.cds--overflow-menu-options__btn:focus` draws its outline from `--cds-focus` directly, so they keep the blue ring visible on their light popover in your third screenshot.

e2e/specs/mark-patient-deceased.spec.ts
dkayiwa: Worth adding a companion assertion in this step, because the tag half of the recolour still has nothing watching it. Against the head commit I dropped just the two `[class*='cds--tag']` exclusions from the `:is()` list, which is the regression this step would need to catch, and re-ran it: the step still returns 7.10 and passes, while the OpenMRS ID pill falls to 1.20:1 (`rgb(244, 244, 244)` on its own `rgb(224, 224, 224)` tag) and the Active Visit tag to 1.19:1 on `rgb(208, 226, 255)`.

Reading the identifier pill's `.cds--tag__label` colour and comparing it against the pill's own background rather than the band would close that off. Today the pair is `rgb(22, 22, 22)` on `rgb(224, 224, 224)`, 13.71:1. Not something I'd hold the merge for if you'd rather do it separately.

VeronicaMuthee:
Re: the description — trimmed the intro and the sticky/unit-test bullets so it matches what ships (deceased differentiation only, e2e-validated). The always-visible header is in the Other section as a follow-up. Thanks @dkayiwa.

dkayiwa:
@VeronicaMuthee you have started coders habbits of working late in the night. 😊

packages/esm-patient-banner-app/src/banner-tags/deceased-patient-tag.scss
```diff
@@ -9,7 +9,10 @@
...
 
+// The deceased tag only renders on the deceased patient banner, which is now a dark grey
+// ($gray-70) band. A dark chip ($gray-80) nearly disappears against it (1.48:1 edge), so use a
+// light pill with dark text — consistent with the identifier pill and reading clearly as a chip.
 .tagOverride {
-  background-color: colors.$gray-80;
-  color: colors.$gray-20;
+  background-color: colors.$gray-20;
+  color: colors.$gray-100;
 }
```
dkayiwa: This wants scoping to the chart's band before it goes in. `.tagOverride` sits on the tag itself, so it travels with the extension to every banner that renders it, not only the one this PR darkens. `deceased-patient-tag` is registered against `patient-banner-tags-slot`, and `PatientBannerPatientInfo` in esm-styleguide renders that slot in four places over in patient-management: the search results page, the compact search dropdown, the appointments patient header, and the create-queue-entry workspace. All four are light surfaces.

I measured the pill on those surfaces against the compiled Carbon CSS plus this file. On the white search row it goes from 11.55:1 today (`#393939`) to 1.32:1 (`#e0e0e0`), and on the `$ui-01` ones from 10.5:1 to 1.2:1. The label stays perfectly legible either way (13.71:1 on its own pill), so nothing becomes unreadable, the chip just stops reading as a chip. That bites hardest in the search results, where patient-management gives deceased patients no band of their own (`.container` and `.deceasedPatientContainer` are both `$ui-02` on its main), so this tag is the only cue a clinician gets while picking a patient out of a list. Nothing throws and no test looks at it, so it lands on users.

My pointer at `.tagOverride` last round was too broad. Leaving the tag dark here and lifting it in `patient-banner.scss` gives the identical result on the chart while leaving the other four screens where they are. With the scoped rule in place I measure the pill at `#e0e0e0` with a `#161616` label, 13.71:1 on the pill and 5.92:1 against the band, the same numbers you have now, and the white search row back at 11.55:1.

```suggestion
.tagOverride {
  background-color: colors.$gray-80;
  color: colors.$gray-20;
}
```

and inside `.deceasedPatientContainer` in `patient-banner.scss`:

```scss
// The Deceased tag stays $gray-80 so it reads as a firm chip on the light banners this same
// extension renders in (patient search, appointments, queue entry). That chip is only 1.48:1
// against this band, so lift it to a light pill here, and only here.
[data-extension-id='deceased-patient-tag'] [class*='cds--tag'] {
  background-color: colors.$gray-20;
  color: colors.$gray-100;
}
```

e2e/specs/mark-patient-deceased.spec.ts
```diff
@@ -49,4 +49,62 @@ test('Mark a patient as deceased', async ({ page, patient }) => {
...
+
+      return {
+        valueContrast: contrast(band, getComputedStyle(value).color),
+        minTagContrast: Math.min(...tagContrasts),
+      };
+    });
+
+    // Contact-details values read against the dark band, and tags stay readable on their own pills.
+    expect(valueContrast).toBeGreaterThanOrEqual(4.5);
+    expect(minTagContrast).toBeGreaterThanOrEqual(4.5);
```
dkayiwa, line 112: Whatever selector the tag fix lands on, this step won't tell you whether it took. I put `deceased-patient-tag.scss` back to its base colours and re-ran the step against the head: it still returns 8.75 and passes, while the pill sits at 1.48:1 against the band, which is the thing that change exists to fix. Comparing the pill against the band as well would close that off.

It touches the destructure too, so as three edits rather than a suggestion block:

```ts
// line 63
const { valueContrast, minTagContrast, minTagVsBand } = await banner.evaluate((el) => {

// in the returned object, alongside minTagContrast
minTagVsBand: Math.min(...tags.map((tag) => contrast(getComputedStyle(tag).backgroundColor, band))),

// alongside the other expects
expect(minTagVsBand).toBeGreaterThanOrEqual(3);
```

Not something I'd hold the merge for by itself.

VeronicaMuthee:
@dkayiwa @denniskigen, is there further feedback pending on this PR?

e2e/specs/mark-patient-deceased.spec.ts
```diff
@@ -49,4 +49,67 @@ test('Mark a patient as deceased', async ({ page, patient }) => {
...
+        // scoped Deceased-tag lift exists to guarantee (a dark chip would sit ~1.48:1 here).
+        minTagVsBand: Math.min(...tags.map((tag) => contrast(getComputedStyle(tag).backgroundColor, band))),
+      };
+    });
+
+    // Contact-details values read against the dark band, tags stay readable on their own pills, and
+    // tags read as distinct chips against the band (WCAG 1.4.11 non-text contrast).
+    expect(valueContrast).toBeGreaterThanOrEqual(4.5);
+    expect(minTagContrast).toBeGreaterThanOrEqual(4.5);
+    expect(minTagVsBand).toBeGreaterThanOrEqual(3);
```
dkayiwa, line 113: Worth one more assertion in this step, though I wouldn't hold the merge for it: the two focus-ring rules are the part of this change that nothing watches. I reverted each one against the head, measured on the compiled @carbon/styles 1.107.0 this PR resolves, and re-ran this step. Green both times.

The trigger rule is the one I'd want covered. Without `[class*='cds--overflow-menu__trigger']:focus`, the focused Actions trigger computes `outline-color: rgb(15, 98, 254)` and screenshots as 592 of 592 ring pixels blue, with none of the white box-shadow showing through, so the whole indicator sits at 1.56:1 on this band against the 3:1 WCAG 1.4.11 asks for. Its only reason for existing is that Carbon's outline paints over the box-shadow the property above recolours, which the next reader won't rediscover from the code. Without `--cds-button-focus-color` it's milder but still a regression: Show more/less comes back 656 blue pixels to 316 white, an outer blue line with a 1px white one inside it, where the head gives 972 of 972 white. Neither throws and no test looks at focus, so a later cleanup takes them out quietly.

The ring sits in a different property on each button, an outline on the trigger and an inset box-shadow on the toggle, so it's one read each. `toHaveCSS` polls, so Carbon's 110ms outline transition needs no wait.

```suggestion
    expect(minTagVsBand).toBeGreaterThanOrEqual(3);

    // Carbon draws the Actions trigger's focus ring as an outline and the toggle's as an inset
    // box-shadow, so each needs its own read. The default blue (#0f62fe) is 1.56:1 against this
    // band, below the 3:1 WCAG 1.4.11 wants for non-text contrast.
    const actionsTrigger = banner.getByRole('button', { name: /actions/i });
    await actionsTrigger.focus();
    await expect(actionsTrigger).toHaveCSS('outline-color', 'rgb(255, 255, 255)');

    const showLessToggle = banner.getByRole('button', { name: /show less/i });
    await showLessToggle.focus();
    await expect(showLessToggle).toHaveCSS('box-shadow', /^rgb\(255, 255, 255\)/);
```

chibongho:
Sorry, can we hold off on this and discuss in the next UI design review? I have some RDE mockups that also changes the patient chart background color, and might conflict with this.

In O2, we have a yellow bar at the top to indicate a deceased patient. I was thinking maybe we can do something like that in O3 as well?

image

VeronicaMuthee:
> Sorry, can we hold off on this and discuss in the next UI design review? I have some RDE mockups that also changes the patient chart background color, and might conflict with this.
> 
> In O2, we have a yellow bar at the top to indicate a deceased patient. I was thinking maybe we can do something like that in O3 as well?
> 
> image

Okay, sounds good to me.

denniskigen:
Did we make any traction on this, @chibongho?

chibongho:
> Did we make any traction on this, @chibongho?

We haven't. RDE probably will need at least 1 or 2 more UX reviews. Let's get this in and not hold it up for now.

VeronicaMuthee:
> > Did we make any traction on this, @chibongho?
> 
> We haven't. RDE probably will need at least 1 or 2 more UX reviews. Let's get this in and not hold it up for now.

Thanks, @chibongho @denniskigen. Dennis, do you have any outstanding comments?

denniskigen (approved):
LGTM. Thanks, @VeronicaMuthee!

denniskigen:
@VeronicaMuthee I took the liberty of pushing a fixup commit 07b29834c to your branch that:

- Moves the contrast parsing and calculation into shared E2E helpers
- Adds a stable test hook for the contact-details panel
- Replaces the manual DOM discovery with visible Playwright locators

The product behavior is unchanged. The test still checks every visible banner tag, but it now snapshots the tag count and rejects transparent computed colors instead of treating them as black. I also trimmed comments that were restating the code.

PR #3524, opened 2026-07-29
(feat) O3 5840 : Add Smart Notifications  for (Labs) 

tendomart:
## Requirements

- [ ] My work is based on designs, which are linked or shown either in the Jira ticket  https://openmrs.atlassian.net/browse/O3-5840

## Summary

A bell icon appears in the top-nav header (next to search/account). It shows an unread count for the currently-open patient's lab results that need attention.

Clicking it opens a slide-over inbox listing those results (priority tag, test — value, reference range, timestamp). Routine in-range results do not appear here.

Clicking a row opens a detail modal: test, result/flag, ordering clinician, location, order number, with "Mark as reviewed" (records provider + timestamp, removes it from the list) and "View in chart" (navigates to Results without clearing it).

The test-order form gains a "Notify me when this result is ready" checkbox (with an explanatory callout) so a clinician can opt in for an otherwise-routine test.

Updates appear without a manual refresh (polling), and are scoped to the location where the order was placed.

## Screenshots

Screenshot From 2026-07-30 00-40-26
Screenshot From 2026-07-30 00-39-44
Screenshot From 2026-07-30 00-39-27
Screenshot From 2026-07-30 00-39-04

https://github.com/user-attachments/assets/33ceba2a-6bcd-4da4-a142-eb5e72ee0f99

## Related Issue

[](https://openmrs.atlassian.net/browse/O3-5840)


tendomart:
@VeronicaMuthee  please check this out

VeronicaMuthee:
@tendomart thanks for the progress on this work - something worth a look:

The "Notify me when resulted" opt-in may not persist. setOptIn() only runs inside prepTestOrderPostData, but the basket's Sign and close may resolve a different prep (the orders-app's generic panel registers one too), so it can be skipped and the opt-in silently dropped. Could we call setOptIn from the form's save-to-basket handler instead — where the flag, patient, and concept are already in scope — and keep the prep a pure mapper? Quick check: order a routine test with the box ticked, result it, confirm the notification shows.

packages/esm-patient-tests-app/src/test-orders/api.ts
```diff
@@ -94,12 +96,17 @@ export function useOrderReasons(conceptUuids: Array<string>) {
...
+  order: SmartTestOrderBasketItem,
   patientUuid,
   encounterUuid,
   orderingProviderUuid,
 ): TestOrderPost => {
   if (order.action === 'NEW' || order.action === 'RENEW') {
+    // The "Notify me when resulted" opt-in can't ride along in the POST — core has no order-level
+    // notify field — so it is recorded client-side here, at the one point that sees the final
+    // submitted order. Keyed by patient + concept; see opt-in-store for the consequences.
+    setOptIn(patientUuid, order.testType?.conceptUuid, Boolean(order.notifyWhenResulted));
```
dkayiwa: An unticked order cancels an earlier opt-in here, and I think that needs fixing before this merges. Because this runs on every NEW/RENEW submit with `Boolean(order.notifyWhenResulted)`, the sequence is: Dr. A orders a CD4 with "Notify me when resulted" on, so `optIns[patient:cd4]` is recorded. Later, anyone on that browser orders another CD4 for the same patient and leaves the toggle off (its default). This call then executes `setOptIn(..., false)` and deletes the record, so when the first order's result comes back in range, `classifyNotification` sees `optedIn: false` and files it silently. The notification Dr. A explicitly asked for never fires, and nothing throws. I verified this at runtime: two sequential `prepTestOrderPostData` calls (ticked, then unticked) leave `isOptedIn()` false.

The opt-in store's own doc says ordering the same test twice should err toward notifying, so I'd make this write assert-only:

```suggestion
    if (order.notifyWhenResulted) {
      setOptIn(patientUuid, order.testType?.conceptUuid, true);
    }
```

A small test pinning that (two prep calls, opt-in survives) would keep it from regressing.

This composes with @VeronicaMuthee's suggestion to move the write out of the prep. For what it's worth, I traced the registration path: the skip she describes needs a non-default `orderTypes` config, since the generic panel only registers preps for configured order types, and the orders dashboard's details table registers this same function. The cancellation above happens in the default config. Wherever the write ends up living, it should be assert-only.

packages/esm-patient-tests-app/src/smart-notifications/smart-notifications.resource.ts
```diff
@@ -0,0 +1,257 @@
...
+  const { labOrderTypeUuid } = config.orders;
+  const session = useSession();
+  const reviewed = useReviewedNotifications();
+  const read = useReadNotifications();
+  const optIns = useOptIns();
+
+  const shouldFetch = Boolean(enabled && patientUuid);
+
+  const ordersUrl = shouldFetch
+    ? `${restBaseUrl}/order?patient=${patientUuid}&careSetting=${careSettingUuid}&status=any&orderType=${labOrderTypeUuid}&v=full`
```
dkayiwa: Nothing on this path bounds how old an order can be and still alert, and with `smartNotifications.enabled` defaulting to true I'd want that addressed before merge. `status=any` returns the newest page of the patient's lab orders (the REST default is typically 50, newest first, which for a chronic patient can span years). Every DECLINED order in that page raises SAMPLE_REJECTED with no result needed, and every resulted STAT order whose observation is still among the newest 100 raises STAT, however long ago it was dealt with. Since review state is per user per browser in localStorage, the same backlog reappears for every clinician on every workstation: opening a returning patient's chart shows a bell full of old alerts each person has to clear by hand, and the badge starts at 9+ on day one. For a feature whose promise is "only what needs your attention", teaching users to ignore the badge is the failure mode, and it lands hardest on exactly the sites where this matters (the ones with lab integrations that set fulfiller statuses).

A cheap fix that keeps the design intact: drop orders older than a configurable horizon before classification, next to the reviewed/location filters in the useMemo below (say `notificationWindowDays`, defaulting to a week). If that feels out of scope here, I'd flip `enabled` to default false until it exists, so distros opt in rather than inherit the backlog.

packages/esm-patient-tests-app/src/smart-notifications/constants.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Read state is separate from review state, and per-user for the same reason. Opening a
+ * notification marks it read — it stops counting against the bell but stays in the inbox until it
+ * is actually reviewed.
+ */
+export function readStorageKey(userUuid: string): string {
+  return `openmrs:smart-notifications:read:${userUuid}`;
+}
+
+/** Opt-in is keyed by patient + concept; there is no order uuid at the time the checkbox is set. */
+export const optInStorageKey = 'openmrs:smart-notifications:opt-in';
```
dkayiwa: Was it deliberate that opt-in state is browser-global while read and review state are per-user? On a shared workstation, which is the common OpenMRS setup, "Notify me when resulted" effectively means "notify whoever uses this browser": Dr. B inherits Dr. A's routine notifications for this patient, and either can cancel the other's (see my comment on api.ts). Keying this store by user uuid like the other two would make the label honest; the order form has the session in scope to do it. If shared opt-ins are actually intended for team-based care, a sentence here saying so would stop the next reader from "fixing" it.

package.json
```diff
@@ -16,7 +16,9 @@
     "test": "turbo run test --color",
     "coverage": "turbo run coverage --color",
     "postinstall": "husky install",
-    "test-e2e": "playwright test"
+    "test-e2e": "playwright test",
+    "harness": "node ./node_modules/vite/bin/vite.js --config e2e/harness/vite.config.ts",
```
dkayiwa: Three tooling points on the harness, none touching the feature code. First, `vite` isn't declared in package.json (and yarn.lock doesn't change in this PR), so this script works only because vitest pulls vite in transitively and the linker hoists it; a vitest bump that changes that breaks `yarn harness` with nothing in the lockfile to explain why. It should be a devDependency. Second, the spec rewrites `e2e/harness/screenshots/*.png` (about 1.3 MB of committed binaries) on every run, so anyone running it locally gets a dirty tree and binary churn in the next commit. I'd gitignore that directory like `e2e/harness/test-results/` already is. Third, really a maintainer call: CI never runs `test-e2e:screenshots`, so the harness spec gates nothing and its framework mock will drift as esm-framework evolves. If the harness earns a place in-tree it probably deserves a CI job (it needs no backend); otherwise it might be better as its own PR so this one stays reviewable as just the feature.

packages/esm-patient-tests-app/src/smart-notifications/reviewed-banner.component.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/**
+ * Confirms, on the Results dashboard, that a notification was signed off — the other half of the
+ * round trip that starts with "Mark as reviewed" or "View in chart" in the notification inbox.
+ */
+const ReviewedBanner: React.FC<ReviewedBannerProps> = ({ patientUuid }) => {
+  const { t } = useTranslation();
+  const { smartNotifications } = useConfig<ConfigObject>();
+  const review = useLatestReviewForPatient(patientUuid);
```
dkayiwa, line 21: This shows the patient's most recent review indefinitely, so weeks later the Results dashboard still opens with a green "Result reviewed, Reviewed by Dr. A" strip even when newer unreviewed criticals have arrived since. It reads as a statement about the current results rather than about one past notification. Bounding it to a short window after the review (a day?), or naming the test it refers to, would keep it truthful. Not blocking, since it only ever shows on the reviewing user's own browser.

packages/esm-patient-tests-app/src/smart-notifications/use-chart-patient.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Reads the patient whose chart is currently open.
+ *
+ * `usePatientChartStore` can't be used here: it requires the caller to already know the patient
+ * uuid and returns an empty store otherwise. The bell lives in the top nav, outside the chart's
+ * React tree, so it has to discover the open patient rather than be handed one. The chart app
+ * populates this store on mount and clears it on unmount
+ * (see `usePatientChartPatientAndVisit` in esm-patient-chart-app), so an empty store means
+ * "no chart open".
+ */
+const patientChartStore = getGlobalStore<PatientChartStore>('patient-chart-global-store', {
```
dkayiwa, line 14: This duplicates the store name and default shape that esm-patient-common-lib owns privately (`patientChartStoreName` isn't exported). If the chart app renames the store or reworks its shape, this file still compiles and the bell just silently never renders again. Fine to leave for this PR, but a read-only accessor exported from common-lib (the uuid-gated `usePatientChartStore` deliberately doesn't fit this use, as your comment notes) would let the compiler own the contract.

bmamlin:
UI should not render encoded HTML entities in encoded state. Currently shows units as "U&#×2F;L" instead of "U/L".

bmamlin:
If all results (regardless of criticality) are sent to orderer for STAT orders, then I'd expect the notification toggle to be on and disabled when urgency is set to STAT.

image

Assuming the user is automatically notified for critical results, the blue info panel could be removed (becomes unnecessary) if the label for the toggle was changed to "Also notify me for non-critical result"

bmamlin:
I'm not a fan of the toast:

image

Since it's location is distant from the user's focus, the user already gets visual feedback of their action from the toggle button, and the toast adds no information (just repeats what is already in the info panel content):

image

Showing uninformative toasts like this will train users to ignore them.

packages/esm-patient-tests-app/src/smart-notifications/constants.ts
tendomart: I think it should be per specific user not for everyone that uses the workstation, that would send unncessary out of context traffic to Dr B, when it was actually intended for Dr.A  not so cc @VeronicaMuthee  ?

do we need to include team-based care in this use case cc @VeronicaMuthee

tendomart: Good observation. It should be that Opt-in is tied to the particular user who can also Read and Review.  If Dr. A opted in to receive notifications, only he can read and review notifications. He opted in for.

packages/esm-patient-tests-app/src/routes.json
```diff
@@ -35,9 +35,19 @@
       "component": "labOrderPanel",
       "slots": ["order-basket-slot", "result-order-basket-slot"],
       "order": 2
+    },
+    {
+      "name": "smart-notification-bell",
+      "component": "notificationBell",
+      "slot": "top-nav-actions-slot",
```
denniskigen:
```suggestion
      "slot": "notifications-menu-button-slot",
```

The primary navigation app already owns this panel. It gives us [`notifications-menu-button-slot`](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/apps/esm-primary-navigation-app/src/components/navbar/navbar.component.tsx#L72-L78) for the bell and [`notifications-nav-menu-slot`](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/apps/esm-primary-navigation-app/src/components/navbar-header-panels/notifications-menu-panel.component.tsx#L15-L19) for the content, and the panel it renders is already a `HeaderPanel` with its own `Notifications` heading. Both slots are empty today.

So let's put the bell here and the rows in `notifications-nav-menu-slot`, and drop our own `HeaderPanel`, panel heading and `showPanel` state. `app-menu-button.component.tsx` is the pattern to follow - the slot hands you `isActivePanel`, `togglePanel` and `hidePanel`, which is also how the panel becomes mutually exclusive with the other header panels. Keep your close button and Escape handling for now, they aren't in the shell's panel yet.

denniskigen (commented):
Thanks for pushing this forward, @tendomart. I've left a few suggestions. The big one is that the app shell already owns the notifications panel, so let's plug into it rather than build a second one here. Also, let's take the screenshot harness out of this PR.

packages/esm-patient-tests-app/src/smart-notifications/patient-display.ts
```diff
@@ -0,0 +1,17 @@
...
+ * preferred identifier with `use: 'usual'`.
+ */
+export function getPreferredIdentifier(
+  patient: fhir.Patient | undefined | null,
+): { label: string; value: string } | undefined {
+  const identifiers = patient?.identifier;
+  if (!identifiers?.length) {
+    return undefined;
+  }
+  const identifier = identifiers.find((id) => id.use === 'usual') ?? identifiers[0];
```
denniskigen, line 12:
```suggestion
/**
 * The patient's preferred identifier, falling back to the first one present. FHIR marks the
 * preferred identifier with `use: 'official'`; non-preferred ones are `usual`.
 */
export function getPreferredIdentifier(
  patient: fhir.Patient | undefined | null,
): { label: string; value: string } | undefined {
  const identifiers = patient?.identifier;
  if (!identifiers?.length) {
    return undefined;
  }
  const identifier = identifiers.find((id) => id.use === 'official') ?? identifiers[0];
```

fhir2 marks the preferred identifier `official`, not `usual` - see [the translator](https://github.com/openmrs/openmrs-module-fhir2/blob/master/api/src/main/java/org/openmrs/module/fhir2/api/translators/impl/PatientIdentifierTranslatorImpl.java#L65-L71) and the reverse mapping a few lines down. So for any patient with more than one identifier we're currently picking a non-preferred one over the OpenMRS ID. Confirmed against a running backend, where the OpenMRS ID comes back as `"use": "official"`.

packages/esm-patient-tests-app/src/test-orders/api.ts
denniskigen: Agreed. Another reason to move it out of the prep: `PostDataPrepFunction` only gets `orderingProviderUuid`, and it isn't a component, so it can't reach `useSession` for the per-user key you agreed to in the other thread.

`saveLabOrderToBasket` can, so let's drop lines 105-109 and do it there:

```tsx
if (data.notifyWhenResulted) {
  setOptIn(session?.user?.uuid, patient.id, data.testType?.conceptUuid);
}
```

Guarding on the flag instead of passing `Boolean(...)` is what keeps a later unticked order from clearing an earlier opt-in.

Separately, the toggle renders whenever `smartNotifications.enabled` is set, but `REVISE` submits come through this same prep and the `NEW || RENEW` guard skips them. So on an order edit the toggle does nothing. Let's either hide it for `REVISE` or handle that path too.

packages/esm-patient-tests-app/src/smart-notifications/smart-notifications.resource.ts
denniskigen: Agreed. While we're on this URL, `v=full` is heavier than we need - about 3 KB per order versus under 800 bytes for just the fields we read, and this pages 50 by default at a 30s poll.

```suggestion
    ? `${restBaseUrl}/order?patient=${patientUuid}&careSetting=${careSettingUuid}&status=any&orderType=${labOrderTypeUuid}&v=custom:(uuid,orderNumber,dateActivated,urgency,fulfillerStatus,fulfillerComment,display,patient:(uuid),concept:(uuid,display),orderer:(display,person:(display)),encounter:(uuid,location:(uuid,display)))`
```

package.json
denniskigen: Agreed on all three. Let's do your last suggestion and take the harness out of this PR, @tendomart - that's `e2e/harness/**`, `playwright.harness.config.ts`, the `harness` and `test-e2e:screenshots` scripts, and the `e2e/harness/test-results/` gitignore entry. It's 18 files and about 1,560 lines, roughly a third of the diff, and since nothing runs `test-e2e:screenshots` in CI it isn't gating anything while the framework mock drifts. The vitest suites should all stay, they're in good shape.

packages/esm-patient-tests-app/src/smart-notifications/smart-notifications.resource.ts
denniskigen: Ran this locally, and the custom representation turns out to be a correctness fix too.

`v=full` returns `encounter` as a bare ref, so there's no location on it:

```
encounter keys: ['display', 'links', 'uuid']
```

`locationUuid` is therefore always undefined, which makes `!notification.locationUuid` in the filter below always true. So `locationScoped` never filters anything, even though it defaults to `true` and the PR description says notifications are scoped to where the order was placed. Same reason the detail dialog shows "Location --".

Session at Mobile Clinic, both orders at Ubuntu Hospital, both notifications still showed. The custom representation above returns `Ubuntu Hospital` for the same order, so it fixes the filter and trims the payload. Could we add a test for the mismatch case?

denniskigen:
@bmamlin, @VeronicaMuthee, a scoping question that I think decides a few of the smaller ones here.

The bell only renders when a chart is open, and it only ever lists that patient's results, but it sits in the top nav where everything else is user-scoped. So it can't answer "do I have results waiting?" - you have to already be in the chart of the patient whose results you want to know about. I pushed it further that way by suggesting the shell's `notifications-menu-button-slot`, which is the slot meant for global notifications, so it's worth settling the intent before we build on it.

Is this meant to be "results I'm waiting on" across patients, or "this patient's results"? If the former, the patient scoping is the temporary part and the bell belongs in the top nav. If the latter, it probably belongs in the chart rather than the global header.

It also settles whether each row needs the patient's name and avatar. Running this locally with 8 notifications, the patient's name is on screen 9 times at once, once in the banner and once per row, and the avatar is the widest element in each row.

CleanShot 2026-08-04 at 00 35 00@2x

packages/esm-patient-tests-app/src/smart-notifications/constants.ts
VeronicaMuthee: @tendomart this is a great question; the opt-in is a personal preference. But with critical values, STAT, and rejected tests, I would not consider those as personal preferences but as patient safety signals and should be visible to whoever is in the patient chart, regardless of who ordered the test.

denniskigen:
> UI should not render encoded HTML entities in encoded state. Currently shows units as "U&#x2F;L" instead of "U/L".

This comes from i18next's default escaping - the app shell inits i18next without `interpolation.escapeValue`, so a plain `{{value}}` HTML-escapes the slashes in units. The fix is the unescaped form:

```
t('testAndValue', '{{test}} — {{- value}}', ...)
```

Same for `{{- range}}`. Two gotchas: the shared react-i18next mock only substitutes `{{key}}`, so it needs to learn the `{{- key}}` form too, and i18next-parser keeps existing en.json values, so after changing the inline default you need to delete the two keys from en.json and re-run extract-translations. en.json wins over the inline default at runtime, so without that step the fix silently doesn't ship.

packages/esm-patient-tests-app/src/smart-notifications/notification-panel.component.tsx
```diff
@@ -0,0 +1,192 @@
...
+            <ul className={styles.list}>
+              {notifications.map((notification) => (
+                <li key={notification.id}>
+                  <button
+                    className={classNames(styles.row, { [styles.rowUnread]: !read[notification.id] })}
+                    onClick={() => handleOpenDetail(notification)}
+                    type="button"
+                  >
+                    <span className={styles.avatar}>
+                      <PatientPhoto patientName={patientName} patientUuid={notification.patientUuid} />
```
denniskigen: `PatientPhoto` gets `notification.patientUuid` but `patientName` is computed once from the chart patient, so if this list ever shows a second patient, a row would pair one patient's photo with another's name. Probably better to read both from the notification, or both from the chart patient.

packages/esm-patient-tests-app/src/smart-notifications/notification-tag.component.tsx
```diff
@@ -0,0 +1,47 @@
...
+  }
+}
+
+const PriorityTag: React.FC<PriorityTagProps> = ({ tag }) => {
+  const label = usePriorityTagLabel(tag);
+
+  return (
+    <Tag className={styles.priorityTag} size="sm" type={tagTypes[tag]}>
+      {label}
+    </Tag>
```
denniskigen: Carbon's `Tag` wraps its label in a `DefinitionTooltip` (a button) whenever it measures the label as truncated, and inside these row buttons that nests interactive elements. With 8 notifications I counted 16 tab stops, each row announced its tag twice, and "Stat" rendered uppercase while "Sample rejected" didn't, since Carbon resets `text-transform` only on the wrapped variant. The label here is static and always visible, so a plain span with the same Carbon classes avoids all of this and renders identically:

```suggestion
    <span
      className={classNames('cds--tag', 'cds--tag--sm', 'cds--layout--size-sm', `cds--tag--${tagTypes[tag]}`, styles.priorityTag)}
    >
      {label}
    </span>
```

With `import classNames from 'classnames';` replacing the `Tag` import up top.

VeronicaMuthee:
> @bmamlin, @VeronicaMuthee, a scoping question that I think decides a few of the smaller ones here.
> 
> The bell only renders when a chart is open, and it only ever lists that patient's results, but it sits in the top nav where everything else is user-scoped. So it can't answer "do I have results waiting?" - you have to already be in the chart of the patient whose results you want to know about. I pushed it further that way by suggesting the shell's `notifications-menu-button-slot`, which is the slot meant for global notifications, so it's worth settling the intent before we build on it.
> 
> Is this meant to be "results I'm waiting on" across patients, or "this patient's results"? If the former, the patient scoping is the temporary part and the bell belongs in the top nav. If the latter, it probably belongs in the chart rather than the global header.
> 
> It also settles whether each row needs the patient's name and avatar. Running this locally with 8 notifications, the patient's name is on screen 9 times at once, once in the banner and once per row, and the avatar is the widest element in each row.
> 
> CleanShot 2026-08-04 at 00 35 00@2x

@denniskigen a patient-scoped-only bell just makes the clinician hunt chart by chart, which is the opposite of what a notification surface is for. Can we aim for notifications acting as a provider inbox and therefore global in nature? Are there constraints to achieving this?

tendomart:
> a patient-scoped-only bell just makes the clinician hunt chart by chart, which is the opposite of what a notification surface is for. Can we aim for notifications acting as a provider inbox and therefore global in nature? Are there constraints to achieving this?

So @VeronicaMuthee @bmamlin @denniskigen  @dkayiwa  shall then say notifications for all STAT / CRITICAL orders done by the provider should float around in the bell until opened, regardless of the patient he is seeing ?

tendomart:
Also @VeronicaMuthee 
Screenshot From 2026-08-04 14-27-01
Are we to inlude other options here for  CRITICAL and RESULTS_REJECTED , something like  

```
export interface OrderCreate {
  uuid?: string;
  encounter: string;
  orderType: string;
  action: 'NEW' | 'REVISE' | 'DISCONTINUE' | 'RENEW';
  accessionNumber?: string;
  patient: string;
  concept: string;
  careSetting: string;
  previousNumber?: string;
  instructions: string;
  urgency: 'ROUTINE' | 'STAT' | 'CRITICAL' | ' RESULTS_DECLINED' |'ON_SCHEDULED_DATE';
  dateActivated?: Date | string;
  dateStopped?: Date | string;
  // TODO: Fill as required.
}

I have noticed that for any basic routine check, a notification is still fired
```

packages/esm-patient-tests-app/src/smart-notifications/notification-model.ts
ibacher, line 1: Almost all of the concerns in this file (calculating reference ranges and the like) is _massively_ beyond the scope of what a notifications app should be worried about. The notifications app should rely on `obs.interpretation` and flag for the user if there is no interpretation. It should _not_ being trying to normalize bounds into ranges, especially because the logic here doesn't seem to be based on the actual underpinning ranges (for example, something outside of the normal range is not, by itself, critical).

packages/esm-patient-tests-app/src/smart-notifications/constants.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Opt-in is keyed by patient + concept; there is no order uuid at the time the checkbox is set.
+ *
+ * Per-user for the same reason as review and read state. "Notify me when resulted" is a promise to
+ * the clinician who ticked it, and a browser-global store would quietly turn that into "notify
+ * whoever signs in at this workstation next" — the next clinician would inherit the routine
+ * notification and be able to clear it by re-ordering the same test with the toggle off.
+ */
+export function optInStorageKey(userUuid: string): string {
```
ibacher, line 30: Under no circumstances should we be pushing _patient uuids_ (which constitute patient data) into local-storage. User-specific settings like this belong _only_ in the backend.

packages/esm-patient-tests-app/src/smart-notifications/constants.ts
```diff
@@ -0,0 +1,48 @@
...
+  return `openmrs:smart-notifications:opt-in:${userUuid}`;
+}
+
+export function optInKey(patientUuid: string, conceptUuid: string): string {
+  return `${patientUuid}:${conceptUuid}`;
```
ibacher, line 35: Opt-in should be per-order, which cannot be correctly determined by only this data.

packages/esm-patient-tests-app/src/config-schema.ts
```diff
@@ -89,6 +90,36 @@ export const configSchema = {
...
+      _default: false,
+      _description:
+        'Also raise a notification for abnormal-but-not-critical values. Critical values always notify and cannot be turned off.',
+    },
+    locationScoped: {
+      _type: Type.Boolean,
+      _default: true,
+      _description: 'Only surface a notification at the location where the order was placed.',
+    },
+    pollingIntervalMs: {
```
ibacher, line 111: Configurations are not intended to be read solely by developers. This should be expressed in seconds or minutes (or an adjustable language like `1s`), not `ms`. Most non-programmers have no idea what a millisecond is.

packages/esm-patient-tests-app/src/config-schema.ts
```diff
@@ -89,6 +90,36 @@ export const configSchema = {
...
+      _type: Type.Number,
+      _default: 30000,
+      _description: 'How often, in milliseconds, to poll for newly resulted orders.',
+    },
+    notificationWindowDays: {
```
ibacher, line 116: Remember that these configurations are global. This feels more like something that needs to be configured at a test level or something. And what if I want the notification to still be visible _until the next one_?

packages/esm-patient-tests-app/src/config-schema.ts
```diff
@@ -89,6 +90,36 @@ export const configSchema = {
     _description: 'List of various order types, each associated with the Java class name `org.openmrs.TestOrder`.',
     _default: [],
   },
+  smartNotifications: {
+    enabled: {
+      _type: Type.Boolean,
+      _default: true,
```
ibacher, line 96:
```suggestion
      _default: false,
```

ibacher (changes requested):
I have a lot more comments on this than are here, but I'm stopping with just a few because:

1. I don't think the technical architecture here is completely done. There are a lot of concerns here pushed to the frontend (how long are notifications shown by default? Did the user opt into the notification?) that really ought to be backend concerns.
2. As Dennis pointed out, the app shell (or primary nav) owns the notification window. I think this means the frontend part needs to be split up quite a bit into: (a) a straight-forward notifications app and (b) the lab-specific plumbing for the notifications app. As-is I wouldn't be comfortable having this merged because its stuffing a lot of concerns into the test app.
3. Even as a tests app, this app should not be trying to interpret lab results. Again, this is a concern that we handle when the obs is entered. This app's responsibility should be displaying that interpretation correctly.
4. This app pushes patient identifiers (UUIDs) into local storage. This is not safe.
5. Stuff like this _must_ be behind a feature flag before it's mergable.

I'm saying this to say that _this PR_ cannot be merged as written for the above reasons. I really appreciate the work that you've done on this, though, and I'm excited to get such a feature into OpenMRS!

tendomart:
Hi all moved the conversation to https://github.com/openmrs/openmrs-esm-smart-notifications-app/pull/5

cc @VeronicaMuthee  @bmamlin  @denniskigen @dkayiwa @ibacher

PR #3525, opened 2026-07-30
(fix) O3-5841: Sort Vitals and Biometrics tables in the direction the header arrow indicates

sudeshkar:
## Requirements

## Summary

The Vitals and Biometrics tables sorted in the opposite direction to the one indicated by the column header arrow: an ascending (↑) arrow put the newest reading at the top, and a descending (↓) arrow put the oldest reading at the top.

Both `PaginatedVitals` and `PaginatedBiometrics` bypass Carbon's own row sorting and re-sort the full dataset themselves, so that sorting applies across pages rather than just the visible page. There were two bugs in that arrangement.

### 1. The sort direction was inverted

```js
return sortParams.sortDirection === 'DESC' ? sortingNum : -sortingNum;
```

Every `sortFunc` in `vitals-overview.component.tsx` and `biometrics-base.component.tsx` is an ascending comparator (`a - b`, i.e. negative when `a` sorts first). Carbon treats `ASC` the same way — its `defaultSortRow` returns `compare(cellA, cellB)` for `ASC` and `compare(cellB, cellA)` for `DESC` — and renders the header's `ArrowUp` icon rotated 180° only when the state is `DESC`. So `↑` = `ASC` = smallest/oldest first. The comparator therefore maps directly onto `ASC` and needs negating for `DESC`, which is the reverse of what the code did.

This aligns the rendered order with both the arrow and the header's `aria-sort` value, for every sortable column in these two tables rather than just `Date and time`.

### 2. The sort state was read from the wrong place

Both components recorded the sort state from inside `sortRow`, returning `0` so that Carbon would not reorder anything itself. `sortRow` is the wrong hook for that, for three reasons:

- **Carbon does not treat it as a pure comparator.** `getDerivedStateFromProps` calls `getSortedState` — and therefore `sortRow` — whenever `sortHeaderKey` is already set. Setting state from inside it closed a cycle: a new `sortParams` object → a new `sortedData` array → a new `rows` reference → Carbon's `useEffect([headers, rows])` → `getDerivedStateFromProps` → back into `sortRow`. Before the first click `sortHeaderKey` is `null`, so the cycle never started and the table looked fine, which is why this went unnoticed. Measured with a counter on `usePagination`, a single sort click produced 300+ renders, climbing until the heap was exhausted.
- **Carbon skips it when a header transitions to `NONE`.** The `sortDirection === 'NONE'` branch never fired, so the third click on a header left the rows sorted while `aria-sort` had already reset to `none`.
- **Carbon never invokes it when the page holds a single row**, because `Array.prototype.sort` performs no comparisons on a one-element array. On such a page, clicking a header moved `aria-sort` but left the rows untouched.

The sort state is now captured from the header's `onClick` callback, which Carbon calls exactly once per click with the state it is transitioning to:

```tsx
const handleSortHeaderClick = (
  _event: unknown,
  { sortHeaderKey, sortDirection }: { sortHeaderKey: string; sortDirection: DataTableSortState },
) => {
  setSortParams({ key: sortDirection === 'NONE' ? '' : sortHeaderKey, sortDirection });
};
```

`sortRow` is left as a pure `() => 0`. That covers all three sort states, fixes both of the gaps above, and removes the render loop at its source rather than damping it with a previous-state bailout.

Both tables carried their own copy of all of this — the sort state, the click handler and the block that re-sorts the full dataset were identical line for line, so any future change would have had to be made twice. They now share a `useTableSorting` hook in the package's `common` folder, which also exports the inert comparator as `noopSortRow`, and which types its sort direction with Carbon's `DataTableSortState` rather than restating `'ASC' | 'DESC' | 'NONE'` by hand.

## Screenshots

Patient `Grace Garcia`, Vitals & Biometrics dashboard, sorting on `Date and time`.

### Ascending (↑) arrow

**Before** — newest reading first:

before-ascending

**After** — oldest reading first:

after-ascending

### Descending (↓) arrow

**Before** — oldest reading first:

before-descending

**After** — newest reading first:

after-descending

## Related Issue

https://openmrs.atlassian.net/browse/O3-5841

## Other

`biometrics-overview.test.tsx` already declared the expected ascending and descending orders but never asserted them — the `sortRowsButton` and `expectedAscendingOrder` bindings were unused, which is why the inversion went unnoticed. Those assertions are now wired up and check `aria-sort` alongside the row order. An equivalent test was added for the Vitals table, which had no sorting coverage at all.

Each table also gets two further tests, covering the cases where Carbon never calls `sortRow`:

- **A single-row page still sorts the whole dataset.** Vitals renders with `pageSize={1}`; Biometrics paginates six readings five to a page and navigates to the second. In both, clicking `Date and time` swaps the visible row for the oldest reading.
- **The third click restores the unsorted order.** Sorting by `Pulse` / `Weight` through `ASC` → `DESC` → `NONE` and asserting the rows come back to their original order, with `aria-sort` at `none`.

All four fail against the previous `sortRow`-based implementation.

While in `paginated-biometrics.component.tsx`, the column headers now take an explicit `key={header.key}` rather than receiving it through the `getHeaderProps` spread, which React 19 warns about. `PaginatedVitals` already did this; the biometrics line predates the PR.

Two related things I left alone to keep this focused, both worth a follow-up:

- `tableHeaders` is rebuilt as a new array literal on every render in both `biometrics-base.component.tsx` and `vitals-overview.component.tsx`. It is a dependency of the `sortedData` memo and is passed straight to Carbon's `headers` prop, so each parent render currently costs an extra `DataTable` state sync. Memoising it would remove that churn; it is not required for the fix above.
- The same inverted direction mapping exists in `esm-generic-patient-widgets-app`'s `obs-table` and in `esm-patient-conditions-app`'s conditions sorting. In `obs-table` the date comparator is itself written backwards (`rowB - rowA`), so the two inversions cancel out for that one column while the other columns stay wrong.

### DIFF
--- a/packages/esm-patient-vitals-app/src/common/use-table-sorting.ts
+++ b/packages/esm-patient-vitals-app/src/common/use-table-sorting.ts
@@ -1,0 +1,65 @@
+import { useCallback, useMemo, useState } from 'react';
+import { type DataTableSortState } from '@carbon/react';
+
+interface SortableTableHeader<TableRow> {
+  key: string;
+  sortFunc: (rowA: TableRow, rowB: TableRow) => number;
+}
+
+/**
+ * Carbon only ever sorts the rows it was handed, which is a single page. Pass this
+ * as the table's `sortRow` to keep its comparator inert so that the hook can sort
+ * the full dataset instead.
+ */
+export const noopSortRow = () => 0;
+
+/**
+ * Sorts a paginated table's full dataset rather than just its visible page.
+ *
+ * The sort state is read from the header's click handler, which Carbon calls
+ * exactly once per click with the state it is transitioning to. `sortRow` is the
+ * wrong hook for that: Carbon re-runs it while deriving its own state, skips it
+ * when a header transitions to `NONE`, and never invokes it when the page holds a
+ * single row, because `Array.prototype.sort` performs no comparisons on a
+ * single-element array.
+ */
+export function useTableSorting<TableRow>(
+  tableRows: Array<TableRow>,
+  tableHeaders: Array<SortableTableHeader<TableRow>>,
+) {
+  const [sortParams, setSortParams] = useState<{ key: string; sortDirection: DataTableSortState }>({
+    key: '',
+    sortDirection: 'NONE',
+  });
+
+  const handleSortHeaderClick = useCallback(
+    (
+      _event: unknown,
+      { sortHeaderKey, sortDirection }: { sortHeaderKey: string; sortDirection: DataTableSortState },
+    ) => {
+      setSortParams({ key: sortDirection === 'NONE' ? '' : sortHeaderKey, sortDirection });
+    },
+    [],
+  );
+
+  const sortedData: Array<TableRow> = useMemo(() => {
+    if (sortParams.sortDirection === 'NONE') {
+      return tableRows;
+    }
+
+    const header = tableHeaders.find((header) => header.key === sortParams.key);
+
+    if (!header) {
+      return tableRows;
+    }
+
+    // `sortFunc` compares in ascending order, so it maps directly onto the ASC
+    // sort direction and needs inverting for DESC.
+    return tableRows.slice().sort((rowA, rowB) => {
+      const sortingNum = header.sortFunc(rowA, rowB);
+      return sortParams.sortDirection === 'DESC' ? -sortingNum : sortingNum;
+    });
+  }, [tableHeaders, tableRows, sortParams]);
+
+  return { handleSortHeaderClick, sortedData };
+}
--- a/packages/esm-patient-vitals-app/src/vitals/paginated-vitals.component.tsx
+++ b/packages/esm-patient-vitals-app/src/vitals/paginated-vitals.component.tsx
@@ -54,11 +54,19 @@ const PaginatedVitals: React.FC<PaginatedVitalsProps> = ({
     _cellB: any,
     { key, sortDirection }: { key: string; sortDirection: DataTableSortState },
   ) => {
-    if (sortDirection === 'NONE') {
-      setSortParams({ key: '', sortDirection });
-    } else {
-      setSortParams({ key, sortDirection });
-    }
+    // Carbon re-runs this comparator while deriving its own state, so keep the
+    // previous object when nothing changed. Returning a new one every time would
+    // re-render, produce a fresh `sortedData` reference, and re-enter this
+    // callback through Carbon's rows effect, looping indefinitely.
+    setSortParams((prevSortParams) => {
+      const nextKey = sortDirection === 'NONE' ? '' : key;
+
+      if (prevSortParams.key === nextKey && prevSortParams.sortDirection === sortDirection) {
+        return prevSortParams;
+      }
+
+      return { key: nextKey, sortDirection };
+    });
     return 0;
   };
 
@@ -73,9 +81,11 @@ const PaginatedVitals: React.FC<PaginatedVitalsProps> = ({
       return tableRows;
     }
 
+    // `sortFunc` compares in ascending order, so it maps directly onto the ASC
+    // sort direction and needs inverting for DESC.
     const sortedRows = tableRows.slice().sort((rowA, rowB) => {
       const sortingNum = header.sortFunc(rowA, rowB);
-      return sortParams.sortDirection === 'DESC' ? sortingNum : -sortingNum;
+      return sortParams.sortDirection === 'DESC' ? -sortingNum : sortingNum;
     });
 
     return sortedRows;
--- a/packages/esm-patient-vitals-app/src/vitals/paginated-vitals.component.tsx
+++ b/packages/esm-patient-vitals-app/src/vitals/paginated-vitals.component.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from 'react';
+import React, { useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import {
   DataTable,
@@ -12,10 +12,10 @@ import {
   TableHead,
   TableHeader,
   TableRow,
-  type DataTableSortState,
 } from '@carbon/react';
 import { NumericObservation, useLayoutType, usePagination } from '@openmrs/esm-framework';
 import { PatientChartPagination } from '@openmrs/esm-patient-common-lib';
+import { noopSortRow, useTableSorting } from '../common';
 import type { VitalsTableHeader, VitalsTableRow } from './types';
 import { VitalsAndBiometricsActionMenu } from '../components/action-menu/vitals-biometrics-action-menu.component';
 import styles from './paginated-vitals.scss';
@@ -31,8 +31,6 @@ interface PaginatedVitalsProps {
   patient: fhir.Patient;
 }
 
-const noopSortRow = () => 0;
-
 const PaginatedVitals: React.FC<PaginatedVitalsProps> = ({
   isPrinting,
   pageSize,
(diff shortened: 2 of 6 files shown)


packages/esm-patient-vitals-app/src/vitals/paginated-vitals.component.tsx
```diff
@@ -54,11 +54,19 @@ const PaginatedVitals: React.FC<PaginatedVitalsProps> = ({
...
-    if (sortDirection === 'NONE') {
-      setSortParams({ key: '', sortDirection });
-    } else {
-      setSortParams({ key, sortDirection });
-    }
+    // Carbon re-runs this comparator while deriving its own state, so keep the
+    // previous object when nothing changed. Returning a new one every time would
+    // re-render, produce a fresh `sortedData` reference, and re-enter this
+    // callback through Carbon's rows effect, looping indefinitely.
+    setSortParams((prevSortParams) => {
```
denniskigen: Thanks for stabilizing the repeated comparator calls. Could we capture the sort state from `getHeaderProps`'s `onClick` callback instead of from `sortRow`? Carbon skips `sortRow` when transitioning to `NONE` (the `sortDirection === 'NONE'` branch here never fires), and it's also never invoked when the current page contains only one row, because `Array.sort` performs no comparisons. In both cases `sortParams` keeps its previous value even though the header state changes. I reproduced both: on a one-row page, clicking Date changed `aria-sort` to `ascending` but the newest row remained visible. Clicking Weight three times also left the rows sorted by weight while `aria-sort` was `none`.

Carbon passes the next sort state to the header's click handler, so something like this covers every state and avoids the render loop without the bailout:

```tsx
const handleSortHeaderClick = (
  _event: unknown,
  { sortHeaderKey, sortDirection }: { sortHeaderKey: string; sortDirection: DataTableSortState },
) => {
  setSortParams({ key: sortDirection === 'NONE' ? '' : sortHeaderKey, sortDirection });
};
```

with `sortRow` left as a pure `() => 0` and the header rendered via `getHeaderProps({ header, onClick: handleSortHeaderClick })`. Could we add coverage for these two cases? Ditto for `paginated-biometrics.component.tsx`.

sudeshkar: Good catch, thank you — you're right on both counts, and the click handler is the correct place to read this from. Done in cae94982 for both tables.

`sortRow` is now a pure `() => 0`, and the state comes from `getHeaderProps({ header, onClick: handleSortHeaderClick })` exactly as you sketched. That also removes the need for the previous-state bailout: Carbon calls `onClick` once per click rather than repeatedly from `getDerivedStateFromProps`, so there is no cycle left to damp. I've rewritten that section of the PR description accordingly — the three ways `sortRow` was the wrong hook (re-entrancy, the skipped `NONE` transition, and the uninvoked single-row case) now read as one problem rather than two unrelated ones.

Coverage for the two cases you reproduced, in `vitals-overview.test.tsx` and `biometrics-overview.test.tsx`:

- **Single-row page.** Vitals renders with `pageSize={1}`; Biometrics paginates six readings five to a page and navigates to the second, since its `pageSize` is fixed at 5 in `biometrics-overview.component.tsx`. In both, clicking `Date and time` sets `aria-sort` to `ascending` *and* swaps the visible row for the oldest reading.
- **Third click.** Sorting by `Pulse` / `Weight` through `ASC` → `DESC` → `NONE`, asserting the row order at each step and that the rows return to their original order once `aria-sort` is `none`.

I checked all four fail against the previous implementation — the single-row tests kept showing the newest row, and the third-click tests kept the rows sorted under `aria-sort="none"`, matching what you saw.

VeronicaMuthee:
@sudeshkar, thanks for working on this. These are some minor, non-blocking suggestions

cc: @denniskigen 

1. Shared sorting logic (non-blocking, fine as a follow-up)

The sorting logic is now identical, line for line, in both paginated-vitals.component.tsx and paginated-biometrics.component.tsx — the do-nothing noopSortRow function, the handleSortHeaderClick handler, and the block that re-sorts the full dataset. Since any future change would need to be made in both files (and could easily be made in only one), consider extracting this into a single shared hook in the package that both components use. The duplication existed before this PR, so this doesn't need to hold it up.

2. Reuse the imported sort-direction type

The sortParams state declares its direction by writing out the three values 'ASC' | 'DESC' | 'NONE', while the click handler a few lines below already uses Carbon's DataTableSortState type, which stands for exactly those three values. Using the imported type in both places (one line in each file) avoids maintaining a hand-written copy of something the library already defines.

3. Define the test helper once per file

Each new test defines its own identical copy of the getRowDates helper (the small function that reads the dates shown in the table so the test can check their order). Defining it once at the top of the describe block and reusing it in every test would remove roughly 30 repeated lines without changing what the tests verify.

4. Explicit key on the biometrics table headers (pre-existing)

In PaginatedBiometrics, the column headers get their React key only indirectly, bundled inside the props spread from getHeaderProps(...). PaginatedVitals already passes it explicitly (key={header.key}). React 19 warns when a key arrives via a spread, so while you're in this file, adding the explicit key={header.key} to the biometrics header would silence that and make the two tables consistent. This line predates the PR, so it's purely a convenience fix.

sudeshkar:
Thanks for the read-through @VeronicaMuthee — all four are in, and I've rebased on `main` since the branch had drifted a fair way behind.

**1. Shared sorting logic.** You're right that leaving it duplicated was asking for a one-sided fix later, so I did it now rather than as a follow-up. The sort state, the click handler and the full-dataset re-sort moved into a `useTableSorting` hook under the package's `common` folder, with the inert comparator exported next to it as `noopSortRow`. Both components are down to:

```tsx
const { handleSortHeaderClick, sortedData } = useTableSorting<VitalsTableRow>(tableRows, tableHeaders);
```

The hook is generic over the row type and accepts any header carrying a `key` and a `sortFunc`, so `VitalsTableHeader` and `BiometricsTableHeader` both satisfy it without either table's shape leaking in.

**2. Sort-direction type.** Done — `DataTableSortState` in the state declaration instead of the hand-written union. Now that the state lives in the hook there's one declaration covering both tables rather than two copies to keep in step.

**3. Test helper.** `getRowDates` is defined once per file now. The single copy matches the date without the trailing time, which meant dropping the `, 12:00 AM` suffix from the existing `expectedDescendingOrder` / `expectedAscendingOrder` arrays in `biometrics-overview.test.tsx` — every mock reading is at midnight, so nothing is lost from the assertion; it still pins the order.

**4. Explicit key.** Added `key={header.key}` to the biometrics header so it no longer arrives via the spread, which lines it up with `PaginatedVitals`.

Suite is green — 69 tests across the vitals package — and `typescript` passes. I've also updated the description to cover the hook.

sudeshkar:
@denniskigen @VeronicaMuthee — could one of you approve the workflow run here when you get a moment? `OpenMRS CI`, `E2E Tests` and `Report bundle size` are all sitting at `action_required`, so only the title and description checks have reported so far. They've been in that state since the first push rather than since the rebase, so it looks like the fork approval gate rather than anything having failed.

Locally `yarn verify` (lint, typescript and test across all packages) is green, and the vitals package suite passes on its own, but it'd be good to see that confirmed on CI before this goes any further.

packages/esm-patient-vitals-app/src/common/use-table-sorting.ts
```diff
@@ -0,0 +1,65 @@
...
+ * Sorts a paginated table's full dataset rather than just its visible page.
+ *
+ * The sort state is read from the header's click handler, which Carbon calls
+ * exactly once per click with the state it is transitioning to. `sortRow` is the
+ * wrong hook for that: Carbon re-runs it while deriving its own state, skips it
+ * when a header transitions to `NONE`, and never invokes it when the page holds a
+ * single row, because `Array.prototype.sort` performs no comparisons on a
+ * single-element array.
+ */
+export function useTableSorting<TableRow>(
```
denniskigen, line 26: Non-blocking, nothing to change here. Keeping the hook in the vitals package is right for this PR.

For the follow-up, I'd move it into `esm-patient-common-lib` next to [`PatientChartPagination`](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-common-lib/src/pagination/pagination.component.tsx#L18) rather than let it get copied a third time. The other call sites already match your contract exactly: headers of `{ key, sortFunc(rowA, rowB) }`, with `usePagination(sortedData, pageSize)` under a Carbon `DataTable`. Conditions has it as [`useConditionsSorting`](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-conditions-app/src/conditions/conditions.resource.ts#L301), and [`obs-table`](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-generic-patient-widgets-app/src/obs-table/obs-table.component.tsx#L199) has it inline. Both invert the direction the same way you're fixing here, and obs-table has since [grown a `setTimeout` and an `isMountedRef`](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-generic-patient-widgets-app/src/obs-table/obs-table.component.tsx#L170) to defer the `setState` rather than move it off `sortRow`.

common-lib is the natural home: `PatientChartPagination` is what these tables render underneath, and `compare` already lives there.

Could you file that as a ticket? Worth naming the `setTimeout` in it, so whoever picks it up knows there's a workaround to delete.

packages/esm-patient-vitals-app/src/vitals/vitals-overview.test.tsx
```diff
@@ -46,6 +46,13 @@ mockUseConfig.mockReturnValue({
   mockVitalsConfig,
```
denniskigen:
```suggestion
  ...mockVitalsConfig,
```

nit, pre-existing, but the spread operator `...` is missing here, so the mock lands under a `mockVitalsConfig` key instead of being merged in, and these tests have only ever seen the schema defaults. `biometrics-overview.test.tsx` spreads it.

denniskigen (approved):
LGTM, @sudeshkar. Thanks for the thorough write-up, and for wiring up the assertions that were sitting unused in the biometrics test. I've verified the direction fix, the `NONE` transition, and the single-row page against a local backend.

PR #3526, opened 2026-07-30
(feat) O3-5837: Add opt-in toggle for lab result notifications

sudeshkar:
## Requirements

## Summary

Sub-task of the Smart Notification Feature epic ([O3-5751](https://openmrs.atlassian.net/browse/O3-5751)): give the ordering clinician a way to say up front that they want to hear about a lab order once it is resulted.

Today a routine test order is filed silently to the chart. Unless the value happens to come back critical, nothing tells the orderer that it has arrived, so they have to remember to go back and look. This adds the opt-in affordance from the ticket designs to the **Test Order** form, between *Priority* and *Additional instructions*:

- A **`Notify me when resulted`** toggle, off by default. Its state is held on the order basket item as `notifyWhenResulted`.
- An **inline info notification** beneath the toggle explaining what turning it on does, with a muted note that delivery is planned for a future release. Switching the toggle on also raises a confirmation snackbar.
- A **priority summary** at the foot of the form describing how the currently selected priority is notified today (routine / stat / on scheduled date), so it is clear whether opting in adds anything beyond what the priority already does.
- The whole affordance sits behind a new **`showNotifyWhenResultedToggle`** config property (default `true`), following the existing `showReferenceNumberField` pattern, so distributions without the notification backend can hide it.

### A note on scope

The opt-in is deliberately **not** posted to the backend. The `Order` data model has no field for it yet, and `prepTestOrderPostData` whitelists the properties it sends, so the preference stays in the order basket for now. Persisting it and routing the actual notification is backend work under phases 2 and 3 of the parent epic — which is why the inline notification tells the user that delivery is still to come.

Two things I would happily change on review: the confirmation snackbar repeats the inline copy, so if that reads as noise I can drop it and keep the inline notification alone. The priority summary is also strictly an addition beyond the toggle itself; it is in both mockups and it is what makes the opt-in legible, but it can be split out if you would rather keep this sub-task to the toggle.

## Screenshots

The form follows the two mockups attached to [O3-5837](https://openmrs.atlassian.net/browse/O3-5837) — the toggle sits directly below *Priority* with the info notification under it, and the priority summary closes out the form.

## Related Issue

https://openmrs.atlassian.net/browse/O3-5837

## Other

- Four new cases in `add-test-order.test.tsx` cover the default-off state, opting in (both the basket value and the confirmation), the summary switching with the selected priority, and the config gate hiding the field. The full `esm-patient-tests-app` suite passes (216 tests), and repo-wide `typescript` and `lint` are clean.
- `translations/en.json` was regenerated with `yarn extract-translations` rather than hand-edited.
- `config-schema.ts` also carries a two-line prettier reformat of the neighbouring `orderTypeUuid` description — the pre-commit hook applies that automatically to the staged file.

[O3-5751]: https://openmrs.atlassian.net/browse/O3-5751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[O3-5837]: https://openmrs.atlassian.net/browse/O3-5837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sudeshkar:
Closing this — O3-5837 belongs in the dedicated `openmrs-esm-smart-notifications-app` repo rather than here. The opt-in will be delivered as an extension contributed into the existing `top-of-lab-order-form-slot`, which keeps the notification feature decoupled from the patient chart and avoids adding notification-specific config to `esm-patient-tests-app`.

Sorry for the noise.

PR #3529, opened 2026-07-31
(fix) O3-5844: Allow recording vaccinations on a patient's birth date

solomonfortune:
## Requirements

## Summary
Vaccination date validation compared full `Date` timestamps instead of calendar days, so a vaccination recorded on the patient's actual birth date was rejected whenever local midnight preceded the backend's UTC birth date timestamp (e.g. any timezone ahead of UTC). This blocked recording routine birth-dose vaccines like BCG, Hepatitis B, and OPV-0.

Replaced the `.min()` timestamp comparison with a `.refine()` that normalizes both dates to start-of-day before comparing, mirroring the pattern already used for the future-date check.

## Screenshots
Screenshot (534)
Screenshot (535)

## Related Issue
https://openmrs.atlassian.net/browse/O3-5844

## Other

### DIFF
--- a/packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.test.tsx
+++ b/packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.test.tsx
@@ -27,6 +27,7 @@ import { savePatientImmunization } from './immunizations.resource';
 import { FHIR_NEXT_DOSE_DATE_EXTENSION_URL } from './immunization-mapper';
 import { useImmunizations } from '../hooks/useImmunizations';
 import ImmunizationsForm from './immunizations-form.workspace';
+import { fireEvent } from '@testing-library/react';
 
 const mockCloseWorkspace = vi.fn();
 const mockSavePatientImmunization = savePatientImmunization as Mock;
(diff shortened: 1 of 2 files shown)


solomonfortune:
@VeronicaMuthee you can please review this or help me request a review from the maintainer in charge of the repo

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.test.tsx
```diff
@@ -594,6 +595,75 @@ describe('Immunizations Form', () => {
       expect.any(AbortController),
     );
   });
+
+  it("should accept a vaccination date that matches the patient's birth date", async () => {
```
denniskigen: I checked out the branch, restored the old `.min(new Date(patient.birthDate))` implementation, and reran these tests. All three new tests still pass. `tools/vitest.shared.ts` pins `process.env.TZ` to `UTC`, so the birthday input and FHIR birth date resolve to equal timestamps, which Zod's `.min()` accepts. The regression only appears in timezones ahead of UTC.

Could we scope these three boundary tests to a nested `describe` that runs under a timezone such as `Africa/Nairobi` and restores the original timezone afterward (adding `beforeAll` and `afterAll` to the vitest import)? I verified that the birthday test then fails against the old code and passes with this fix, while the other tests remain unaffected:

```ts
describe('Vaccination date boundaries', () => {
  const originalTimezone = process.env.TZ;

  beforeAll(() => {
    // The birth-date boundary bug only reproduces in timezones ahead of UTC
    process.env.TZ = 'Africa/Nairobi';
  });

  afterAll(() => {
    process.env.TZ = originalTimezone;
  });

  // the three new tests go here
});
```

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.test.tsx
```diff
@@ -594,6 +595,75 @@ describe('Immunizations Form', () => {
...
+
+    expect(mockSavePatientImmunization).toHaveBeenCalledTimes(1);
+  });
+
+  it("should reject a vaccination date one day before the patient's birth date", async () => {
+    const user = userEvent.setup();
+    render(<ImmunizationsForm {...testProps} />);
+
+    const vaccinationDateField = screen.getByRole('textbox', { name: /vaccination date/i });
+    fireEvent.change(vaccinationDateField, { target: { value: '03/04/1972' } }); // one day before birth date
```
denniskigen: The `OpenmrsDatePicker` mock parses this with plain `dayjs()`, which reads `03/04/1972` as March 4, 1972. That is a month before the birth date rather than one day before. To exercise the exact boundary the old code mishandled:

```suggestion
    fireEvent.change(vaccinationDateField, { target: { value: '1972-04-03' } }); // one day before birth date
```

An ISO string avoids depending on how the mock resolves ambiguous `DD/MM` versus `MM/DD` input. Probably better to use ISO strings in the other two new tests as well (`'1972-04-04'` and `'2099-01-01'`).

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.workspace.tsx
```diff
@@ -49,9 +49,17 @@ const ImmunizationsForm: React.FC<PatientWorkspace2DefinitionProps<{}, {}>> = ({
...
+          (date) => {
+            // Normalize both dates to start of day in local timezone
+            const inputDate = dayjs(date).startOf('day');
+            const birthDate = dayjs(patient.birthDate).startOf('day');
+            return inputDate.isSame(birthDate) || inputDate.isAfter(birthDate);
+          },
+          {
+            message: t('vaccinationDateCannotBeBeforeBirthDate', 'Vaccination date cannot precede birth date'),
+          },
+        )
```
denniskigen: One edge case this introduces is that `birthDate` is optional in FHIR. When it is absent, `dayjs(undefined)` resolves to today, so every past vaccination date is rejected with a misleading birth-date error. The previous `.min()` received an invalid date in this case and did not reject the vaccination.

A guard preserves the previous behavior, and `isBefore` with day granularity removes the need for explicit `startOf('day')` normalization. This matches the comparison already used in [programs-form.workspace.tsx](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-programs-app/src/programs/programs-form.workspace.tsx#L61):

```suggestion
        .refine((date) => !patient.birthDate || !dayjs(date).isBefore(patient.birthDate, 'day'), {
          message: t('vaccinationDateCannotBeBeforeBirthDate', 'Vaccination date cannot precede birth date'),
        })
```

packages/esm-patient-immunizations-app/src/immunizations/immunizations-form.test.tsx
```diff
@@ -27,6 +27,7 @@ import { savePatientImmunization } from './immunizations.resource';
 import { FHIR_NEXT_DOSE_DATE_EXTENSION_URL } from './immunization-mapper';
 import { useImmunizations } from '../hooks/useImmunizations';
 import ImmunizationsForm from './immunizations-form.workspace';
+import { fireEvent } from '@testing-library/react';
```
denniskigen: ESLint reports this as an `import/no-duplicates` error. Could you merge `fireEvent` into the existing `@testing-library/react` import on line 12?

denniskigen (commented):
Thanks @solomonfortune, this is a well-diagnosed fix, and the PR description explains the timezone mechanism clearly. I found one production edge case for patients without a birth date, and the new tests currently pass against the old implementation, so my main asks are below.

solomonfortune (commented):
Thanks for the review, @denniskigen! I have pushed a follow-up commit addressing all four points. in it, i have guarded the birth-date check for patients with no recorded birth date, scoped the boundary tests to run under Africa/Nairobi, switched to ISO date strings, and merged the duplicate import. Ready for another look.

denniskigen (approved):
LGTM! Thanks, @solomonfortune.

Confirmed the boundary tests now fail against the old `.min()` implementation, so the regression is properly pinned.

PR #3546, opened 2026-08-05
(feat) O3-5850: Support concept reference ranges in the Angular form engine

Nahid-NHB:
## Requirements

## Summary

Concept reference ranges (Platform 2.7+) never reached forms rendered by the Angular form engine (AFE). The vitals widget and the results viewer both use them already, but a numeric question in an AFE form was still bounded only by whatever `questionOptions.min` / `questionOptions.max` the schema author happened to hardcode. So a form would accept a value that the backend then rejects, and none of the age/sex/pregnancy based ranges applied.

The cause is straightforward: nothing in `esm-form-entry-app` ever fetched `/conceptreferencerange`. `QuestionFactory.addValidators` in ngx-formentry builds the min/max validators purely from the schema, and we hand it the schema untouched.

### What changed

- `ConceptReferenceRangeResourceService` fetches the ranges for the concepts of every numeric question in the schema, for the patient the form is being filled for.
- `FeWrapperComponent` loads them alongside the previous encounter, once the schema is known.
- `FormCreationService` narrows each numeric question to the absolute bounds of the range that applies, right before the form is built.

A few decisions worth calling out:

- Only `lowAbsolute` / `hiAbsolute` are used. That mirrors what the backend validates against, and matches the [docs](https://openmrs.atlassian.net/wiki/spaces/docs/pages/178782250) ("only absolute values are used"). Normal and critical bounds are for flagging, which the Angular engine has no UI for.
- A bound already in the schema is never widened. If the schema says `max: 50` and the range says 43, you get 43; if the schema says 30 and the range says 25, you get 30. Whichever is stricter wins, same idea as core applying the strictest of several matching ranges.
- Questions can name their concept either by UUID or by a reference like `CIEL:5089`, so those get resolved through the schema's `conceptReferences` map before matching.
- Bounds are written back as strings, which is how the [form schema](https://github.com/openmrs/openmrs-contrib-json-schemas/blob/main/form.schema.json) declares `min` and `max`. Matters more than it looks: a `lowAbsolute` of 0 written as a number would be falsy and the engine would then skip both validators.

Nothing changes on servers that don't have the resource. A failed request, an older backend, or offline mode all fall back to an empty map and the form loads exactly as it does today.

### For form authors and implementers

No schema changes are required. Numeric `obs` questions using `number` or `decimal` automatically use patient-specific absolute reference ranges when their concept resolves by UUID or mapping. Existing `min` and `max` values still apply when they are stricter.

This only applies to implementations using `@openmrs/esm-form-entry-app`, the Angular Form Engine integration. The default O3 React Form Engine does not use this code. The backend must also have concept reference ranges configured, and both effective bounds must be available.

### Known limitation

The engine only wires up the min and max validators when both are present (`if (questionOptions.max && questionOptions.min)` in `QuestionFactory.addValidators`). A concept with only one absolute bound therefore still gets no validation. Left alone here since fixing it means a change in ngx-formentry, happy to file that separately.

## Screenshots

No UI of its own. The out-of-range message is the engine's existing `min` / `max` validation message.

## Related Issue

https://openmrs.atlassian.net/browse/O3-5850

## Other

Tested with `yarn test` in `packages/esm-form-entry-app` (84 passing, 17 new). The `FormCreationService` tests build a real form through `FormFactory` and assert the control rejects 50 °C and 20 °C but accepts 37 °C for a range of 25–43; both fail without the change. Also ran `yarn lint`, `tsc --noEmit` and `yarn build`.


packages/esm-form-entry-app/src/app/openmrs-api/concept-reference-range-resource.service.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (!patientUuid || !conceptReferences?.length) {
+      return of(new Map<string, ConceptReferenceRange>());
+    }
+
+    const params: HttpParams = new HttpParams()
+      .set('patient', patientUuid)
+      .set('concept', conceptReferences.join(','))
+      .set('v', 'full');
+
+    return this.http.get<ConceptReferenceRangeResponse>(this.getUrl(), { params }).pipe(
```
denniskigen: The batching here is useful. Can we also handle the REST page limit? `/conceptreferencerange` returns `NeedsPaging`, and REST defaults to 50 results. AMRS already has forms with 57 unique numeric concepts, so later ranges can be silently omitted when more than 50 concepts resolve. It's probably best to follow `links.next` or split this into bounded requests.

One possible starting point is to add `links` to `ConceptReferenceRangeResponse`, then use RxJS `expand` to follow the link where `rel === "next"` and `reduce` to combine the results before building the map.

denniskigen (commented):
Good start, @Nahid-NHB. I've left some feedback on the REST request handling.

packages/esm-form-entry-app/src/app/openmrs-api/concept-reference-range-resource.service.ts
```diff
@@ -0,0 +1,61 @@
...
+    patientUuid: string,
+    conceptReferences: Array<string>,
+  ): Observable<Map<string, ConceptReferenceRange>> {
+    if (!patientUuid || !conceptReferences?.length) {
+      return of(new Map<string, ConceptReferenceRange>());
+    }
+
+    const params: HttpParams = new HttpParams()
+      .set('patient', patientUuid)
+      .set('concept', conceptReferences.join(','))
```
denniskigen: The single request is efficient. Can we stop one unresolved concept from discarding all the valid ranges? The current backend throws on the first unknown reference, and `catchError` then returns an empty map for the whole form. A hidden or unused stale question could therefore disable range validation everywhere. It's probably best to filter unresolved mappings before the request and isolate any remaining failures in bounded batches.

A useful first step would be to resolve and deduplicate the concepts before making the request. We can skip mappings that are present in `conceptReferences` but have no UUID, while keeping valid UUIDs and resolved mappings. Stale raw UUIDs may still need failed requests to be isolated into smaller batches.

packages/esm-form-entry-app/src/app/openmrs-api/concept-reference-range-resource.service.ts
Nahid-NHB: Good catch, that's a real hole. `doSearch` returns `NeedsPaging` and REST caps at 50 by default, so anything past the 50th concept just disappears with no error and no signal that it happened.

I'll follow `links.next` rather than chunking. Client side chunking works, but it bakes the server's page size into the frontend, and `expand` stays correct whatever `webservices.rest.maxResultsAbsolute` is set to.

packages/esm-form-entry-app/src/app/openmrs-api/concept-reference-range-resource.service.ts
Nahid-NHB: @denniskigen 
u r both suggestion applied in [4472e77f](https://github.com/openmrs/openmrs-esm-patient-chart/pull/3546/changes/4472e77fb29758dd4e2d4a258c20c7b5de71afbf). Concepts are resolved and deduped against `conceptReferences` before the request, so references the schema marks as unresolved never get sent. The rest goes out in batches of 25 with `catchError` inside the per-batch pipe, so a failing batch only costs its own ranges. Requests follow `links.next` now too.

One question on the batching. `ConceptReferenceRangeResource2_7.doSearch` looks like it skips references it can't resolve rather than throwing

packages/esm-form-entry-app/src/app/form-schema/concept-reference-ranges.ts
```diff
@@ -0,0 +1,127 @@
...
+ * Questions can refer to their concept either by UUID or by a concept reference such as `CIEL:5089`,
+ * while the reference ranges are always keyed by UUID.
+ *
+ * Returns nothing for a reference the schema itself already tells us does not resolve, which is how
+ * a stale question ends up being left out of the request instead of taking a whole batch down.
+ */
+function resolveConceptUuid(formSchema: FormSchema, concept: string): string | undefined {
+  const conceptReferences = formSchema.conceptReferences;
+
+  if (conceptReferences && concept in conceptReferences) {
```
denniskigen:
```suggestion
  if (conceptReferences) {
    return conceptReferences[concept]?.uuid ?? undefined;
  }

  return concept;
```

[`/o3/forms` omits unresolved concepts from `conceptReferences`](https://github.com/openmrs/openmrs-module-o3forms/blob/a60eb63ec1bf70800690b014ae6e8cfbdf9632b9/omod/src/main/java/org/openmrs/module/o3forms/api/impl/O3FormsServiceImpl.java#L182-L206); it doesn't add entries with `uuid: null`. The current check therefore sends the stale reference to `/conceptreferencerange`, where it [fails the whole batch](https://github.com/openmrs/openmrs-module-webservices.rest/blob/6f5471d1aaba5201e322adcad94788e62c5d92b1/omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_8/ConceptReferenceRangeResource2_8.java#L96-L100). Can we cover the omitted-key shape in the test too?

Nahid-NHB:
@denniskigen 
You're right, I had the wrong shape in mind. I assumed /o3/forms would return the key with a null uuid, but O3FormsServiceImpl only does result.put when the concept resolves, so a stale reference just isn't there. The concept in conceptReferences check then sends it straight through, which defeats the point of pre-resolving. Applying the suggestion.

packages/esm-form-entry-app/src/app/form-schema/concept-reference-ranges.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  if (conceptReferences) {
+    return conceptReferences[concept]?.uuid ?? undefined;
+  }
+
+  return concept;
+    return conceptReferences[concept]?.uuid ?? undefined;
+  }
+
+  return concept;
```
denniskigen:
```suggestion
```

This block was left behind when the suggestion was applied, so the file no longer parses. Remove it.

packages/esm-form-entry-app/src/app/form-schema/concept-reference-ranges.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+      [
+        { id: 'stale', type: 'obs', questionOptions: { rendering: 'number', concept: 'CIEL:00000' } as any },
+        { id: 'alsoStale', type: 'obs', questionOptions: { rendering: 'number', concept: 'CIEL:11111' } as any },
+        { id: 'weight', type: 'obs', questionOptions: { rendering: 'decimal', concept: 'CIEL:5089' } as any },
+      ],
+      {
+        'CIEL:00000': { uuid: null, display: null },
+        'CIEL:11111': {},
+        'CIEL:5089': { uuid: 'weight-uuid', display: 'Weight (kg)' },
+      },
```
denniskigen, line 83:
```suggestion
      {
        'CIEL:5089': { uuid: 'weight-uuid', display: 'Weight (kg)' },
      },
```

[`/o3/forms` omits unresolved references from this map](https://github.com/openmrs/openmrs-module-o3forms/blob/a60eb63ec1bf70800690b014ae6e8cfbdf9632b9/omod/src/main/java/org/openmrs/module/o3forms/api/impl/O3FormsServiceImpl.java#L182-L206); it doesn't return null or empty entries. Can we make the fixture match that response shape so this test covers the regression?

packages/esm-form-entry-app/src/app/openmrs-api/concept-reference-range-resource.service.ts
denniskigen: Both, depending on the platform version. You're reading [`ConceptReferenceRangeResource2_7`](https://github.com/openmrs/openmrs-module-webservices.rest/blob/6f5471d1aaba5201e322adcad94788e62c5d92b1/omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_7/ConceptReferenceRangeResource2_7.java#L170-L178), which covers `2.7.0 - 2.7.8, 2.8.0 - 2.8.4` and does skip:

```java
Concept concept = conceptService.getConceptByReference(conceptReference.trim());
if (concept != null) { ... }
```

[`ConceptReferenceRangeResource2_8`](https://github.com/openmrs/openmrs-module-webservices.rest/blob/6f5471d1aaba5201e322adcad94788e62c5d92b1/omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_8/ConceptReferenceRangeResource2_8.java#L96-L100) overrides `doSearch` for `2.7.9 - 2.7.999, 2.8.5 - 9.*` and throws instead:

```java
if (concept == null) {
    throw new ObjectNotFoundException("No concept with ref: " + conceptRef);
}
```

The reference application pins Platform 2.8.7 and webservices.rest 3.5.0, so 2_8 is the one running. Against a local instance of it:

```
GET /conceptreferencerange?patient=…&concept=CIEL:5089,CIEL:5090&v=full
  → 2 results ("resourceVersion":"2.8"), Weight hiAbsolute 250.0, Height hiAbsolute 272.0

GET /conceptreferencerange?patient=…&concept=CIEL:5089,CIEL:00000,CIEL:5090&v=full
  → ObjectNotFoundException: No concept with ref: CIEL:00000
    at ConceptReferenceRangeResource2_8.doSearch(ConceptReferenceRangeResource2_8.java:100)
```

Both valid ranges lost to the one stale reference, which is what the batching now bounds.

denniskigen (approved):
LGTM. Thanks, @Nahid-NHB!

PR #3548, opened 2026-08-05
(fix) O3-5875: Guard missing FHIR array entries in chart widgets

omeriinnocent:
## Requirements

## Summary

Adds optional chaining to nine nested FHIR accesses across six files:

- `identifier.type.coding[0].code` in the medications, orders, and vitals views
- `patient.address[0].city` in those three views and the test-results print modal
- `observation.code.coding[0].display` in `usePanelData`
- `observation.code.coding[0].code` in `getConceptUuid`

This is defensive hardening rather than a fix for a reproducible crash. The current OpenMRS FHIR2 translation path does not emit these shapes: an addressless patient has no `address` key, while identifier and observation
codings are populated. These guards still make the UI tolerant of non-standard or client-constructed data.

The test-results print modal already used the guarded identifier access. This change brings the other copies in line with it.

The duplicated `patientDetails` memo remains a follow-up, as discussed in the review.

## Related Issue

https://openmrs.atlassian.net/browse/O3-5875


Josephkagimu1 (commented):
Nice work @omeriinnocent  , A couple of observations:

1. **DRY opportunity:** The `patientDetails` useMemo block (name, age, gender, location, identifiers) is now duplicated across 4 files with identical logic. This could be extracted into a shared `usePatientDetails(patient)` hook in `@openmrs/esm-patient-common-lib` in a follow-up PR. Not a blocker for this one though.

2. **Minor:** In `usePanelData.tsx`, the guard on `observation?.code?.coding?.[0]?.display` — is `observation.code` actually undefined in some cases, or is this just defensive? If the FHIR Observation response is missing `code`, that might indicate an upstream data issue worth investigating separately.

other words LGTM  👍

omeriinnocent:
Thank you for the thorough review, @Josephkagimu1, I appreciate the thoughtful feedback.

On the DRY opportunity:- You're absolutely right — extracting `usePatientDetails()` into a shared hook would clean this up nicely. I'll create a follow-up ticket to refactor this across the codebase. For this PR, I wanted to keep scope minimal and focused on the defensive guard.

On the `observation.code` validation:-Great question. In our current implementation, the FHIR observations **should** always have a `code` field, but I added the guard defensively to handle incomplete or malformed data gracefully rather than crashing. This prevents runtime errors when the API response is incomplete (which can happen during edge cases or data migrations). If we see this trigger in production, we can investigate upstream. For now, the guard keeps the UI stable.

VeronicaMuthee:
@omeriinnocent, thanks for the bug fix. For hardening purposes, could you add a regression test covering the crash case — e.g., render VitalsOverview (or the print modal) with a patient whose address is [] and an identifier whose type.coding is [], and assert that it renders without throwing, with the location displayed as blank? The existing tests all use fully populated mock patients, so the crash path this PR fixes is currently untested and could regress unnoticed.
cc: @denniskigen

packages/esm-patient-tests-app/src/test-results/individual-results-table-tablet/usePanelData.tsx
```diff
@@ -126,7 +126,7 @@ export default function usePanelData(patientUuid: string) {
         const meta = conceptData[conceptUuid];
         const interpretation = meta?.getInterpretation(value);
 
-        const name = observation?.code.coding[0].display;
+        const name = observation?.code?.coding?.[0]?.display;
```
denniskigen: This handles missing or empty `coding`, but a missing `code` still throws in `getConceptUuid` before reaching this line. If we want to cover that case too, can we update the helper to:

```ts
export const getConceptUuid = (obs: FHIRObservationResource | ObsRecord) => obs?.code?.coding?.[0]?.code;
```

denniskigen (approved):
LGTM. Thanks, @omeriinnocent!

denniskigen (approved):

omeriinnocent:
Glad and thanks much @VeronicaMuthee @denniskigen

PR #3563, opened 2026-08-09
(fix) O3-5845: Preserve task due dates across edit-save cycles

solomonfortune:
## Requirements

## Summary

A task with a Date-type due date could shift one day earlier whenever it was opened for editing and saved. Repeating the cycle could move the date back again each time. This affected dates whose local calendar date differed from their UTC date, including local-midnight selections in time zones ahead of UTC.

The edit form converted the stored date to `YYYY-MM-DD` using `.toISOString().split('T')[0]`. This selected the UTC calendar date and discarded the original time and offset. On save, `parseDate` interpreted that date-only value at local midnight, persisted the shifted instant, and allowed the drift to repeat on the next edit.

The form now retains the full ISO instant using `.toISOString()`, matching the value produced by the date picker. Saving an unchanged task therefore preserves the exact stored instant and displays the correct local calendar date.

Tests cover:

- Displaying the correct local date for an instant near a UTC day boundary in `Africa/Nairobi`.
- Preserving the exact due-date instant when an existing task is saved without changing its due date.

## Screenshots

**Before:** The edit form shows the shifted due date, `19/08/2026`.

Before the fix: the edit task form shows the shifted due date as 19 August 2026

**After:** The edit form shows the correct due date, `20/08/2026`.

After the fix: the edit task form shows the correct due date as 20 August 2026

## Related Issue

[O3-5845: Task due date shifts one day earlier every time an existing task is edited and saved](https://openmrs.atlassian.net/browse/O3-5845)

## Other

This fix prevents further drift but does not repair due dates on tasks that already went through one or more affected edit-save cycles.


solomonfortune:
@VeronicaMuthee could you please review this PR in relation to "Task due date shifts one day earlier every time an existing task is edited and saved" or request for a review from a person responsible

packages/esm-patient-task-list-app/src/workspace/add-task-form.test.tsx
```diff
@@ -387,6 +387,47 @@ describe('AddTaskForm', () => {
...
+            editTaskUuid={editTaskUuid}
+          />,
+        );
+
+        await waitFor(() => {
+          expect(screen.getByLabelText(/task name/i)).toHaveValue('Existing Task');
+        });
+
+        await waitFor(() => {
+          expect(screen.getByRole('textbox', { name: /due date/i })).toHaveValue('20/01/2024');
```
denniskigen, line 426: Good find on the root cause, @solomonfortune, and thanks for adding a regression test!

This covers the display side. Can we also assert that the saved value remains unchanged in `should preserve existing task properties when updating` just above?

```ts
dueDate: expect.objectContaining({
  type: 'DATE',
  date: baseTask.dueDate.date,
}),
```

I verified this fails before the fix because the timestamp is replaced with midnight, and passes with this change. It’s worth covering both since the ticket symptom happens after saving.

solomonfortune: Wow, thanks for this. Let me work on it. probably i failed to catch that

solomonfortune (commented):
@denniskigen Turns out the bug also flattened the saved time to local midnight, not just the day (10:00:00Z → 00:00:00Z), since the date-only string had nowhere to recover the original time from. Confirmed the assertion fails pre-fix with that exact midnight-collapse and passes post-fix. Pushed. Hope this is what i had to keep looking out for.

denniskigen (approved):
LGTM. Thanks, @solomonfortune!

PR #3565, opened 2026-08-11
(fix) UVL emr Orthanc Dicom consideration under 'authorisation' plugin patch

Virlein:
## Requirements

- [x] Tested locally against a live OpenMRS instance with `orthanctoken` installed
- [ ] Tested against a genuine Orthanc-backed attachment (both a valid session and an expired one)
- [ ] Confirmed the correct redirect target for expired sessions (see "Other" below — currently unresolved)
- [ ] Non-Orthanc attachment types (images, PDFs, downloads) confirmed unaffected

## Summary

Attachment links pointing at Orthanc's Stone Web Viewer (`…/pacs/stone-webviewer/index.html?study=<UID>&orthancId=<id>`) were being opened directly via `window.open(attachment.description, ...)`. The `orthancId` in that URL exists so a frontend widget can exchange it for a real access token — but nothing was calling the `orthanctoken` module's `/ws/rest/v1/orthanc/token` endpoint to do that exchange. The backend module is installed and healthy on the deployment; it was simply never invoked from the frontend.

`openAttachment` in `attachments-overview.component.tsx` now detects an Orthanc attachment link (by its `:8889` host marker), extracts the study identifier from either a `study=` or `orthancId=` query parameter (or a `#/studies/<id>` fragment, for URL formats that use it), and calls `/openmrs/ws/rest/v1/orthanc/token` with that identifier before opening anything. Only the token-bearing URL returned by that endpoint is opened — the raw stored link is never navigated to directly anymore.

Non-Orthanc attachments (images, PDFs, downloads) are unaffected — that branch of `openAttachment` is unchanged.

**Error handling:**
- A `401` from the token endpoint is treated as an expired session: the user is redirected to log back in, with a `redirect` parameter set so they land back on the current page afterward.
- Any other non-`OK` response, or a network failure, surfaces a snackbar (`dicomTokenError`) rather than failing silently or opening a broken link.
- A URL that doesn't contain a recognizable study identifier at all surfaces its own snackbar (`dicomStudyIdError`) rather than attempting the fetch with an empty/undefined value.

## Screenshots

Captura de pantalla 2026-08-11 a las 15 59 20

## Related Issue

The related issue is the required OrthancToken Module for OpenMRS as the Orthanc Authorization plugin requires resource-specific token to access the StoneWebViewer.

## Other

**Open question for reviewers:** the `401` handling currently redirects to `/openmrs/login.htm` — the standard OpenMRS 2.x login page. Given this deployment authenticates through Keycloak SSO, should this instead redirect through the Keycloak login flow? Flagging this explicitly rather than assuming either answer, since getting it wrong would silently break the expired-session recovery path specifically.

The contribution has been proven to work on LocalHost / It will soon be tested under UAT deployment — needs a real expired/valid session test against an environment with `orthanctoken` installed and a genuine Orthanc-backed attachment present, to confirm both the happy path and the `401` recovery path work as written.

This currently doesn't have 'Jira' tickets as the deployment of this solution is under review by @mks-d, @VaishSiddharth and @rbuisson on the Slack 'Ozone' channel.


packages/esm-patient-attachments-app/src/attachments/attachments-overview.component.tsx
```diff
@@ -83,17 +83,58 @@ const AttachmentsOverview: React.FC<AttachmentsOverviewProps> = ({ patientUuid }
...
+      const orthancUrl = attachment.description ?? '';
+      if (orthancUrl.includes(':8889')) {
+        const normalizedUrl = orthancUrl.replace(/&amp;/g, '&');
+        const studyIdMatch = normalizedUrl.match(/[?&]study=([^&#+]+)/) || normalizedUrl.match(/#\/studies\/([^?&]+)/);
+        const orthancIdMatch = normalizedUrl.match(/[?&]orthancId=([^&]+)/);
+        const studyId = orthancIdMatch ? orthancIdMatch[1] : studyIdMatch ? studyIdMatch[1] : null;
+        if (!studyId) {
+          showSnackbar({
+            title: t('dicomError', 'DICOM Viewer Error'),
+            subtitle: t('dicomStudyIdError', 'Could not extract study ID from DICOM attachment.'),
```
EDSONZ-WASSWA, line 101: Hay since the key already exists in en.json, i18next resolves to that value rather than falling back to the inline default — so when the token exchange fails (bad response, network error, orthanctoken down, etc.), the user sees the literal text dicomTokenError in the snackbar instead of any useful message. This is distinct from the dicomStudyIdError snackbar just above it, which is written correctly and shows a real message — so it's easy to test the "can't parse study ID" path, see a sensible error, and miss that the token-failure path is silently broken.

I Suggest this --->give it a real, translatable string with an interpolation slot, consistent with how dicomStudyIdError is done:

json
"dicomTokenError": "Could not open DICOM viewer: {{error}}"

and call it as:

js
subtitle: t('dicomTokenError', 'Could not open DICOM viewer: {{error}}', { error: (err as Error).message }),

rather than building the interpolated string in JS and passing it as the default. This also has the side benefit of making the string properly translatable later — as written now, translators localizing this file would have no idea the value is supposed to say anything at all.

Since this is exactly the failure path the PR's own description calls out as intentionally handled ("surfaces a snackbar... rather than failing silently"), I'd treat this as a small but real regression against that stated goal, worth fixing before merge rather than as a follow-up.

EDSONZ-WASSWA (changes requested):

EDSONZ-WASSWA:
hi @Virlein  thank you for taking up this could you please attach a link of the Jira ticket this responds to .
Then also have a look at this document of how you can raise the pr
[https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477199/Pull+Request+Tips](https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477199/Pull+Request+Tips)

packages/esm-patient-attachments-app/src/attachments/attachments-overview.component.tsx
Virlein: Hi @EDSONZ-WASSWA!

I would like to thank you for the time you took to review the code! I did the changes that you advised me, it was right on the point! Concerning the JIRA ticket, in the 'Other' section of the PR description I've added that this implementation is currently under review by the '#ozone' community on Slack, until then I did not create any JIRA tickets.

Virlein: I however did the modifications as suggested, it should work as intended now.

ibacher:
Thanks for this. I appreciate that integrating Orthanc with OpenMRS is something that people want to do. I have some qualms with this implementation as written:

1. It doesn't feel correct to have this much Orthanc-specific business logic embedded directly into the attachments app. It might be better to consider a sort of generic "attachment points to a link" implementation that could do the URL unescaping and maybe support for URL-specific handlers.
2. It makes too many assumptions about end points that aren't part of the RefApp strictly, e.g., that there's a `/orthanc/token` REST endpoint. That makes it hard to merge this directly as-written. Both that _and_ `/oauth2login` (and the assumed need to flow between them) are hard-coded assumptions that don't apply to _most_ OpenMRS instances.
3. Rather than a direct fetch like what you're doing, we normally use `openmrsFetch()` which should handle the details around authentication if the user isn't authenticated correctly.

Basically, I see what you're trying to do and it makes sense, but there's no way we can add _this_ as written to the system. It would probably be helpful if you could open a ticket in our Jira system with the requirements so we have a forum to discuss what the implementation could look like.

PR #3569, opened 2026-08-12
(fix) O3-5763: Let hosts handle encounter editing outside the chart

UjjawalPrabhat:
## Requirements

## Summary

Follow-up to #3509. `EncountersTable` launches the chart's edit workspaces, which belong to the `patient-chart` group — scoped to chart URLs, with group props only the chart populates. Hosted elsewhere (Service Queues, via the `visit-summary` extension), the window opens with null group props and every action button in the group throws at once, one per registered icon.

Null-safe buttons wouldn't help, as the group is out of scope there regardless; nor would overriding the workspace names, since the exported notes form takes the patient context as workspace props while exported form entry takes it as window
props. Hosts have to do the launching, hence the optional callback. Absent it, chart behaviour is unchanged.

The `.expandedRow > div` rule removed here was already dead — a `<tr>` has no `<div>` child. Repairing it instead would apply `$ui-02` and leave the expanded rows a white block inside their grey gutters.

## Related Issue

[O3-5763](https://openmrs.atlassian.net/browse/O3-5763)

## Other

Related PR [here](https://github.com/openmrs/openmrs-esm-patient-management/pull/2600).

[O3-5763]: https://openmrs.atlassian.net/browse/O3-5763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jwnasambu (commented):
Thanks @UjjawalPrabhat for this progress. Kindly do you mind adding a test for the `onEditEncounte`r callback path? The existing tests only cover the launchWorkspace2 fallback.

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/encounters-table.test.tsx
```diff
@@ -340,6 +342,80 @@ describe('Encounter editability', () => {
   });
 });
 
+describe('Edit Encounter', () => {
+  const admissionRowName = /Select row 18-Jan-2022, 04:25 PM Facility Visit Admission POC Consent Form -- Options/i;
+  const visitNoteRowName = /Select row 03-Aug-2021, 12:47 AM Facility Visit Visit Note -- User One Options/i;
+  const [mockAdmissionEncounter, mockVisitNoteEncounter] = mockEncountersAlice;
```
NethmiRodrigo: Since none of these mocks has a `encounter.form`, a Visit Note encounter that has a form isn't covered. Lets add a case with` { ...mockVisitNoteEncounter, form: mockAdmissionEncounter.form }` and assert the callback gets false.

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/encounters-table.test.tsx
```diff
@@ -340,6 +342,80 @@ describe('Encounter editability', () => {
   });
 });
 
+describe('Edit Encounter', () => {
```
NethmiRodrigo: Nit - The describe block just above is for `Encounter editability`, which sounds the same as this...

NethmiRodrigo: I think we need to make it more clear that these test cases are meant to cover the paths that would trigger if the prop was provided. We could add these tests to the describe block above and make the test description more clearer to state this

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/visit-summary.component.tsx
```diff
@@ -165,10 +166,10 @@ const VisitSummary: React.FC<VisitSummaryProps> = ({ visit, patientUuid }) => {
             <MedicationSummary medications={medications} />
           </TabPanel>
           <TabPanel>
-            <VisitCompletedFormsTable visit={visit} patientUuid={patientUuid} />
+            <VisitCompletedFormsTable visit={visit} patientUuid={patientUuid} onEditEncounter={onEditEncounter} />
```
NethmiRodrigo, line 169: If this `onEditEncounter={onEditEncounter}` is ever removed, nothing catches it. Can we add a test in `visit-summary.test.tsx` that renders this with the callback and asserts that clicking edit inside one of the tabs calls it?

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/encounters-table.test.tsx
```diff
@@ -381,6 +457,14 @@ function renderEncountersTable(props: Partial<EncountersTableProps> = {}) {
...
 }
 
+async function clickEditEncounter(rowName: RegExp) {
+  const user = userEvent.setup();
+  const row = screen.getByRole('row', { name: rowName });
+  await user.click(within(row).getByRole('button', { name: /expand current row/i }));
+  const expandedRow = row.nextElementSibling as HTMLElement;
+  await user.click(within(expandedRow).getByRole('button', { name: /edit this encounter/i }));
+}
+
```
NethmiRodrigo, line 475: Since all of the tests are testing the edit functionality through the edit button when the row is expanded, for at least one of the tests, lets test it through the edit button in the overflow menu.

```suggestion

async function clickEditEncounterViaOverflowMenu(rowName: RegExp) {
   const user = userEvent.setup();
   const row = screen.getByRole('row', { name: rowName });
   await user.click(within(row).getByRole('button', { name: /options/i }));
   const overflowMenu = screen.getAllByText('Focus sentinel')[0].parentElement;
   await user.click(within(overflowMenu).getByText(/edit this encounter/i));
 }
```

NethmiRodrigo (changes requested):
Thanks @UjjawalPrabhat!

NethmiRodrigo (approved):
Thanks @UjjawalPrabhat!

PR #3578, opened 2026-08-17
(fix) Flag inverted same-day visit end times on the time field

pirupius:
## Requirements

## Summary

The visit form's stop-before-start rule always attached its error to `visitStopDate`. When only the times were inverted on a single day, the date picker was flagged red while the offending time picker looked valid. The error now lands on the end time field when the start and stop fall on the same day, and on the end date field when the dates themselves are out of order.

Visits whose stop datetime equals their start datetime stay valid: core rejects only a stop before the start, the EMR API closes visits with no non-voided encounters by setting the stop equal to the start, and the form's hh:mm fields drop seconds, so a short real visit collapses onto a single minute when loaded for editing. Tests cover the field routing and the equal-endpoints cases.

## Screenshots
Screenshot 2026-08-17 at 21 56 29


packages/esm-patient-chart-app/src/visit/visit-form/visit-form.resource.ts
```diff
@@ -344,11 +344,14 @@ export function useVisitFormSchemaAndDefaultValues(visitToEdit: Visit, earliestA
               message: t('futureEndTime', 'End time cannot be in the future'),
               path: ['visitStopTime'],
             });
-          } else if (visitStopDateTime < visitStartDateTime) {
+          } else if (visitStartDateTime !== null && visitStopDateTime <= visitStartDateTime) {
```
denniskigen:
```suggestion
          } else if (visitStartDateTime !== null && visitStopDateTime < visitStartDateTime) {
```

Can we keep equality valid here? Core rejects only a stop before the start, and EMR API deliberately closes visits with no non-voided encounters by setting `stopDatetime` equal to `startDatetime`. With `<=`, those visits can't be re-saved in this form without changing their timestamps. The same applies to a positive-duration visit whose endpoints are truncated into the same `hh:mm`.

I verified both cases against the backend and this branch. It's probably better to keep the field-path fix, add a new key such as `endTimeCannotBeBeforeStartTime` with “End time cannot be before start time”, run `yarn turbo run extract-translations`, and flip the equality test to assert acceptance.

pirupius: Thanks @denniskigen for this catch

denniskigen (approved):
LGTM. Thanks, @pirupius!

PR #3582, opened 2026-08-18
(test) O3-5888: Add e2e test coverage for entering lab results via the Laboratory app

VeronicaMuthee:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number.
- [ ] My work is based on designs, which are linked or shown either in the Jira ticket or the description below. 
- [x] My work includes tests or is validated by existing tests.

## Summary

The patient-chart e2e suite covered adding lab results via the **Orders app** ("Add results") but had no coverage for the **Laboratory app** worklist flow. This PR adds that coverage.

- New spec `e2e/specs/lab-app-results.spec.ts` driving the full lifecycle: seed a lab order via REST → **Pick lab request** from *Tests ordered* (→ `IN_PROGRESS`) → **Add lab results** from *In progress* → assert the order moves to *Completed* → verify the saved value in the Results Viewer.
- Enables `@openmrs/esm-laboratory-app` in the e2e docker frontend assembly (`e2e/support/github/run-e2e-docker-env.sh`) so the Laboratory app is present in CI.
- Hoists a shared lab-order fixture (`e2e/core/lab-fixtures.ts`) and a shared Results-Viewer assertion helper (`e2e/commands/test-helpers.ts`), consumed by both lab specs, removing prior duplication in `e2e/specs/lab-orders.spec.ts`.
- Adds a `LaboratoryPage` page object.

All 5 lab specs pass locally against a running O3 instance; the new spec is stable across repeated and concurrent runs.

## Screenshots

N/A — test-only change (no UI changes).

## Related Issue

https://openmrs.atlassian.net/browse/O3-5888

## Other

The Laboratory app is not part of this monorepo, so it is pulled from npm (`@openmrs/esm-laboratory-app: "next"`) alongside the existing home/primary-navigation apps in the e2e assemble step. Note `next` is a floating tag — a future upstream publish could affect this suite; happy to pin to a fixed version if reviewers prefer.

PR #3583, opened 2026-08-18
(chore) Typecheck and lint the E2E suite in CI

denniskigen:
## Requirements

## Summary

The e2e directory is currently excluded from every static check in CI. `yarn verify` runs `turbo run lint typescript test`, which only covers workspaces under `packages/`, and the E2E workflow transpiles the specs without typechecking. So a type error in a spec, fixture, command, or page object never fails a build.

This PR adds an `e2e/tsconfig.json` that extends the root config and covers everything under `e2e/` plus `playwright.config.ts`, and registers two root-level turbo tasks (`//#typescript` and `//#lint`) so `tsc -p e2e` and `eslint e2e` run as part of the existing verify pipeline on every PR. No workflow changes are needed since `turbo run lint typescript test` picks up the root tasks automatically.

The e2e program follows `@openmrs/esm-patient-common-lib` into its source through the package's `src/index.ts` main entry, so `e2e/declarations.d.ts` mirrors the per-package `*.scss` module declaration to keep those imports resolvable.

A few notes on validation:

- The new program covers all 53 TS files under `e2e/` plus `playwright.config.ts` (verified with `tsc --listFilesOnly`), and an injected type error in a spec fails the build.
- The e2e code is already clean, so both checks pass as-is. Same for `eslint e2e`, which the `e2e/**` override in `eslint.config.mjs` was already set up for but nothing ran in CI.
- `yarn verify` passes locally with the two root tasks in the graph.

## Other

Spun out of reviewing #3582, but independent of it.

### DIFF
--- a/e2e/declarations.d.ts
+++ b/e2e/declarations.d.ts
@@ -1,0 +1,4 @@
+declare module '*.scss' {
+  const content: { [className: string]: string };
+  export default content;
+}
--- a/e2e/tsconfig.json
+++ b/e2e/tsconfig.json
@@ -1,0 +1,4 @@
+{
+  "extends": "../tsconfig.json",
+  "include": ["**/*", "../playwright.config.ts"]
+}
--- a/package.json
+++ b/package.json
@@ -7,6 +7,8 @@
   ],
   "scripts": {
     "start": "openmrs develop --sources packages/esm-patient-chart-app/",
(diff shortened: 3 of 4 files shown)


NethmiRodrigo (approved):
Thanks @denniskigen!

PR #3586, opened 2026-08-18
(feat) O3-5889: Support fullWidth configuration for the obs-by-encounter widget

denniskigen:
## Requirements

## Summary

The obs-by-encounter widget is registered with `"fullWidth": false` in its routes.json meta, and there was no way for implementers to override that. This PR adds a `fullWidth` property (default `false`) to the widget's config schema, so each configured instance can opt into spanning the full width of the dashboard:

```json
"patient-chart-summary-dashboard-slot": {
  "add": ["obs-by-encounter-widget#partograph"],
  "configure": {
    "obs-by-encounter-widget#partograph": { "fullWidth": true }
  }
}
```

To make that work, `DashboardView` now resolves each widget's width from its extension config first, falling back to the extension's routes.json meta.

This also fixes a latent bug. `DashboardView` looked up meta by extension instance id, but `useExtensionSlotMeta` keys its map by extension name. For instances added via config with a `#` suffix, such as `obs-by-encounter-widget#partograph`, the lookup always missed and the widget rendered half width regardless of its registered meta. Reading meta straight off the `AssignedExtension` fixes that, so `#` suffixed instances of other widgets now honor their registered `fullWidth` value too.

## Related Issue

https://openmrs.atlassian.net/browse/O3-5889

## Other

I deliberately didn't add `fullWidth` to the horizontal obs table's config schema. That schema serves both `obs-table-horizontal-widget-small` (meta `false`) and `obs-table-horizontal-widget` (meta `true`), and a schema default would override the meta for both registrations, flipping the full width one to half width.


packages/esm-generic-patient-widgets-app/src/config-schema-obs-switchable.ts
```diff
@@ -97,6 +97,11 @@ export const configSchemaSwitchable = {
     _description: 'Display Encounter type column',
     _default: false,
   },
+  fullWidth: {
```
ibacher, line 100: I don't love surfacing what's really an implicit schema like this, but I suppose it works.

ibacher (approved):

packages/esm-generic-patient-widgets-app/src/config-schema-obs-switchable.ts
denniskigen: Agreed it's a bit implicit. It's probably better to handle this where meta gets assembled rather than in each widget's schema. I've filed [O3-5891](https://openmrs.atlassian.net/browse/O3-5891) proposing a reserved "Meta overrides" section alongside "Display conditions" and "Translation overrides". If that lands, this schema key should eventually go away. What do you think of that approach?

ibacher: I think at one point I had even drafted a version just using "meta", but yeah, "Meta overrides" is consistent.

Muta-Jonathan:
Thanks @denniskigen @ibacher

PR #3591, opened 2026-08-20
(feat) O3-5907: Render medication side effects extension slot on the drug order form

druchniewicz:
## Requirements

## Summary

 Adds a single `drug-order-form-side-effects-slot` extension slot to the drug order form, mounted with the selected drug's uuid when one is set. The host only declares the slot - all rendering lives in the `openmrs-esm-medication-side-effects` microfrontend, which plugs its panel in.

## Screenshots

image

## Related Issue

  https://openmrs.atlassian.net/browse/O3-5907

## Other

  Related PRs:
  * https://github.com/openmrs/openmrs-esm-medication-side-effects/pull/1
  * https://github.com/openmrs/openmrs-esm-dispensing-app/pull/889
  * https://github.com/openmrs/openmrs-module-medicationsideeffects/pull/1
  * https://github.com/mekomsolutions/openmrs-module-initializer/pull/333
  * https://github.com/openmrs/openmrs-content-referenceapplication-demo/pull/85

### DIFF
--- a/packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
+++ b/packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
@@ -512,7 +512,7 @@ export function DrugOrderForm({
                 unitValue={watchedUnitValue}
               />
             </div>
-            {drug?.uuid && <ExtensionSlot name="order-form-side-effects-slot" state={{ drugUuid: drug.uuid }} />}
+            {drug?.uuid && <ExtensionSlot name="drug-order-form-side-effects-slot" state={{ drugUuid: drug.uuid }} />}
             <section className={styles.formSection}>
               <Grid className={styles.gridRow}>
                 <Column lg={12} md={6} sm={4}>
--- a/packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
+++ b/packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
@@ -512,6 +512,7 @@ export function DrugOrderForm({
                 unitValue={watchedUnitValue}
               />
             </div>
+            {drug?.uuid && <ExtensionSlot name="order-form-side-effects-slot" state={{ drugUuid: drug.uuid }} />}
             <section className={styles.formSection}>
               <Grid className={styles.gridRow}>
                 <Column lg={12} md={6} sm={4}>
--- a/packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.test.tsx
+++ b/packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.test.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { vi, describe, it, expect, beforeEach, afterEach } from 'vitest';
 import userEvent from '@testing-library/user-event';
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
-import { getDefaultsFromConfigSchema, OpenmrsDatePicker, useConfig, useSession } from '@openmrs/esm-framework';
+import { ExtensionSlot, getDefaultsFromConfigSchema, OpenmrsDatePicker, useConfig, useSession } from '@openmrs/esm-framework';
 import { type DrugOrderBasketItem } from '@openmrs/esm-patient-common-lib';
 import { mockPatient } from 'tools';
 import { mockDrugSearchResultApiData, mockSessionDataResponse } from '__mocks__';
@@ -13,6 +13,7 @@ import DrugOrderForm, { getOrderStartDateForSubmission } from './drug-order-form
 const mockUseConfig = vi.mocked(useConfig<ConfigObject>);
 const mockUseSession = vi.mocked(useSession);
 const mockOpenmrsDatePicker = vi.mocked(OpenmrsDatePicker);
+const mockExtensionSlot = vi.mocked(ExtensionSlot);
 const mockUseMedicationOrders = vi.fn().mockReturnValue({
   futureOrders: [],
   activeOrders: [],
@@ -205,6 +206,18 @@ describe('DrugOrderForm - auto-calculation of dispense quantity', () => {
       expect(quantityInput).toHaveValue(14);
     });
     expect(screen.getByText(/auto-calculated/i)).toBeInTheDocument();
+  });
+
+  it('mounts the side effects extension slot with the selected drug uuid', () => {
+    mockExtensionSlot.mockClear();
+    const item = createNewOrderBasketItem();
+    renderDrugOrderForm(item);
+
+    const slotCall = mockExtensionSlot.mock.calls.find(
+      (call) => call[0].name === 'drug-order-form-side-effects-slot',
+    );
+    expect(slotCall).toBeTruthy();
+    expect(slotCall?.[0].state).toEqual({ drugUuid: item.drug.uuid });
   });
 
   it('auto-calculates with weekly duration units', async () => {
--- a/packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.test.tsx
+++ b/packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.test.tsx
@@ -2,7 +2,13 @@ import React from 'react';
 import { vi, describe, it, expect, beforeEach, afterEach } from 'vitest';
 import userEvent from '@testing-library/user-event';
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
-import { ExtensionSlot, getDefaultsFromConfigSchema, OpenmrsDatePicker, useConfig, useSession } from '@openmrs/esm-framework';
+import {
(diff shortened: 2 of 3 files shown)


druchniewicz:
@denniskigen @ibacher @dkayiwa can you take a look?

druchniewicz:
@denniskigen @ibacher @dkayiwa bumping this

dkayiwa (commented):
The change itself looks fine to me. What's holding this PR up is CI rather than the code, and I don't think the reason is obvious from the PR page: `pr-title / Validate PR title` and `pr-description / Validate PR description` are both required status checks in the active "Protect main" ruleset, so the branch stays unmergeable while they're red no matter what a reviewer does. Two edits, no code involved:

* The title needs the type prefix the shared workflow expects, so `(feat) O3-5907: Render medication side effects extension slot on the drug order form`. It also got truncated when the PR was opened and currently ends mid-word at "on the drug or...".
* The description has to follow the repo template. The check looks for `Requirements`, `Summary`, `Screenshots`, `Related Issue` and `Other` as markdown headings, the first Requirements checkbox ticked, and real prose under `Summary`. Right now the body is the ticket link plus the related-PR list with no headings at all, so it fails on the missing sections.

Sort those two out and this is mergeable from my side.

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
```diff
@@ -512,6 +512,7 @@ export function DrugOrderForm({
                 unitValue={watchedUnitValue}
               />
             </div>
+            {drug?.uuid && <ExtensionSlot name="order-form-side-effects-slot" state={{ drugUuid: drug.uuid }} />}
```
dkayiwa: Worth qualifying this slot name while both sides are still unmerged. Everywhere else the repo says which order form a slot belongs to: `drug-order-form-actions-slot` further down this same file, `top-of-lab-order-form-slot` for the test and lab order forms. The companion slot for this same feature is `dispensing-prescription-side-effects-slot`, so this ends up the one name in the set that doesn't name its host. `drug-order-form-side-effects-slot` reads consistently and still fits inside prettier's 120 columns (118).

Timing is the real argument. The name is a cross-repo contract, bound in `src/routes.json` in openmrs/openmrs-esm-medication-side-effects#1, so today it's a two-line change across two open PRs. Once both land and the MFE is released it's a change in two repos plus a release. Both sides have to move in the same breath, which is why I've written the name out rather than leaving a committable suggestion. Take it or leave it, but it's much cheaper now than later.

dkayiwa: Could you add a test pinning the slot name and the state key? `drug-order-form.test.tsx` already has the `renderDrugOrderForm(createNewOrderBasketItem())` helper, and the established pattern for this in the repo is `vi.mocked(ExtensionSlot)` plus a lookup over the recorded calls: see `encounters-table.test.tsx:145`, and `form-entry.test.tsx:70` which also asserts on the `state` that was passed.

`order-form-side-effects-slot` and `drugUuid` are a contract with openmrs-esm-medication-side-effects' `routes.json` and `SideEffectsPanelProps`. If either side drifts, a rename here or a typo there, the panel simply doesn't render. Nothing throws, no test fails, and the only way to find out is to open the form against a drug that has side effects recorded. Three lines of assertion catches that class of regression, so it feels worth having even though nothing is wrong today. Not a merge blocker.

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
```diff
@@ -512,6 +512,7 @@ export function DrugOrderForm({
...
             <section className={styles.formSection}>
```
dkayiwa, line 516: `.orderForm` in `drug-order-form.scss` sets `counter-reset: section` and numbers its sections through a descendant rule on bare `h3` (`counter-increment: section` plus a `::before` carrying the counter). That's where "1. Dosage instructions" and the two headings below it get their numbers. The slot on the line above sits inside that `<Form>`, so an extension that renders an `h3` will bump the counter, shift these numbers, and pick up a stray "1. " in front of its own heading.

Nothing trips it today: the panel in openmrs-esm-medication-side-effects#1 uses `<span>` for "Known side effects" and for the group labels. It is an invisible constraint on a public extension point though, and all three numbered headings here already carry `.sectionHeader`, so hanging the counter rule off that class instead of bare `h3` would close it. Deliberate, or worth a follow-up?

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
druchniewicz: renamed to drug-order-form-side-effects-slot

druchniewicz: test added

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.component.tsx
druchniewicz: Moved the counter-increment/::before rule from the bare h3 onto .sectionHeader

druchniewicz:
@dkayiwa changes applied

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.test.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { vi, describe, it, expect, beforeEach, afterEach } from 'vitest';
 import userEvent from '@testing-library/user-event';
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
-import { getDefaultsFromConfigSchema, OpenmrsDatePicker, useConfig, useSession } from '@openmrs/esm-framework';
+import { ExtensionSlot, getDefaultsFromConfigSchema, OpenmrsDatePicker, useConfig, useSession } from '@openmrs/esm-framework';
```
dkayiwa: This file is no longer prettier-clean. With the repo's pinned prettier (3.3.2) and `printWidth: 120`, this import is 126 characters, so prettier wraps it across lines, and the `mockExtensionSlot.mock.calls.find(...)` at 216-218 collapses back onto a single 119-character line. The file was clean on main, so both come from this PR.

`yarn prettier --write packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.test.tsx` fixes both. CI only runs eslint, so nothing goes red today, but `lint-staged` runs `prettier --write` over `*.tsx`, which means whoever commits this file next picks up the two reformats in their own diff. Your call whether that is worth a push.

dkayiwa (commented):
The only thing blocking merge is the description check, and the message it prints points in the wrong direction. Every `##` heading in the PR body starts with two spaces. The shared workflow (`.github/workflows/frontend-pr-description-check.yml` in openmrs-contrib-gha-workflows) tests each one with `re.match(r'^#{1,6}\s+' + header + r'\s*$', line)`, which anchors at column 0, so `  ## Requirements` never matches and all five sections get reported missing even though they are visibly there.

Deleting the leading spaces on those five heading lines is the whole fix. The checkbox rule calls `line.strip()` first, so the indented checkboxes are fine as they stand, and the Summary-content rule passes as soon as `## Summary` matches. It is a required check in the Protect main ruleset, so the branch stays unmergeable while it is red.

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.test.tsx
druchniewicz: fixed

druchniewicz:
@dkayiwa changes applied

packages/esm-patient-medications-app/src/add-drug-order/drug-order-form.test.tsx
```diff
@@ -207,6 +214,16 @@ describe('DrugOrderForm - auto-calculation of dispense quantity', () => {
     expect(screen.getByText(/auto-calculated/i)).toBeInTheDocument();
   });
 
+  it('mounts the side effects extension slot with the selected drug uuid', () => {
+    mockExtensionSlot.mockClear();
```
dkayiwa: `tools/vitest.shared.ts` sets `clearMocks: true`, so vitest already calls `.mockClear()` on every spy before each test and this line is a no-op. Asserting that `mockExtensionSlot.mock.calls.length` is `0` as the first statement of the test passes exactly as the file stands; flipping `clearMocks` to `false` turns that same assertion into `expected 40 to be +0`, which is the leakage the line looks like it is guarding against.

The other `vi.mocked(ExtensionSlot)` tests in the repo (`form-entry.test.tsx`, `encounters-table.test.tsx`, `visit-timeline.test.tsx`) all just lean on the config, so this is the only place that reads as though mocks survive between tests here. Only worth deleting if you end up pushing for something else.

druchniewicz:
@dkayiwa changes applied

PR #3596, opened 2026-08-25
(fix) O3-5908: Prevent duplicate orders from pinned drugs

abheeee03:
## Requirements

## Summary

This fixes inconsistent duplicate-order handling between drug search and pinned drug orders. Pinned drugs now use the same active and scheduled prescription check as search. When a matching order exists, the pinned item is tagged “Already prescribed,” disabled, and does not open the order form.

The pinned section shows a skeleton while prescription status is loading. If the lookup fails, it keeps pinned items visible and unpinnable while disabling ordering.

The drug order form also shows a warning when “Save order” is disabled because an active or scheduled order already exists.

## Screenshots

Pinned drug order marked Already prescribed

Drug order form warning for an existing prescription

## Related Issue

https://openmrs.atlassian.net/browse/O3-5908

## Other

Component tests cover eligible, already-prescribed, loading, error, empty, and unpin behavior.


packages/esm-patient-orders-app/src/order-favorites/drug-favorites.resource.ts
```diff
@@ -1,8 +1,34 @@
 import { useMemo } from 'react';
 import useSWR from 'swr';
 import { getLoggedInUser, openmrsFetch, restBaseUrl, setUserProperties } from '@openmrs/esm-framework';
+import { careSettingUuid, drugCustomRepresentation, type Order } from '@openmrs/esm-patient-common-lib';
 import type { DrugFavoriteOrder, UserResponse } from './types';
 
+export function useActiveDrugOrders(patientUuid: string | undefined) {
+  const url = patientUuid
+    ? `${restBaseUrl}/order?patient=${patientUuid}&careSetting=${careSettingUuid}&v=${drugCustomRepresentation}&excludeDiscontinueOrders=true`
```
denniskigen: This request returns HTTP 400 when the response includes a lab or test order. Without an `orderTypes` filter, the search returns every order type in the outpatient care setting, and `drugCustomRepresentation` asks for `DrugOrder`-only properties, so the REST module throws while converting a `TestOrder`:

```json
{"error":{"message":"[converting class org.openmrs.TestOrder ... => Unknown property 'dosingType' on class 'org.openmrs.TestOrder']"}}
```

I reproduced this against a local backend with the demo data. Since the error is ignored here, the hook returns an empty set and the favorites stay clickable. So for exactly these patients, the bug this PR fixes is still there, plus a failed request on every basket open.

Can we drop this hook and compute the check in `AddDrugOrder` instead? We can call `useMedicationOrders(patientUuid)` once there, derive the active and future drug UUIDs, and pass them to search, browse, the form, and the favorites extension through slot state. That reuses the configured drug order type and the existing active and future semantics, avoids duplicate requests with different cache keys, and lets every ordering path handle loading and errors consistently. It's less code overall.

packages/esm-patient-orders-app/src/order-favorites/drug-favorites-list.component.tsx
```diff
@@ -16,51 +16,69 @@ interface DrugFavoritesListExtensionProps {
...
+          disabled={alreadyPrescribed}
+          aria-disabled={alreadyPrescribed}
+        >
+          <div className={styles.itemContent}>
+            <p className={styles.itemTitle}>{favorite.displayName}</p>
+            {favorite.attributes.strength && <p className={styles.itemDetails}>{favorite.attributes.strength}</p>}
+          </div>
+          {alreadyPrescribed && (
+            <Tag type="green" size="sm">
+              {t('drugAlreadyPrescribed', 'Already prescribed')}
```
denniskigen, line 51: These strings need to be extracted. `drugAlreadyPrescribedNotificationBody` isn't in this app's `en.json`, and `drugAlreadyPrescribed` isn't in the orders app's. Running `yarn turbo run extract-translations` from the repo root should pick both up.

packages/esm-patient-orders-app/src/order-favorites/drug-favorites-list.component.tsx
```diff
@@ -16,51 +16,69 @@ interface DrugFavoritesListExtensionProps {
...
   visit,
   daysDurationUnit,
+  patientUuid,
 }) => {
   const { t } = useTranslation();
   const { enableDrugOrderFavorites } = useConfig<ConfigObject>();
   const isTablet = useLayoutType() === 'tablet';
   const { favorites, error, isLoading, deleteMultipleFavorites } = useFavoritesActions();
 
+  const { prescribedDrugUuids } = useActiveDrugOrders(patientUuid);
```
denniskigen: Can we add focused tests for both ticket outcomes? The existing medication tests cover search-result behavior, while `drug-favorites.resource.test.ts` only covers add/remove. We should prove active and future favorites are tagged and cannot open the form, the duplicate warning appears in the form, and unresolved/error order state remains disabled.

denniskigen (changes requested):
Good start, @abheeee03. Thanks for picking this up. There's one blocking issue with the new favorites order fetch that I've left inline, plus a few smaller things.

packages/esm-patient-orders-app/src/order-favorites/drug-favorites-list.component.tsx
```diff
@@ -16,51 +16,69 @@ interface DrugFavoritesListExtensionProps {
...
-  );
-});
+    return (
+      <div className={styles.favoriteItem}>
+        <button
+          type="button"
+          className={styles.itemButton}
+          onClick={() => onClick(favorite)}
+          disabled={alreadyPrescribed}
+          aria-disabled={alreadyPrescribed}
```
denniskigen: Nit: `aria-disabled` is redundant when `disabled` is set, and the `prescribedDrugUuids.has()` guard in `handleFavoriteClick` cannot be reached once the button is disabled by the same condition. Fine to keep just the `disabled` prop.

packages/esm-patient-orders-app/src/order-favorites/drug-favorites-list.component.tsx
denniskigen, line 51: Can we get a screenshot of a disabled pinned item with the tag? The PR screenshot only shows the form warning.

abheeee03: here it is, updating the same in PR.
Add drug order

packages/esm-patient-orders-app/src/order-favorites/drug-favorites.resource.ts
denniskigen: Thanks for moving the medication-order lookup into `AddDrugOrder`. This is still open because `DrugFavoritesListExtension` does not accept `prescribedDrugUuids` or `isLoadingOrders`; it still calls `useActiveDrugOrders(patientUuid)`, so the new slot state is ignored and the failing unfiltered request remains in use.

Can we finish the wiring by accepting the shared values in the extension, removing `patientUuid` and `useActiveDrugOrders`, and deleting that hook? It’s probably better to pass the error separately too, rather than combining it with loading.

packages/esm-patient-medications-app/src/add-drug-order/drug-search/drug-search.component.tsx
```diff
@@ -72,7 +76,15 @@ export default function DrugSearch({
...
       <ExtensionSlot
         name="drug-search-slot"
-        state={{ openOrderForm, isSearching: Boolean(debouncedSearchTerm), visit, daysDurationUnit }}
+        state={{
+          openOrderForm,
+          isSearching: Boolean(debouncedSearchTerm),
+          visit,
+          daysDurationUnit,
+          patientUuid: patient?.id,
+          prescribedDrugUuids,
```
denniskigen, line 86: Can we add coverage for this extension boundary? The `AddDrugOrder` tests mock `ExtensionSlot` to `null`, so they cannot verify that the favorites extension consumes this state. A component test that renders `DrugFavoritesListExtension` with these values and checks prescribed, loading, and error behavior would catch this wiring gap.

packages/esm-patient-orders-app/src/order-favorites/drug-favorites-list.component.tsx
denniskigen: This is another screenshot of the form warning. Can we get the pinned orders list with the disabled item and its “Already prescribed” tag visible?

abheeee03: the pinned orders list screenshot was already attached earlier in the PR. It shows the disabled pinned item with the “Already prescribed” tag; please let me know if anything needs to be changed.

file-01594e3fd56e25b235b21e7094083562

packages/esm-patient-orders-app/src/order-favorites/drug-favorites-list.component.tsx
```diff
@@ -16,45 +16,66 @@ interface DrugFavoritesListExtensionProps {
...
-    </div>
-  );
-});
+    return (
+      <div className={styles.favoriteItem}>
+        <button
+          type="button"
+          className={styles.itemButton}
+          onClick={() => onClick(favorite)}
+          disabled={isLoadingOrders || hasOrdersError || alreadyPrescribed}
```
denniskigen: Can we handle the medication-order request at the section level instead of per item? [`LabResults`](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-orders-app/src/lab-results/lab-result.component.tsx) is the pattern for section-level handling of dependent requests. The important distinction here is that the favorites themselves stay valid when the order lookup fails: only ordering them becomes unsafe, and unpinning still works. So the list shouldn't disappear. Concretely:

- Loading: show the section skeleton.
- Error: show a section-level `InlineNotification`, keep the favorites visible with ordering disabled, and keep unpin available.
- Loaded: show the normal list and disable only already-prescribed drugs.

That explains every reachable state without discarding valid data, and it lets us drop the per-item `isLoadingOrders` prop. Keep the `favorites.length === 0` early return above the new branches so users with nothing pinned don't see a skeleton or an error for an empty section.

Relatedly, `.itemButton` needs a `:disabled` style. `.itemTitle` sets an explicit color and the button keeps `cursor: pointer`, so a disabled favorite currently looks identical to an enabled one and silently ignores clicks:

```scss
&:disabled {
  cursor: not-allowed;

  .itemTitle,
  .itemDetails {
    color: colors.$gray-40;
  }
}
```

packages/esm-patient-orders-app/src/order-favorites/drug-favorites-list.component.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+    const favoriteButton = screen.getByRole('button', { name: /aspirin 81mg/i });
+    expect(favoriteButton).toBeDisabled();
+    expect(screen.queryByText(/already prescribed/i)).not.toBeInTheDocument();
+
+    await user.click(favoriteButton);
+    expect(defaultProps.openOrderForm).not.toHaveBeenCalled();
+  });
+
+  it('disables favorites when the shared order lookup fails', async () => {
```
denniskigen: These two tests would then assert the new behavior: no favorite buttons while the lookup is loading; on failure, the notification plus the favorites still rendered but disabled, with unpin still enabled.

packages/esm-patient-orders-app/src/order-favorites/drug-favorites-list.component.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+    mockUseFavoritesActions.mockReturnValue({
+      favorites: [favorite],
+      error: undefined,
+      isLoading: false,
+      deleteMultipleFavorites: vi.fn(),
+      persistFavorites: vi.fn(),
+    });
+  });
+
+  it('marks and disables a favorite supplied as already prescribed', async () => {
```
denniskigen, line 41: nit: these all cover the disabled paths. Can we add one that clicks an enabled favorite and asserts `openOrderForm` is called? Without it, a regression that disables every favorite would still pass this suite. Fine to ignore or do as a follow-up.

denniskigen (changes requested):
Thanks for finishing the wiring and adding the tests, @abheeee03 - the new tests pass for me locally and the failing favorites fetch is gone. This is very close. I've left a suggestion on how the pinned section should present the order lookup's loading and error states, plus a couple of notes on the tests.

packages/esm-patient-orders-app/src/order-favorites/drug-favorites-list.component.tsx
abheeee03: thanks for the guidance!
image

denniskigen (approved):
LGTM. Thanks for working through the loading and error states, @abheeee03!

PR #3601, opened 2026-08-26
(feat) O3-5934: Map fhir person-attribute extensions to patient attributes in form-entry

donaldkibet:
## Requirements

## Summary
This PR exposes OpenMRS person attributes on the patient object available to O3 Form Engine expressions.

Previously, form logic could access only a limited set of patient properties, such as age and gender. Person attributes—including phone number, health facility, and marital status—were unavailable for conditional rendering, calculated values, and validation. With this change, form authors can reference person attributes alongside existing patient properties, for example

```json
{
    "hide": {
        "hideWhenExpression": "attributes['Telephone contact'] !== ''"
    }
}
```

## Related Issue
 https://openmrs.atlassian.net/browse/O3-5934

### DIFF
--- a/packages/esm-form-entry-app/src/app/form-data-source/form-data-source.service.ts
+++ b/packages/esm-form-entry-app/src/app/form-data-source/form-data-source.service.ts
@@ -39,6 +39,8 @@ function getConceptCode(concept: Concept, conceptSourceUuid: string): string | u
 
 @Injectable()
 export class FormDataSourceService {
+  private readonly personAttributeUrl = 'http://fhir.openmrs.org/ext/person-attribute';
+
   constructor(
     private providerResourceService: ProviderResourceService,
     private locationResourceService: LocationResourceService,
@@ -388,7 +390,7 @@ export class FormDataSourceService {
         return 'U';
     }
   }
-  public getPatientObject(patient): PatientModel {
+  public getPatientObject(patient: fhir.Patient): PatientModel {
     return {
       ...patient,
       patientUuid: patient.id,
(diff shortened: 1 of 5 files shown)


packages/esm-form-entry-app/src/app/form-data-source/form-data-source.service.ts
```diff
@@ -397,9 +399,29 @@ export class FormDataSourceService {
       age: this.calculateAge(patient.birthDate),
       gendercreatconstant: patient.gender === 'female' ? 0.85 : patient.gender === 'male' ? 1 : undefined,
       identifiers: this.mapFHIRPatientIdentifiersToOpenMRSIdentifiers(patient),
+      attributes: this.mapFHIRPatientExtensionsToPersonAttributes(patient),
     };
   }
 
+  private mapFHIRPatientExtensionsToPersonAttributes(patient: fhir.Patient): Record<string, string> {
+    const extensions = patient?.extension?.filter((ext) => ext.url?.startsWith(this.personAttributeUrl)) ?? [];
```
ibacher: The URL here should be an exact match. These URLs are system identifiers so any sub-URL or even an unexpected hash indicates a different meaning.

packages/esm-form-entry-app/src/app/form-data-source/form-data-source.service.ts
```diff
@@ -397,9 +399,29 @@ export class FormDataSourceService {
...
+      if (attributeType && value) {
+        attributes[attributeType] = value;
+      }
+
+      return attributes;
+    }, {});
+  }
+
+  private getNestedExtensionValue(parent: fhir.Extension, name: string): string | undefined {
+    return parent.extension?.find((ext) => ext.url?.endsWith(name))?.valueString;
```
ibacher: Similarly here... Since URLs are identifiers we should be checking explicitly for `http://fhir.openmrs.org/ext/{name}`

ibacher (commented):
Basically LGTM with two exactness clamps because the "URLs" in this context are strict identifiers.

ibacher (approved):

PR #3605, opened 2026-08-28
(fix) Patient commons lib should import from framework not styleguide

ibacher:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/en-US/docs/frontend-modules/contributing/#contributing-guidelines) label. See existing PR titles for inspiration.
- [ ] My work is based on designs, which are linked or shown either in the Jira ticket or the description below. (See also: [Styleguide](http://om.rs/o3ui))
- [x] My work includes tests or is validated by existing tests.

## Summary

`@openmrs/esm-patient-commons-lib` accidentally imports directly from esm-styleguide instead of esm-framework, which results in duplication of framework code in the patient-commons-lib.

PR #3607, opened 2026-08-28
(fix) Clean build outputs before bundle size comparison

denniskigen:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a conventional commit label.
- [ ] My work is based on designs. Not applicable because this is a CI-only change.

## Summary

The compressed-size action builds the PR and base revisions sequentially in the same workspace. When the base build is restored from the Turborepo cache, Rspack does not run and stale PR chunks remain in each package `dist` directory. This causes renamed chunks to be counted as removed instead of replaced and substantially inflates the reported bundle-size reduction.

This change uses the action-supported `clean-script` hook to run `clean:dist` between measurements. Each of the 22 buildable packages owns a `clean` script scoped to its ignored `dist` directory, while the root script orchestrates those package tasks through Turborepo. The Turborepo `clean` task has caching disabled so cleanup always executes.

The bundle-size workflow validated that all 22 cleanup tasks ran successfully and now reports a 0 B size change for this CI-only PR.

## Related Issue

Follow-up to #3605.

## Other


ibacher:
I like the idea. I know this is more work, but I think it would be preferential to add a `clean` script to each package's `package.json`?

ibacher (approved):
Suggested a different implementation, but this also works.

denniskigen:
Thanks, @ibacher. Updated this to make cleanup package-owned: each of the 22 buildable packages now defines a `clean` script, and the root script runs them through a non-cacheable Turborepo task. The bundle-size workflow confirms all 22 cleanup tasks ran successfully and now reports a 0 B change.

PR #3616, opened 2026-09-02
Speckit setup and 1MB attachment limit increase

jonesg504:


ibacher:
Thanks for the contribution! However, there's no way we can take this PR with this much cruft into this repo. While SpecKit can be great, it's not something we've adopted as a community. There are also a bunch of what look like UI fixes for a screen-size below our supported sizes and then a dialing up of a configuration default.

Unfortunately, the configuration default by itself isn't the only barrier here; we don't yet have a unified setting to manage this; it's set as low as it is not because we think a 1 MB photo is the ideal size, but because that's the minimum size allowed by all of: our nginx configuration, our default Tomcat configuration, and [the maximum size our attachments module allows by default](https://github.com/openmrs/openmrs-module-attachments/blob/86251a3f03e93338cdf1f312a4f9a9642d84018f/omod/src/main/resources/config.xml#L93). Happy to discuss how we can work to get a better default, though!

jonesg504:
So sorry! Did not mean to open a PR in this repository! Just experimenting with my own fork! I will make sure this doesn't happen in the future!

PR #3619, opened 2026-09-03
(fix) O3-5953: Use the active or selected visit when renewing medication orders

donaldkibet:
## Requirements

## Summary
When submitting a batch of orders on a new encounter, the encounter datetime was pinned to the *earliest* explicit `dateActivated`. Because a RENEW order can be batched alongside (or carry) a previous order's activation date, this backdated the new encounter to the original prescription date instead of the renewal date.

This flips the logic to pin the encounter to the *latest* explicit `dateActivated`. Any earlier orders in the same batch are raised to the encounter datetime, preserving the backend invariant `dateActivated >= encounterDatetime`.

## Screenshots
https://github.com/user-attachments/assets/21e25a20-ffe8-427c-930f-b555771009f2

### DIFF
--- a/packages/esm-patient-chart-app/src/visit/visit-prompt/start-visit-dialog.modal.tsx
+++ b/packages/esm-patient-chart-app/src/visit/visit-prompt/start-visit-dialog.modal.tsx
@@ -8,6 +8,7 @@ import styles from './start-visit-dialog.scss';
 interface StartVisitDialogProps {
   patientUuid: string;
   closeModal: () => void;
+  onCancel?: () => void;
   launchPatientChart?: boolean;
   onVisitStarted?: () => void;
 }
@@ -15,10 +16,12 @@ interface StartVisitDialogProps {
 const StartVisitDialog: React.FC<StartVisitDialogProps> = ({
   patientUuid,
   closeModal,
+  onCancel,
   launchPatientChart,
   onVisitStarted,
 }) => {
   const { t } = useTranslation();
+  const handleCancel = onCancel ?? closeModal;
 
   const handleStartNewVisit = useCallback(() => {
     if (launchPatientChart) {
@@ -43,14 +46,14 @@ const StartVisitDialog: React.FC<StartVisitDialogProps> = ({
 
   return (
     <div>
-      <ModalHeader closeModal={closeModal}>
+      <ModalHeader closeModal={handleCancel}>
         <span className={styles.header}>{modalHeaderText}</span>
       </ModalHeader>
       <ModalBody>
         <p className={styles.body}>{modalBodyText}</p>
       </ModalBody>
       <ModalFooter>
-        <Button kind="secondary" onClick={closeModal}>
+        <Button kind="secondary" onClick={handleCancel}>
           {t('cancel', 'Cancel')}
         </Button>
         <Button kind="primary" onClick={handleStartNewVisit}>
--- a/packages/esm-patient-medications-app/src/api/api.test.ts
+++ b/packages/esm-patient-medications-app/src/api/api.test.ts
@@ -260,18 +260,6 @@ describe('buildMedicationOrder', () => {
     expect(result.scheduledDate).not.toBe(medicationOrder.dateActivated);
   });
 
-  it.each(['NEW', 'RENEW'] as const)('does not copy the previous encounter onto a %s basket item', (action) => {
-    const result = buildMedicationOrder(medicationOrder, action);
-
-    expect(result.encounterUuid).toBeUndefined();
-  });
-
-  it.each(['REVISE', 'DISCONTINUE'] as const)('copies the previous encounter onto a %s basket item', (action) => {
-    const result = buildMedicationOrder(medicationOrder, action);
-
-    expect(result.encounterUuid).toBe(medicationOrder.encounter.uuid);
-  });
-
   it('defaults REVISE basket items to now instead of the previous activation or scheduled date', () => {
     const before = Date.now();
     const scheduledDate = '2026-05-01T00:00:00.000+0000';
--- a/packages/esm-patient-medications-app/src/api/api.test.ts
+++ b/packages/esm-patient-medications-app/src/api/api.test.ts
@@ -260,6 +260,18 @@ describe('buildMedicationOrder', () => {
     expect(result.scheduledDate).not.toBe(medicationOrder.dateActivated);
   });
 
+  it.each(['NEW', 'RENEW'] as const)('does not copy the previous encounter onto a %s basket item', (action) => {
+    const result = buildMedicationOrder(medicationOrder, action);
+
+    expect(result.encounterUuid).toBeUndefined();
+  });
+
+  it.each(['REVISE', 'DISCONTINUE'] as const)('copies the previous encounter onto a %s basket item', (action) => {
+    const result = buildMedicationOrder(medicationOrder, action);
+
+    expect(result.encounterUuid).toBe(medicationOrder.encounter.uuid);
+  });
+
   it('defaults REVISE basket items to now instead of the previous activation or scheduled date', () => {
     const before = Date.now();
     const scheduledDate = '2026-05-01T00:00:00.000+0000';
--- a/packages/esm-patient-medications-app/src/components/medications-details-table.component.tsx
+++ b/packages/esm-patient-medications-app/src/components/medications-details-table.component.tsx
@@ -441,7 +441,7 @@ function OrderBasketItemActions({
     launchWorkspace2<{}, OrderBasketWindowProps, PatientWorkspaceGroupProps>(
       'order-basket',
       {},
-      { encounterUuid: medication.encounter.uuid },
+      { encounterUuid: '' },
       workspaceGroupProps,
     );
   }, [items, setItems, medication, workspaceGroupProps]);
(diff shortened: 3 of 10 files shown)


ibacher:
This still seems to me like it's covering over a bug in the renewals process. Renewals should have their own activated date for when the renewal itself is activated. It shouldn't be carrying the "previous orders" date.

makombe:
> This still seems to me like it's covering over a bug in the renewals process. Renewals should have their own activated date for when the renewal itself is activated. It shouldn't be carrying the "previous orders" date.


@donaldkibet I think if I get @ibacher right, it should be same process like creating a new drug order but with status/order action `RENEW`.

denniskigen:
I think @makombe's reading is right. A renewal should follow the same path as a new order, with `action: RENEW` and `previousOrder` pointing to the original order. It already gets its own start date on `main`: `buildMedicationOrder` defaults an immediate renewal to now, and the payload sends that as `dateActivated`. Future renewals use `scheduledDate`.

It’s probably better to fix this at the renewal launch point. Renew currently opens the basket using the original order’s visit. Can we route it through the same current-visit-aware path used by Add? That’s the remaining mismatch I can see.

We shouldn’t change `earliestDateActivated` or rewrite other orders’ dates. The earliest/latest swap doesn’t change the single-renewal case described by the ticket: both versions clamp the encounter datetime to the old visit’s stop, or use its start when no order date is set. In a mixed basket, the branch can rewrite an earlier explicit NEW order date to the latest activation date.

`postOrdersOnNewEncounter` can only use the visit it is given, so I think the regression test belongs at the Renew launch point. Can we assert that Renew passes the current visit context to the basket, then cover the full workflow in e2e if needed?

makombe:
@ibacher are you able to review this again? Some adjustments have been made. Thanks

packages/esm-patient-medications-app/src/components/medications-details-table.component.tsx
```diff
@@ -441,7 +441,7 @@ function OrderBasketItemActions({
     launchWorkspace2<{}, OrderBasketWindowProps, PatientWorkspaceGroupProps>(
       'order-basket',
       {},
-      { encounterUuid: medication.encounter.uuid },
+      { encounterUuid: '' },
       workspaceGroupProps,
```
denniskigen: The visit-context issue from the earlier review is still here: `workspaceGroupProps` contains `medication.encounter.visit`, so clearing `encounterUuid` still creates the renewal encounter in the original visit.

I reproduced the effect in the local distro using synthetic patient data: renewing an April prescription with an active visit today gives the renewal today’s start date, but it is missing from Active prescriptions in dispensing. All prescriptions shows it with the April encounter date.

https://github.com/user-attachments/assets/a6ff1ab4-230d-485f-9c8b-50979c048f73

Can we use `useStartVisitIfNeeded(patient.id)` here and launch without the medication’s group props, so the basket keeps the chart’s current or deliberately selected visit? Add the renewal only after visit selection succeeds, so cancelling the prompt leaves the basket unchanged.

packages/esm-patient-medications-app/src/api/api.test.ts
```diff
@@ -260,6 +260,18 @@ describe('buildMedicationOrder', () => {
     expect(result.scheduledDate).not.toBe(medicationOrder.dateActivated);
   });
 
+  it.each(['NEW', 'RENEW'] as const)('does not copy the previous encounter onto a %s basket item', (action) => {
+    const result = buildMedicationOrder(medicationOrder, action);
+
+    expect(result.encounterUuid).toBeUndefined();
```
denniskigen: These cover the basket item’s encounter UUID, but not the visit used when signing. Can we add a Renew interaction test with an old closed visit and a current visit, asserting that the renewal encounter uses the current visit? Please also cover cancelling visit selection, which should add nothing to the basket.

For end-to-end validation, use an original encounter older than dispensing’s cutoff, 90 days by default, and verify that the renewal appears in Active prescriptions and can be dispensed.

denniskigen (dismissed):
Thanks for removing the date rewrite, @donaldkibet. The visit-context issue from the earlier review is still open; I’ve left follow-ups on the launch path and regression coverage. Can we also update the title and summary? They still describe the earliest/latest date change that’s now been removed.

packages/esm-patient-medications-app/src/api/api.test.ts
donaldkibet: [Kooha-2026-09-10-17-09-56.webm](https://github.com/user-attachments/assets/8da237bc-2f12-40ad-88cf-595aa53469e1)
cc @denniskigen updated

packages/esm-patient-chart-app/src/visit/visit-prompt/start-visit-dialog.modal.tsx
chibongho, line 34: As it's possible to also do cancel on the start visit workspace, I think we need to pass the `onCancel` callback into it and call it when the user clicks cancel there.

packages/esm-patient-medications-app/src/components/medications-details-table.component.tsx
```diff
@@ -436,15 +438,16 @@ function OrderBasketItemActions({
     );
   }, [medication, workspaceGroupProps]);
 
-  const handleRenewClick = useCallback(() => {
+  const handleRenewClick = useCallback(async () => {
+    const canProceed = await startVisitIfNeeded();
+    if (!canProceed) {
+      setItems([]);
```
ibacher, line 444: It doesn't seem _great_ to clear the order basket completely here...

packages/esm-patient-orders-app/src/utils/index.ts
```diff
@@ -99,7 +99,7 @@ export function buildMedicationOrder(order: Order, action?: OrderAction): DrugOr
           valueCoded: order.quantityUnits.uuid,
         }
       : null,
-    encounterUuid: order.encounter?.uuid,
+    encounterUuid: action === 'REVISE' || action === 'DISCONTINUE' ? order.encounter?.uuid : undefined,
```
ibacher, line 102: This, and the other, similar line, seem like they're just _always_ going to create a new encounter for renewals, but ideally, wouldn't we be using the existing orders encounter from the visit?

packages/esm-patient-medications-app/src/components/medications-details-table.component.tsx
denniskigen: Agreed. Cancelling here should leave the basket exactly as it was, since the renewal was never added:
```suggestion
    if (!canProceed) {
      return;
    }
```
It'd be good to have the cancel test start with an unrelated item already in the basket and assert it's still there afterwards.

packages/esm-patient-orders-app/src/utils/index.ts
denniskigen: These fields don't determine which encounter the basket uses. With visits enabled, it already creates a new encounter; with visits disabled, it can reuse today's encounter independently of the item's `encounterUuid`. It's probably better to revert these unrelated changes and keep this PR focused on visit selection. Encounter reuse can be discussed separately.

packages/esm-patient-chart-app/src/visit/visit-prompt/start-visit-dialog.modal.tsx
denniskigen: Related: `visit-form.workspace.tsx` only reads `openedFrom` and `showPatientHeader` from the workspace props, so the supplied `onVisitStarted` is dropped too. Saving the new visit leaves `startVisitIfNeeded` unresolved and Renew never continues. The wrapper issue predates this PR, but the new Renew flow now depends on it. Can we wire both callbacks through and test successful visit creation as well as cancellation?

denniskigen: Thanks for wiring `onCancel` through, @ibacher. The wrapper still drops the caller's `onVisitStarted`, so Renew keeps waiting after the visit saves. #3688 forwards it and also waits to resume the action until the workspace group relaunches, which otherwise closes the basket. It's probably simplest to leave the success path to #3688 and rebase whichever lands second.

denniskigen (commented):
Thanks for preserving the basket on cancellation and adding regression coverage, @ibacher. #3688 handles the remaining success path, so we should coordinate the two PRs and rebase whichever lands second.

packages/esm-patient-medications-app/src/components/medications-details-table.test.tsx
```diff
@@ -0,0 +1,275 @@
...
+  test('discontinuing an order opens the basket against the order’s own encounter and visit', async () => {
+    const user = userEvent.setup();
+    renderMedicationsDetailsTable();
+
+    await clickMenuItem(user, 'Discontinue');
+
+    expect(mockLaunchWorkspace2).toHaveBeenCalledWith(
+      'order-basket',
+      {},
+      { encounterUuid: medicationFixture.encounter.uuid },
```
denniskigen: Is this what we want for Discontinue? With visits enabled, it also calls `postOrdersOnNewEncounter` with the order’s visit, so this preserves the same old-visit behavior we’re fixing for Renew. It’s probably better to leave that visit-context assertion out and track Discontinue separately.

packages/esm-patient-chart-app/src/visit/visit-prompt/start-visit-dialog.modal.tsx
ibacher: Yeah, that's fair.

packages/esm-patient-medications-app/src/components/medications-details-table.test.tsx
ibacher: No. This is very much the wrong behavior. It's here because my commit undid the line that blank the encounter on discontinue or cancel.

PR #3621, opened 2026-09-04
(feat) O3-5823: Capture diagnosis order and certainty in the visit note

VeronicaMuthee:
## Requirements

## Summary

The visit note form previously hardcoded `certainty: 'PROVISIONAL'` on every diagnosis it saved, and encoded rank in which of two search boxes (primary/secondary) the clinician typed into. Neither rank nor certainty was ever an explicit clinical choice, and neither was visible anywhere after saving.

This PR makes both values deliberate on entry and visible on display:

- **One diagnosis search box** replaces the primary/secondary pair. Each added diagnosis renders as a compact row with two exception checkboxes — **Primary** and **Confirmed**. Unticked, a diagnosis is recorded as secondary and provisional, and the row spells those presumed values out ("Secondary · Provisional"), so an untouched diagnosis is immediately saveable and its recorded state is always explicit. (The diagnosis controls went through design iteration during review: per-diagnosis toggle groups with forced explicit choice were replaced by this defaults-plus-exceptions model as too heavy for the highest-volume form.)
- **When a primary diagnosis is required** (`isPrimaryDiagnosisRequired`, default on), a newly added diagnosis defaults to Primary whenever the note has none — so a single-diagnosis note needs no extra clicks. The requirement is stated in the helper text up front, and the requirement is validated at the group level: the message renders once below the diagnosis list (not beneath the search input), the checkboxes stay neutral, ticking any Primary clears it live, and a failed save focuses the first Primary checkbox.
- **Editing pre-selects stored values.** Rank or certainty values outside the known enums (possible from other REST writers) fall back to the same secondary/provisional presumption rather than being silently rewritten; the `mapEncounter` `|| 'PROVISIONAL'` fallback is removed for the same reason. Previously, diagnoses with rank outside {1,2} were silently *deleted* on edit. Non-coded diagnoses now survive the edit round-trip, and rows are keyed by a client-side draft id so duplicate concepts or non-coded entries no longer collide.
- **Visit-level aggregation**: the visits-widget representation now fetches `certainty`, and `dedupeDiagnoses` merges duplicates across a visit's encounters with documented, order-independent rules — rank = minimum across duplicates; certainty by deterministic precedence (CONFIRMED > PROVISIONAL > other stored values > unset, lexicographic tie-break) — returning a distinct `DedupedDiagnosis` type so the display aggregate cannot be mistaken for an encounter diagnosis.
- **Display defers to the styleguide:** the visits timeline and visit summary continue to use the styleguide's DiagnosisTags; certainty display lights up with no further patient-chart changes once DiagnosisTags gains certainty support in a follow-up esm-core PR.

Notes for reviewers:

- This is a behavior change to the diagnosis section of the visit note form: single search box, explicit-and-visible rank/certainty, and secondary/provisional as stated defaults. The wire format is unchanged (`DiagnosisPayload` with `rank` and `certainty`).
- Several translation keys were removed/renamed (the `primaryDiagnosisSearch`-era keys), which will affect distro translation overrides; `en.json` is the exact output of `yarn extract-translations`.
- Testing: 63 unit tests across the notes app (validation and auto-primary behavior, presumed-values display, edit round-trips including CONFIRMED preservation through the delete-and-recreate path, multi-row independence, out-of-enum fallback) and 164 across the chart app's visit suites (including the deterministic dedupe rules); the e2e spec covers the new flow end to end — auto-primary on the first diagnosis, presumed defaults on the second, and a new assertion that diagnoses survive the note-edit round-trip.

### Before
image

### After
https://github.com/user-attachments/assets/0c2e1c76-5dbd-42a6-9f6d-7529dc3a7d6b

## Related Issue
https://openmrs.atlassian.net/browse/O3-5823


packages/esm-patient-chart-app/src/visit/visit-diagnosis-tags.component.tsx
```diff
@@ -0,0 +1,66 @@
...
+          diagnosis.rank === 1 ? diagnosisTags?.primaryColor ?? 'red' : diagnosisTags?.secondaryColor ?? 'blue';
+        const tag = (
+          <span className={classNames('cds--tag', 'cds--tag--md', 'cds--layout--size-md', `cds--tag--${color}`)}>
+            <span className="cds--tag__label">{diagnosis.display}</span>
+            {certaintyLabel && <span className={styles.visuallyHidden}>{` (${certaintyLabel})`}</span>}
+          </span>
+        );
+
+        return certaintyLabel ? (
+          <span key={diagnosis.uuid} className={styles.certaintyTooltip} data-certainty={certaintyLabel} tabIndex={0}>
```
denniskigen: Can we show certainty alongside the diagnosis name? The hover/focus tooltip isn’t reliably available through touch. It also replaces the full-name tooltip, so clinicians can no longer reveal the complete name when a diagnosis is truncated.

packages/esm-patient-notes-app/src/notes/selected-diagnosis-card.component.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <span className={styles.diagnosisCardTitle}>{diagnosis.display}</span>
+      <div className={styles.diagnosisCardControls}>
+        <Checkbox
+          checked={diagnosis.rank === 1}
+          id={`diagnosis-${diagnosis.draftId}-primary`}
+          labelText={t('primary', 'Primary')}
+          onChange={(_, { checked }) => onUpdate(diagnosis, { rank: checked ? 1 : 2 })}
+        />
+        <Checkbox
+          checked={diagnosis.certainty === 'CONFIRMED'}
```
denniskigen, line 69: Can we show “Secondary” and “Provisional” on the row when those are the recorded values? The unchecked boxes don’t explicitly show what will be saved. Clinicians have to refer back to the helper text, including when reviewing an existing note.

denniskigen (commented):
Thanks, @VeronicaMuthee. The single search and per-diagnosis controls make this easier to work through. I’ve left a few suggestions around making the recorded state clearer and keeping it accessible on tablets.

Can we update the summary to match the current implementation? It still describes unset radio groups and a separate certainty-display follow-up. The code now uses checkboxes with secondary/provisional defaults and includes the timeline tooltip.

packages/esm-patient-notes-app/src/notes/selected-diagnosis-card.component.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export default function SelectedDiagnosisCard({ diagnosis, onRemove, onUpdate }: SelectedDiagnosisCardProps) {
+  const { t } = useTranslation();
+
+  return (
+    <div className={styles.diagnosisCard} role="group" aria-label={diagnosis.display}>
+      <span className={styles.diagnosisCardTitle}>{diagnosis.display}</span>
+      <div className={styles.diagnosisCardControls}>
+        <Checkbox
+          checked={diagnosis.rank === 1}
```
denniskigen, line 62: When a primary diagnosis is required, can we default the first diagnosis added to Primary and let the clinician change it? This would avoid an extra step for a note with one diagnosis. Can we also explain the requirement before save and show any validation beside the Primary controls, rather than beneath the search input?

packages/esm-patient-notes-app/src/notes/selected-diagnosis-card.component.tsx
```diff
@@ -0,0 +1,65 @@
...
+        />
+        <Checkbox
+          checked={diagnosis.certainty === 'CONFIRMED'}
+          id={`diagnosis-${diagnosis.draftId}-confirmed`}
+          labelText={t('confirmed', 'Confirmed')}
+          onChange={(_, { checked }) => onUpdate(diagnosis, { certainty: checked ? 'CONFIRMED' : 'PROVISIONAL' })}
+        />
+        <Button
+          hasIconOnly
+          iconDescription={t('removeDiagnosisNamed', 'Remove {{diagnosis}}', { diagnosis: diagnosis.display })}
```
denniskigen, line 76: Can we keep the remove-button tooltip within the workspace? It currently gets clipped at the right edge, cutting off part of the diagnosis name.

CleanShot 2026-09-11 at 11 54 18@2x

packages/esm-patient-notes-app/src/notes/visit-notes-form.component.tsx
```diff
@@ -651,15 +559,41 @@ const VisitNotesForm: React.FC<VisitNotesFormProps> = ({
...
+                        <SelectedDiagnosisCard
+                          key={diagnosis.draftId}
+                          diagnosis={diagnosis}
+                          onRemove={handleRemoveDiagnosis}
+                          onUpdate={handleUpdateDiagnosis}
+                        />
+                      ))}
+                    </>
+                  ) : (
+                    <p className={styles.diagnosesText}>
```
denniskigen, line 607: nit: Can we add a little space above “No diagnosis selected”? It currently sits directly against the search input’s bottom border.

packages/esm-patient-chart-app/src/visit/visit-diagnosis-tags.scss
```diff
@@ -0,0 +1,56 @@
+@use '@carbon/colors';
+@use '@carbon/layout';
+@use '@carbon/type';
+
+.container {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
```
denniskigen, line 7: nit: Can we add a small horizontal gap between diagnosis pills? They look crowded in the timeline. The gap should separate the pills rather than increase their internal padding.

denniskigen:
Here’s a walkthrough of the UI issues covered in the review above:

https://github.com/user-attachments/assets/0a57c87b-15fd-43a0-b953-9b3133b33c60

packages/esm-patient-chart-app/src/visit/visit-diagnosis-tags.component.tsx
VeronicaMuthee: > Can we show certainty alongside the diagnosis name? The hover/focus tooltip isn’t reliably available through touch. It also replaces the full-name tooltip, so clinicians can no longer reveal the complete name when a diagnosis is truncated.

Done in 486ebbf3 — certainty now renders as visible text beside the name (Malaria (Provisional)), so it reaches touch devices, and the label's title is restored so truncated names are revealable again. The hover-only tooltip is removed as redundant.

packages/esm-patient-notes-app/src/notes/selected-diagnosis-card.component.tsx
VeronicaMuthee: > Can we show “Secondary” and “Provisional” on the row when those are the recorded values? The unchecked boxes don’t explicitly show what will be saved. Clinicians have to refer back to the helper text, including when reviewing an existing note.

Done in 486ebbf3 — each row now spells out the presumed values beneath the name ("Secondary · Provisional", shrinking to just "Provisional" once Primary is ticked, and disappearing when both exceptions are ticked). It's derived from the diagnosis state, so it also shows correctly when reviewing an existing note.

packages/esm-patient-notes-app/src/notes/selected-diagnosis-card.component.tsx
VeronicaMuthee: > When a primary diagnosis is required, can we default the first diagnosis added to Primary and let the clinician change it? This would avoid an extra step for a note with one diagnosis. Can we also explain the requirement before save and show any validation beside the Primary controls, rather than beneath the search input?

Done in 486ebbf3 — a newly added diagnosis defaults to Primary whenever the note has none (changeable, and it also re-applies if the primary is removed and another diagnosis is added), the helper text states the requirement up front, and the validation message + invalid state now render beside the Primary checkboxes and clear live when one is ticked. The search-input error styling and its focus-steal are removed.

packages/esm-patient-notes-app/src/notes/selected-diagnosis-card.component.tsx
VeronicaMuthee: > Can we keep the remove-button tooltip within the workspace? It currently gets clipped at the right edge, cutting off part of the diagnosis name.

Done in 486ebbf3 — the tooltip is end-aligned, so it opens inward from the workspace edge

packages/esm-patient-notes-app/src/notes/visit-notes-form.component.tsx
VeronicaMuthee: > nit: Can we add a little space above “No diagnosis selected”? It currently sits directly against the search input’s bottom border.

Done in 486ebbf3 — added a spacing-03 top margin so the empty-state line no longer sits against the search input's border.

packages/esm-patient-chart-app/src/visit/visit-diagnosis-tags.scss
VeronicaMuthee: > nit: Can we add a small horizontal gap between diagnosis pills? They look crowded in the timeline. The gap should separate the pills rather than increase their internal padding.

Done in 486ebbf3 — the tag container now uses a spacing-02 gap between pills (padding untouched). Screenshot of the timeline attached.

packages/esm-patient-chart-app/src/visit/dedupe-diagnoses.ts
```diff
@@ -1,24 +1,59 @@
...
+        merged.set(key, entry);
       }
-      seen.add(key);
-      return true;
-    });
+      result.push(entry);
+    } else {
+      existing.rank = minRank(existing.rank, rank);
+      existing.certainty =
+        existing.certainty === 'CONFIRMED' || certainty === 'CONFIRMED' ? 'CONFIRMED' : existing.certainty ?? certainty;
```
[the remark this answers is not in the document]
VeronicaMuthee: Done in fc6b086a — deterministic precedence CONFIRMED > PROVISIONAL > other stored values > unset, with a lexicographic tie-break; order-independence is unit-tested including the PRESUMED/PROVISIONAL case.

packages/esm-patient-notes-app/src/notes/selected-diagnosis-card.component.tsx
```diff
@@ -0,0 +1,165 @@
...
+            key={option.label}
+            onClick={option.onSelect}
+            onKeyDown={(event) => {
+              if (['ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown'].includes(event.key)) {
+                event.preventDefault();
+                const other = options[(index + 1) % options.length];
+                other.onSelect();
+                (event.currentTarget.parentElement?.children[(index + 1) % options.length] as HTMLElement)?.focus();
+              }
+            }}
```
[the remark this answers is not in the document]
VeronicaMuthee: Outdated — the radio toggle groups were replaced with checkboxes in f566a587, so this code no longer exists.

VeronicaMuthee:
@denniskigen, thanks for your review - I have addressed your comments. Please take another look.

denniskigen:
Looks like this has accrued conflicts, mind sorting them out?

VeronicaMuthee:
> Looks like this has accrued conflicts, mind sorting them out?

Resolved in 49e21385 — merged main and ported #3645's keyboard-access work into the redesigned diagnosis search (results are focusable buttons with arrow-key navigation and focus return) along with the Add/Edit workspace titles.

packages/esm-patient-notes-app/src/notes/selected-diagnosis-card.component.tsx
denniskigen: Having tried this with multiple diagnoses, I think the error belongs to the group. Can we show it once below the diagnosis list and leave unchecked Primary boxes neutral? Selecting any primary should clear it, and a failed save should focus the first Primary checkbox.

### Before
CleanShot 2026-09-14 at 22 41 10@2x

### After
CleanShot 2026-09-14 at 22 34 13@2x

packages/esm-patient-chart-app/src/visit/visit-diagnosis-tags.component.tsx
```diff
@@ -0,0 +1,66 @@
...
+        return (
+          <span
+            key={diagnosis.uuid}
+            className={classNames('cds--tag', 'cds--tag--md', 'cds--layout--size-md', `cds--tag--${color}`)}
+          >
+            {/* title keeps the full name revealable when a long diagnosis is truncated */}
+            <span className="cds--tag__label" title={diagnosis.display}>
+              {diagnosis.display}
+            </span>
+            {certaintyLabel && <span className={styles.certaintySuffix}>({certaintyLabel})</span>}
```
denniskigen: Can we prevent the certainty suffix from shrinking and use Carbon’s `Tooltip` on the whole pill? I tested this locally: `flex-shrink: 0` keeps certainty inside the tag, while the name truncates and remains available in full on hover and keyboard focus.

### Before

CleanShot 2026-09-14 at 22 39 51@2x

### After
CleanShot 2026-09-14 at 22 38 03@2x

packages/esm-patient-notes-app/src/notes/selected-diagnosis-card.component.tsx
```diff
@@ -0,0 +1,86 @@
...
+  // The values presumed by the unticked checkboxes, spelled out so the row always states
+  // what will be recorded without referring back to the helper text
+  const presumedValues = [
+    diagnosis.rank === 2 ? t('secondary', 'Secondary') : null,
+    diagnosis.certainty === 'PROVISIONAL' ? t('provisional', 'Provisional') : null,
+  ].filter(Boolean);
+
+  return (
+    <div className={styles.diagnosisCard} role="group" aria-label={diagnosis.display}>
+      <span className={styles.diagnosisCardTitle}>
```
denniskigen, line 47: nit: It’s probably better to give the diagnosis name the full row and put the controls underneath. Long names currently get squeezed between the controls. The two-row layout was easier to read at desktop and tablet widths.

denniskigen (commented):
Thanks, @VeronicaMuthee. The automatic primary selection and visible certainty make this clearer. I’ve left two remaining suggestions around group validation and long diagnosis tags.

packages/esm-patient-chart-app/src/visit/visit-diagnosis-tags.component.tsx
VeronicaMuthee: Done — the certainty suffix now has flex-shrink: 0 so it never collapses, and the whole pill shows the full diagnosis name in a tooltip. One deviation from the suggestion: plain Carbon Tooltip didn't survive real chart conditions — the pills sit inside .cds--data-table-content (overflow: auto), which clips any in-place popover, and the chart's transformed ancestors break autoAlign's fixed positioning. So the tooltip is a small body-portaled element positioned from the pill's bounding rect (9e369492, refined in 3cc4dc2e and 10b315dc). It shows on hover and keyboard focus, closes on scroll, and is aria-hidden since screen readers already get the full name from the pill text.

packages/esm-patient-notes-app/src/notes/selected-diagnosis-card.component.tsx
VeronicaMuthee: Done in 9e369492 — the diagnosis name now takes the full row (wrapping freely), with the Primary/Confirmed controls and remove button on a row underneath.

VeronicaMuthee:
@denniskigen, I have addressed the tooltip and two-row layout nit comments.

packages/esm-patient-notes-app/src/notes/selected-diagnosis-card.component.tsx
denniskigen: Thanks for the layout updates. I checked 10b315dc locally, and the validation message still appears above the diagnosis list, with both Primary checkboxes marked invalid. Can we address the group-validation suggestion above? The attached “After” screenshot shows the intended placement.

VeronicaMuthee: Done in 5d2340dd — the requirement is now group-level validation exactly as suggested: the message renders once below the diagnosis list, the unchecked Primary checkboxes stay neutral (the per-checkbox invalid state is removed), ticking any Primary clears it live, and a failed save moves focus to the first Primary checkbox (or to the search input when no diagnosis has been selected yet).

denniskigen:
Thanks for iterating on this, @VeronicaMuthee. The diagnosis controls and validation feedback are much clearer now.

Can we have the form track diagnosis changes and validation together? Currently these are handled separately, which makes it harder to keep the error message, Save button, and unsaved-changes warning consistent when someone edits or undoes a change.

[This prototype](https://gist.github.com/denniskigen/bc4d384c2d3cc607fe6ef0cb42df50e7) is a starting point. It makes `diagnoses` a react-hook-form value and puts the primary-required rule in the Zod schema. This removes the custom resolver and the separate flags for diagnosis edits and the displayed error. It also drops the "Error transforming diagnoses" guard, since core rejects a null diagnosis on save, and reads the encounter's diagnoses once when the form mounts.

The intended behaviour is:

- Show the primary-required error after someone tries to save, then update it as they change their selection.
- If someone undoes all their changes, disable Save and clear the unsaved-changes state.
- Preserve diagnosis certainty when saving and reopening a note, including after editing only its text.

All 23 form tests and the TypeScript check pass with this prototype.

VeronicaMuthee:
> Thanks for iterating on this, @VeronicaMuthee. The diagnosis controls and validation feedback are much clearer now.
> 
> Can we have the form track diagnosis changes and validation together? Currently these are handled separately, which makes it harder to keep the error message, Save button, and unsaved-changes warning consistent when someone edits or undoes a change.
> 
> [This prototype](https://gist.github.com/denniskigen/bc4d384c2d3cc607fe6ef0cb42df50e7) is a starting point. It makes `diagnoses` a react-hook-form value and puts the primary-required rule in the Zod schema. This removes the custom resolver and the separate flags for diagnosis edits and the displayed error. It also drops the "Error transforming diagnoses" guard, since core rejects a null diagnosis on save, and reads the encounter's diagnoses once when the form mounts.
> 
> The intended behaviour is:
> 
> * Show the primary-required error after someone tries to save, then update it as they change their selection.
> * If someone undoes all their changes, disable Save and clear the unsaved-changes state.
> * Preserve diagnosis certainty when saving and reopening a note, including after editing only its text.
> 
> All 23 form tests and the TypeScript check pass with this prototype.

Done in 967c8cd8 — applied your prototype as-is (thanks, it was a clean cut on 5d2340dd): diagnoses is now a form value with the primary-required rule as a Zod refine, replacing the custom resolver, the phantom-error cast, and the separate touched flag. Undoing all diagnosis edits now re-disables Save and clears the unsaved-changes guard, the encounter's diagnoses are read once at mount, and the transform guard + its translation key are removed. All 23 form tests pass and I've verified the three intended behaviours end to end on a local instance.

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/encounters-table.resource.ts
```diff
@@ -115,13 +115,10 @@ export function mapEncounter(encounter: Encounter): MappedEncounter {
...
-    diagnoses:
-      encounter.diagnoses
-        ?.filter((diagnosis) => !diagnosis.voided)
-        .map((diagnosis) => ({
-          ...diagnosis,
-          certainty: diagnosis.certainty || 'PROVISIONAL',
-        })) || [],
+    // Certainty is passed through as stored (no PROVISIONAL fallback): the visit note
+    // form renders a missing/unknown certainty as unset so the clinician chooses
+    // explicitly rather than a fabricated value being silently re-saved (O3-5823).
```
ibacher: I think long-term this comment can go. It describes how things were before this PR, but not in a way that I think we need to worry about. A not on the `diagnosis` type that a missing certainty should be interpreted as "PROVISIONAL" would be a better way of capturing the same.

packages/esm-patient-chart-app/src/visit/visit-diagnosis-tags.component.tsx
```diff
@@ -0,0 +1,123 @@
...
+    <div className={styles.container}>
+      {diagnoses.map((diagnosis) => {
+        // The aggregate certainty is a passthrough of stored values, so anything outside
+        // the known enum gets no label rather than a guessed one
+        const certaintyLabel =
+          diagnosis.certainty === 'CONFIRMED'
+            ? t('confirmed', 'Confirmed')
+            : diagnosis.certainty === 'PROVISIONAL'
+              ? t('provisional', 'Provisional')
+              : null;
```
ibacher: Shouldn't `null` be treated as provisional?

packages/esm-patient-chart-app/src/visit/visit-diagnosis-tags.component.tsx
```diff
@@ -0,0 +1,123 @@
...
+      return;
+    }
+    const close = () => setTooltipPosition(null);
+    window.addEventListener('scroll', close, true);
+    return () => window.removeEventListener('scroll', close, true);
+  }, [tooltipPosition]);
+
+  return (
+    <>
+      <span
```
ibacher: Is there a reason we aren't still using the existing Carbon tag here?

ibacher (commented):
Couple of nits. My overall concern here is that this is re-inserting a patient-chart specific diagnosis tag. We moved `DiagnosisTag` into the styleguide so we'd have a consistent display across the application and what's primarily added here (certainty to the tag, etc.) feels like information we want to display everywhere so it might be best to break this into two PRs: one that updates the `DiagnosisTag` in the styleguide and the other (this one) that updates the visit note to capture certainty and rank.

packages/esm-patient-chart-app/src/visit/visit-diagnosis-tags.component.tsx
VeronicaMuthee: Null will display as Provisional. Handling it in the style guide PR, since the local component is removed from this one.

packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/encounters-table/encounters-table.resource.ts
VeronicaMuthee: Removed. Agreed that the note belongs on the Diagnosis type — that's in esm-core, so I'll include it in the styleguide PR.

packages/esm-patient-chart-app/src/visit/visit-diagnosis-tags.component.tsx
VeronicaMuthee: Tag force-writes title only for string children, and its built-in truncation tooltip renders in place, where the visits table's overflow: auto clips it — that drove the custom markup. Moot now: the component is removed, and the styleguide PR will work within DiagnosisTags' existing Tag.

VeronicaMuthee:
> Couple of nits. My overall concern here is that this is re-inserting a patient-chart specific diagnosis tag. We moved `DiagnosisTag` into the styleguide so we'd have a consistent display across the application and what's primarily added here (certainty to the tag, etc.) feels like information we want to display everywhere so it might be best to break this into two PRs: one that updates the `DiagnosisTag` in the styleguide and the other (this one) that updates the visit note to capture certainty and rank.

Done in 131df216: this PR is now capture-only. The local tag component is removed and both call sites are back on the styleguide's DiagnosisTags; the certainty fetch and visit-level dedupe stay, so the display lights up with no further patient-chart changes once the styleguide PR lands. Opening the esm-core PR for DiagnosisTags next.

packages/esm-patient-chart-app/src/visit/visit-diagnosis-tags.component.tsx
VeronicaMuthee: > Can we show certainty alongside the diagnosis name?

@denniskigen, per Ian's review, certainty display is moved out of this PR into the style guide's DiagnosisTags (esm-core follow-up) — 131df216.

packages/esm-patient-chart-app/src/visit/visit-diagnosis-tags.component.tsx
VeronicaMuthee: > Can we prevent the certainty suffix from shrinking and use Carbon’s `Tooltip` on the whole pill?

@denniskigen, we've removed the pill component in 131df216 (display deferred to the style guide per Ian's review); the suffix/tooltip treatment moves to the esm-core PR.

VeronicaMuthee:
@ibacher @denniskigen, thanks both for your feedback, please review the revised PR

denniskigen:
Styleguide PR augmenting the existing DiagnosisTag component with the new behaviour https://github.com/openmrs/openmrs-esm-core/pull/1881

denniskigen:
Can we drop the "{{count}} diagnoses on this note" line and replace the helper text with:

> Unticked diagnoses are saved as secondary and provisional.

The count sits directly above the cards it's counting. "Tick Primary and Confirmed where they apply" describes checkboxes that are right there, and "at least one diagnosis must be marked primary" is redundant now that the first diagnosis defaults to primary.

nit: It's probably better to move the remove button to the top right of the card, level with the diagnosis name, so the bottom row is just Primary and Confirmed.

CleanShot 2026-09-17 at 8 15 29 PM@2x

denniskigen:
nit: In the spirit of trying to simplify the UI, can we consider dropping the "Search for a diagnosis to add" caption above the search input? It repeats the "Diagnosis" column label and the placeholder, and none of the other search fields in the chart have one. Carbon's Search hides its label on purpose and relies on the placeholder as the visible prompt, so the input keeps its accessible name without it.

CleanShot 2026-09-17 at 8 34 24 PM@2x

packages/esm-patient-notes-app/src/notes/visit-notes-form.component.tsx
```diff
@@ -570,46 +534,18 @@ const VisitNotesForm: React.FC<VisitNotesFormProps> = ({
...
-                <FormGroup legendText={t('searchForPrimaryDiagnosis', 'Search for a primary diagnosis')}>
+                <FormGroup legendText={t('searchForDiagnosis', 'Search for a diagnosis to add')}>
                   <DiagnosisSearch
-                    name="primaryDiagnosisSearch"
+                    name="diagnosisSearch"
                     control={control}
-                    labelText={t('enterPrimaryDiagnoses', 'Enter Primary diagnoses')}
-                    placeholder={t('primaryDiagnosisInputPlaceholder', 'Choose a primary diagnosis')}
+                    labelText={t('searchForDiagnosis', 'Search for a diagnosis to add')}
+                    placeholder={t('diagnosisInputPlaceholder', 'Choose a diagnosis')}
```
denniskigen, line 547:
```suggestion
                    placeholder={t('diagnosisInputPlaceholder', 'Search for a diagnosis')}
```
nit: "Choose" reads like picking from a list. Since this becomes the only prompt once the caption goes, it's probably better to match the other search fields in the chart ("Search for a drug or orderset", "Search for a cause of death"). The e2e spec and one unit test look the input up by placeholder, so they'd change too.

One thing to watch on the placeholder suggestion: `extract-translations` keeps the existing value for a key that already exists, so `en.json` needs the same edit by hand or the UI will keep showing "Choose a diagnosis".

denniskigen:
nit: Same spirit, can we also consider dropping the "No diagnosis selected — Enter a diagnosis above" line? It only shows when the list is visibly empty, and the search placeholder right above already says what to do. Two tests assert on it; asserting that no diagnosis group is left does the same job.

That whittles the input down to just the bare essentials:

CleanShot 2026-09-18 at 2 37 56 PM@2x

PR #3626, opened 2026-09-06
(fix) O3-5957: Correct MUAC color code age boundaries

Faisal77666:
## Requirements

## Summary

`getMuacColorCode` (in `vitals-biometrics-form.utils.ts`) had inconsistent age-bracket boundary conditions. Patients aged exactly 10 or 18 got no color code set at all, for any MUAC value, because adjacent brackets both excluded the boundary age.

**Update after review:** the initial fix only closed the gap without checking which side of each boundary was correct. Per [@denniskigen's review](#), age brackets and MUAC cut-offs now follow the Uganda Ministry of Health IMAM Guidelines (Jan 2016), Table 2: Summary of Classification of Acute Malnutrition (https://platform.who.int/docs/default-source/mca-documents/policy-documents/guideline/UGA-CH-38-03-GUIDELINE-2016-eng-IMAM-Guidelines-for-Uganda-Jan-2016.pdf).

Since `age` is the whole number of years extracted from the patient's birth date, each bracket's **lower bound is inclusive** (e.g. a patient turning exactly 5 falls into the "5 to <10 years" bracket, not "6-59 months"). The brackets are now:

| Age | SAM (red) | MAM (yellow) | Healthy (green) |
|---|---|---|---|
| under 5y | ≤11.5cm | 11.5–<12.5cm | ≥12.5cm |
| 5 to <10y | ≤13.5cm | 13.5–<14.5cm | ≥14.5cm |
| 10 to <15y | ≤16.0cm | 16.0–<18.5cm | ≥18.5cm |
| 15 to <18y | ≤18.5cm | 18.5–<21.0cm | ≥21.0cm |
| 18y+ (adults) | ≤19.0cm | 19.0–<22.0cm | ≥22.0cm |

This also splits the old combined "10 to <18" bracket into the two separate brackets (10-<15 and 15-<18) that the guideline actually specifies, and corrects their MUAC thresholds (previously both used 16.5/19.0, which matched neither bracket).

Verified against the reviewer's own example: a 5-year-old with MUAC 13.0cm now correctly classifies as red/SAM (previously green under the first pass, and previously no color at all before any fix).

Before any fix, 3/3 boundary-age tests fail:
getMuacColorCode (3)
× sets a color code for a healthy 5-year-old (age boundary)
× sets a color code for a 10-year-old (age boundary)
× sets a color code for an 18-year-old (age boundary)

After the guideline-aligned fix, all tests pass, including a new test for the 15-year-old boundary and the reviewer's 5y/13.0cm example:
getMuacColorCode (4)
✓ classifies a 5-year-old with MUAC 13.0cm as SAM (red), per the 5-<10y bracket
✓ sets a color code for a 10-year-old (age boundary)
✓ sets a color code for a 15-year-old (age boundary)
✓ sets a color code for an 18-year-old (age boundary), classified with adults

Full esm-patient-vitals-app test suite also passes.

## Screenshots

N/A  this is a pure logic fix in a utility function, not directly rendered UI. The `useMuacColors` config flag is off by default, so behavior is covered by unit tests rather than a visual screenshot.

## Related Issue

https://issues.openmrs.org/browse/O3-5957

## Other

Found via direct code review of the vitals/biometrics module. Boundary-age direction and bracket split corrected per review feedback, using the Uganda MoH IMAM Guidelines as the source of truth.

### DIFF
--- a/packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.test.ts
+++ b/packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.test.ts
@@ -2,27 +2,35 @@ import { describe, it, expect } from 'vitest';
 import { getMuacColorCode } from './vitals-biometrics-form.utils';
 
 describe('getMuacColorCode', () => {
-  it('sets a color code for a healthy 5-year-old (age boundary)', () => {
+  it('classifies a 5-year-old with MUAC 13.0cm as SAM (red), per the 5-<10y bracket', () => {
     let result = '';
-    getMuacColorCode(5, 13, (color) => {
+    getMuacColorCode(5, 13.0, (color) => {
       result = color;
     });
-    expect(result).toBe('green');
+    expect(result).toBe('red');
   });
 
   it('sets a color code for a 10-year-old (age boundary)', () => {
     let result = '';
-    getMuacColorCode(10, 20, (color) => {
+    getMuacColorCode(10, 17.0, (color) => {
       result = color;
     });
-    expect(result).toBe('green');
+    expect(result).toBe('yellow');
   });
 
-  it('sets a color code for an 18-year-old (age boundary)', () => {
+  it('sets a color code for a 15-year-old (age boundary)', () => {
     let result = '';
-    getMuacColorCode(18, 20, (color) => {
+    getMuacColorCode(15, 20.0, (color) => {
       result = color;
     });
-    expect(result).toBe('green');
+    expect(result).toBe('yellow');
+  });
+
+  it('sets a color code for an 18-year-old (age boundary), classified with adults', () => {
+    let result = '';
+    getMuacColorCode(18, 20.0, (color) => {
+      result = color;
+    });
+    expect(result).toBe('yellow');
   });
 });
--- a/packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.test.ts
+++ b/packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.test.ts
@@ -1,0 +1,28 @@
+import { describe, it, expect } from 'vitest';
+import { getMuacColorCode } from './vitals-biometrics-form.utils';
+
+describe('getMuacColorCode', () => {
+  it('sets a color code for a healthy 5-year-old (age boundary)', () => {
+    let result = '';
+    getMuacColorCode(5, 13, (color) => {
+      result = color;
+    });
+    expect(result).toBe('green');
+  });
+
+  it('sets a color code for a 10-year-old (age boundary)', () => {
+    let result = '';
+    getMuacColorCode(10, 20, (color) => {
+      result = color;
+    });
+    expect(result).toBe('green');
+  });
+
+  it('sets a color code for an 18-year-old (age boundary)', () => {
+    let result = '';
+    getMuacColorCode(18, 20, (color) => {
+      result = color;
+    });
+    expect(result).toBe('green');
+  });
+});
--- a/packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
+++ b/packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
@@ -54,44 +54,44 @@ export function extractNumbers(str: string) {
 
 export function getMuacColorCode(age: number, muac: number, setColorCode: (color) => void) {
   switch (true) {
-    // children 5 years and below with a muac equal to 14
+    // children 5 years and below (0-5 inclusive)
     case age <= 5 && muac <= 11.5 && muac > 0:
       setColorCode('red');
       break;
-    case age < 5 && muac > 11.5 && muac < 12.5:
+    case age <= 5 && muac > 11.5 && muac < 12.5:
       setColorCode('yellow');
       break;
-    case age < 5 && muac > 12.5:
+    case age <= 5 && muac >= 12.5:
       setColorCode('green');
       break;
-    // above 5 but less than 10
-    case age > 5 && age < 10 && muac <= 13.5 && muac > 0:
+    // above 5, up to and including 10
+    case age > 5 && age <= 10 && muac <= 13.5 && muac > 0:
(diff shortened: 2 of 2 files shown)


packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    });
+    expect(result).not.toBe('');
+  });
+
+  it('sets a color code for an 18-year-old (age boundary)', () => {
+    let result = '';
+    getMuacColorCode(18, 20, (color) => {
+      result = color;
+    });
+    expect(result).not.toBe('');
```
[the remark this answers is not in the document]
Faisal77666: Good catch  fixed in the latest commit. Both the age-10 and age-18 tests now assert the specific expected color (`green`) instead of just checking that a color was returned, so a regression in bracket selection would actually be caught.

Faisal77666:
@denniskigen  would appreciate a review when you have a moment! This is a small, well-tested boundary-condition fix in the vitals module  happy to make any changes needed.

packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
```diff
@@ -54,44 +54,44 @@ export function extractNumbers(str: string) {
 
 export function getMuacColorCode(age: number, muac: number, setColorCode: (color) => void) {
   switch (true) {
-    // children 5 years and below with a muac equal to 14
+    // children 5 years and below (0-5 inclusive)
     case age <= 5 && muac <= 11.5 && muac > 0:
```
denniskigen: Before we settle on which bracket the boundary ages fall into, can we check this against the intended protocol? For children aged two and above, the caller extracts whole years from `age(birthDate)`, so `age === 5` is a child anywhere from 5 years 0 months to 5 years 11 months. The [Uganda guideline linked here](https://platform.who.int/docs/default-source/mca-documents/policy-documents/guideline/UGA-CH-38-03-GUIDELINE-2016-eng-IMAM-Guidelines-for-Uganda-Jan-2016.pdf) brackets ages as 6 to 59 months, 5 to <10 years, 10 to <15, 15 to <18, and 18 and above. Under that, a 5-year-old belongs in the 13.5/14.5 bracket, a 10-year-old in the next one up, and an 18-year-old with the adults. This PR puts each of them in the younger bracket, which has the lower cut-offs. A 5-year-old with a MUAC of 13.0 cm shows green here but is SAM (<13.5) per that guideline.

It's probably better to confirm the intended protocol with a clinician before changing the comparisons and expected colors. The guideline also splits ages 10 to <15 and 15 to <18, which the current function combines, so changing the boundary direction alone wouldn't fully align it.

packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.test.ts
```diff
@@ -0,0 +1,28 @@
+import { describe, it, expect } from 'vitest';
+import { getMuacColorCode } from './vitals-biometrics-form.utils';
+
+describe('getMuacColorCode', () => {
```
denniskigen, line 4: nit: this PR also moves the MUAC-value boundaries (exactly 12.5, 14.5, 19.0 and 22.0 in their respective age brackets now get green instead of nothing). Worth a line in the PR description. It'd be nice to parameterize these cases to cover the exact MUAC thresholds as well, with expectations based on whichever protocol we agree on above.

denniskigen (commented):
Thanks for the careful writeup, @Faisal77666. The gap at 10 and 18 is real and the fix is tidy. One question below on which side of the boundary each age should land, since that's the thing we'd be locking in, plus a nit on the tests.

Not for this PR, but worth a follow-up ticket: the caller does `extractNumbers(age(birthDate))`, which pulls the first number out of a display string. For a 13-month-old, `age()` can return a months-and-days display string, so that child comes through here as age 13. I'll file it.

packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.ts
Faisal77666: Confirmed and fixed in the latest push  you're right about the direction. Age brackets now use the lower bound as inclusive (age >= 5, age >= 10, etc.), matching the "5 to <10 years" phrasing in the guideline. Also split the combined 10-18 bracket into the two separate 10-<15 and 15-<18 brackets from Table 2, each with its own correct MUAC thresholds.

Verified against your example: a 5-year-old with MUAC 13.0cm now correctly shows red/SAM instead of green.

packages/esm-patient-vitals-app/src/vitals-biometrics-form/vitals-biometrics-form.utils.test.ts
Faisal77666: Good catch, updated the PR description to call out the MUAC-value boundary changes explicitly.

For the parameterized tests happy to convert these to test.each once we're settled on the final bracket agreement, so I'm not rewriting them twice if anything shifts. Let me know if you'd rather have that done now.

Faisal77666:
@denniskigen please take a look

VeronicaMuthee:
@Faisal77666 The SAM/MAM boundary seems to be on the wrong side - please cross-check. The guideline says SAM is MUAC < 11.5 and MAM is ≥ 11.5 to < 12.5 (same pattern in every bracket). The code uses muac <= 11.5 → red, so a patient at exactly 11.5 (or 13.5, 16.0, 18.5, 19.0) shows red when the guideline says yellow.

image

Faisal77666:
@VeronicaMuthee  Good catch  you're right, the guideline uses strict `<` for SAM and `≥` for the MAM lower bound, not the inclusive `<=` I had. Fixed across all five brackets (all "red" cases now use `<`, all "yellow" lower bounds now use `>=`). Added a test asserting MUAC exactly at a cut-off (13.5 for the 5-<10y bracket) classifies as yellow, not red  matching the guideline's own image you shared.

Faisal77666:
@denniskigen @VeronicaMuthee  please review

PR #3627, opened 2026-09-06
(fix) O3-5801: Display location names for location observations

pinankshah8-Aa:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a conventional commit label.

## Summary

Fixes [O3-5801](https://openmrs.atlassian.net/browse/O3-5801) by displaying the human-readable name for reference-valued observations such as locations.

Encounter observations previously derived the displayed answer from `obs.display`. For reference-valued observations, the actual readable resource name is available as `obs.value.display`, while `obs.display` may contain the referenced resource ID.

This change uses `obs.value.display` when the observation value is a reference resource with a UUID, while preserving the existing `obs.display` parsing for primitive values and file/image values that do not carry a resource UUID. The same handling is applied to grouped observation members.

Tests cover:
- reference-valued observations such as locations
- primitive observation values
- file values without a reference UUID, preserving the original filename
- reference-valued group members

### Before
The Location observation shows the numeric resource ID (`9`) instead of the location name.

Before — location observation displays numeric ID

### After
Using a fresh dev3 test patient, the Ward Admission form was completed with **Inpatient Ward**, then verified from **Visits → All encounters** with this branch loaded locally. The encounter now displays the human-readable location name.

After — location observation displays Inpatient Ward

## Related Issue

https://openmrs.atlassian.net/browse/O3-5801

## Other

Local validation completed:
- patient-chart package tests: 192 passed, 1 skipped
- focused encounter-observation regression tests: 4 passed
- TypeScript: passed
- ESLint/Prettier pre-commit checks: passed
- repository pre-push `yarn verify`: passed
- manual end-to-end reproduction on dev3 with a fresh test patient: passed
- git diff --check: passed

[O3-5801]: https://openmrs.atlassian.net/browse/O3-5801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/esm-patient-chart-app/src/visit/visits-widget/encounter-observations/encounter-observations.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit/visits-widget/encounter-observations/encounter-observations.component.tsx
@@ -25,6 +25,8 @@ const EncounterObservations: React.FC<EncounterObservationsProps> = ({ observati
     if (
       obs.value !== null &&
       typeof obs.value === 'object' &&
+      'uuid' in obs.value &&
+      typeof obs.value.uuid === 'string' &&
       'display' in obs.value &&
       typeof obs.value.display === 'string'
     ) {
@@ -57,7 +59,7 @@ const EncounterObservations: React.FC<EncounterObservationsProps> = ({ observati
               <span className={styles.parentConcept}>{obs.concept.display}</span>
               <span />
               {obs.groupMembers.map((member) => (
-                <React.Fragment key={index}>
+                <React.Fragment key={member.uuid}>
                   <span className={styles.childConcept}>{member.concept.display}</span>
                   <span>{getAnswer(member)}</span>
                 </React.Fragment>
--- a/packages/esm-patient-chart-app/src/visit/visits-widget/encounter-observations/encounter-observations.component.tsx
+++ b/packages/esm-patient-chart-app/src/visit/visits-widget/encounter-observations/encounter-observations.component.tsx
@@ -21,6 +21,19 @@ const EncounterObservations: React.FC<EncounterObservationsProps> = ({ observati
     }
   }
 
+  function getAnswer(obs: Obs): string {
+    if (
+      obs.value !== null &&
+      typeof obs.value === 'object' &&
+      'display' in obs.value &&
+      typeof obs.value.display === 'string'
+    ) {
+      return obs.value.display;
+    }
+
+    return getAnswerFromDisplay(obs.display);
+  }
+
   const filteredObservations = !!obsConceptUuidsToHide.length
     ? observations?.filter((obs) => {
         return !obsConceptUuidsToHide.includes(obs?.concept?.uuid);
@@ -46,7 +59,7 @@ const EncounterObservations: React.FC<EncounterObservationsProps> = ({ observati
               {obs.groupMembers.map((member) => (
                 <React.Fragment key={index}>
                   <span className={styles.childConcept}>{member.concept.display}</span>
-                  <span>{getAnswerFromDisplay(member.display)}</span>
+                  <span>{getAnswer(member)}</span>
                 </React.Fragment>
               ))}
             </React.Fragment>
@@ -55,7 +68,7 @@ const EncounterObservations: React.FC<EncounterObservationsProps> = ({ observati
           return (
             <React.Fragment key={index}>
               <span>{obs.concept.display}</span>
-              <span>{getAnswerFromDisplay(obs.display)}</span>
+              <span>{getAnswer(obs)}</span>
             </React.Fragment>
           );
         }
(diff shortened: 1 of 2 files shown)


denniskigen:
Thanks, @pinankshah8-Aa, for fixing the location names and adding tests for grouped observations too! I’ve left one suggestion to preserve file names for image and file observations.

packages/esm-patient-chart-app/src/visit/visits-widget/encounter-observations/encounter-observations.component.tsx
```diff
@@ -21,6 +21,19 @@ const EncounterObservations: React.FC<EncounterObservationsProps> = ({ observati
     }
   }
 
+  function getAnswer(obs: Obs): string {
+    if (
+      obs.value !== null &&
+      typeof obs.value === 'object' &&
+      'display' in obs.value &&
```
denniskigen, line 30: This fixes the location names, thanks! Can we also check that `obs.value` has a `uuid` before using its `display`? Image and file observations have `display: 'raw file'`, so this change replaces their file names with “raw file”. Checking for a UUID lets those observations keep using the existing fallback. Can we add a test for that case too?

packages/esm-patient-chart-app/src/visit/visits-widget/encounter-observations/encounter-observations.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+    },
+    value: {
+      uuid: 'location-uuid',
+      display: 'Inpatient Ward',
+    },
+    ...overrides,
+  }) as Obs;
+
+beforeEach(() => {
+  mockUseConfig.mockReturnValue({ obsConceptUuidsToHide: [] } as any);
```
denniskigen: nit: can we use `vi.mocked(useConfig<ChartConfig>)` and `getDefaultsFromConfigSchema(esmPatientChartSchema)` here, as in [visit-summary.test.tsx](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-chart-app/src/visit/visits-widget/past-visits-components/visit-summary.test.tsx)? That removes the `as any` type casts and follows the existing test setup.

packages/esm-patient-chart-app/src/visit/visits-widget/encounter-observations/encounter-observations.component.tsx
```diff
@@ -46,7 +59,7 @@ const EncounterObservations: React.FC<EncounterObservationsProps> = ({ observati
               {obs.groupMembers.map((member) => (
                 <React.Fragment key={index}>
```
denniskigen: nit: pre-existing, but since you're touching this hunk: this code currently reuses the parent's `index`, so every member in the group gets the same key. Can we change it to use the unique UUID instead?

```suggestion
                <React.Fragment key={member.uuid}>
```

packages/esm-patient-chart-app/src/visit/visits-widget/encounter-observations/encounter-observations.component.tsx
pinankshah8-Aa: Updated in b3a583cf. `obs.value.display` is now used only when the object also has a string `uuid`, so file/image values without a resource UUID keep the existing `obs.display` fallback. I also added a regression test that verifies a file value with `display: 'raw file'` still renders its filename.

packages/esm-patient-chart-app/src/visit/visits-widget/encounter-observations/encounter-observations.test.tsx
pinankshah8-Aa: Updated in b3a583cf to use `vi.mocked(useConfig<ChartConfig>)` with `getDefaultsFromConfigSchema(esmPatientChartSchema)`, so the `as any` config mock is removed and the test setup follows the existing pattern.

packages/esm-patient-chart-app/src/visit/visits-widget/encounter-observations/encounter-observations.component.tsx
pinankshah8-Aa: Updated in b3a583cf. Grouped observation members now use `member.uuid` as the fragment key.

packages/esm-patient-chart-app/src/visit/visits-widget/encounter-observations/encounter-observations.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+    value: {
+      uuid: 'location-uuid',
+      display: 'Inpatient Ward',
+    },
+    ...overrides,
+  }) as Obs;
+
+beforeEach(() => {
+  mockUseConfig.mockReturnValue(getDefaultsFromConfigSchema(esmPatientChartSchema));
+});
```
denniskigen, line 27:
```suggestion
});

```

denniskigen (approved):
LGTM. Thanks, @pinankshah8-Aa, for addressing the feedback and adding the file-name regression test!

PR #3657, opened 2026-09-15
O3-5976: Fix Angular locale import requesting non-existent .mjs files

Kalyangotimothy:
## Summary

Fixes [O3-5976](https://openmrs.atlassian.net/browse/O3-5976): Form engine never loads Angular locale data because the locale import requests `.mjs` files that do not exist.

## Problem

`FeWrapperComponent.loadAllFormDependencies()` lazily imports Angular locale data using a webpack context restricted to `.mjs`:

```ts
import(
  /* webpackInclude: /\.mjs$/ */
  `../../../node_modules/@angular/common/locales/${locale}.mjs`
)
```

`@angular/common/locales/` ships only `.js` files, so the webpack context is empty, `dist/assets/l10n/locales/` is never emitted, and the import rejects on every form launch with `Cannot find module './en.mjs'`.

`registerLocaleData()` has therefore never run. The engine falls back to Angular's built-in `en-US` data, so dates and numbers in every form are formatted as US English regardless of the user's locale.

### `fe-wrapper.component.ts`
- **`webpackInclude`**: Changed from `/\.mjs$/` to `/[\/\\\\][a-z]{2}\.js$/` — matches the `.js` files Angular actually ships, and narrows to two-letter base language files only (~30 chunks instead of all 1,731 CLDR files / ~2.6 MB)
- **Import path**: Changed from relative `node_modules` path with `.mjs` to `@angular/common/locales/${locale}.js`
- **Error handling**: Added `.catch()` so unsupported locales degrade gracefully without logging unhandled rejections

## Test Results

The existing test suite passes. The component's existing test is marked `xit` (skipped) due to missing providers — this is pre-existing and unrelated to this change.

[O3-5976]: https://openmrs.atlassian.net/browse/O3-5976?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


solomonfortune (commented):
@Kalyangotimothy Could you please revise the title and description to match the OpenMRS standard????

packages/esm-form-entry-app/src/app/fe-wrapper/fe-wrapper.component.ts
```diff
@@ -137,11 +137,15 @@ export class FeWrapperComponent implements OnInit, OnDestroy {
     this.translateService.use(locale);
 
     import(
-      /* webpackInclude: /\.mjs$/ */
+      /* webpackInclude: /[\/\\][a-z]{2}\.js$/ */
```
ibacher: The ticket specifically asked this to be changed to `\.js$`. Just use that.

ibacher: But see Greptile's note below.

packages/esm-form-entry-app/src/app/fe-wrapper/fe-wrapper.component.ts
```diff
@@ -137,11 +137,15 @@ export class FeWrapperComponent implements OnInit, OnDestroy {
...
       /* webpackChunkName: "./assets/l10n/locales/[request]"*/
       /* webpackMode: "lazy" */
-      `../../../node_modules/@angular/common/locales/${locale}.mjs`
-    ).then((module) => registerLocaleData(module.default));
+      `@angular/common/locales/${locale}.js`
```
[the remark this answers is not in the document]
ibacher: The fix suggested in the ticket was premised on still using the raw path. Like Greptile says, if you switch to using the import style, you need to actually respect the package.json _exports_ which are `@angular/common/locales/en`, not `en.js`.

packages/esm-form-entry-app/src/app/fe-wrapper/fe-wrapper.component.ts
```diff
@@ -137,11 +137,15 @@ export class FeWrapperComponent implements OnInit, OnDestroy {
...
+    )
+      .then((module) => registerLocaleData(module.default))
+      .catch(() => {
+        // Unsupported locale — Angular falls back to en-US automatically.
+      });
```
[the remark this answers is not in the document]
ibacher: Right, why are we adding this catch?

ibacher (changes requested):

packages/esm-form-entry-app/src/app/fe-wrapper/fe-wrapper.component.ts
ibacher: Basically, because the export does not include the file path, this rule can no longer be used at all.

ibacher:
@Kalyangotimothy Thanks for the contribution. A couple of notes. Please do not ignore our PR template. PRs that ignore our template will not be merged. That template is for everyone's benefit as it also documents things like the format we expect from PR titles and commits, which this PR also needs to be updated to respect.

packages/esm-form-entry-app/src/app/fe-wrapper/fe-wrapper.component.ts
ibacher, line 140: My earlier review was wrong on this, but the ticket assumed we'd be loading the files with the extension pattern, so asking for this change sort of no longer makes a lot of sense with the export change. I think this winds up just being:

```
/\/[a-z]{2}$/ 
```

Note that we do not need to support `\` in the file separator since that's not a resolvable path by the resolution rules we're using here.

PR #3673, opened 2026-09-18
(fix) O3-5987: Let callers restrict the attachment dialog's file types and quiet its upload toast

denniskigen:
## Requirements

## Summary

The Add attachment dialog took an `allowedExtensions` prop but never read it: every step fetched the backend's `attachments.allowedFileExtensions` itself, so the visit note form's request to leave out PDFs did nothing. The helper text advertised pdf, a PDF went through to the review step, and the note ended up with a blank thumbnail. The dialog also told the user "Upload complete" the moment the caller's `saveFile` resolved, which in the note form only stages the file for upload on save.

The dialog now puts the caller's list on its context and falls back to the backend list only when the caller passes nothing. The upload and review steps read from the context instead of fetching on their own. The success snackbar can be switched off with `showUploadSnackbar`; the note form does that, and passes `undefined` instead of an empty list while the backend list is loading, since an empty list means nothing is allowed. Building the error message no longer pops the last extension off the shared list, which mutated it for the next attempt.

The staged thumbnail's alt now falls back to the file name when there's no description, and the visit-note e2e spec checks that thumbnail instead of the removed toast.

Verified against a local backend: the note form's dialog lists only jpeg, jpg and png, refuses a PDF before the review step, and staging a PNG shows no toast. The Attachments page still passes the backend list and behaves as before.

## Screenshots

PDF refused in the note form's dialog:

o3-5987-pdf-refused

PNG staged with no "Upload complete" toast:

o3-5987-no-toast

## Related Issue

https://openmrs.atlassian.net/browse/O3-5987

## Other

Found while working on O3-5986 (#3668). Related: O3-5988 (the dialog's size limit isn't the backend's), O3-5989 (#3670).


packages/esm-patient-notes-app/src/notes/visit-notes-form.component.tsx
```diff
@@ -349,12 +349,13 @@ const VisitNotesForm: React.FC<VisitNotesFormProps> = ({
       closeModal: () => {
         close();
       },
-      allowedExtensions:
-        allowedFileExtensions && Array.isArray(allowedFileExtensions)
-          ? allowedFileExtensions.filter((ext) => !/pdf/i.test(ext))
-          : [],
+      // The note only takes images. Leave this undefined until the backend list has loaded so the
+      // dialog falls back to it instead of treating an empty list as "nothing allowed".
+      allowedExtensions: allowedFileExtensions?.filter((ext) => !/pdf/i.test(ext)),
```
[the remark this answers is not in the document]
denniskigen: Agreed. The form now passes an image allowlist (jpg, jpeg, png, gif, webp, bmp) intersected with the backend's list, rather than dropping pdf alone.

packages/esm-patient-attachments-app/src/camera-media-uploader/camera-media-uploader.component.tsx
```diff
@@ -77,7 +83,7 @@ const CameraMediaUploaderModal: React.FC<CameraMediaUploaderModalProps> = ({
   return (
     <CameraMediaUploaderContext.Provider
       value={{
-        allowedExtensions: allowedFileExtensions,
+        allowedExtensions: allowedExtensions ?? allowedFileExtensions,
```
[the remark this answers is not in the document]
denniskigen: Done. The dialog lower-cases and strips a leading dot once, when it builds its context, so 'PNG' and '.png' both accept scan.PNG and the accept list reads .png.

packages/esm-patient-attachments-app/src/camera-media-uploader/camera-media-uploader.component.test.tsx
```diff
@@ -0,0 +1,13 @@
...
+import CameraMediaUploaderModal from './camera-media-uploader.component';
+
+test("normalises the caller's extensions before using them for the picker", () => {
+  render(
+    <CameraMediaUploaderModal allowedExtensions={['.PNG', 'Jpg', ' jpg ']} closeModal={vi.fn()} saveFile={vi.fn()} />,
+  );
+
+  expect(screen.getByText(/supported files are png, jpg\./i)).toBeInTheDocument();
+  expect(screen.getByLabelText(/drag and drop files here/i)).toHaveAttribute('accept', '.png,.jpg');
+});
```
NethmiRodrigo: Can we add a case for the fallback too, so `allowedExtensions ?? allowedFileExtensions` is covered from both sides? E.g. `uses the backend list when the caller passes none`:

```
vi.mock('@openmrs/esm-patient-common-lib', () => ({
  useAllowedFileExtensions: vi.fn(() => ({ allowedFileExtensions: ['pdf', 'jpg'], error: undefined, isLoading: false })),
  useMaxAttachmentFileSize: () => ({ maxFileSize: 5, error: undefined, isLoading: false, isValidating: false, retry: vi.fn() }),
}));

test('uses the backend list when the caller passes none', () => {
  render(<CameraMediaUploaderModal closeModal={vi.fn()} saveFile={vi.fn()} />);

  expect(screen.getByLabelText(/drag and drop files here/i)).toHaveAttribute('accept', '.pdf,.jpg');
});
```

packages/esm-patient-attachments-app/src/camera-media-uploader/media-uploader.component.test.tsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
```
NethmiRodrigo: Nit but since `media-uploader.test.tsx` already exists, should we consider adding these tests there?

packages/esm-patient-attachments-app/src/camera-media-uploader/media-uploader.component.tsx
```diff
@@ -35,17 +38,18 @@ const MediaUploaderComponent = () => {
             )} ${maxFileSize} MB.`,
           });
         } else if (!isFileExtensionAllowed(file.name, allowedFileExtensions)) {
-          const lastExtension = allowedFileExtensions.pop();
+          const otherExtensions = allowedFileExtensions.slice(0, -1);
```
NethmiRodrigo: Nit:
When only one type is allowed (for example the note form passes `['png']`), `otherExtensions` is empty and the message reads "extensions: , or png." Fixing it means changing the English string in `en.json`, which resets its translations.

packages/esm-patient-notes-app/src/notes/visit-notes-form.component.tsx
```diff
@@ -139,7 +142,13 @@ const VisitNotesForm: React.FC<VisitNotesFormProps> = ({
...
   const [saveError, setSaveError] = useState<string>();
   const [error, setError] = useState<Error>(null);
-  const { allowedFileExtensions } = useAllowedFileExtensions();
+  const {
+    allowedFileExtensions,
+    error: allowedFileExtensionsError,
+    isLoading: isLoadingAllowedFileExtensions,
+  } = useAllowedFileExtensions();
+  const isImageCaptureDisabled =
+    isLoadingAllowedFileExtensions || Boolean(allowedFileExtensionsError) || !allowedFileExtensions;
```
NethmiRodrigo: Doesn't an empty `attachments.allowedFileExtensions` mean the backend allows every type? Because when it is empty the [hook gives `undefined`](https://github.com/openmrs/openmrs-esm-patient-chart/blob/74146327c577e7ffa9e712f58470ea9617227bf2/packages/esm-patient-common-lib/src/useAllowedFileExtensions.ts#L18-L22) for it and `!allowedFileExtensions` then disables Add image for good, while a list with no image type gives the dialog `[]`, which refuses everything with "Supported files are .". Should we offer every image type in the first case and disable the button in the second?
```suggestion
  const allowedImageExtensions = useMemo(
    () =>
      allowedFileExtensions
        ? allowedFileExtensions.filter((ext) => imageExtensions.has(ext.toLowerCase()))
        : [...imageExtensions],
    [allowedFileExtensions],
  );
  const isImageCaptureDisabled =
    isLoadingAllowedFileExtensions || Boolean(allowedFileExtensionsError) || allowedImageExtensions.length === 0;
```

packages/esm-patient-notes-app/src/notes/visit-notes-form.component.tsx
```diff
@@ -356,14 +369,13 @@ const VisitNotesForm: React.FC<VisitNotesFormProps> = ({
...
-          ? allowedFileExtensions.filter((ext) => !/pdf/i.test(ext))
-          : [],
+      allowedExtensions: allowedFileExtensions?.filter((ext) => imageExtensions.has(ext.toLowerCase())),
       collectDescription: true,
       multipleFiles: true,
+      // Files are only staged here; they upload when the note is saved.
+      showUploadSnackbar: false,
     });
-  }, [allowedFileExtensions, getValues, setValue]);
+  }, [allowedFileExtensions, getValues, isImageCaptureDisabled, setValue]);
```
NethmiRodrigo: Because of the comment ^
```suggestion
      allowedExtensions: allowedImageExtensions,
      collectDescription: true,
      multipleFiles: true,
      // Files are only staged here; they upload when the note is saved.
      showUploadSnackbar: false,
    });
  }, [allowedImageExtensions, getValues, isImageCaptureDisabled, setValue]);
```

packages/esm-patient-notes-app/src/notes/visit-notes-form.component.tsx
```diff
@@ -776,6 +788,7 @@ const VisitNotesForm: React.FC<VisitNotesFormProps> = ({
                   </p>
                   <Button
                     className={styles.uploadButton}
+                    disabled={isImageCaptureDisabled}
```
NethmiRodrigo: When the request for the extension list fails the button just stays grey with no hint why. Should we say so, like we do for `savedImagesError` below?

```
{allowedFileExtensionsError && (
  <InlineNotification
    className={styles.savedImagesLoading}
    kind="error"
    lowContrast
    hideCloseButton
    title={t('allowedFileExtensionsLoadError', "Couldn't load the allowed image formats")}
    subtitle={t('allowedFileExtensionsLoadErrorHint', 'Reload the page to add images to this note.')}
  />
)}
```

NethmiRodrigo (commented):
Thanks @denniskigen! A couple of suggestions -

